PR #55, opened 2026-04-04
fix: prevent false 'waiting' state during tool execution

### LINKED ISSUE #53: False 'waiting' state during tool execution
## Bug

Sessions that are continuously processing (running tools like Bash, Read, Write, etc.) briefly flash as **waiting** for several seconds before returning to **working**. The session should stay **working** the entire time.

## Root Cause

`NeedsUserAttention()` returns true whenever `HasOpenToolCall` is true and the tool isn't `Agent` or `mcp__*`. Between a `tool_use` being written in the transcript and the corresponding `tool_result` arriving, every standard tool (Bash, Read, Write, Edit, Grep, Glob, etc.) triggers a false **waiting** transition that lasts for the duration of tool execution.

Irrlicht can't distinguish between:
- **Tool awaiting user approval** (truly waiting)
- **Tool executing automatically** (still working)

Both look identical in the transcript: `tool_use` written, `tool_result` not yet written.

## Observed

Session `7728f143` was running continuously with background processing. The UI showed it flipping to **waiting** (orange) for a few seconds multiple times during the run, then back to **working** (purple).

## Expected

Session should stay **working** (purple) the entire time tools are executing. Only transition to **waiting** when the agent is genuinely blocked on user input (e.g. `AskUserQuestion`).

## Proposed Fix

Flip `isAutoExecutingTool` logic in `core/domain/session/session.go`: instead of allowlisting `Agent` and `mcp__*` as auto-executing, treat **all** tools as auto-executing except `AskUserQuestion` (the one tool that always blocks for user input regardless of permission settings).

**Key files:**
- `core/domain/session/session.go:74-78` — `isAutoExecutingTool()`
- `core/domain/session/session.go:54-72` — `NeedsUserAttention()`
- `core/application/services/session_detector.go:286` — state machine branch using `NeedsUserAttention()`

ingo-eichhorst:
## Summary
- Flips `isAutoExecutingTool()` allowlist → `isUserBlockingTool()` blocklist in session state detection
- Only `AskUserQuestion` and `ExitPlanMode` trigger the "waiting" state — all other tools (Bash, Read, Write, Agent, MCP, etc.) keep the session "working"
- Eliminates false orange flashes in the menu bar during continuous tool execution

Closes #53

## Test plan
- [x] Unit tests updated and passing (10 cases in `TestNeedsUserAttention`)
- [x] Integration tests updated (`TransitionsToWaiting` tests now use user-blocking tool names)
- [x] Manual test: built and installed on macOS, session stays purple (working) during tool execution

### DIFF
--- a/core/application/services/session_detector_test.go
+++ b/core/application/services/session_detector_test.go
@@ -103,8 +103,9 @@ func TestSessionDetector_Activity_TransitionsToWaiting_WhenToolUseOpen(t *testin
 		UpdatedAt:      time.Now().Unix(),
 		EventCount:     1,
 		Metrics: &session.SessionMetrics{
-			LastEventType:   "tool_use",
-			HasOpenToolCall: true,
+			LastEventType:     "tool_use",
+			HasOpenToolCall:   true,
+			LastOpenToolNames: []string{"AskUserQuestion"},
 		},
 	}
 
@@ -184,8 +185,9 @@ func TestSessionDetector_Activity_TransitionsToWaiting_WhenAssistantButOpenTools
 		FirstSeen:      time.Now().Unix(),
 		UpdatedAt:      time.Now().Unix(),
 		Metrics: &session.SessionMetrics{
-			LastEventType:   "assistant",
-			HasOpenToolCall: true,
+			LastEventType:     "assistant",
+			HasOpenToolCall:   true,
+			LastOpenToolNames: []string{"ExitPlanMode"},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


ingo-eichhorst (commented):
## Review

Clean, well-scoped fix. The allowlist→blocklist flip is the right architectural choice — there are ~2 tools that always block for user input vs. dozens that may auto-execute.

**Observations:**

- The "no names" default flip (`{HasOpenToolCall: true, LastOpenToolNames: []}` → `false`) is the right call — false "waiting" flashes are worse UX than occasionally missing a prompt
- `ExitPlanMode` inclusion alongside `AskUserQuestion` is correct
- Accepted trade-off: permission prompts (e.g. Bash awaiting approval) will now show "working" instead of "waiting", since the transcript can't distinguish the two states
- Tests are comprehensive (10 cases covering all scenarios including mixed tools)

**Minor suggestion:** Consider a comment on `isUserBlockingTool` noting the list should be extended if Claude Code adds more always-blocking tools in the future.

LGTM — ship it.

PR #71, opened 2026-04-05
fix: skip local commands in tailer to prevent orphan sessions

### LINKED ISSUE #62: Session stays working after ! shell escape command
## Bug

When executing a shell command via the `!` prefix in Claude Code (e.g. `! ls`), the session transitions to **working** and never returns to **ready**. It should remain in the **ready** state since the `!` command is a user-side shell escape, not agent activity.

## Steps to Reproduce

1. Start a Claude Code session (irrlicht detects it as ready)
2. Type `! ls` or any `!` prefixed command in the prompt
3. Observe the session transitions to **working** in irrlicht
4. The session never transitions back to **ready**

## Expected Behavior

The session should stay in **ready** state — `!` commands are user-initiated shell escapes, not agent turns.

## Likely Cause

The `!` command writes a user event to the transcript, which triggers `onActivity` in the session detector. The `ClassifyState` logic sees transcript activity and transitions to working, but no subsequent agent turn (assistant message or turn_done) follows to transition back to ready.

### LINKED ISSUE #64: Orphan session remains after /clear
## Description

After running `/clear` in Claude Code, the previous session is not properly cleaned up and remains as an orphan session in irrlicht.

## Expected behavior

When `/clear` is issued, the previous session should transition to `ready` state and be cleaned up.

## Actual behavior

The previous session persists as an orphan — it is no longer associated with an active Claude Code conversation but remains visible in irrlicht.

ingo-eichhorst:
## Summary

- Local commands (`/clear`, `/context`, `/model`, `! <cmd>`) wrote user messages with XML-tagged content to the JSONL transcript. The tailer treated these as normal user input, triggering `working` state with no agent turn to transition back — leaving orphan sessions.
- Skip user messages starting with `<local-command-caveat>`, `<command-name>`, or `<local-command-stdout>` to preserve previous state
- Handle `system subtype=local_command` as `turn_done` so `/clear` actively transitions to `ready`
- Skill invocations (`/seo`, `/simplify`) start with `<command-message>` and pass through unaffected
- **Session merge on /clear**: when the same PID appears on a new transcript, the old session adopts the new transcript and the new session is deleted — preserving session identity instead of creating duplicates

Closes #62, closes #64

## Test plan

- [x] 5 new tailer tests covering `/clear`, `/context`, `!` shell escape, skill pass-through, and `/clear` mid-turn
- [x] 2 updated PID assignment tests verifying merge behavior (old session survives, new session deleted)
- [ ] Manual: build daemon, run `/clear` in Claude Code, verify old session persists with new transcript
- [ ] Manual: run `/context` in Claude Code, verify session stays `ready`
- [ ] Manual: run `! ls` in Claude Code, verify session stays `ready`


ingo-eichhorst:
won't merge. issues with /clear. we use anothere appraoch in another pr

PR #127, opened 2026-04-10
feat: record and replay full session lifecycle events

### LINKED ISSUE #107: Record and replay full session lifecycle events, not just transcript activity
## Summary

The replay tooling should be able to reproduce the full lifecycle a coding agent session can go through, not just transcript-driven state transitions.

Today `core/cmd/replay-session` replays `.jsonl` transcript lines through the tailer + classifier with virtual time. That is useful for state-classifier regressions, but it does not cover the rest of the inputs that actually shape a session's lifecycle in production.

We need a session-scoped event log that captures all relevant events and side effects so a later replay can reproduce the whole session, including process lifecycle and non-transcript signals.

## Problem

Several session behaviors depend on more than transcript content:

- process start / PID assignment
- transcript file creation
- transcript activity / debounce behavior
- transcript removal
- process exit
- parent/child linking for subagents via `ParentSessionID`
- other files or signals that influence session detection, cleanup, or state transitions

With the current transcript-only replay, we can reproduce classifier output for one narrow path, but we cannot reliably replay full detector behavior when the bug depends on watcher events, process lifecycle, file removal, startup/cleanup behavior, or ordering between those signals.

## Desired Outcome

Introduce a capture format and replay path that can record and later replay the full event stream relevant to a session.

The log should be append-only, timestamped, and assigned to the session so we can replay:

- the main transcript events
- process lifecycle events
- file-level events relevant to the session
- detector-relevant side signals
- parent/child session relationships

The replay should then drive the same production lifecycle handling we use live, so side events are visible during replay instead of being reconstructed heuristically from the transcript alone.

## Acceptance Criteria

- Add a session-scoped event log format for replay capture.
- Every captured event includes enough metadata to correlate it to the session and preserve ordering.
- Captured events cover at least:
  - process start
  - PID assignment / discovery
  - transcript create / write / remove
  - process exit
  - parent session linkage for child sessions
  - any additional file or signal inputs that materially affect session processing
- Replay can consume this log and reconstruct the full session lifecycle, not only transcript-derived state transitions.
- Replay preserves ordering and timestamps well enough to reproduce debounce-sensitive and race-sensitive behavior.
- Multi-session logs still retain per-session assignment so child/background sessions can be replayed alongside the parent.
- The replay output exposes the side events as part of the report so we can inspect what happened, not only the final state transitions.

## Implementation Notes

A reasonable direction is to treat the replay input as a normalized event stream for the `SessionDetector` boundary rather than a transcript-only file.

That likely means introducing explicit event types for:

- watcher events
- process watcher callbacks
- transcript filesystem events
- session creation / deletion side effects
- child-session linkage events

This should stay compatible with the current three-state model (`working`, `waiting`, `ready`) and keep session ownership explicit, especially for subagents.

## Why This Matters

If we want replay to cover "all the scenarios a coding agent can run through", then transcript-only fixtures are not enough. We need to be able to replay the full stream of lifecycle inputs that produced a given session, including side events that happen outside the transcript itself.

ingo-eichhorst:
## Summary

Closes #107. Ships lifecycle event recording + sidecar-driven regression testing, plus a series of detector bug fixes surfaced by the new visibility. 17 commits.

### 1. Lifecycle event recording (`irrlichd --record`)
- New `lifecycle.Event` domain type and `EventRecorder` outbound port
- JSONL recorder adapter with periodic 5s flush + random filename suffix so sub-second restarts don't collide
- SessionDetector + PIDManager instrumentation at every lifecycle point: transcript new/activity/removed, state transitions, parent linkage, PID discovery, process exit, debounce coalescing
- Opt-in via `--record` flag or `IRRLICHT_RECORD=1`; zero overhead when disabled
- Recordings land in `~/.local/share/irrlicht/recordings/<timestamp>-<suffix>.jsonl`, one file per daemon run
- `ir:test-mac` skill updated to always pass `--record` in dev launches

### 2. `replay-lifecycle` report tool
- New CLI for ad-hoc timeline analysis of raw recordings
- Per-session state durations, PID discovery lag, debounce coalescing stats, events-by-kind breakdown

### 3. Sidecar-driven regression testing via `replay-session`
- When a `<fixture>.events.jsonl` sidecar is present next to a transcript, `replay-session` drives the tailer directly from the sidecar's `transcript_activity` events (each with file_size) — byte-identical reproduction of the daemon's state transitions
- Applies the daemon's debounce state machine over the sidecar events, respects `process_exited` to cancel pending debounce timers
- Any mismatch exits non-zero → CI catches classifier/tailer/debounce drift
- `curate-lifecycle-fixture.sh` turns a raw recording + transcript + subagents dir into a committable fixture bundle; `replay-fixtures.sh` walks them automatically

### 4. Committed fixtures with extended check
- `10-full-lifecycle-839f0678` — single-session fixture with pre-session detection, question-waiting, process exit. **10/10 ordered match ✓**
- `11-background-agents-189ff529` — new bundle format with 12 in-process subagents, 323-event lifecycle log, 12 subagent transcripts, process_exited signal. **10/10 ordered match ✓**

### 5. Detector bugs surfaced + fixed by the recording feature

| Commit | Bug | Fix |
|---|---|---|
| `84fbeea` | fswatcher missed subagent transcripts created alongside their parent directory | Recursive `addSubtree` walk when new dirs appear at runtime (was already correct at startup) |
| `fd28197` | Parent stuck in `working` after liveness sweep deletes its last child | `onChildDeleted` callback from PIDManager → SessionDetector.`reevaluateParent` |
| `a7a38dc` | Parent held for 2 min waiting for sweep when subagent ends with `stop_reason: null` | New `finishOrphanedChildren` promotes children with no open tool calls when parent's turn is done |
| `bc065b4` | Above fix was too aggressive for background agents between tool calls — they got falsely promoted and the daemon re-created them on the next write, causing UI flicker | Added `os.Stat` + mtime check in `finishOrphanedChildren`: skip children whose transcripts were written in the last N seconds |
| `66f5981` | N=2s window was still too tight for background Task agents with 5-15s API response gaps | Bumped N to 30s, comfortably above normal API latency, still 4× faster than the 2-min sweep fallback |

## Test coverage

- **Recorder**: 5 unit tests (round-trip, concurrent writes, nested dir creation, sub-second restart collision, Close vs flush race)
- **Replay-session**: 6 unit tests (golden fixture per-transition assertion, drift detection, non-sidecar `Replay()` golden, adapter detection, malformed sidecar line handling, multi-session sidecar rejection)
- **Session detector**: 3 new parent-child tests (orphaned subagents fast-forward, active subagents not promoted, background subagents not fast-forwarded, parent released when child swept by liveness)
- **fswatcher**: 1 new test for nested-subtree Create events
- **Statement coverage on `replay-session/main.go`**: 46.8% → 76.3%

## Files

**New:**
- `core/domain/lifecycle/event.go`
- `core/adapters/outbound/recorder/jsonl.go` + test
- `core/ports/outbound/ports.go` — `EventRecorder` interface
- `core/cmd/replay-lifecycle/main.go`
- `core/cmd/replay-session/main_test.go`
- `scripts/curate-lifecycle-fixture.sh`
- `testdata/replay/claudecode/10-full-lifecycle-839f0678.{jsonl,events.jsonl}`
- `testdata/replay/claudecode/11-background-agents-189ff529.{jsonl,events.jsonl,subagents/*}`

**Modified:**
- `core/application/services/session_detector.go` — recorder field, 11+ instrumentation points, `finishOrphanedChildren`, `onChildDeleted` wiring, `RunPIDLivenessSweepForTest` seam
- `core/application/services/pid_manager.go` — recorder field, 2 instrumentation points, `SetChildDeletedHandler`, sweep fires callback
- `core/application/services/session_detector_test.go` — 6 new tests
- `core/adapters/inbound/agents/fswatcher/watcher.go` — `addSubtree` recursive helper for runtime Create events
- `core/adapters/inbound/agents/fswatcher/watcher_test.go` — nested-subtree regression test
- `core/cmd/replay-session/main.go` — `ReplayWithSidecar`, extended check, null-timestamp fix, primary-session filter for multi-session sidecars
- `core/cmd/irrlichd/main.go` — `--record` flag + env var wiring
- `scripts/replay-fixtures.sh` — skip `*.events.jsonl` and `*.subagents/` during the walk
- `.claude/skills/ir:test-mac/SKILL.md` — pass `--record` in dev launches

## Test plan
- [x] `go test ./core/...` — all suites pass, replay-session coverage at 76.3%
- [x] `scripts/replay-fixtures.sh` — 13 fixtures replay, exit 0, sidecar fixtures both 10/10 ✓
- [x] Verified the detector regression tests fail on the pre-fix code (temporarily disabling each fix, observing the matching test fail, restoring)
- [x] Live-tested the daemon with real Claude Code background-agent runs: UI shows circles throughout, parent transitions to ready within ~2s of the last subagent finishing


ingo-eichhorst:
Merged locally to main in commit 46cbcdf. PR branch stays on origin for reference.

PR #165, opened 2026-04-19
refactor(capacity): LiteLLM as single source of truth

ingo-eichhorst:
## Summary

Deletes the hand-maintained `core/pkg/capacity/model-capacity.json` and reworks capacity lookups to trust only the LiteLLM remote fetch.

### Why
The embedded table and LiteLLM data were drifting. When LiteLLM lacked a direct entry for a newer model (e.g. `claude-opus-4-7`), the fuzzy fallback matched `claude-4.1-opus` (200K, no 1m beta), producing the 200K/1M flip-flop depending on whether the transcript happened to carry the `[1m]` marker. Hand-patching the embedded table (as in #164) doesn't address the root issue: we were maintaining a table LiteLLM already owns.

### What changed
- `model-capacity.json` removed; `//go:embed` removed.
- `DefaultCapacityManager()` returns a process-wide singleton backed only by `~/.local/share/irrlicht/model-capacity-cache.json`, hot-reloading when the file's mtime advances. Existing tailers now see the daily refresh without a daemon restart.
- `GetModelCapacity` is a direct map lookup. No fuzzy patterns, no family defaults, no global fallback. Absent model → zero value → UI shows tokens only (both macOS and web already degrade gracefully).
- Content-based token estimation dropped (`char_to_token_ratio` removed). Transcript without token data keeps `total_tokens = 0`.
- `BetaFeatures[context_1m]` check removed from tailer — LiteLLM already reports the real window directly.
- `[1m]` suffix override (parser.go:273, tailer.go:451) kept — it's the only per-request override.

### Behaviour changes
- `claude-opus-4-7` now reliably shows 1M (was 200K/1M flip). Verified live.
- `claude-sonnet-4-6` now shows 1M (was conservatively 200K in the embedded table). Verified live.
- Offline first-run: raw tokens only until the first LiteLLM fetch succeeds.

### Supersedes
Closes #164.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` green.
- [x] Deterministic test fixture in `newTestTailer` (via new `capacity.NewForTest` helper) so tailer tests don't depend on the real on-disk cache.
- [x] New hot-reload test (`TestMaybeReload_PicksUpNewCache`) and concurrent-access test.
- [x] Manual: rebuild daemon, confirm opus-4-7 and sonnet-4-6 sessions show `context_window=1000000`.
- [x] Manual: daemon restart re-fetches LiteLLM cache and singleton reloads transparently.


ingo-eichhorst:
Live-verified against the dev stack after rebuild. Both previously-flipping models now report 1M directly from LiteLLM:

\`\`\`
model=claude-sonnet-4-6  context_window=1000000  pressure=safe
model=claude-opus-4-7    context_window=1000000  pressure=safe
\`\`\`

Daemon fetch on cold start also confirmed — singleton hot-reloads the cache file without restart.

PR #255, opened 2026-04-28
feat: add OpenCode adapter (issue #100)

### LINKED ISSUE #100: Support for OpenCode
Add session monitoring support for [OpenCode](https://github.com/anomalyco/opencode), an open-source AI coding agent.

This includes:
- Detecting OpenCode sessions
- Parsing OpenCode transcript/state files
- Tracking session state (working/waiting/ready)

## References
- https://github.com/anomalyco/opencode

gimmebytes:
## Summary

Closes #100 — adds session monitoring support for OpenCode (https://github.com/anomalyco/opencode).

### New: core/adapters/inbound/agents/opencode/
| File | Purpose |
|---|---|
| adapter.go | Package constants: AdapterName, ProcessName, dbRelPath |
| watcher.go | fsnotify-based WAL watcher — polls session + part tables, emits EventNewSession / EventActivity / EventRemoved |
| parser.go | Maps OpenCode part.data rows to ParsedEvent: step-finish (reason=stop/interrupted/length/error) -> turn_done, text -> assistant_message/user_message, tool -> function_call/function_call_output |
| metrics.go | SQLite-backed ComputeMetrics that queries parts directly |
| pid.go | CWD-based PID discovery via pgrep -x opencode |
| parser_test.go | 30 unit tests covering all part types and edge cases |
| watcher_test.go | 8 integration tests for scan logic, cursor dedup, EventRemoved, parent-child |
| metrics_test.go | 8 tests for parseTranscriptPath format parsing |

### Fixes applied during review
- step-finish reasons interrupted (Ctrl+C), length, error now emit turn_done instead of assistant_message so cancelled/errored sessions properly transition to ready
- EventRemoved is emitted when session.time_archived is set, preventing stale sessions lingering until TTL sweep
- Cursor race condition fixed: uses >= with part-ID dedup to avoid missing parts with identical millisecond timestamps
- PerTurnContribution now accounts for all non-tool-calls reasons (interrupted/length/error also consumed tokens)

### Framework extension: agents.Config.ComputeMetrics (optional)
Backwards-compatible MetricsProvider field — all existing adapters leave it nil.

### Wiring
- core/cmd/irrlichd/main.go: registers opencode.Config() + dedicated opencode.New() watcher
- core/cmd/replay/main.go: registers opencode adapter and detectAdapter() path

### New dependency: modernc.org/sqlite
Pure-Go SQLite driver — no CGo, no system libraries.

### Replay fixtures
- replaydata/agents/opencode/capabilities.json
- replaydata/agents/opencode/scenarios/baseline-hello/

## Testing
All three test layers pass:
go test ./core/... -race -count=1
tools/replay-fixtures.sh


gimmebytes:
Hi @ingo-eichhorst !

I've finished the work with OpenCode (Go) and tested the Backend only, the WebUI and the Mac-native app.

If you would like to test OpenCode with Irrlicht yourself, you can do the following:

OpenCode kurz testen (auch mit free Modellen):

1. Installieren — via npm:

npm install -g @anomalyco/opencode oder direkt: npx @anomalyco/opencode

2. Free Model einrichten — OpenCode nutzt Anthropic/OpenAI default, aber Du kannst auf OpenRouter (free credits) oder lokale Modelle via Ollama wechseln:

opencode config set modelProvider openrouter
opencode config set model openrouter/anthropic/claude-sonnet-4-5

opencode config set modelProvider ollama
opencode config set model ollama/qwen2.5-coder:7b

3. Starten:
cd ~/dein-projekt
opencode

4. Irrlicht sieht's dann automatisch — der OpenCode-Adapter überwacht ~/.local/share/opencode/opencode.db.

And I saw your dark-factory design idea: would you like to let Claude to automatically discover new agents, or hint claude and let claude do the work then?

ingo-eichhorst:
First of all thanks for the contribution. It's motivating and exiting to form a community.

I started checking the PR and most things just work. That is the experience i want for the users. Thanks!

In my first manual test I stumbled over some minor issues. Worth fixing before merge. If your LLM is multi-model it should be able to read and understand the images and reason about a fix :-)

## 1. For Permission Requests the state is not changing to waiting state:
image

## 2. Looks like the waiting text detection is based on the beginning of the output text. It should only be the end (see claude code for reference)
image

## 3. For this model I have specified the context length and it's also shown in opencode, but its not transmitted to the swift app:
image


... Continuing now with review ...

ingo-eichhorst (changes requested):
## Overview

Adds a SQLite-backed adapter for OpenCode. Architecturally clean: hexagonal layers respected (`domain → ports → adapters`), parser kept inside the adapter package per project convention, three-state model preserved, `ParentSessionID` used for subagent linking. Solid test coverage on the parser (30 cases) and watcher (8 cases).

A few issues to address before merge.

## Blockers

### 1. `cleanupChildren` deletes all OpenCode subagents on every parent turn-end

The PR adds a `?session=` exemption inside `CheckPIDLiveness`'s child-cleanup branch (`pid_manager.go:482-503`) so DB-backed-adapter children aren't reaped by the readyTTL sweep — the documented invariant is *"DB-backed adapters... manage their own session lifetime via maxAge; their subagents are persistent historical records."*

But the same protection is missing from `cleanupChildren` (`pid_manager.go:233-245`), which is invoked unconditionally from `session_detector_activity.go:378-380` whenever a top-level parent transitions to `ready`:

```go
if state.State == session.StateReady && state.ParentSessionID == "" {
    d.pidMgr.cleanupChildren(state.SessionID)
}
```

Trace:
1. Parent finishes a turn — `step-finish reason=stop` → `turn_done` → `ClassifyState` → `StateReady`.
2. `processActivity` calls `cleanupChildren(parent.id)`.
3. `cleanupChildren` deletes every child with that `ParentSessionID`. The `?session=` check in `CheckPIDLiveness` never even runs here.
4. The watcher's `cursors` map still holds those session IDs, so `if !known` is skipped on the next scan and **`EventNewSession` is never re-emitted**. Subagents are gone from the UI until daemon restart.

This is the exact regression the readyTTL exemption was added to prevent — applied in one of the two places that delete DB-backed children, missed in the other. Fix: add the same guard to the `cleanupChildren` loop, or extract an `isDBBackedTranscript(path)` helper used by both sites and `helpers.go:25`.

### 2. `gofmt` fails on 4 files

```
core/application/services/pid_manager.go
core/adapters/outbound/metrics/adapter.go
core/adapters/inbound/agents/opencode/metrics.go
core/adapters/inbound/agents/opencode/parser_test.go
```

`pid_manager.go:482-504` has the entire `ParentSessionID` branch indented one tab too few — likely a copy/paste mishap. The struct in `adapter.go:28-33` is unaligned. `gofmt -w` fixes everything.

### 3. Misleading `MetricsProvider` contract

`core/adapters/outbound/metrics/adapter.go:102`:

```go
return provider(transcriptPath, adapter)  // passes adapter NAME as sessionID
```

The provider's second parameter is documented as `sessionID`, but the adapter passes the adapter name. It only works because OpenCode encodes the session ID in `transcriptPath` as `?session=...` and `parseTranscriptPath` ignores the second arg in that case. If `transcriptPath` ever lacks the query suffix, the provider treats `"opencode"` as a session ID and silently returns `nil`. Either:
- Pass `""` as second arg (and rename it to clarify it's a fallback hint), or
- Rework the port to take an explicit `sessionID` parameter.

### 4. Hardcoded user path in golden fixture

`replaydata/agents/opencode/scenarios/baseline-hello/transcript.jsonl.replay.json.golden:3`:

```json
"source_transcript": "/Users/marvin.kruse/Source/Irrlicht/replaydata/agents/opencode/scenarios/baseline-hello/transcript.jsonl"
```

This will fail `tools/replay-fixtures.sh` on every other contributor's machine. Use a relative or stripped path.

## Maintainer-flagged regressions (confirmed with code citations)

### 5. Assistant text truncated from the wrong end

`parser.go:166-170` uses `text[:200]`, keeping the **first** 200 bytes. Assistant questions live at the **end** of long messages, so this throws away the question and leaves the preamble. The canonical `tailer.ExtractAssistantText` (`core/pkg/tailer/parser.go:310-313`) takes the **last** 200 runes with a leading "…":

```go
runes := []rune(text)
if len(runes) > 200 {
    return "…" + string(runes[len(runes)-200:])
}
return text
```

Also a byte-vs-rune issue: `text[:200]` on UTF-8 can split CJK / accented / emoji characters mid-rune, producing invalid UTF-8. Replace with the rune-aware tail-slicing helper, or factor it out so OpenCode and the JSONL adapters share one implementation. Update `TestParser_TextPart_LongTruncated` to assert leading "…" + tail content.

### 6. Permission requests never trigger `waiting`

`parseToolPart` (`parser.go:188-217`) emits a bare `function_call` with the raw tool name (`bash`, `read`, ...) for any pending/running tool — no special-casing for OpenCode's permission part. None of `ClassifyState`'s three paths to `StateWaiting` fires:

- Rule 0 (`PermissionPending`) is hook-only and there's no `/api/v1/hooks/opencode` endpoint.
- Rule 1 (`isUserBlockingTool`) only matches Claude's `AskUserQuestion` / `ExitPlanMode`.
- Rule 2 (`IsAgentDone() && IsWaitingForUserInput()`) requires a `turn_done`; permission prompts arrive mid-turn.

Fix: surface OpenCode's permission part shape explicitly — either set `ev.PermissionPending`, push a synthetic blocking tool name into `LastOpenToolNames`, or extend `isUserBlockingTool` with OpenCode's permission tool names. `replaydata/agents/opencode/capabilities.json` declares `"permission_hooks": true`, so the data exists in OpenCode's schema; need to identify the exact part shape.

### 7. Context window / pressure level never computed

`querySessionMetrics` (`metrics.go:75-79`) initializes `PressureLevel: "unknown"` and never derives `ContextWindow` / `ContextUtilization` / `PressureLevel` from `(modelName, totalTokens)`. The capacity-manager logic only exists as a method on `TranscriptTailer.computeContextUtilization` (`core/pkg/tailer/tailer_metrics.go:314`), which the OpenCode path bypasses entirely.

Result: every OpenCode session shows an empty progress bar and a `?` pressure pill in the WebUI and macOS app.

Fix: factor a helper `(modelName, totalTokens) → (window, util%, pressure, unknown)` out of `tailer_metrics.computeContextUtilization` and call it from both sites. Set `ContextWindowUnknown=true` when no capacity entry exists so the macOS tokens-only fallback still renders.

## Significant issues

### 8. Errors silently swallowed

`watcher.go:1396, 1424, 1556` and `metrics.go:181, 221`: `if err != nil { return }` / `return nil, nil`. The project convention (per `CLAUDE.md`) is to log via the `Logger` interface, not silently absorb. SQL open errors, scan errors, and query failures will be invisible in production.

### 9. `agent.Event.TranscriptPath` semantics quietly changed

`core/domain/agent/event.go:23` still documents this field as "Absolute path to the .jsonl file" but it now sometimes carries `?session=…` query strings, and downstream code (`helpers.go:isStaleTranscript`, `pid_manager.go`) uses `strings.Contains(..., "?session=")` to branch behavior. Either:
- Update the comment to reflect the new convention and centralize the parsing in an `isDBBackedTranscript(path)` helper, or
- Add an explicit `agent.Event.DBSessionID` (or similar) field instead of overloading the path.

The current scheme leaks SQLite-adapter implementation details into shared core code and will multiply if a second DB-backed adapter shows up. Bonus: the two predicates in use today disagree (`helpers.go:25` uses `IndexByte('?')`, `pid_manager.go:489` uses `Contains("?session=")`); functionally equivalent now, drift waiting to happen.

### 10. PID discovery via CWD only

`pid.go` uses `pgrep -x opencode` + CWD match. If two OpenCode instances run from the same directory (e.g. background agent + interactive), the disambiguator can't distinguish them since OpenCode doesn't hold the DB file open. Worth a comment acknowledging this trade-off (or a follow-up issue).

## Nits

- **Trailing-edge debounce missing** (`watcher.go:178-195`). 500 ms debounce is leading-edge only; the final write of a burst (e.g. the `step-finish reason=stop` row) is silently dropped if it arrives within the window. Recovers via `refreshStaleSessions` ~5 s later for `working` sessions, but `ready→working` transitions are not covered. Standard trailing-edge `time.AfterFunc(minScanGap-elapsed, scanSessions)` fix. Side note: the stated `.db-shm` feedback-loop rationale is moot — the event filter at line 147 already drops anything but `opencode.db` / `opencode.db-wal`.
- **`ElapsedSeconds` always 0** (`metrics.go:110-130`). `lastTS` is updated in the loop but never read back into `metrics.ElapsedSeconds`. Active sessions hide it via the client-side `firstSeen` clock, but ready/idle OpenCode rows show blank elapsed. Track `firstTS` alongside and assign `metrics.ElapsedSeconds = int64(lastTS.Sub(firstTS).Seconds())`.
- **First `emitRemovedForArchivedSessions` swallows archives** (`watcher.go:240-260`). First call returns early after stamping `lastArchivedCheck = time.Now()`. A session archived in the microsecond window between the SELECT and the stamp is permanently missed, leaking a cursor entry. Initialize `lastArchivedCheck` in `New()` / `NewWithDBPath()`.
- **Initial-scan race against `Subscribe`** (`watcher.go:112-120`). 200 ms `time.Sleep` to let `detector.Subscribe()` "win the race" before broadcasting. Replace with synchronization (subscribe before launching the goroutine, or expose a Ready channel from `detector.Run`). The author's own comment acknowledges the race.
- **`NewWithDBPath` ignores its arg** (`watcher.go:150-165, 459`). Constructor takes an arbitrary path but `Watch()` / `waitForDB` hardcode `opencode.db` / `opencode.db-wal`. Zero impact today (production always uses `opencode.db`, tests bypass `Watch()`), but the constructor's signature lies. Two-line fix: `dbName := filepath.Base(w.dbPath)`.
- **`_ = cur` dead code** (`watcher.go:1529`). Either use it or don't fetch it.
- **`ParentSessionID` lacks struct alignment** (`event.go:25`). Trivial; gofmt fixes.
- **`home, _ := os.UserHomeDir()`** (`watcher.go:1277`). If `$HOME` is unset, you get a relative path silently. `LogWarn` is cheap.
- **`minScanGap = 500ms` magic number** (`watcher.go:1386`). Worth a named constant.
- **`metrics.go:179` opens a fresh DB connection per `ComputeMetrics`** call. `sql.DB` is a connection pool; cache one per DB path. Acceptable for now.
- **Replay fixture format diverges from real OpenCode storage** — synthetic `_role`/`_cwd`/`_ts` keys mixed in. Note this in a README so readers don't expect 1:1 schema fidelity.
- **`SessionState.swift` missing trailing newline** after the new function.

## Follow-up: drop the dashboard UI changes

The `core/cmd/irrlichd/ui/index.html` rework (per-row history bars, granularity toggle, view-mode toggle, server-backed history polling) has been extracted into #258 (tracking issue: #257). Could you drop those `index.html` changes from this branch (e.g. `git checkout main -- core/cmd/irrlichd/ui/index.html` + commit) so #255 reviews cleanly as the OpenCode adapter alone?

## Follow-up: revisit `modernc.org/sqlite` dependency

Concrete numbers: `modernc.org/sqlite` adds **+5 MB binary growth** (12 MB → 17 MB on the daemon, ~42%) and **8 transitive dependencies** (`modernc.org/{sqlite,libc,mathutil,memory}` + `dustin/go-humanize`, `google/uuid`, `mattn/go-isatty`, `ncruces/go-strftime`, `remyoudompheng/bigfft`). `modernc.org/libc` is a transpiled C runtime — most of the weight. Every other adapter in this project (claudecode, codex, pi, aider) ships zero DB deps.

OpenCode does not write a file-based shadow of session content (verified: `storage/session_diff/*.json` is 2 bytes empty arrays; `log/*.log` only contains event-type announcements without payloads; `snapshot/<hash>` is content-addressed file snapshots, not session content). The DB is the only canonical store, so we can't avoid reading SQLite — but we *can* avoid the Go driver.

Suggestion for a follow-up PR (not blocking this one): replace `_ "modernc.org/sqlite"` + `sql.Open` with `exec.Command("/usr/bin/sqlite3", "-json", dbPath, query)`. Eliminates all 8 deps and the 5 MB. ~10 ms subprocess spawn per call, invisible behind the existing 500 ms debounce. macOS-first project + macOS ships `sqlite3` = no real downside. Tests need `sqlite3` on PATH (CI already has it) and a tempdir DB instead of `:memory:`. Bonus: `/usr/bin/sqlite3` follows the OS, so it tracks OpenCode's SQLite version more closely than a vendored Go driver does.

There's also a more ambitious option (hit OpenCode's local HTTP API for live sessions, fall back to DB for historical) — visible from the daemon logs as `service=server method=GET path=/session ...` — but that's a bigger refactor and best filed as a separate issue.

## Strengths

- Cursor race fix is a nice piece of work — the `>=` + per-timestamp `seenIDs` set is the right pattern for the shared-millisecond case, and `TestScanSessions_CursorRace` exercises both new and re-scan paths.
- Step-finish reason handling (`stop` / `interrupted` / `length` / `error` → `turn_done`; `tool-calls` → `assistant_message`) is correct and well-tested.
- `EventRemoved` on `time_archived` matches the existing fswatcher contract — sessions don't linger until the TTL sweep.
- `ComputeMetrics` accumulating cost/tokens across all non-`tool-calls` reasons is the right call for cost accuracy.
- Pure-Go SQLite (`modernc.org/sqlite`) avoids CGo — keeps cross-compilation simple (though see follow-up above for whether the trade-off is right).
- Capability declaration follows the canonical scenario matrix convention.

## Verdict

Not ready to merge. Fix the four blockers above (cleanupChildren regression, gofmt, MetricsProvider contract, golden-fixture path), address the three maintainer-flagged regressions (text truncation / permission state / context pressure) with concrete pointers above, drop the UI changes (now in #258), and address the silent-error swallowing. The `TranscriptPath` / sessionID overloading is the architectural concern worth discussing — either commit to the convention with proper docs or add a dedicated field.

core/cmd/irrlichd/ui/index.html
ingo-eichhorst, line 1: This is a unrelated change. Moved it to: #257 
Please delete changes here and use new pr: #258

core/adapters/inbound/agents/opencode/metrics.go
```diff
@@ -0,0 +1,184 @@
+package opencode
+
+import (
+	"database/sql"
+	"encoding/json"
+	"strings"
+	"time"
+
+	_ "modernc.org/sqlite"
```
ingo-eichhorst, line 10: I really like the WAL sqlite approach. It's pretty clever and the perfect fit for OpenCode.

However, I really dislike the increased bundle size (+40%). I see 2 options. Both have tradeoffs.

---

1. Status quo — modernc.org/sqlite
- +5 MB binary, 8 transitive deps forever
- Pure Go, easy cross-compile
- Works for live + historical sessions
- One code path

2. Shell out to /usr/bin/sqlite3 -json
- Zero new Go deps
- ~10 ms subprocess spawn per call, invisible behind the existing 500 ms debounce
- Single code path, ~1:1 translation from sql.Query to exec.Command
- macOS ships sqlite3 by default; CI does too
- Tests need sqlite3 on PATH (already true) and a tempdir DB instead of :memory:
- Slightly more robust against OpenCode SQLite version bumps than modernc, since /usr/bin/sqlite3 follows the OS

---

Maybe another alternative would be to ship both a small and a large version of the tool and the installer picks the one or the other based on the available tools.

I don't know?! It's late.

What do you think?

core/go.mod
```diff
@@ -12,9 +12,18 @@ require (
 
 require (
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
+	github.com/dustin/go-humanize v1.0.1 // indirect
```
ingo-eichhorst, line 15: Do we really need all this new dependencies?
Please recheck.

---

@ingo-eichhorst  
Maybe we can also get rid of some existing ones while I'm looking at:
`bigfft v0.0.0` -> that looks fishy (created a new issue for checking: #259)

platforms/macos/Irrlicht/Models/SessionState.swift
ingo-eichhorst, line 1: @ingo-eichhorst 
new icons should live in adapter: #260

gimmebytes:
@ingo-eichhorst thanks for double checking! Will fix the mentioned things and will get back to you when things are ready and btw: thanks for Irrlicht! Will promote it to our engineers :-)

gimmebytes:
@ingo-eichhorst Sorry, it took me a while to fix the PR (long weekend), but now everything should be ready.

core/adapters/inbound/agents/opencode/metrics.go
ingo-eichhorst: #276

ingo-eichhorst (approved):

ingo-eichhorst:
> @ingo-eichhorst Sorry, it took me a while to fix the PR (long weekend), but now everything should be ready.

@gimmebytes 

Many thanks for your contribution. And also thanks for the nice words. It would be wonderful if more and more people use the tool. 

I will also position it as a learning project for ai driven contributions more and more.

If you have any feedback or ideas for improvement, I'm more than happy to jump on a call or chat on LinkedIn or https://github.com/ingo-eichhorst/Irrlicht/discussions

And, BTW: I'll have a workshop with ~100 students on Friday about AI Coding with OpenCode and we'll use your adapter to keep track of the agents. Great timing!

core/cmd/irrlichd/ui/index.html
gimmebytes: All changes to index.html have been reverted :-)

core/adapters/inbound/agents/opencode/metrics.go
gimmebytes: @ingo-eichhorst I usally ship modernc.org/sqlite with my apps directly, despite the slighty larger distributable. When moving to /usr/bin/sqlite3 I could imagine installing things on windows, when this target should be supported in the future, might be cumbersome, compared to the backed in sqlite support? Both paths would be fine for me.

core/go.mod
gimmebytes: go-humanize comes from modernc.org/sqlite. So we can only get rid of it, when shifting out to sqlite3 locally.

platforms/macos/Irrlicht/Models/SessionState.swift
gimmebytes: Anything to do here from my end?

gimmebytes:
@ingo-eichhorst happy, to support here :-)

I really like the idea of 

> I will also position it as a learning project for ai driven contributions more and more.

because having a real world ai-first project with a real purpose and benefit for the end-user is the ultimate showcase for Claude and Friends :-)

> And, BTW: I'll have a workshop with ~100 students on Friday about AI Coding with OpenCode and we'll use your adapter to keep track of the agents. Great timing!

:-)

Will catch up on LinkedIn with you, the latest after my week off next week.

Regarding my contribution: anything left before merging? I can assist also on the follow ups. 

A similar tool I stumbled over is 

https://github.com/getagentseal/codeburn

It would be great if irrlicht could also extend a little bit on the actual costs burned. Afaik Irrlicht has support alredy, but the codeburn statistics seem more sophisticated.

Have a good one!

core/adapters/inbound/agents/opencode/metrics.go
ingo-eichhorst: agree. let's leave the overhead for now

core/go.mod
ingo-eichhorst: agree. it's ugly, but we cannot change it

platforms/macos/Irrlicht/Models/SessionState.swift
ingo-eichhorst: nope. that was just a clean up reminder for me. thanks

PR #279, opened 2026-05-06
ci(coverage): make badge update non-blocking, surface HTTP errors

ingo-eichhorst:
## Summary
- Coverage badge update failed post-merge of #255 with `curl` exit 22 (HTTP ≥400) — see [run 25460501651](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/25460501651/job/74701154054). Most likely the `GIST_SECRET` PAT expired (last successful run was 2026-05-02).
- Badge updates are non-critical; shouldn't show a red ✗ on main.
- Adds `continue-on-error: true`, logs HTTP status + response body, and emits a workflow warning naming the likely root cause.

## Follow-up
Rotate `GIST_SECRET` in repo secrets (PAT with `gist` scope) to restore the badge. The workflow itself will now succeed regardless.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -57,15 +57,24 @@ jobs:
           echo "COLOR=$COLOR" >> $GITHUB_ENV
 
       - name: Update Gist with coverage
+        continue-on-error: true
         env:
           GIST_SECRET: ${{ secrets.GIST_SECRET }}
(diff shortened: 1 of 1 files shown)


ingo-eichhorst:
I want any issue to be visible. That is part of the pipeline and if it has issues, we need to get notified so we can fix. No workarounds for issues.

PR #284, opened 2026-05-09
refactor(agents): co-locate display name + icons with Go adapters (#260)

### LINKED ISSUE #260: refactor: co-locate adapter display name + icon with the Go adapter, drop Swift switch statements
Adding a new adapter today requires changes on both sides of the codebase:

1. Go: `core/adapters/inbound/agents/<adapter>/` — the actual session-monitoring code.
2. Swift: `platforms/macos/Irrlicht/Models/SessionState.swift` — three places:
   - `adapterName(...)` switch (display string)
   - `adapterIcon(...)` switch (which SVG to render)
   - A new `<adapter>SVG(dark:)` function carrying the inline SVG markup

The Swift side is hand-edited every single time. From the recent OpenCode PR (#255):

```swift
case "opencode": return "OpenCode"
...
case "opencode":
    svg = SessionState.openCodeSVG(dark: isDark)
...
private static func openCodeSVG(dark: Bool) -> String {
    let bg = dark ? "#0D1117" : "#F0F6FF"
    let fg = "#3B82F6"
    return """<svg ...>..."""
}
```

This is a foot-gun: a Go-only contributor can land an adapter whose sessions are nameless and icon-less in the macOS app, with no compiler/test signal that anything is missing.

## Proposal

The adapter is the source of truth for its own display info. Co-locate display name + icon SVG with the Go package, expose it via an API endpoint, and drop the Swift switches.

### Go side

Extend `agents.Config` (or a sibling `Branding` struct) with:

```go
type Config struct {
    Name         string
    DisplayName  string   // "OpenCode"
    IconSVGLight string   // raw SVG
    IconSVGDark  string   // raw SVG (or one with CSS-driven theming)
    // ...existing fields
}
```

Each adapter's `Config()` constructor fills these alongside the fields it already populates.

### Daemon

Add `GET /api/v1/agents` returning the registered adapters with their branding:

```json
[
  {"name":"claude-code","display_name":"Claude Code","icon_svg_light":"<svg…>","icon_svg_dark":"<svg…>"},
  {"name":"opencode","display_name":"OpenCode","icon_svg_light":"<svg…>","icon_svg_dark":"<svg…>"},
  ...
]
```

### Swift side

`SessionState` fetches the registry once on startup, caches it, and looks up `name`/`icon` by adapter name. The three hardcoded switches in `SessionState.swift` go away. Unknown adapter → fall back to a generic icon + the raw adapter name (no special case for Claude Code).

## Benefits

- Adding an adapter is a one-side change (Go only).
- Icon and display name live next to the adapter code that knows the brand.
- Other future frontends (web dashboard already inlines its own SVGs in `core/cmd/irrlichd/ui/index.html` — same problem) can consume the same registry.
- New contributors can ship an adapter without needing to touch Xcode at all.

## Scope

- [ ] Define the `agents.Config` branding fields (or sibling struct)
- [ ] Migrate the existing five SVGs (claude-code, codex, pi, aider, opencode) into their Go packages
- [ ] Add `GET /api/v1/agents` endpoint
- [ ] Swift app: fetch + cache + lookup; remove `adapterName` / `adapterIcon` switches and the per-adapter SVG functions
- [ ] WebUI: same migration so it consumes the registry instead of inlining SVGs in `index.html`
- [ ] Update `AGENTS.md` / contributor docs to reflect the one-side adapter onboarding

Trigger: review of #255.

ingo-eichhorst:
## Summary
- New `agents.Config` branding fields (`DisplayName`, `IconSVGLight`, `IconSVGDark`) and a `GET /api/v1/agents` endpoint serving them, so adding a new adapter is a Go-only change. Closes #260.
- macOS app and web dashboard now look up adapter name + icon from the registry instead of hardcoded switches; the five Swift `<adapter>SVG` functions and the two switch statements in `SessionState.swift` are gone.
- `TestGate_GetAgents` wires the real production Configs into the test stack — a future adapter that forgets to fill branding fails the gate.

## Why this shape
- The registry lives in a file-scope `AgentRegistry` enum (now `@MainActor`-isolated as of `dd0e2a2`) so it stays Swift-6 strict-concurrency clean while remaining accessible from the SwiftUI views that call `adapterName`/`adapterIcon`.
- Two pre-baked SVGs per adapter (light/dark) instead of one with `currentColor` — matches the issue spec, and Swift's appearance picker already runs at render time.
- Unknown-adapter fallback in Swift is a neutral question-mark icon (gray `#9CA3AF`, deliberately distinct from the cancelled-state gray); web UI just hides the slot.
- Web dashboard renders adapter SVGs via `` so the browser image-loading sandbox blocks scripts even if the daemon binary is tampered with.

## Test plan
- [x] `go test ./core/...` — `TestGate_GetAgents` passes; all other packages pass.
- [x] `swift build --arch arm64` — clean.
- [x] `swift build --arch arm64 -Xswiftc -strict-concurrency=complete` — no new warnings involving `AgentRegistry` / `adapterName` / `adapterIcon`.
- [x] `swift test --arch arm64 --filter AgentRegistryTests` — 5/5 pass (registry-miss/hit paths for both `adapterName` and `adapterIcon`).
- [x] `tools/replay-fixtures.sh` — clean.
- [x] Manual: launched the macOS app via `/ir:test-mac`, verified all 5 adapter icons render via the registry-lookup path. See [verification comment](https://github.com/ingo-eichhorst/Irrlicht/pull/284#issuecomment-4413007272).
- [ ] Manual: open the web dashboard and confirm session rows show the adapter icon (with display-name tooltip) for each adapter.
- [ ] Manual: smoke-test the web XSS-hardening path — in DevTools, run `agentRegistry['fake'] = {name:'fake', display_name:'Fake', icon_svg_light:'<svg onload="alert(1)"></svg>'}` and inject a fake session. The alert should NOT fire (`` sandbox blocks script).


ingo-eichhorst:
## Manual verification — passed

Ran `/ir:test-mac` against `dd0e2a2`:

- **Daemon endpoint**: `GET /api/v1/agents` returns all 5 adapters with their `display_name`, `icon_svg_light`, `icon_svg_dark` populated end-to-end (Claude Code, Codex, Pi, Aider, OpenCode).
- **macOS app**: launched the dev `.app` bundle against the dev daemon, menu rendered all adapter icons through the registry-lookup path (no more Swift switches).
- **No crash on cold launch**: the defensive `NSApp?` change (latent force-unwrap fix in `adapterIcon`) verified by `AgentRegistryTests` running cleanly in headless XCTest, where `NSApp` is nil.

Ready for review/merge from my side.

Issue #503, opened 2026-05-27
opencode: concurrent same-cwd sessions never emit a working transition (headless start_session path)


ingo-eichhorst:
## Summary

When two opencode sessions run concurrently in the **same cwd** (the `multiple-sessions-same-cwd` scenario, driven via the headless `opencode run` + driver `start_session` path), the daemon correctly tracks **two distinct `ses_…` rows** keyed by SQLite `session.id` with no flap/merge — but **never emits a `working` state_transition for either session**. Every turn is observed only as `transcript_activity` followed by `state_transition → ready`.

## Evidence

Recording: `replaydata/agents/opencode/scenarios/multiple-sessions-same-cwd/events.jsonl` (committed as `known_failing`).

```
:02.363 state_transition ses_…X5 ready      (slot 1 born)
:02.754 process_exited    ses_…X5            (turn-1 'opencode run' subprocess exits)
:02.755 transcript_activity ses_…X5
:08.388 state_transition ses_…Ks ready      (slot 2 born, distinct ses_)
:08.389 transcript_activity ses_…Ks
:13.989 transcript_activity ses_…X5          (slot 1 re-driven via --session while slot 2 alive)
:14.024 state_transition ses_…X5 ready       <-- ready, NO preceding working
```

`grep -c '"new_state":"working"'` = **0**; `ready` = 4. Reproduced byte-identically across two runs.

## Contrast — single-session headless DOES emit working

`opencode/basic-turn` (also headless `opencode run`, same slow local model) emits a clean `working`→`ready` pair at turn completion:

```
:24:12.856 state_transition ses_… working
:24:12.856 state_transition ses_… ready
:24:12.895 process_exited   ses_…
```

## Root cause (hypothesis)

opencode is `ProcessOwnedStore` (SQLite). The daemon's watcher surfaces `working` at the store-flush boundary of an in-flight message. Each headless `opencode run` turn is a **separate short-lived process**; under the concurrent two-slot pattern, slot 1's first subprocess `process_exited` fires (:02.754) **before** the working-window is observed, and the later store-flush for the completed turn only yields the terminal `ready` (the row's settled state) without the `working` blip. The single-session case (`basic-turn`) catches it because the process lives across the flush.

## What the spec requires

[...]

PR #505, opened 2026-05-28
follow-up #504: opencode reconciliation + late-arriving recordings + token-quota daemon finding

ingo-eichhorst:
Follow-up to #504 (which squash-merged before this work). Starting the viewer to verify "everything is solved" surfaced several things the original PR missed; this PR closes those loops and lands one additional GREEN recording.

## Net change

- **opencode GREEN: 22 / 41** (was ~20 in #504's state at merge time).
- Completeness gate COMPLETE; catalog-drift OK; replay byte-identity green.
- **No `core/` daemon code changes** — fixtures, recipes, assessments, skill, and `tools/` only.

### One more cell recorded green
- **`turn-aborted-by-error`** ✅ — recorded via `reason=length` output-overflow at a tuned 22.5k gemma context. The prior "unobservable / zero-part" verdict was a **4096-context false negative** (same class as #503): every error at 4096 was a pre-inference prompt-overflow with no parts. A `length` abort emits a `step-start` then a terminal `step-finish reason=length`, observable end-to-end (driver, daemon, parser all agree). Assessment + rollup flipped to `yes/full/ready`.

### Re-record + #503 re-scoped
- **`multiple-sessions-same-cwd`** re-recorded clean at 32k. The prior recording was a 4096-context artifact (3 `UnknownError` turns from prompt-overflow, no real work). Clean turns prove the daemon *does* emit `working` for both concurrent same-cwd sessions — **#503's "zero working transition" premise is refuted**. The cell stays `known_failing` on a *narrower* real gap: the per-turn closing `ready` edge is deferred for concurrent short-lived `opencode run` turns. Posted a re-scope comment on #503 (not closed — left for editorial review).

### Post-record assessment reconciliation (the loop the pipeline still doesn't close)
The viewer derives matrix verdicts from per-cell `assessment.json`, but `implement` never wrote its result back. Four cells were stale:
- `subagent-orphan-cleanup`, `mid-turn-message-queued`, `long-agentic-session-stress` → `gap:` → `ready` (each had passing recording).
- `multiple-sessions-same-cwd` → `gap:start_session` → `bug/ready` (matches re-record outcome).

### False-greens caught + downgraded
- `checkpoint-rewind` and `user-esc-interrupt` rendered `yes/full/ready` from pre-record optimistic assessments despite **4 failed recordings each**. Downgraded to honest `gap:` with explicit scope notes naming the actual blockers (run_live slash→send composer race; local-model interrupt non-elicitation).

### Stale-assessment correction
- `oversized-transcript-line` → `N/A` for opencode. Scenario `requires_transport=line_based`; opencode is `structured_store` — there is no transcript line to oversize, so the cell genuinely doesn't apply. (The prior `yes/full/ready` was a phantom that rendered as recordable.)

### token-quota: two distinct daemon findings via a new mock
Built `tools/agent-onboarding/recording/mock-openai-429/main.go` (the openai-compatible analogue of `mock-anthropic-429`), with `--error-status` (default 403 non-retryable). Live probes documented **two opencode behaviours**:
1. With **HTTP 429** (real OpenAI quota status): the AI-SDK retries indefinitely (24+ requests over 180s) — opencode never writes `message.data.error`.
2. With a **non-retryable 403**: opencode persists the error terminally as `message.data.error` (`APIError isRetryable=false`) — but on a **zero-part** assistant message. `watcher.go scanParts` JOINs `part→message` (watcher.go:482-489), so a zero-part errored message can't carry `msgData` to `isErrorMessage`. That's the load-bearing blocker — the narrow zero-part edge already noted at `watcher.go:521-527`. Records when the watcher broadens to scan messages with `.data.error` and no terminal part, or when opencode emits a `step-start` before an API error.

Cell verdict: `partial/incapable/n/a` with the two findings as caveats. Driver + mock are correct; the gap is in the daemon scan path.

## Remaining opencode "open" (4 cells, each with one named external lever)

| Cell | Lever |
|---|---|
| `model-switch-midsession` | a working 2nd chat model under opencode (`functiongemma-270m` broken by a LM Studio jinja template bug; 26B/30B too slow/RAM-tight on a 32GB host; `IONOS_API_TOKEN` unset) |
| `token-quota-exhausted` | broaden `watcher.go scanParts` to handle zero-part errored messages, or wait for opencode to emit step-start before API errors |
| `checkpoint-rewind` | harden run_live slash→send composer hand-off, or a faster model |
| `user-esc-interrupt` | a model that streams promptly and honors mid-stream cancel |

## Verification
`tools/replay-fixtures.sh` ✓ · `go test ./core/cmd/replay/... -count=1` ✓ · `go test ./tools/agent-onboarding/...` ✓ · catalog-drift OK · completeness-gate `opencode` COMPLETE.


ingo-eichhorst:
Superseded by the clean cherry-pick PR — #505 was opened from the original (post-merge) branch and conflicted with main's squash of #504, so the diff was wrong. The same content is now in the new follow-up PR (see comments), with each commit cleanly applied on top of main.

Issue #597, opened 2026-06-05
GroupViewSnapshotTests fail on machines in light mode — host() doesn't pin appearance


ingo-eichhorst:
All 5 `GroupViewSnapshotTests` fail on a clean `main` checkout when the Mac runs in **light mode** ("Snapshot does not match reference").

Cause: `GroupViewSnapshotTests.host()` (platforms/macos/Tests/GroupViewSnapshotTests.swift:60) builds the `NSHostingView` without pinning `hosting.appearance`, so `Color(NSColor.windowBackgroundColor)` adapts to the current system appearance while the committed reference PNGs were recorded in dark mode.

`SessionRowSnapshotTests.host()` already solves this:

```swift
// Pin to dark aqua so snapshots don't depend on the current system
// appearance (Color(NSColor.windowBackgroundColor) adapts otherwise).
hosting.appearance = NSAppearance(named: .darkAqua)
```

Fix: apply the same one-liner to the GroupView test host. References may need re-recording once if the pinned rendering differs from the committed PNGs.

Noticed while verifying #596 — fails identically with and without that change.

PR #598, opened 2026-06-05
fix(macos): clamp ready state icon to its size × size layout box (#596)

### LINKED ISSUE #596: macOS dashboard: leading state icons not pixel-aligned, shifting all following row content
## Symptom

In the macOS dashboard session list, the left-most state icons (working dot / ready checkmark / waiting pause) are not pixel-perfectly aligned across rows. Because the row is a plain leading `HStack`, every element after the icon (agent number, badge, session title, context bar) inherits the offset, so whole rows read as slightly shifted against each other. Clearly visible when a ready row (green checkmark) sits next to working rows (purple dots).

## Suspected cause

`SessionStateIcon` sizes its three states inconsistently (`platforms/macos/Irrlicht/Views/SessionStateIcon.swift`):

- **working**: `Circle().frame(width: size, height: size)` — exactly 12×12.
- **waiting**: bar HStack wrapped in `.frame(width: size, height: size)` — exactly 12×12.
- **ready**: `Image(systemName: "checkmark.circle.fill").font(.system(size: size))` — **no fixed frame**. The SF Symbol's layout box comes from font metrics, not from `size`, so its width/height ≠ 12 and includes glyph-dependent padding. Its horizontal center also doesn't match the framed 12×12 box of the other two states.

Row usage: `SessionRowView` at `platforms/macos/Irrlicht/Views/SessionListView.swift:1143` (`HStack(spacing: 6)` with the icon first), same pattern in the child-row at :1655 and the menu-bar icon strip at :1046.

Secondary suspect: fractional point sizes in `WaitingIcon` (bar width 1.875pt, gap 2.25pt at size 12) land off the pixel grid, which can make the pause glyph look optically off-center even when its frame is correct.

## Expected

All three state glyphs occupy an identical fixed `size × size` layout box, centered, so the column after the icon starts at the same x for every row.

## Likely fix sketch

Wrap the ready branch in the same fixed frame, e.g.:

```swift
Image(systemName: "checkmark.circle.fill")
    .resizable()
    .scaledToFit()
    .frame(width: size, height: size)
    .foregroundColor(IrrColors.ready)
```

(or `.font(...)` + `.frame(width: size, height: size)` if the optical size should stay font-derived), and consider rounding the waiting-bar metrics to half-pixel boundaries.

## Environment

- Observed in the menu-bar dashboard session list (screenshot shows ready row `587-rebra…` vs working rows `592-kiro-…` / `593-st…`, with the task-progress dots row also inheriting the offset).

ingo-eichhorst:
Fixes #596.

## Problem

The leading state icons in the macOS dashboard session list weren't pixel-aligned across rows: working and waiting glyphs are framed to exactly `size × size` (12×12), but the ready branch rendered the `checkmark.circle.fill` SF Symbol via `.font(.system(size:))` with **no fixed frame** — its font-derived layout box measures **14×14** at size 12. Since the row is a leading `HStack`, every element after the icon (agent number, title, context bar) in ready rows sat 2 pt right of neighbouring working/waiting rows.

## Fix

Add an explicit `.frame(width: size, height: size)` to the ready branch. The glyph stays font-sized — the working dot's optical size is deliberately tuned to match the check's visible circle — only the layout box is clamped, so the visual appearance of the icon itself is unchanged.

## Verification

- New `SessionStateIconLayoutTests` hosts `SessionStateIcon` for **every** state and asserts `fittingSize == size × size`. Red before the fix (`ready` measured 14×14), green after.
- Full Swift suite: green, except the 5 pre-existing `GroupViewSnapshotTests` failures that reproduce identically on a clean `main` checkout in light mode — filed separately as #597 (test host doesn't pin appearance; unrelated to this change).
- `go test ./core/... -race -count=1`: green (untouched layer, ran as the full-suite gate).


ingo-eichhorst:
Review follow-up (98c69199) — hardened the coverage:

- `SessionStateIconLayoutTests` now asserts the layout box at sizes **12 and 16** (font-metric scaling isn't linear, so a second size guards ratio-dependent regressions).
- New `testStateIconAlignment_AcrossStates` snapshot stacks one row per state (working/waiting/ready) and locks visually that the leading icons — and the number/title/context-bar columns after them — start at the same x in every row. Reference recorded under the dark-aqua-pinned host, so it's immune to the #597 appearance issue.

Full Swift suite re-run: green except the 5 pre-existing #597 failures.

Issue #578, opened 2026-06-04
Flaky test: TestSessionDetector_Activity_SubagentCompletion_TransitionsChildToReady fails under full-suite load


ingo-eichhorst:
## Observed

During a full `go test ./core/... -race -count=1` run (on `worktree-554`, base a8c911d2):

```
--- FAIL: TestSessionDetector_Activity_SubagentCompletion_TransitionsChildToReady (0.08s)
    session_detector_test.go:2344: child state: got "working", want ready (parent task-notification should transition child)
FAIL
FAIL	irrlicht/core/application/services	13.183s
```

## Not reproducible in isolation

`go test ./core/application/services/ -race -count=3 -run 'TestSessionDetector_Activity_SubagentCompletion_TransitionsChildToReady'` → passes all 3 runs. An immediate full-suite re-run was also green.

Looks timing-sensitive under parallel package load — the parent task-notification's child transition appears to race the assertion.

## Context

Unrelated to the change being tested at the time (#554, which only touches `core/cmd/irrlicht-ls/` + docs).

PR #618, opened 2026-06-05
fix(detector): synchronize PID assignment with the detector loop (#606)

### LINKED ISSUE #606: Data race: PIDManager.assignPIDLocked vs SessionDetector.processActivity (plus two flaky Removed_* detector tests)
Surfaced nondeterministically under full-suite `-race` load while gating #599 (whose branch does not touch `core/application/services/` — `git diff main...HEAD --stat` over the package is empty). Passes clean on targeted reruns (`-count=2` green), so it needs scheduling pressure to reproduce.

## Race

`TestSessionDetector_CWDFallback_CleansUpOldSessionOnClear`:

```
Write  by goroutine 582:
  services.(*PIDManager).assignPIDLocked        pid_manager.go:449
  services.(*PIDManager).HandlePIDAssigned      pid_manager.go:389
  services.(*PIDManager).TryDiscoverPID         pid_manager.go:562
  services.(*SessionDetector).processActivity.gowrap1   session_detector_activity.go:352  (spawned goroutine)

Previous write by goroutine 579:
  services.(*SessionDetector).processActivity   session_detector_activity.go:494
  services.(*SessionDetector).onActivity        session_detector_activity.go:262
  services.(*SessionDetector).handleTranscriptEvent     session_detector_activity.go:25
  services.(*SessionDetector).Run               session_detector.go:391
```

The PID-discovery goroutine spawned at `session_detector_activity.go:352` writes session state (via `assignPIDLocked`, which despite the name doesn't synchronize with the detector loop's own write at `:494`) racing the main `processActivity` path.

## Flaky siblings (same family, seen in the same full-suite runs)

- `TestSessionDetector_Removed_TransitionsToReady` — `state: got "working", want ready`
- `TestSessionDetector_Removed_PrunesMetricsLedger` — `PruneEntry calls: got [], want [...]`

Both pass in isolation and at package level; they fail only under parallel full-suite load (timing-sensitive waits). Probably worth fixing alongside since the race fix may change the timing they depend on.

## Repro

`go test ./core/... -race -count=1` on a loaded machine (two hits out of ~four full-suite runs on 2026-06-05); targeted package runs stay green.

ingo-eichhorst:
## The race

`processActivity` runs on the single detector event-loop goroutine. When a session has no PID yet it spawns a background discovery goroutine (`go d.pidMgr.DiscoverPIDWithRetry` in `onNewSession`, `go d.pidMgr.TryDiscoverPID` in `processActivity`). That goroutine reaches `HandlePIDAssigned → assignPIDLocked`, which loads the session and writes `state.PID` / `state.UpdatedAt` / `state.Launcher` on the **shared `*SessionState` pointer**. Meanwhile the same `processActivity` pass writes `state.State` / `state.UpdatedAt` on that same pointer.

`assignMu` — despite the "Locked" in `assignPIDLocked` — only serialized the PID-assignment critical sections **against each other** (two concurrent discoveries) and against `claimedPIDs`. It did **not** synchronize with the detector loop's own load-modify-save. So under `-race` scheduling pressure the two writers collided:

```
Write by goroutine X:  assignPIDLocked            pid_manager.go
  ← HandlePIDAssigned ← TryDiscoverPID
  ← processActivity.gowrap1 (spawned goroutine)   session_detector_activity.go
Previous write by goroutine Y: processActivity     session_detector_activity.go
  ← onActivity ← handleTranscriptEvent ← Run
```

It reproduced only under full-suite `-race` load; targeted reruns passed.

## The fix (synchronization, no sleeps)

Extend `assignMu` to also guard the detector loop's per-event load-modify-save. `processActivity`'s post-load body moved into `processActivityLocked`, which now runs under the new `PIDManager.WithSessionStateLock` — the helper that holds `assignMu`, the same lock `assignPIDLocked` takes. The spawned discovery goroutine and the event loop are now mutually exclusive on the shared state: the goroutine simply blocks on `assignMu` until the loop releases it.

- No new mutex, no widened state — reuses `assignMu`, whose doc is updated to reflect the broadened responsibility.
- No lock inversion: none of the inline helpers `processActivityLocked` calls (`cleanupChildren`, `reevaluateParent`, `finishOrphanedChildren`, `ConsumePendingPID`, `captureLauncher`, …) take `assignMu`; the only `assignMu` consumer it reaches is the *spawned* `go TryDiscoverPID`, which runs on its own goroutine.
- Scope matches the existing lock: `assignPIDLocked` already held `assignMu` across `repo.Save` + `repo.ListAll`, so holding it across the detector's save is consistent.

## Test robustness (the flaky siblings)

The two tests named in the issue plus several same-family tests used fixed `time.Sleep` waits and, in some cases, read shared state a *detached* discovery goroutine still owned — flaking (and racing) only under parallel full-suite load. Replaced the fixed waits with poll-for-condition helpers, mirroring the existing `waitForSessionState` pattern (#578):

- `Removed_TransitionsToReady`, `Removed_PrunesMetricsLedger` (named in #606)
- `CWDFallback_DoesNotAssignDuplicatePID`, `CWDFallback_CleansUpOldSessionOnClear`, `ClearWithStaleMetadata_DeletesOldSessionImmediately`, `OrphanedSubagentsFinishWhenParentTurnDone`, `Activity_NoSubstantiveActivity_HoldsState`, `Activity_SubstantivePass_StillClassifies`

New helpers: `waitForPID`, `waitForSessionDeleted`, `waitForCondition`, locked `pidOf`/`eventCountOf` probes, and a mutex-guarded `mockMetrics` with `prunedSnapshot`. None of these change production behavior.

## Test evidence

- `go test ./core/application/services/ -race -count=10` — green (was a reliable race repro before the fix).
- Same package green under **3 concurrent test processes + 8 `yes` CPU burners** (heavy scheduling pressure); the touched tests green at `-count=40`.
- `go test ./core/... -race -count=1` — green **three times** (all 35 packages, including the daemon smoke test and e2e; no `TestGastownReplay`/#586 flake observed in these runs).
- `go vet ./core/...` clean.

Note: under deliberately absurd oversubscription (4 concurrent processes + 10 burners on 10 cores) one unrelated test, `TestSessionDetector_ParentBadgeCleared_WhenChildSweptWhileParentWaiting`, flaked once with **no data race** — a pre-existing tight-`readyTTL` (1s) vs wall-clock fragility untouched by this PR (passes targeted at `-count=30`). Out of scope here.

Fixes #606

### DIFF
--- a/core/application/services/pid_manager.go
+++ b/core/application/services/pid_manager.go
@@ -82,13 +82,16 @@ type PIDManager struct {
 	pendingMu   sync.Mutex
 	pendingPIDs map[string]int
 
-	// assignMu serializes the PID-assignment critical sections that read and
-	// write the repo's shared *SessionState pointers — HandlePIDAssigned's
-	// load-modify-save + same-PID cleanup scan, and claimedPIDs' scan. Without
-	// it, two concurrent discoveries (two same-named agents starting together)
-	// race on state.PID: one writes its session's PID while the other reads it
-	// in a ListAll scan. Never held across detector callbacks (see
-	// HandlePIDAssigned) so it can't invert with the SessionDetector lock.
+	// assignMu serializes every critical section that reads or writes the repo's
+	// shared *SessionState pointers: HandlePIDAssigned's load-modify-save +
+	// same-PID cleanup scan, claimedPIDs' scan, AND the SessionDetector event
+	// loop's own load-modify-save in processActivity (via WithSessionStateLock).
+	// The PID-discovery goroutine spawned by processActivity runs concurrently
+	// with that same processActivity call; without a shared lock, assignPIDLocked
+	// writes state.PID while processActivity writes state.State on the same
+	// pointer (issue #606). Never held across detector callbacks (Watch/delete
+	// run after assignPIDLocked releases it) so it can't invert with any
+	// SessionDetector lock.
 	assignMu sync.Mutex
 
 	// recorder captures lifecycle events for offline replay (optional).
@@ -498,6 +501,19 @@ func (pm *PIDManager) ConsumePendingPID(sessionID string) (int, bool) {
 		delete(pm.pendingPIDs, sessionID)
 	}
 	return pid, ok
+}
+
+// WithSessionStateLock runs fn while holding assignMu, the lock that also
+// guards assignPIDLocked's write to the shared *SessionState. The SessionDetector
+// event loop wraps its own load-modify-save of session state in this so the
+// PID-discovery goroutine it spawns can't write state.PID concurrently with the
+// loop writing state.State on the same pointer (issue #606). fn must not call
+// back into any PIDManager method that takes assignMu (assignPIDLocked,
+// claimedPIDs) — those run only from the spawned goroutine, never inline.
+func (pm *PIDManager) WithSessionStateLock(fn func()) {
+	pm.assignMu.Lock()
+	defer pm.assignMu.Unlock()
+	fn()
 }
 
 // claimedPIDs returns the set of PIDs already assigned to sessions other than
--- a/core/application/services/session_detector_activity.go
+++ b/core/application/services/session_detector_activity.go
@@ -157,22 +157,28 @@ func (d *SessionDetector) onNewSession(id agent.Identity, ev agent.Event) {
 		// processActivity fallback (debounce/refresh) created the session
 		// without an identity. The watcher's identity on the next real
 		// transcript event is the authoritative source.
-		changed := false
-		if existing.TranscriptPath == "" {
-			existing.TranscriptPath = ev.TranscriptPath
-			changed = true
-		}
-		if existing.Adapter == "" && id.Name != "" {
-			existing.Adapter = id.Name
-			changed = true
-		}
-		if changed {
-			existing.UpdatedAt = now
-			if err := d.repo.Save(existing); err != nil {
-				d.log.LogError("session-detector", ev.SessionID,
-					fmt.Sprintf("failed to update existing session: %v", err))
+		//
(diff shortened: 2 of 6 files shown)


ingo-eichhorst:
**Review follow-up (181cfd57):** the original fix locked `processActivity`'s load-modify-save, but two more event-loop paths write the same shared `*SessionState` pointer while a discovery goroutine can still be in flight:

- `onRemoved`'s ready transition (`state.State/UpdatedAt/Confidence/LastEvent`) — `UpdatedAt` collides directly with `assignPIDLocked`'s write; this is the exact path the issue's flaky `Removed_*` tests exercise.
- `onNewSession`'s existing-session backfill (`TranscriptPath/Adapter/UpdatedAt`) — races a `DiscoverPIDWithRetry` goroutine spawned by the earlier event for the same session.

Both now wrapped in `WithSessionStateLock`, mirroring `processActivityLocked`. The isNew branch is fine as-is (all writes happen before that session's discovery goroutine spawns). Deadlock check: `assignMu` is only taken by `assignPIDLocked`/`claimedPIDs` (spawned goroutines only) and `WithSessionStateLock` (event loop); no code path holds `d.mu`/`bgMu`/`permMu` while taking `assignMu`, so the new `assignMu → d.mu` edge (via `deleteSession → removeFromProjectSessions`) cannot form a cycle.

Verified: `go build`, `go vet`, `go test ./core/application/services/ -race -count=10`, full `go test ./core/... -race -count=1` — all green.

Residual (out of scope, pre-existing): PIDManager sweep paths (`SweepDeadPIDs`, `CleanupZombies`, `handleAlivePIDState`, `dedupeByPID`) also touch shared state without `assignMu` if they run off the event loop — worth a separate audit issue if the race detector ever points there.

PR #648, opened 2026-06-07
fix(claudecode): exclude cc-daemon infra processes (daemon run, --bg-pty-host, --bg-spare) from session matching

### LINKED ISSUE #644: claudecode: Claude Code 2.1.168 daemon infra processes (daemon run, --bg-pty-host, --bg-spare) minted as permanent proc- ghost sessions
## Symptom

After daemon restart on 2026-06-07 (v0.5.0), the dashboard shows ghost pre-sessions that never resolve:

- `proc-84783` — cwd `/Users/ingo` → actual process: `claude daemon run --origin transient --spawned-by {"label":"claude","cwd":"…/besenkammer","pid":16298}`
- `proc-84902` — cwd `…/besenkammer` → actual process: `…/ClaudeCode.app/Contents/MacOS/claude --bg-pty-host /tmp/cc-daemon-501/…/pty/ab5f01f8.sock 86 34 -- …/versions/2.1.168 --resume bb9f6ebf-… --allowed-tools mcp__computer-use__… --permission-mode auto`

Also running but invisible (binary `…/versions/2.1.168`, name doesn't match): `--bg-spare` claim/pty hosts (84899/84913) and the actual resumed session process (84992).

## Root cause

Claude Code 2.1.168 introduced a background daemon architecture (`cc-daemon`): a long-lived `claude daemon run`, PTY-host wrappers (`--bg-pty-host`), and pre-warmed spares (`--bg-spare`). The wrapper/daemon processes run the `claude` binary, so the claudecode process matcher accepts them and `processlifecycle/scanner.go` mints `proc-<pid>` pre-sessions (`scanner.go:282`). These processes never own a transcript, so the pre-sessions can never be superseded — they're permanent ghosts until the process exits.

## Suggested fix

Exclude daemon-infra argv shapes in the claudecode matcher: argv containing `daemon run`, `--bg-pty-host`, or `--bg-spare` is infrastructure, not a session. (The TUI/session processes either run plain `claude …` or are children spawned *by* these hosts; the hosted session is observable via its transcript as usual.)

Separately worth a look: the *hosted* resumed process (84992, binary `versions/2.1.168`) did get PID-bound to its session `bb9f6ebf` via PID discovery, so monitoring of the real session works — the bug is purely matcher noise from the wrappers.

This is exactly the upstream-release-impact category `/ir:agent-releases` tracks — 2.1.168 changed the process landscape under us.

ingo-eichhorst:
## Root cause

Claude Code 2.1.168 introduced a background-daemon architecture (**cc-daemon**): a long-lived `claude daemon run`, PTY-host wrappers (`--bg-pty-host`), and pre-warmed spares (`--bg-spare`). All of these run the `claude` binary, so the claudecode `ExactName{"claude"}` process matcher accepts them and `processlifecycle/scanner.go` mints `proc-<pid>` pre-sessions for them. These infrastructure processes never own a transcript, so the pre-sessions can never be superseded — they're **permanent ghosts** in the dashboard until the process exits.

Observed ghost shapes:
- `claude daemon run --origin transient --spawned-by {...}` (the cc-daemon, cwd `$HOME`)
- `.../ClaudeCode.app/Contents/MacOS/claude --bg-pty-host /tmp/cc-daemon-501/<id>/pty/<x>.sock 86 34 -- .../versions/2.1.168 --resume <sid> ...` (PTY-host wrapper)
- `.../versions/<ver> --bg-pty-host ...` / `--bg-spare ...` (pre-warmed spares)

## Fix

Add a per-adapter **argv exclusion seam**, keeping format-specific logic in the adapter and the scanner generic:

- `ports.ProcessObserver` gains `ArgvOf(pid) ([]string, error)` — implemented per platform (darwin: `KERN_PROCARGS2` argv parse; linux: `/proc/<pid>/cmdline`; other: stub). A nil argv (hardened-runtime processes strip it) is not an error.
- `agent.Process` gains an optional `ExcludeArgv func(argv []string) bool` predicate.
- The scanner consults it (`WithArgvFilter`) at the top of its per-PID loop and skips matched-but-excluded PIDs entirely — no pre-session minted, never tracked (so it can't later be reported as "exited").
- The claudecode package declares the predicate `IsInfraArgv` and wires `ExcludeArgv: IsInfraArgv` in `Agent()`.

PID-discovery for the *real* hosted session (resumed under `versions/<ver>`) is unaffected — this PR only removes the wrapper/daemon matcher noise.

## Exclusion rule

Positional / flag-position based, **never a blanket substring scan**, so a user prompt that merely mentions these tokens is not excluded:

- `argv[1] == "daemon" && argv[2] == "run"` (the cc-daemon itself), **OR**
- any element of `argv[1:]` exactly equal to `--bg-pty-host` or `--bg-spare` (a flag occupies its own argv slot; a prompt string is a single slot that won't equal the flag exactly).

A nil/empty/unreadable argv is treated as a **session** — we never exclude on absence of evidence. So `claude -p "explain --bg-pty-host"`, plain `claude`, `claude --resume <sid> ...`, and `claude --model ... --dangerously-skip-permissions` all stay matched.

## Tests

- `claudecode/process_test.go` — table-driven `IsInfraArgv`: the 3 infra shapes (excluded) + plain/resume/model-flag sessions, prompt-string false-positive guards, and degenerate argv (matched).
- `processlifecycle/scanner_argvfilter_test.go` — fake `ProcessObserver`; verifies exactly one `new_session` for the real PID and that the 3 infra PIDs are neither emitted nor tracked.
- `processlifecycle/osutil_test.go` — `TestParseProcargs2Argv` for the darwin-shape argv parser (cross-platform).

Suite results (all green):
- `go test ./core/... -race -count=1`
- `tools/replay-fixtures.sh`
- `go run ./tools/onboarding-factory/cmd/of validate`
- `GOOS=linux` and `GOOS=windows` cross-compiles (build-tag coverage)

Fixes #644

### DIFF
--- a/core/adapters/inbound/agents/claudecode/agent.go
+++ b/core/adapters/inbound/agents/claudecode/agent.go
@@ -56,11 +56,15 @@ func Agent() agent.Agent {
 				Kind:            permission.KindObserve,
 				Title:           "Read session transcripts",
 				FeatureUnlocked: "Session list, timeline, cost & token metrics",
-				Touches:         "Reads session transcripts under ~/.claude/projects/",
+				Touches: "Reads session transcripts under ~/.claude/projects/ and " +
+					"basic info (working directory, command line) of running claude processes",
 				Detail: "Tails *.jsonl transcript files under ~/.claude/projects/ " +
-					"to derive session state, cost, and token metrics. Read-only — " +
-					"no transcript file is ever modified. Toggling off stops all " +
-					"reading immediately.",
+					"to derive session state, cost, and token metrics. Also scans " +
+					"for running claude processes and reads their working directory " +
+					"and command line — to show a session before its first message " +
+					"and to skip Claude Code's background daemon processes. " +
+					"Read-only — no transcript file is ever modified. Toggling off " +
+					"stops all reading immediately.",
 			},
 			{
 				Key:             PermissionKeyHooks,
--- a/core/adapters/inbound/agents/claudecode/agent.go
+++ b/core/adapters/inbound/agents/claudecode/agent.go
@@ -42,6 +42,7 @@ func Agent() agent.Agent {
 		Process: agent.Process{
 			Match:         agent.ExactName{Name: ProcessName},
 			PIDForSession: DiscoverPID,
+			ExcludeArgv:   IsInfraArgv,
 		},
 		Source: agent.FilesUnderRoot{
 			Dir: transcriptsDir(),
--- a/core/adapters/inbound/agents/claudecode/process.go
+++ b/core/adapters/inbound/agents/claudecode/process.go
@@ -1,0 +1,37 @@
+package claudecode
+
+// IsInfraArgv reports whether a `claude`-binary process is Claude Code's
+// background infrastructure rather than an interactive session. Claude Code
+// 2.1.168 introduced a background-daemon architecture ("cc-daemon"): a
+// long-lived `claude daemon run`, PTY-host wrappers (`--bg-pty-host`), and
+// pre-warmed spares (`--bg-spare`). All run the `claude` binary, so the
+// ExactName process matcher accepts them and the lifecycle scanner would mint
+// permanent `proc-<pid>` ghost pre-sessions for them — they never own a
+// transcript, so nothing ever supersedes them (issue #644).
+//
+// The match is intentionally positional / flag-position based, never a blanket
+// substring scan, so a user prompt that merely mentions these tokens — e.g.
+// `claude -p "explain --bg-pty-host"` — is not excluded:
+//
+//   - argv[1]=="daemon" && argv[2]=="run"  (the cc-daemon itself)
+//   - any argv element (after argv[0]) equal to "--bg-pty-host" or "--bg-spare"
+//     (the PTY-host wrappers and pre-warmed spares; a flag occupies its own
+//     argv slot, whereas a prompt string is a single slot that won't equal the
+//     flag exactly)
+//
+// A nil/empty argv (unreadable, e.g. a hardened-runtime process) is treated as
+// a session — we never exclude on the absence of evidence.
+func IsInfraArgv(argv []string) bool {
+	if len(argv) < 2 {
+		return false
+	}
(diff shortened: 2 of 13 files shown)


ingo-eichhorst:
## Code review (xhigh) → fixes pushed in c7103488

9 findings (4 confirmed / 5 plausible); 6 addressed in the follow-up commit:

| # | Finding | Status |
|---|---|---|
| 1 | **Pre-existing persisted ghosts survive the upgrade** — nothing retired an alive-PID `proc-` pre-session that never gets a superseding session | ✅ fixed: first excluded verdict emits one retirement `EventRemoved` for `proc-<pid>` (detector deletes pre-sessions, no-ops if absent) |
| 2 | `ArgvOf` sysctl per matched PID per poll forever; nil-argv could mint a 1-poll ghost flap | ✅ fixed: per-PID verdict cache (argv is immutable), nil never cached (retried), entries pruned on PID exit |
| 3 | `parseProcargs2Argv` duplicated `parseProcargs2`'s preamble | ✅ fixed: shared `procargs2ArgvOffset` |
| 4 | Seam test carried a drifting inline copy of `IsInfraArgv` (direct import would cycle) | ✅ fixed: stub predicate + rationale comment |
| 5 | Transcripts permission text didn't disclose process scanning / argv (command lines can carry secrets) | ✅ fixed: `Touches`/`Detail` now disclose cwd + command-line reads |
| 6 | Denylist rots when upstream ships the next `--bg-*` flag | ✅ documented: rot note on `IsInfraArgv` (allowlist verified infeasible — `-p` prompts are unbounded) |
| 7 | Truncated `KERN_PROCARGS2` buffer (> ARG_MAX env) → partial argv slips the filter | 📝 documented as fail-open in the parser doc; trigger not realistically reachable |
| 8 | `Match` + `ExcludeArgv` are two separately-threaded policy channels | ⏭️ not changed — single production wiring site + seam test mitigate; unifying would churn the matcher API |

New regression tests: verdict cached across polls (1 `ArgvOf` per PID), nil-argv mint→retire without flapping, one-shot retirement for persisted pre-filter ghosts. Full suite green: `go test ./core/... -race`, factory, replay fixtures, `of validate`.

Issue #641, opened 2026-06-07
claudecode: session stuck in 'working' after /compact — compact summary user event parsed as a real user turn


ingo-eichhorst:
## Symptom

Session `bb9f6ebf-f1b0-46d7-b955-59b10cfdd525` (project besenkammer) ran a manual `/compact`. After compaction finished, Claude Code sat idle at the prompt — but irrlicht shows the session as **working** indefinitely. It should be **ready**.

Live daemon state (v0.4.8+a8d1a35.dirty, ~6 min after compaction):

```json
{"session_id":"bb9f6ebf","state":"working","last_event":"transcript_activity","last_event_type":"user","has_open_tool":false}
```

## Transcript shape (manual /compact completion)

`~/.claude/projects/-Users-ingo-projects-besenkammer/bb9f6ebf-…525.jsonl` lines 1137–1141:

| line | type | flags | content |
|---|---|---|---|
| 1137 | `system` / `compact_boundary` | | "Conversation compacted" |
| 1138 | `user` | `isCompactSummary: true`, `isVisibleInTranscriptOnly: true` | "This session is being continued from a previous conversation…" |
| 1139 | `user` | `isMeta: true` | `<local-command-caveat>` |
| 1140 | `user` | | `<command-name>/compact</command-name>…` |
| 1141 | `user` | | `<local-command-stdout>Compacted…` |

Then attachments / file-history-snapshot / mode entries. **No assistant event and no `turn_done` ever follows** — a manual compaction doesn't start a turn.

## Root cause

`core/adapters/inbound/agents/claudecode/parser.go` `handleUserEvent` skips `isMeta`, task-notifications, and `<command-name>`/`<local-command-*>` wrappers — so lines 1139–1141 are correctly dropped. But the **compact summary user event (line 1138) is not skipped**: it has no `isMeta`, plain string content with no command prefix, so it falls through as a real `user` turn event.

`handleSystemEvent` skips `compact_boundary` (only `turn_duration`/`stop_hook_summary` promote to `turn_done`), so nothing signals end-of-activity.

The classifier (`core/application/services/state_classifier.go`) then hits rule 4 (default: transcript activity → working): `ready → working`. Since no `turn_done`/assistant event ever arrives, the session is stuck in `working` until the user actually prompts again.

## Suggested fix

[...]

PR #658, opened 2026-06-12
fix(claudecode): manual /compact working-during, ready-after lifecycle (#656, #657)

### LINKED ISSUE #656: claudecode: session stuck in 'working' after manual /compact when prior turn was mid tool-use (#641 follow-up)
## Summary

After a **manual `/compact`**, a session whose last pre-compact turn was *mid tool-use* (assistant `tool_use` → `tool_result`, no `turn_done`) is stranded in **`working`** forever. It never transitions to `ready`, even after the session goes idle.

This is the mirror image of #641: that fix stopped the compact summary from stranding a *ready* session in *working*; this is a *working* session that never gets *released* to ready.

## Repro / evidence

Observed live session: `fa5751bf-5a78-4235-a63c-993afdc85d96` (Claude Code 2.1.168, manual `/compact`, `durationMs` 161046, `preTokens` 655773).

Transcript tail (file order; note timestamps are **non-monotonic** — Claude Code writes the whole compaction block in a burst when compaction *finishes* and back-dates the `/compact` invocation):

```
1314: assistant  stop=tool_use      ts=19:57:42   blocks:tool_use      <- last real assistant
1315: user        (tool_result)      ts=19:57:48   blocks:tool_result   <- last NON-skipped event, NO turn_done
1316: system      compact_boundary   ts=17:21:23   (trigger:"manual")   <- skipped
1317: user        isCompactSummary   ts=17:21:23                        <- skipped (#641)
1318: user        isMeta caveat      ts=17:18:42                        <- skipped
1319: user        <command-name>/compact</command-name>                <- skipped
1320: user        local-command-stdout "Compacted…"                    <- skipped
1321-1327: attachment                                                   <- skipped
1328: file-history-snapshot                                            <- skipped
1333: system      away_summary       ts=17:39:44                        <- skipped (proves session went idle)
```

## Root cause

Every entry after the last real turn is skip-able:
- `handleSystemEvent` (`core/adapters/inbound/agents/claudecode/parser.go:191`) drops `compact_boundary` into its default `ev.Skip = true` branch.
- `isCompactSummary`, the `isMeta` caveat, and the `<command-name>` / `local-command-stdout` wrappers are all skipped (`parser.go:204-227`, the #641 fix).

So `LastEventType` stays pinned at the mid-turn `tool_result`, `IsAgentDone()` returns false (no `turn_done`), and the per-pass `NoSubstantiveActivity=true` makes the detector ignore the compaction entirely. Result: stuck in `working`.

## Proposed fix

Treat `compact_boundary` with `compactMetadata.trigger == "manual"` as a **substantive `turn_done`** in the parser. A manual `/compact` definitively ends the prior turn (the context was replaced), so:
- working → ready (fixes this bug)
- ready → ready (no-op; preserves #641's observable outcome)

Gate on `manual` so that *auto*-compaction (which fires mid-turn and continues) doesn't get a spurious ready-blip + false "ready" notification.

`turn_done` also sweeps any lingering open tool call (`tailer.go:980`), so `HasOpenToolCall` clears correctly.

### Test impact
`TestTailer_ManualCompactTail_NoSubstantiveActivity` (`compaction_test.go`) and the paired `session_detector_no_substantive_test.go` assert `NoSubstantiveActivity=true` for the compaction tail. With this change the boundary becomes substantive (`NoSubstantiveActivity=false`) while the resulting state is unchanged for the ready→ready case — update those assertions to reflect the new semantics and add a working→ready case.

Related: #641 (the ready→working half).

### LINKED ISSUE #657: claudecode: manual /compact window never shows as 'working' (compaction invisible to tailer)
## Summary

Running a **manual `/compact`** never shows the session as **`working`**. The session sits in its pre-compact state (typically `ready`) for the entire compaction window — even though the agent is busy for tens of seconds to minutes.

## Repro / evidence

Observed live session: `7e2775d8-cc3f-4507-bf48-d49ea244e705` (Claude Code 2.1.168, manual `/compact`, `durationMs` 61426).

The session's last real turn was a `turn_duration` (= `turn_done`, state `ready`). The user then ran `/compact`, which took **61s**. During those 61s, **nothing is appended to the transcript** — Claude Code flushes the entire compaction block at the *end* and back-dates the `/compact` invocation:

```
60: assistant  stop=end_turn      ts=10:17:37          <- last real assistant
61: system     turn_duration      ts=10:17:37          <- turn_done → ready
... (bash-escape wrappers, all skipped) ...
81: system     compact_boundary   ts=18:16:54.734      <- written at END of compaction
82: user       isCompactSummary   ts=18:16:54.732
83: user       isMeta caveat      ts=18:15:53.308      <- /compact invoked here (back-dated)
84: user       <command-name>/compact</command-name>  ts=18:15:53.308
85: user       local-command-stdout "Compacted…"      ts=18:16:55.684
```

The `<command-name>/compact</command-name>` line is timestamped 18:15:53 (invocation) but physically written *after* the 18:16:54 boundary. So between 18:15:53 and 18:16:54 the transcript file does not change, and the tailer has no signal to flip to `working`.

## Root cause

Transcript-only detection **cannot** cover this: a manual `/compact` writes nothing to the JSONL during the compaction window. The state stays frozen at the pre-compact value (`ready`), so the user never sees `working`.

## Proposed fix

Add a **`PreCompact` hook** (Claude Code fires it before both manual and auto compaction, with a `trigger` field). The hook POSTs to the daemon, which forces the session to `working` for the compaction window. The `compact_boundary` → `turn_done` change (see sibling issue) then returns it to `ready` when compaction finishes — giving the full lifecycle: **working during, ready after**.

### Effort notes
- No "force working from a hook" path exists today — the hook receiver currently only routes to `HandlePermissionHook` (waiting set/clear). A new handler is needed (`core/adapters/inbound/agents/claudecode/hooks.go`).
- `PreCompact` must be registered in the hook installer (`hookinstaller.go`) and written into `~/.claude/settings.json` — **existing installs need re-onboarding** to pick up the new hook.
- Add hook-receiver + installer tests.

Depends on / pairs with the `compact_boundary` → `turn_done` fix (sibling issue) for the working→ready return. Related: #641.

ingo-eichhorst:
## Summary

A manual `/compact` was invisible to irrlicht's state machine in two complementary ways. This PR restores the full lifecycle you'd expect — **working for the whole compaction window, then ready once it finishes** — by fixing both halves together (the issues cross-reference each other).

| Phase | Transcript | State | Delivered by |
|-------|-----------|-------|--------------|
| `/compact` invoked | (nothing yet) | **working** | #657 PreCompact hook |
| compaction running (~60–160s) | (nothing written) | **working** (holds) | #657 overlay |
| compaction finishes | burst-write `compact_boundary{manual}` | **working → ready** | #656 parser → `turn_done` |

### #656 — release `working → ready` after compaction
When the last pre-compact turn was stranded mid tool-use (`tool_use → tool_result`, no `turn_done`), the burst-written compaction tail was entirely skip-able, so the parser never emitted `turn_done`; the lingering open tool call was never swept and the session stuck in **working** forever.

`handleSystemEvent` now promotes a `compact_boundary` whose `compactMetadata.trigger == "manual"` to `turn_done` (auto stays skipped), reusing the existing open-tool-call sweep so `IsAgentDone()` reports done.

### #657 — show `working` during compaction
The compaction window writes nothing to the transcript, so transcript-only detection fundamentally can't surface it. A new **`PreCompact` hook** (installed with matcher `"manual"`) forces the session to working via `HandleCompactHook`, which sets a `compactPending` hold; `processActivity` overlays `CompactInProgress` and `ClassifyState` rule 0b holds working until the manual `compact_boundary` lands (surfaced as `SawManualCompactBoundary`), which clears the hold and routes to ready.

Both halves gate **strictly on manual** compaction — auto-compaction fires mid-turn while the session is already working, so it must not blip.

> ⚠️ **Existing installs** pick up the "working during compaction" half automatically on the next daemon restart — `EnsureHooksInstalled` is idempotent and adds the missing `PreCompact` entry to `~/.claude/settings.json`. No manual re-onboarding needed.

This is a follow-up to #641 (the ready→working half), whose observable outcome is preserved.

## Test plan
- `go test ./core/... -race -count=1` — full suite green.
- `tools/replay-fixtures.sh` — passes, no golden drift (incl. `2-12_context-compaction`).
- New coverage:
  - Parser/tailer: working→ready, ready→ready (#641 preserved), auto-skipped — plus a faithful `fa5751bf`-shaped regression (non-monotonic back-dated burst + trailing `away_summary`) **proven to fail without the parser fix**.
  - Hook dispatch: manual routes to `HandleCompactHook`, auto ignored.
  - Installer: `PreCompact` matcher `"manual"` + idempotent add to an existing four-event install.
  - Classifier rule 0b; end-to-end detector hold→release.

Closes #656
Closes #657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ beyond), see the [Roadmap](https://irrlicht.io/docs/roadmap.html).
   - **Existing installs** pick up the "working during compaction" half on the
     next daemon restart — `EnsureHooksInstalled` adds the new `PreCompact` entry
     to `~/.claude/settings.json` automatically.
+  - An interrupted compaction that never finishes is bounded by a timeout, so a
+    cancelled `/compact` can't strand the session in "working" either.
 
 ## [0.5.1] — 2026-06-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ beyond), see the [Roadmap](https://irrlicht.io/docs/roadmap.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- A manual `/compact` now shows the full lifecycle: the session goes **working**
+  for the whole compaction window and returns to **ready** when it finishes.
+  Previously the compaction window (which writes nothing to the transcript)
+  stayed frozen at the pre-compact state, and a `/compact` invoked mid tool-use
+  was stranded in "working" forever. The fix adds a Claude Code `PreCompact`
+  hook (forces working during) and treats the manual `compact_boundary` as a
+  turn boundary (releases to ready after). (#656, #657; follow-up to #641)
+  - **Existing installs** pick up the "working during compaction" half on the
+    next daemon restart — `EnsureHooksInstalled` adds the new `PreCompact` entry
+    to `~/.claude/settings.json` automatically.
+
 ## [0.5.1] — 2026-06-07
 
 ### Ghost sessions and stuck states fixed: /compact strandings, Claude Code 2.1.168's background daemon, and restart amnesia.
--- a/core/adapters/inbound/agents/claudecode/compaction_test.go
+++ b/core/adapters/inbound/agents/claudecode/compaction_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"irrlicht/core/application/replayengine"
 	"irrlicht/core/pkg/tailer"
 )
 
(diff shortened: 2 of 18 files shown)


ingo-eichhorst:
Addressed the review's primary finding (the working-hold's only clear paths were the boundary + process exit, so an **interrupted/failed `/compact` with no boundary could re-strand the session in `working` forever** — a narrower version of #656).

**`aff0d430`** bounds the hold with a timeout:
- `compactPending` now stores the hook-fire time (`map[string]int64`).
- The overlay is extracted into `applyCompactHold` (mirroring `markStalledEditTool`'s injected-`now` shape) and clears the hold on the first of:
  - the manual `compact_boundary` landing — normal release → ready (#656), or
  - `compactHoldTimeout` (**5 min**, well beyond the ~161s worst-case real compaction from the #656 evidence) elapsing — an orphaned hold is dropped and the session re-classifies normally on the next `refreshStaleSessions` tick instead of sticking.
- New white-box `TestApplyCompactHold`: fresh→working, boundary→release, **timeout→drop**, plus no-pending / nil-metrics no-ops.

Full `./core/... -race` green.

The two minor notes (daemon-restart-mid-compaction reverting to persisted state; `compactMetadata.trigger` being an external-contract assumption) are left as documented/accepted — happy to revisit if you'd prefer.

PR #671, opened 2026-06-13
fix(geminicli): DiscoverPID honors claimed-aware disambiguator (#664)

### LINKED ISSUE #664: gemini-cli: foreground subagent transcript under chats/<parentId>/ is unwatched → child unlinked
**Found during gemini-cli onboarding** (scenario 4.1 multiple-sessions-same-cwd). Recorded as a `known_failing` fixture: `replaydata/agents/gemini-cli/scenarios/4-1_multiple-sessions-same-cwd`.

**Symptom:** two gemini sessions launched in the same cwd both bind to a single (lowest) PID → cross-contaminated PID-to-session binding (the second session's turns don't produce a working transition on its own row).

**Root cause:** `geminicli.DiscoverPID` is cwd-only and hardcodes `lowestPID`, deliberately ignoring the PID-manager's claimed-aware disambiguator (`pid_manager.go:584`), so both same-cwd launchers collapse to one binding.

**Expected:** each same-cwd session binds its own launcher PID.
**Fix direction:** have `DiscoverPID` use the claimed-aware disambiguator instead of `lowestPID`.

ingo-eichhorst:
Fixes #664.

> **Issue title is shifted vs. its body** — this PR addresses the body: `DiscoverPID` collapsing same-cwd sessions onto the lowest PID (fixture `4-1_multiple-sessions-same-cwd`).

## Root cause
`geminicli.DiscoverPID(cwd, transcriptPath, disambiguate)` already received a claimed-aware `disambiguate` callback (from `pid_manager.go`) but **ignored it**, hardcoding `lowestPID`. So two sessions launched in the same cwd both bound to the single lowest PID — the second session's turns never produced a `working` transition on its own row.

## Fix
`DiscoverPID` now forwards the caller's `disambiguate` to `DiscoverPIDByCWDAndCmdLine`, falling back to `lowestPID` only when `disambiguate` is nil (behaviour-preserving for nil callers). A `discoverByCWDAndCmdLine` package-var seam (mirrors claudecode's `discoverByCWD`) lets the test stub the OS call.

## Verification
- `TestDiscoverPID_HonorsPassedDisambiguator` (verified red on pre-fix: `got pid=100, want 200`) + `TestDiscoverPID_NilDisambiguatorFallsBackToLowest`; `go test ./core/... -race` green.
- Replay can't exercise live PID binding, so this is **unit-test-proven** (the `4-1` replay golden is unchanged, as expected).

## Note
`isHeapBumpWorker`'s doc comment in `agent.go` still says DiscoverPID binds "the lowest-PID ancestor" — now true only on the nil path. Left untouched to stay surgical; trivial follow-up doc tweak if desired.


ingo-eichhorst:

PR #672, opened 2026-06-13
fix(geminicli): parent-link nested chats/<parentId>/ subagent (#663)

### LINKED ISSUE #663: gemini-cli: write_todos tool-call emits no TaskDelta → Tasks stays empty
**Found during gemini-cli onboarding** (scenario 3.4 subagent-orphan-cleanup). Recorded as a `known_failing` fixture: `replaydata/agents/gemini-cli/scenarios/3-4_subagent-orphan-cleanup`.

**Symptom:** a foreground subagent's session never surfaces / is never parent-linked; on parent death it can't be cleaned up as a child.

**Root cause:** gemini writes the child transcript to a **nested** `chats/<parentId>/session-<childId>.jsonl`. The `FilesUnderRoot` Source watches `chats/*.jsonl` at the top level only, and no `deriveParentSession` rule (helpers.go) reads gemini's nested layout, so the child is neither observed nor linked.

**Expected:** the child session is observed and `parent_linked` to its parent.
**Fix direction:** watch the nested `chats/<parentId>/` dir + add a gemini parent-link rule in `deriveParentSession`.

ingo-eichhorst:
Fixes #663.

> **Issue title is shifted vs. its body** — this PR addresses the body: the nested `chats/<parentId>/` subagent transcript being unlinked (fixture `3-4_subagent-orphan-cleanup`).

## Root cause
gemini writes a foreground subagent's transcript to a **nested** `chats/<parentId>/session-<childId>.jsonl`. No `deriveParentSession` rule read that layout, so the child was never `parent_linked` and couldn't be cleaned up as a child on parent death.

## Fix
Added `deriveGeminiParentSessionID` and wired it into `deriveParentSession` (helpers.go): when a transcript path's grandparent dir is named `chats`, the parent session id is the dir under `chats/`. Slotted between the Claude-Code path rule and the Pi-header rule.

## Scope correction on the issue's fix-direction
Part (a) — "watch the nested `chats/<parentId>/` dir" — was **already satisfied**: `FilesUnderRoot` carries no top-level-only glob; the `fswatcher` recursively walks the tree and dynamically watches subdirs created at runtime. So the nested child file is already observed; only the parent-link rule (part b) was missing. No watcher change needed.

## Verification
- `TestDeriveParentSessionGeminiNested` (3 cases: nested→parent, top-level→"", non-gemini lookalike unaffected); red before, green after; `go test ./core/... -race` green.
- Cross-session parent-linking isn't exercised by single-transcript replay, so this is **unit-test-proven** (`3-4` replay golden unchanged, as expected).

## Adjacent (untouched)
`extractSessionID` strips only `.jsonl`, so a child surfaces as `session-<childId>` (prefix retained) vs. the bare UUID in the header — orthogonal to parent-linking; flagged, not fixed.


ingo-eichhorst:

PR #673, opened 2026-06-13
fix(geminicli): map write_todos snapshots to TaskDelta (#662)

### LINKED ISSUE #662: gemini-cli: no background-process tracking → session settles ready while a detached process is alive
**Found during gemini-cli onboarding** (scenario 2.3 task-list). Recorded fixture (state arc passes 5/5; carries a `known_failing` TaskDelta gap): `replaydata/agents/gemini-cli/scenarios/2-3_task-list`.

**Symptom:** gemini's `write_todos` tool-call produces a transcript trace but the API `tasks` field stays empty (no `TaskDelta` emitted).

**Root cause:** the geminicli parser routes `write_todos` through the generic `function_call` path and never emits a `TaskDelta`. `write_todos` is full-list-replace (like codex `update_plan` / opencode `todowrite`), so each snapshot must be reconciled into deltas.

**Note:** `write_todos` is registered only on gemini-2 models (`isGemini2Model`); recorded via `gemini -m gemini-2.5-pro`.
**Fix direction:** map `write_todos` snapshots to `TaskDelta` (mirror the codex `update_plan` handling).

ingo-eichhorst:
Fixes #662.

> **Issue title is shifted vs. its body** — this PR addresses the body: `write_todos` emitting no `TaskDelta` (fixture `2-3_task-list`).

## Root cause
The geminicli parser routed `write_todos` (a gemini-2 full-list-replace todo snapshot) through the generic `function_call` path and never emitted a `TaskDelta`, so the API `tasks` field stayed empty.

## Fix
`write_todos` snapshots (`toolCalls[].args.todos`) now reconcile into `TaskDelta`s — mirroring opencode's `todowrite` handling: Create on first sight, Update on non-pending status, plus a `TaskSnapshot` for reconcile/pruning. Parser state (`todoIDByDesc`, `nextTaskID`) keys todos by their `description` (gemini carries no stable id). Reacts to the tool **name** in the transcript (not gated on `isGemini2Model`, which only governs registration).

## Verification
- `TestParseLine_WriteTodosEmitsTaskDeltas` — three sequential snapshots: 3 creates → 1 in_progress → 1 completed; red before (`got 0`), green after; `go test ./core/... -race` green.
- **Replay-proven**: refreshed `2-3` replay golden now shows a populated `"tasks": [...]` with `status:"completed"`. `of validate` + onboarding-factory replay test green.

## Follow-up
`known_failing` left set (tied to the frozen `events.jsonl`); clearing requires a re-record with the fixed daemon — deferred per the agreed verification approach.


ingo-eichhorst:

Issue #268, opened 2026-05-01
Automated agent/adapter onboarding pipeline (extends ir:onboard-agent)


ingo-eichhorst:
## Problem

`ir:onboard-agent` drives an agent CLI through scenarios and records fixtures, but stops short of doing the real work: a human still has to read the recordings, sniff the transcript format, hand-write a Go parser in `core/adapters/inbound/agents/<name>/`, **and** hand-write an interactive driver in `.claude/skills/ir:onboard-agent/scripts/drive-<agent>-interactive.sh`, **and** hand-author state-classifier rules. That hand-authoring is where adapters break.

The recent **aider regression** ("trailing-`?` contract", `> Tokens:` heuristics, missing `tool_result` events) is the symptom, not the disease. We have no ground-truth labels and no automated rule derivation, so a wrong classifier passes the diff as long as it's wrong consistently.

PR #269 (just merged) makes this clearer: getting claudecode + codex multi-turn / interrupted-turn fixtures recorded required hand-discovering and encoding non-obvious per-agent quirks — trust dialogs, readiness signals (sometimes only visible on the *live* pane, not the pipe-pane log), turn-done event-field shapes, interrupt keys, and even **typing rhythm** (codex needs 0.3s between text and Enter because Ink drops Enter during MCP boot). All of that is agent knowledge the pipeline should either auto-discover or accept declaratively, never re-rediscover by hand.

## Goal

A **fully automated agent-onboarding pipeline** that, given a new agent CLI, independently:

1. Drives it through scenarios with rich multi-sensor recording
2. Synthesizes both the **interactive driver protocol** and the **state classifier rules** from ground-truth labels
3. Emits the committed fixture set, the generated Go adapter package, **and** the generated `drive-<agent>-interactive.sh`
4. Ships a **human-friendly replay viewer** so a maintainer can watch any recording at adjustable speed, scrub backward and forward, and see *why* every state change happened

The aider class of bug should be impossible by construction: a wrong rule fails synthesis instead of slipping past a diff, and when it does fail, the viewer makes it obvious *what signal at what moment* the synthesizer disagreed with.

## Architecture

[...]

PR #675, opened 2026-06-13
fix(geminicli): skip shell-escape preamble so session settles (#660)

### LINKED ISSUE #660: gemini-cli: parser drops type:"info" cancel notice → session sticks in working after ESC interrupt
**Found during gemini-cli onboarding** (scenario 2.15 shell-escape-command). Recorded as a `known_failing` fixture: `replaydata/agents/gemini-cli/scenarios/2-15_shell-escape-command`. Confirmed by a live tmux recording (gemini 0.46.0).

**Symptom:** running a `!<cmd>` shell-escape drives the session ready→working and it never settles — sticks in `working`.

**Root cause:** the shell-escape is persisted as a plain `type:"user"` text message ("I ran the following shell command: …"). `parseUser` classifies it as a normal prompt → opens a turn with no settle. (Contrast claudecode, which wraps it in `<bash-input>/<bash-stdout>` and skips it.)

**Expected:** the shell-escape does not open an unsettled turn.
**Fix direction:** a `parseUser` preamble skip mirroring claudecode's `isLocalCommandContent`.

ingo-eichhorst:
Fixes #660.

## Bug
A `!cmd` shell-escape in gemini-cli runs in a local shell with no LLM round-trip, but Gemini still persists it as an ordinary `type:"user"` message opening with `"I ran the following shell command: …"` — with **no following `gemini` turn**. `parseUser` classified it as a real prompt, opening a turn that never settled, so the session **stuck in `working`**.

## Fix
`parseUser` now skips that preamble (one `HasPrefix` guard, mirroring claudecode's `isLocalCommandContent` skip of `<bash-input>/<bash-stdout>`), plus a unit test. The exact preamble string was verified against gemini-cli's source (`useExecutionLifecycle.ts`) and a live capture.

## Re-recorded the 2-15 fixture
The merged-in fixture couldn't validate the fix — three problems, all fixed:

- **Recipe** drove `! e c h o … Enter` as one tmux burst, so Gemini's `!` shell-mode toggle never engaged and the escape was never persisted (it captured a plain prompt, not the bug). New recipe runs a **warmup turn** to establish the session first — Gemini only persists a shell-escape once a session exists, then flushes it immediately with no following turn (the genuine stuck-working trigger) — then `!`, the command via `send` (literal `-l`; a word-split `keys` burst mangles it and drops Enter), an observation window, and a final turn.
- **Driver** tracks shell mode (`!` on, `Escape` off) so the shell command's `send` is billed no turn (it produces no `gemini` message), keeping `wait_turn` in sync.
- **Spec** re-authored to the warmup arc, dropping `same_session_as` (a transient `proc-<PID>` presession is ready first and would pin `session_birth` to it; turn phases match the real session by order, as in basic-turn).

## Verification
- `parser` + `core/...` + `tools/onboarding-factory/...` test suites — green
- `of verify`: **state 8/8 PASS**, observations PASS (cost/tokens > 0)
- `of validate` OK; replay goldens byte-identical; rig `smoke-test.sh` PASS
- **Discrimination proven**: replaying the same new transcript with the *pre-fix* parser sticks the session in `working` at the shell-escape — the fixture genuinely fails without the fix
- Recorded in coexist mode (isolated daemon on 7838); production untouched

## Notes
- The original `2026-06-12` 2-15 recording is kept alongside the new one (older recordings = informational drift; deleting would trip the #268 guard). Newest wins for validation.
- `replay-fixtures.sh` reports one *pre-existing* nudge — `gemini-cli/2-3_task-list` passes state but is still flagged `known_failing` for an unrelated TaskDelta observation gap (commit `56f255c2`). Untouched here.


ingo-eichhorst:

PR #720, opened 2026-06-21
fix(antigravity): read tokens from the conversation store so the context bar renders (#719)

### LINKED ISSUE #719: Antigravity: context bar (context-window usage) never displays
## Summary

Antigravity sessions never show the **context bar** (context-window usage indicator) in either UI surface. For every Antigravity session the bar is empty (web) or hidden (macOS), regardless of how much work the session has done.

## Root cause

The context bar is driven by two daemon-computed session-metric fields, `context_utilization_percentage` and `pressure_level`. Both come from `ComputeContextUtilization` in `core/pkg/tailer/tailer_metrics.go:479`:

```go
if totalTokens == 0 || modelName == "" {
    return 0, 0.0, "unknown", false
}
// ...
utilizationPercentage := (float64(totalTokens) / float64(effectiveContextWindow)) * 100
```

So the bar requires a **non-zero token count**. The Antigravity adapter never emits one. Its parser (`core/adapters/inbound/agents/antigravity/parser.go`) sets `ParsedEvent.ModelName` but **never** sets `ParsedEvent.Tokens` or `ParsedEvent.ContextWindow`, by design — see the comment at `parser.go:32-34`:

```go
// Tokens/cost are not in the JSONL (they live in the per-conversation SQLite
// protobuf store), so the parser surfaces no usage; state, cwd, model, and
// tool/timeline activity all derive from the transcript alone.
```

Result: every Antigravity session has `totalTokens == 0` → `ComputeContextUtilization` returns `pressure_level = "unknown"`, `context_utilization_percentage = 0`. The UIs then suppress the bar:

- **macOS** — `ContextBar` only renders when `metrics.hasContextData == true` (`platforms/macos/Irrlicht/Views/SessionListView.swift:1231`). `hasContextData` requires `totalTokens > 0 && !modelName.isEmpty && pressureLevel != "unknown"` (`platforms/macos/Irrlicht/Models/SessionState.swift:523`). All three fail → bar hidden, and the `totalTokens > 0` tokens-only fallback also fails → empty space.
- **Web** — the bar element is always rendered (`tools/irrlicht-design-system/ui_kits/dashboard/SessionRow.jsx:35`) but fill width is `Math.min(100, ctx)` = 0% and the percentage label is suppressed because `ctx > 0` is false → an empty/invisible bar.

This is consistent with the recorded fixtures, which already assert `"tokens_nonzero":false,"cost_nonzero":false` for Antigravity scenarios (e.g. `replaydata/agents/antigravity/scenarios/2-7_autonomous-loop/expected.jsonl`).

## Why this is harder than other adapters

Unlike `claudecode` — which reads token usage straight out of the JSONL transcript (`extractClaudeCodeTokens` / `extractClaudeCodeModel` at `core/adapters/inbound/agents/claudecode/parser.go:645-716`) — Antigravity's `transcript.jsonl` contains **no usage data**. Token/cost figures live in a separate per-conversation **SQLite protobuf store** that the transcript-first adapter deliberately does not read. There is no in-transcript field to extract.

### Option A — Read the SQLite protobuf store (highest fidelity)
Read Antigravity's per-conversation SQLite store alongside the transcript, extract cumulative token usage, and emit `ParsedEvent.Tokens` (and let the existing capacity lookup resolve the context window from `ModelName`). This produces a real, accurate utilization %.
- **Pros:** correct numbers, matches `claudecode` fidelity.
- **Cons:** requires locating + reverse-engineering the protobuf schema in the SQLite store; breaks the adapter's transcript-first model by adding a second data source and a DB read on the hot path; needs its own fixtures/recordings. Largest effort.

### Option B — Estimate tokens from transcript text (pragmatic, stays transcript-first)
Maintain a running character/word count over transcript content (user + assistant + thinking + tool I/O) and emit an **estimated** token total (≈ chars/4) as `ParsedEvent.Tokens`. The capacity map already knows Gemini context windows (see below), so utilization becomes computable.
- **Pros:** no new data source; preserves transcript-first design; unlocks the bar immediately.
- **Cons:** approximate — should be surfaced as an estimate (e.g. a flag/tooltip) so it isn't mistaken for exact accounting. (Check whether the metrics pipeline already has an "approximate/estimated" affordance before inventing one.)

### Context-window resolution is already in place (needed by both A and B)
The capacity manager already maps model names to context-window sizes and **already knows Gemini models** — `GetModelCapacity` in `core/pkg/capacity/capacity.go:98`, with Gemini alias entries in `core/pkg/capacity/aliases.go:103-119` (`gemini-3.1-pro`, `gemini-3.1-pro-low`, `gemini-3-flash`, `gemini-3.5-flash`, …). So once a non-zero token count exists, the context window resolves automatically from `ModelName`.

⚠️ **Sub-task / likely gap:** the Antigravity parser harvests the model as a display string like `"Gemini 3.1 Pro (Low)"` (from `<USER_SETTINGS_CHANGE>` blocks). Verify `NormalizeModelName` + the alias map resolve that exact display form to a canonical key with a known context window; if not, add the alias. Without this, even Option A/B would still land in the `"unknown"`/zero-window branch.

## Recommendation
Option B (transcript-text token estimate) preserves the adapter's transcript-first architecture and is the smallest change that makes the bar appear, with Option A as a later fidelity upgrade if exact numbers are wanted. Either way, fix the model-name normalization sub-task and add/record a fixture asserting `tokens_nonzero:true` + a non-`unknown` `pressure_level`.

## Acceptance criteria
- [ ] Antigravity sessions show a non-empty context bar in the web dashboard and macOS app while active.
- [ ] `context_utilization_percentage > 0` and `pressure_level != "unknown"` for an active Antigravity session with real work.
- [ ] `"Gemini 3.1 Pro (Low)"`-style model strings resolve to a known context-window size via the capacity/alias map.
- [ ] A recorded fixture asserts the above (supersedes the current `tokens_nonzero:false` assertion if Option A/B lands), or the deferral is documented if estimation is intentionally left out.

## Key references
- `core/adapters/inbound/agents/antigravity/parser.go:32-34` (no-usage comment), parser sets `ModelName` but not `Tokens`/`ContextWindow`
- `core/pkg/tailer/tailer_metrics.go:479` — `ComputeContextUtilization`, the `totalTokens == 0 → "unknown"` branch
- `core/pkg/capacity/capacity.go:98` + `core/pkg/capacity/aliases.go:103-119` — model→context-window lookup, Gemini entries
- `platforms/macos/Irrlicht/Models/SessionState.swift:523` — `hasContextData` gate
- `platforms/macos/Irrlicht/Views/SessionListView.swift:1231` — macOS `ContextBar` render condition
- `tools/irrlicht-design-system/ui_kits/dashboard/SessionRow.jsx:35` — web context bar render
- `core/adapters/inbound/agents/claudecode/parser.go:645-716` — the usage-extraction pattern Antigravity would mirror (Option A)

ingo-eichhorst:
Fixes #719.

## Problem
Antigravity sessions never show the **context bar**. `ComputeContextUtilization` returns `pressure_level="unknown"` whenever `totalTokens == 0`, and the adapter emitted zero tokens: `transcript.jsonl` carries only the human display model name and no usage at all.

## Root cause
Tokens + the canonical model id live in a sibling per-conversation SQLite store, not the transcript:

- transcript: `~/.gemini/antigravity{,-cli}/brain/<conv-id>/.system_generated/logs/transcript.jsonl`
- store: `~/.gemini/antigravity{,-cli}/conversations/<conv-id>.db` — table `gen_metadata`, one protobuf blob per generation (highest `idx` = current context).

## Fix (kiro-cli sidecar pattern)
- **`dbmetrics.go`** (new) — on `turn_done`, opens the store read-only (WAL-aware, memoized by mtime/size) and decodes the latest `gen_metadata` blob with a dependency-free protobuf wire walker: `#1→#4→#5` = prompt/context tokens, `#1→#19` = canonical model id. A missing/unreadable store degrades to no-usage (pre-#719 behavior).
- **`parser.go`** — opt into `tailer.TranscriptPathAware` (`SetTranscriptPath`) and call `applyDBMetrics` on `turn_done`. The canonical id (e.g. `gemini-3.1-pro-low`) is written back to `p.model` so it stays capacity-resolvable between turns and the bar doesn't flicker off — the transcript's display form (`Gemini 3.1 Pro`) alone does not resolve.
- **`aliases.go`** — a real-data sweep (52 conversations) showed the store also emits `gemini-pro-default` (26), `gemini-3-flash-a` (14), and `gpt-oss-120b-medium` (10); added capacity aliases for those (gemini-3.x → 1M window, gpt-oss-120b → 128k).
- **`agent.go`** — permission `Touches`/`Detail` extended to consent-gate the new `.db` read.

## Verification
- Decoder validated against **real** on-disk stores: `16353 / gemini-3.1-pro-low` and `94240 / gemini-pro-default` → resolve to a `1048576` window → ~1.6% / ~9% "safe".
- New unit tests: protobuf decode, store-path derivation, a real temp-SQLite read (latest-idx + cache hit + graceful absence), and an end-to-end `turn_done` test (token + canonical-model enrichment, no-flicker, store-absent harmless).
- Full `go test ./core/... -race -count=1`, `tools/replay-fixtures.sh` (exit 0), and `of validate` (OK) all pass.

## Scope notes
- **Replay is a deliberate no-op** — recordings capture only the transcript, not the `.db`, so the antigravity `tokens_nonzero:false` goldens stay valid. Proof is the synthetic-store unit tests, not a re-record.
- Cost stays $0 (only `Tokens.Total` is set, no pricing contribution).
- 6/52 sampled conversations carry no canonical model id in the store and remain bar-less (transcript display form doesn't resolve) — a minor tail.

### DIFF
--- a/core/adapters/inbound/agents/antigravity/agent.go
+++ b/core/adapters/inbound/agents/antigravity/agent.go
@@ -63,16 +63,18 @@ func Agent() agent.Agent {
 				Key:             PermissionKeyTranscripts,
 				Kind:            permission.KindObserve,
 				Title:           "Read session transcripts",
-				FeatureUnlocked: "Session list, timeline, and state",
-				Touches: "Reads session transcripts under ~/.gemini/antigravity-cli/ " +
-					"and ~/.gemini/antigravity/ and the working directory of running " +
-					"agy processes",
+				FeatureUnlocked: "Session list, timeline, state, and context-window usage",
+				Touches: "Reads session transcripts and conversation stores under " +
+					"~/.gemini/antigravity-cli/ and ~/.gemini/antigravity/ and the " +
+					"working directory of running agy processes",
 				Detail: "Tails transcript.jsonl files under " +
 					"~/.gemini/antigravity{,-cli}/brain/<conversation>/.system_generated/logs/ " +
-					"to derive session state, model, and timeline. Also scans for " +
-					"running agy CLI processes and reads their working directory to " +
-					"bind a session to its process. Read-only — no file is ever " +
-					"modified. Toggling off stops all reading immediately.",
+					"to derive session state, model, and timeline, and reads the sibling " +
+					"conversation store ~/.gemini/antigravity{,-cli}/conversations/<conversation>.db " +
+					"for token counts and the canonical model name (context-window usage). " +
+					"Also scans for running agy CLI processes and reads their working " +
+					"directory to bind a session to its process. Read-only — no file is " +
+					"ever modified. Toggling off stops all reading immediately.",
 			},
 		},
 	}
--- a/core/adapters/inbound/agents/antigravity/dbmetrics.go
+++ b/core/adapters/inbound/agents/antigravity/dbmetrics.go
@@ -1,0 +1,226 @@
+package antigravity
+
+import (
+	"database/sql"
+	"os"
+	"path/filepath"
+	"time"
+
+	_ "modernc.org/sqlite"
+)
+
+// Antigravity keeps token usage and the canonical model id in a per-conversation
+// SQLite store that sits beside the transcript — the transcript.jsonl itself
+// carries only the human-readable model display name and no usage at all. For a
+// transcript at
+//
+//	<root>/brain/<conv-id>/.system_generated/logs/transcript.jsonl
+//
+// the store is <root>/conversations/<conv-id>.db. Reading it lights up the
+// context bar (#719); without it every Antigravity session reports zero tokens,
+// so ComputeContextUtilization returns pressure "unknown" and both UIs hide the
+// bar. A missing or unreadable store degrades to no-usage — the pre-#719 state.
+
+// dbModelTokens is one read of the conversation store: the canonical model id
+// and the latest generation's context-token count. Either may be zero/empty
+// when the store's shape is unrecognized; the caller then emits no usage.
+type dbModelTokens struct {
+	model         string
+	contextTokens int64
+}
+
+// dbCache memoizes one store read keyed by (mtime, size) so a backfill of a
+// long transcript — every historical turn_done re-reading the same growing
+// .db — costs one query per change, not one per turn (mirrors kirocli's
+// sidecarCache).
+type dbCache struct {
+	mtime time.Time
+	size  int64
+	val   dbModelTokens
+	ok    bool
+}
+
+// storePathForTranscript maps a transcript path to its sibling conversation
+// store, or "" when the path is not a recognized Antigravity transcript. The
+// <conv-id> directory is shared between the brain transcript and the
+// conversations store, so it is the join key.
+func storePathForTranscript(transcriptPath string) string {
+	conv := sessionIDFromPath(transcriptPath)
+	if conv == "" {
+		return ""
+	}
+	// Climb to <root>: transcript.jsonl → logs → .system_generated → <conv-id>
+	// → brain → <root>.
+	root := filepath.Dir(filepath.Dir(filepath.Dir(filepath.Dir(filepath.Dir(transcriptPath)))))
+	return filepath.Join(root, "conversations", conv+".db")
+}
+
+// readStoreModelTokens opens the conversation store read-only and returns the
+// canonical model id plus the latest generation's context-token count, memoized
+// by (mtime, size). The bool is false when the store is missing, locked, or its
+// gen_metadata shape is unrecognized — callers degrade to no-usage.
+func readStoreModelTokens(transcriptPath string, cache *dbCache) (dbModelTokens, bool) {
+	dbPath := storePathForTranscript(transcriptPath)
+	if dbPath == "" {
+		return dbModelTokens{}, false
+	}
+	fi, err := os.Stat(dbPath)
+	if err != nil {
+		return dbModelTokens{}, false
+	}
+	if cache != nil && cache.ok && fi.ModTime().Equal(cache.mtime) && fi.Size() == cache.size {
+		return cache.val, true
+	}
+
+	// Read-only, WAL-aware, short timeout so a live agy write never blocks the
+	// daemon (mirrors the opencode metrics reader).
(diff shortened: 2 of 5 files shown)


ingo-eichhorst:
## Self-review (high-effort, 8 finder angles → verify)

Pushed `c619cfe0` addressing the findings.

### Fixed
- **WAL staleness (correctness, the important one)** — the conversation store runs in WAL mode (`.db-wal` siblings confirmed on disk), so a live `agy` write lands in `-wal` while the main `.db`'s mtime/size don't move until a checkpoint. The cache keyed only on the main file would serve a stale token count and **freeze the bar for an active session** — the exact case the feature targets. Now the `-wal` size is part of the freshness key (size, not mtime, so our own read-only opens don't spuriously invalidate and re-trigger an O(N) backfill). Added cache-invalidation tests (new `gen_metadata` row + grown `-wal`).
- **Model-switch regression test** — added `TestModelSwitchSurvivesStore` proving the store write-back canonicalizes to the *switched-to* model (the switch is always captured by `parseUserInput` first), closing the test gap a finder flagged.

### Reviewed, kept as-is (with rationale)
- **Hand-rolled protobuf walker** — a finder suggested `google.golang.org/protobuf/encoding/protowire`, but that package is **not** a current dependency; adding a module for ~50 trivial, well-tested lines isn't warranted. Kept.
- **Canonical-model write-back to `p.model`** — flagged as a possible clobber of mid-session switches. Verified it's not: `parseUserInput` captures every `<USER_SETTINGS_CHANGE>` first, and write-back only canonicalizes the id at turn boundaries (consistent with how other adapters display model ids, e.g. `claude-opus-4-6`). Locked in by the new test.
- **Shared sidecar-read infra / alias auto-discovery** — out of scope; two adapters with different store formats (kiro JSON vs antigravity SQLite+protobuf) don't yet justify the abstraction, and aliases follow the existing `ir:refresh-aliases` maintenance pattern.

Full `go test ./core/... -race`, `replay-fixtures.sh`, and `of validate` remain green.

Issue #1, opened 2025-09-05
Phase 0: Contracts & Drift Guard (Foundations)


ingo-eichhorst:
## Overview

Phase 0 establishes the foundational contracts and safety mechanisms before implementing the core system. This phase ensures external interfaces are frozen and failures are safe.

## Goal

Freeze external contracts and make failures safe by building robust testing infrastructure and configuration management tools.

### 1. Hook Contract Fixtures
- [ ] Create `fixtures/` directory with JSON samples for each Claude Code hook event:
  - [ ] `SessionStart` event fixture
  - [ ] `UserPromptSubmit` event fixture  
  - [ ] `Notification` event fixture
  - [ ] `Stop` event fixture
  - [ ] `SubagentStop` event fixture
  - [ ] `SessionEnd` event fixture
- [ ] Include edge cases: missing fields, oversized payloads, malformed JSON
- [ ] Document expected field schemas in `fixtures/README.md`

### 2. Replay Testing Tool
- [ ] Implement `irrlicht-replay` test tool that:
  - [ ] Pipes fixture JSON to `irrlicht-hook` via stdin
  - [ ] Supports single event and multi-event scenarios
  - [ ] Includes timing controls for realistic event sequences
  - [ ] Validates output JSON files match expected format
- [ ] Create test scenarios for concurrency (2, 4, 8 sessions)

### 3. Settings Merger Library
- [ ] Implement robust JSON settings merge functionality:
  - [ ] **JSON-aware deep merge** preserving existing structure
  - [ ] **Dry-run mode** to preview changes without applying
  - [ ] **Idempotency** - running multiple times produces same result
  - [ ] **Timestamped backup** of original `~/.claude/settings.json`
  - [ ] **Rollback capability** to restore from backup
  - [ ] **Surgical removal** - remove only Irrlicht hooks without backup
- [ ] Handle edge cases: missing file, malformed JSON, permission errors
- [ ] Unit tests covering all merge scenarios

### 4. Kill Switch Mechanism
- [ ] Environment variable `IRRLICHT_DISABLED=1` to disable all hooks
- [ ] Settings flag in `~/.claude/settings.json` to disable per-hook
- [ ] Graceful degradation when disabled (no-op hook behavior)

### Replay Suite
- [ ] All fixture events replay successfully through `irrlicht-hook`
- [ ] Multi-session scenarios (2, 4, 8 concurrent) execute without conflicts
[...]

Issue #718, opened 2026-06-21
macOS: toggle to publish the local daemon's sessions to a remote relay (funken.io)


ingo-eichhorst:
## Goal

To actually use **funken.io**, a user's local `irrlichd` needs to **publish** its sessions to the hosted relay so they appear remotely. The macOS app should let a user turn that on with a toggle and configure the remote relay URL (funken.io) + the irrlichtrelay device token — i.e. keep serving sessions locally **and** forward them to a configurable remote address, controlled from the app.

## End-to-end operating flow (the loop this issue completes)

This issue is the **last missing piece** to operate Irrlicht against funken.io. The intended journey:

1. **Get a token** — sign in at `admin.funken.io`, issue a device token (the relay "key"). funken writes it (hashed) to the relay's `tokens.json`, which the relay hot-reloads — so the token is valid at `funken.io` immediately. The token is scoped to your account's workspace (`gh:<id>`).
2. **Publish (machine A)** — in the macOS app, enter relay URL `wss://funken.io` + the token and turn **Publish** on → `irrlichd` forwards its sessions to the hosted relay. **← this issue.**
3. **Subscribe (machine B, or the same Mac)** — in the macOS app, enter `wss://funken.io` + a token from the **same funken account** and turn **Relay server** (subscribe) on → the published sessions appear. *(Already implemented.)*

Because both tokens belong to the same funken account, they resolve to the same workspace, so step 3 sees step 2's sessions. Everything in steps 1 and 3 already works; **step 2 is the gap.**

## What already exists (don't rebuild)

Most of the machinery is already there — this is mostly a **wiring / UX** task, not a new daemon protocol.

- **Daemon → relay forwarder already works, but is env-only.** `irrlichd` starts an outbound forwarder when `IRRLICHT_RELAY_URL` is set, authenticating with `IRRLICHT_RELAY_TOKEN` or `<dataDir>/relay-token.json`.
  - Opt-in block: `core/cmd/irrlichd/main.go:180`. Forwarder: `core/adapters/outbound/relay/forwarder.go`.
  - It dials `<url>/api/v1/sessions/stream` (scheme normalized to `ws`/`wss`, stream path appended when absent), so the configured value can be a bare base like `wss://funken.io` — `forwarder.go:23-45`.
[...]

PR #723, opened 2026-06-21
feat(relay): hot-reload the publish forwarder without a daemon relaunch (#722)

### LINKED ISSUE #722: macOS: hot-reload the relay forwarder so "Publish to relay" applies without a daemon relaunch
## Goal

Make the macOS **"Publish to relay"** toggle (shipped in #718, PR #721) reconfigure the **running** `irrlichd` live, instead of relaunching the daemon. Activating/deactivating publish — or changing the relay URL/token — should take effect immediately on the daemon already running, regardless of whether the app spawned it or adopted an existing one.

## Background

#718 wired the toggle by passing `IRRLICHT_RELAY_URL` / `IRRLICHT_RELAY_TOKEN` into the daemon's launch environment and **relaunching the app-owned daemon** to apply a change (`DaemonManager.publishSettingsDidChange()` → `relaunch()`). The forwarder is started once at daemon startup from those env vars.

That has two problems:

1. **Only works when the app spawned the daemon.** `relaunch()` guards on `process != nil` and no-ops otherwise. When the app *adopts* an already-running daemon (e.g. a daemon started by a LaunchAgent, or the `/ir:test-mac` replace-mode harness), toggling Publish on does nothing — `GET /api/v1/relay/publish` stays `{"enabled":false}` and the status dot stays "Publishing — off" even though the toggle is green. Observed live while testing.
2. **Heavyweight even when it works.** Relaunching the daemon to flip one setting tears down and reconnects every session's monitoring just to start/stop an outbound WebSocket.

This was called out as the open question in #718 ("make the forwarder hot-reloadable — larger change in `irrlichd`"). This issue is that larger change.

### Daemon (`scope:daemon`) — a hot-reloadable forwarder
- Introduce a `PublishController` that owns the forwarder's lifecycle:
  - `Apply(enabled, url, token)` — starts the forwarder (cancelable `context`), stops it (cancel the ctx), or reconfigures it (cancel + start a new `relay.NewForwarder` with the new url/token). Idempotent when the effective config is unchanged. Mutex-guarded.
  - Holds the current forwarder (or nil) so the existing `GET /api/v1/relay/publish` reports `{"enabled":false}` when stopped and the live `Forwarder.Status()` when running.
  - **Seeded from `IRRLICHT_RELAY_URL` / `IRRLICHT_RELAY_TOKEN` at startup** so headless/standalone daemons (no macOS app) keep working exactly as today.
- New `PUT /api/v1/relay/publish` accepting `{ "enabled": bool, "url": string, "token": string }` → `controller.Apply(...)` → returns the resulting status (same shape as `GET`). Loopback-only, like the rest of the daemon API.
- Touch points: `core/adapters/outbound/relay/forwarder.go` (forwarder is already ctx-cancelable — reuse), `core/cmd/irrlichd/main.go` (the env opt-in block that currently does `go fwd.Run(relayCtx)`), `core/cmd/irrlichd/handlers.go` (`handleGetPublishStatus` → add the PUT handler).

### macOS app (`scope:macos`) — drive it over loopback
- `DaemonManager.publishSettingsDidChange()` **POSTs** `{enabled,url,token}` to `PUT /api/v1/relay/publish` on the running daemon instead of relaunching. Also POST once on launch, after the daemon becomes reachable, so the daemon's publish state matches the app's settings without waiting for a toggle.
- Remove the relaunch path for publish and the publish-env layering in `DaemonManager.buildDaemonEnv` (keep it for `IRRLICHT_BIND_ADDR`). The `process === terminated` crash-restart guard added in #718 can stay (it's a general correctness fix).
- The existing `PublishStatusMonitor` poll then reflects the live reconfigure within its interval.

## Token-delivery change (decision)

The relay token moves from a process **env var** (set at spawn) to a **loopback HTTP POST body**. The daemon binds `127.0.0.1` only, so this is the same trust boundary as every other daemon endpoint (hooks, permissions/answer, focus), and the token is already readable from the user's own Keychain. Accepted as part of this change; no auth added to the loopback API.

## Acceptance criteria

- Toggling **Publish to relay** on/off, or editing the URL/token, takes effect on the **already-running** daemon with no relaunch — verified both when the app spawned the daemon and when it adopted an existing one (e.g. `/ir:test-mac replace`): `GET /api/v1/relay/publish` flips between `connected` and `enabled:false` accordingly, and the status dot tracks it.
- No session monitoring is interrupted by toggling publish (no daemon restart).
- Headless/standalone daemons still start publishing from `IRRLICHT_RELAY_URL` / `IRRLICHT_RELAY_TOKEN` at startup (backward compatible).
- A bad token still surfaces as `auth_failed` (red dot); turning publish off stops forwarding.
- Tests: Go — `PublishController` start/stop/reconfigure + the `PUT` handler; Swift — the POST client + "toggle does not relaunch the daemon". Re-run the auth-enabled-relay E2E.

## Out of scope
- Any relay-side or funken-side change — unchanged from #718.
- Auth on the daemon's loopback API (separate concern if ever exposed beyond loopback).

## Scope
`scope:daemon` (hot-reload controller + PUT endpoint) + `scope:macos` (POST instead of relaunch). Follow-up to #718 (merged via #721).

ingo-eichhorst:
Closes #722. Follow-up to #718 (merged via #721).

## Problem

The macOS **"Publish to relay"** toggle applied changes by **relaunching the app-owned daemon** (`publishSettingsDidChange()` → `relaunch()`, guarded on `process != nil`). Two problems:

1. **No-op when the app adopted the daemon.** When the app adopts an already-running daemon (LaunchAgent, or `/ir:test-mac replace`), `relaunch()` does nothing — the toggle goes green but `GET /api/v1/relay/publish` stays `{"enabled":false}`.
2. **Heavyweight when it works.** Relaunching tears down and reconnects every session's monitoring just to start/stop one outbound WebSocket.

### Daemon (`scope:daemon`)
- New `relay.PublishController` owns the forwarder lifecycle: `Apply(enabled, url, token)` starts / stops / reconfigures a single forwarder (cancel the ctx + start a fresh `relay.NewForwarder`). Idempotent when the effective config is unchanged; mutex-guarded so concurrent PUTs serialize. A blank URL counts as "off" (mirrors the old env-builder semantics).
- The controller is **always constructed** and **seeded once at startup** from `IRRLICHT_RELAY_URL` / `IRRLICHT_RELAY_TOKEN`, so headless/standalone daemons keep working exactly as before.
- New loopback-only `PUT /api/v1/relay/publish` accepting `{enabled,url,token}` → `Apply(...)` → returns the resulting status (same shape as `GET`). `GET` now reads the controller too.

### macOS app (`scope:macos`)
- `DaemonManager.publishSettingsDidChange()` now **PUTs** `{enabled,url,token}` to the running daemon (new `PublishClient`) instead of relaunching — works whether the app spawned or adopted the daemon.
- Re-syncs publish state **on spawn** (once the daemon answers) and **on adopt**, so the daemon matches the app's settings without a manual toggle.
- `buildDaemonEnv` no longer layers the relay env (keeps `IRRLICHT_BIND_ADDR`). The `process === terminated` crash-restart guard from #718 stays (general correctness).

## Token delivery (decision, per the issue)

The relay token moves from a spawn-time **env var** to the **loopback PUT body**. The daemon binds `127.0.0.1` only — same trust boundary as every other daemon endpoint (hooks, permissions/answer, focus) — and the token is already readable from the user's Keychain. No auth added to the loopback API (out of scope).

## Tests

- **Go** (`go test ./core/... -race -count=1`): `PublishController` start/stop, idempotent re-apply, live reconfigure (URL change reconnects to a new relay), blank-URL-acts-as-off; `PUT` handler toggles publishing on/off end-to-end and rejects a malformed body with 400. Full core suite green (the `TestDebounce_FirstEventFiresImmediately` timing test is a pre-existing flake under full `-race` load — passes in isolation; untouched by this change).
- **Swift**: `PublishClient` PUT wire shape (enabled + disabled bodies); `publishSettingsDidChange` does not spawn/relaunch a daemon.

## Acceptance criteria status

- [x] Toggle/URL/token changes apply to the **already-running** daemon with no relaunch — both spawned and adopted paths (PUT + on-adopt/on-spawn re-sync).
- [x] No session monitoring interrupted by toggling publish (no restart).
- [x] Headless/standalone daemons still start from `IRRLICHT_RELAY_URL`/`IRRLICHT_RELAY_TOKEN` (controller seeded at startup).
- [x] Turning publish off stops forwarding; bad token still surfaces as `auth_failed` (forwarder status path unchanged).
- [ ] Live auth-enabled-relay E2E re-run on a real device — **not run in this worktree** (no relay/Keychain here); flagged for manual verification.

### DIFF
--- a/core/adapters/outbound/relay/controller.go
+++ b/core/adapters/outbound/relay/controller.go
@@ -1,0 +1,112 @@
+package relay
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"irrlicht/core/ports/outbound"
+)
+
+// PublishController owns the relay forwarder's lifecycle so publishing can be
+// turned on, off, or reconfigured on a running daemon without a relaunch
+// (issue #722). The macOS app drives it over the daemon's loopback
+// PUT /api/v1/relay/publish; a headless/standalone daemon seeds it once from
+// IRRLICHT_RELAY_URL / IRRLICHT_RELAY_TOKEN at startup (Apply below), so its
+// behavior is unchanged from the env-only path that preceded this.
+//
+// At most one Forwarder runs at a time. Reconfiguring cancels the current
+// forwarder's context and starts a fresh one — the forwarder is already
+// ctx-cancelable, so this reuses its existing shutdown path.
+type PublishController struct {
+	parentCtx context.Context
+	identity  Identity
+	push      outbound.PushBroadcaster
+	snapshot  SnapshotFunc
+	logger    outbound.Logger
+
+	mu     sync.Mutex
+	fwd    *Forwarder         // non-nil only while publishing is enabled
+	cancel context.CancelFunc // cancels fwd's Run; nil when stopped
+	url    string             // last applied (trimmed) url, for idempotency
+	token  string             // last applied (trimmed) token, for idempotency
+}
+
+// NewPublishController builds a controller in the stopped state. parentCtx
+// bounds every forwarder it starts, so cancelling it (on daemon shutdown)
+// stops publishing. identity, push, and snapshot are the forwarder
+// dependencies that never change across reconfigures; logger may be nil.
+func NewPublishController(parentCtx context.Context, id Identity, push outbound.PushBroadcaster, snapshot SnapshotFunc, logger outbound.Logger) *PublishController {
+	return &PublishController{
+		parentCtx: parentCtx,
+		identity:  id,
+		push:      push,
+		snapshot:  snapshot,
+		logger:    logger,
+	}
+}
+
+// Apply reconciles the running forwarder to the requested publish config.
+// enabled with a non-empty url starts (or reconfigures) the forwarder; either
+// flag falsy stops it. It is idempotent: re-applying the config already in
+// effect is a no-op, so the macOS app can POST on every settings nudge without
+// churning the relay link. Mutex-guarded so concurrent PUTs serialize.
+func (c *PublishController) Apply(enabled bool, url, token string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	url = strings.TrimSpace(url)
+	token = strings.TrimSpace(token)
+	// A blank URL means "off" even when the toggle is on — mirrors the macOS
+	// env builder's old semantics so an empty URL never activates a forwarder.
+	wantOn := enabled && url != ""
+
+	if !wantOn {
+		if c.fwd == nil {
+			return // already stopped — nothing to do
+		}
+		c.cancel()
+		c.cancel = nil
+		c.fwd = nil
+		c.logInfo("relay publishing stopped")
+		return
+	}
+
+	// Running with the same effective config — leave the link untouched.
+	if c.fwd != nil && c.url == url && c.token == token {
+		return
+	}
+
+	// Start fresh: stop any existing forwarder first so only one link runs.
+	if c.cancel != nil {
+		c.cancel()
+	}
+	fwd := NewForwarder(url, c.identity, token, c.push, c.snapshot, c.logger)
+	ctx, cancel := context.WithCancel(c.parentCtx)
+	c.fwd = fwd
+	c.cancel = cancel
+	c.url = url
+	c.token = token
+	go fwd.Run(ctx)
+	c.logInfo(fmt.Sprintf("relay publishing → %s (daemon %q / %s)", fwd.Status().URL, c.identity.DaemonLabel, c.identity.DaemonID))
+}
+
+// Status reports whether publishing is enabled and, when it is, the live link
+// state of the running forwarder. When stopped it returns (false, zero Status)
+// so the endpoint renders {"enabled":false}.
+func (c *PublishController) Status() (enabled bool, status Status) {
+	c.mu.Lock()
+	fwd := c.fwd
+	c.mu.Unlock()
+	if fwd == nil {
+		return false, Status{}
+	}
+	return true, fwd.Status()
+}
+
+func (c *PublishController) logInfo(msg string) {
+	if c.logger != nil {
+		c.logger.LogInfo("relay-publish", "", msg)
+	}
+}
--- a/core/adapters/outbound/relay/controller_test.go
+++ b/core/adapters/outbound/relay/controller_test.go
@@ -1,0 +1,102 @@
+package relay
+
+import (
+	"testing"
+	"time"
+
+	"irrlicht/core/domain/session"
+)
+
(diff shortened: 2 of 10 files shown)


ingo-eichhorst:
## Self-review (high effort) — 2 fixes applied, 2 findings consciously left

Ran a multi-angle review (Go correctness, Swift correctness, cleanup/conventions) over the diff. Two genuine issues fixed in `5a2dc1af`; two findings evaluated and deliberately left.

### Fixed

1. **Spawned daemon inherited stale relay env** — `buildDaemonEnv` passed the app's full environment through, and the daemon *self-seeds* its forwarder from `IRRLICHT_RELAY_URL`/`IRRLICHT_RELAY_TOKEN` at boot. So an app-spawned daemon launched from a shell/LaunchAgent that exported those vars would publish to a relay the user never configured in the UI, until the corrective PUT landed. The original #718 code stripped these (with a dedicated test); the env-layering removal dropped that guard. **Re-added the strip + a focused test** (`testBuildDaemonEnvStripsInheritedRelayVars`).

2. **Single PUT, no retry** — after fix #1, app-spawned daemons start publish-off and depend *entirely* on the loopback PUT to turn publishing on; nothing else re-pushes config until the next settings change. A dropped request would strand the daemon in the wrong state. `pushPublishConfig` now **retries up to 3×** (Apply is idempotent, so it's safe). Covers both the spawn-sync and adopt paths.

### Evaluated, left as-is

3. **Reconfigure overlap** (`controller.go`) — `Apply` cancels the old forwarder asynchronously then starts the new one, so in theory two goroutines briefly coexist. In practice the old `runOnce` wakes on `ctx.Done()` in one scheduler hop while the new forwarder needs a full network RTT to connect, so no two links are ever *connected* at once, and the relay upserts the snapshot idempotently regardless. Not worth the synchronous-handoff complexity.

4. **`LoadOrCreateIdentity` now unconditional** (`main.go`) — every daemon mints `relay-identity.json` at startup even if publishing is never used (pre-#722 it was lazy). Cosmetic (a 0600 file); the identity must be ready for a later PUT anyway, and making it lazy would mean reworking the controller's constructor. Left.

### Verification
- Go: `go test ./core/... -race -count=1` green (the `TestDebounce_FirstEventFiresImmediately` failure is a pre-existing timing flake under full `-race` load — passes 3× in isolation; untouched by this PR).
- Swift: `swift build` clean; `DaemonManagerTests` + `PublishClientTests` pass (9 tests).
- ⚠️ The live auth-enabled-relay E2E (AC final item) was **not** run — no relay/Keychain in this worktree. Needs a manual pass on a real device.

PR #788, opened 2026-07-03
fix(macos): move History's tab switcher into header, collapse controls to one row, fix panel height

### LINKED ISSUE #785: macOS History panel: consolidate controls to header + one line, fix panel height
The macOS menu-bar History panel (`platforms/macos/Irrlicht/Views/HistoryView.swift`, hosted in the 380pt-wide popover — `SessionListView.panelWidth`) has two layout problems.

## A) Controls should collapse to header + one line

Current layout, each on its own row (top to bottom):

1. Header: static **"History"** title (centered) + "‹ Back" button — `header` (lines 158–176)
2. Usage / Yield / Quota segmented picker — `topControls` (lines 186–190)
3. Range + Chart(Cost/Tokens) pickers — `usageControls` row 1 (lines 207–220)
4. Group picker + Forecast toggle — `usageControls` row 2 (lines 224–241)
5. (optional) custom date-range row
6. Provider / Type / Project filter menus — `usageControls` row 3 (lines 248–254)

That's up to 5 stacked control rows above the chart.

Requested:
- Move the chart toggle into the header row, **replacing** the static "History" text — e.g. `‹ Back    [Usage | Yield | Quota]`.
- Collapse the remaining tab-specific controls (Range, Chart metric, Group, Forecast, cross-filters) onto **one line** directly below the header, instead of 3 stacked rows.

*Assumption:* "chart toggle" = the Usage/Yield/Quota segmented control (the only true toggle-style control; "Chart" is a Cost/Tokens dropdown menu, not a toggle). Correct me if the intent was the Cost/Tokens picker instead.

**Known constraint:** the panel is fixed at 380pt wide, and the existing 2-per-row layout was deliberately chosen so nothing overflows it (see the comment at HistoryView.swift:202-205 — "a `.menu` Picker sizes to its widest option, ~90pt each"). Fitting Range + Chart + Group + Forecast + 3 filter menus onto one row will likely need more compact controls (icon-only buttons, shorter labels, etc.) — this may need a design pass rather than a literal one-line drop-in.

## B) Panel height should be fixed, not inherited from the prior view

`content` (lines 353–364) only **caps** height at 560pt (`.frame(maxHeight: Self.contentMaxHeight)`) around a `ScrollView` — it doesn't fix it. Since History swaps into the same popover as the live session list (`SessionListView`), the popover's actual height ends up depending on whatever the session list was showing right before the switch. Result: the chart + legend can render clipped (needing a scroll) or the panel looks too short/tall depending on how many sessions were in the list.

Requested: give the History panel (or its outer container) a **fixed height**, sized so the chart (`HistoryCostChart`, `.frame(height: 200)` at line 605) plus the full legend/summary below it are always fully visible without scrolling — independent of the session list's prior size.

## Acceptance criteria

- [ ] Header row shows the chart/tab toggle in place of the "History" text; Back button unchanged.
- [ ] All Usage-tab controls (Range, Chart, Group, Forecast, filters) render on a single row directly under the header.
- [ ] Opening History right after a long or short session list — or switching between Usage/Yield/Quota tabs — no longer changes the popover's height; it's fixed.
- [ ] The fixed height is tall enough that the chart and the full legend list render without scrolling for the common case (a handful of contributors).
- [ ] `HistoryViewSnapshotTests` / `HistoryPhase2Tests` pass; snapshots regenerated locally if the layout changes (macOS snapshot tests aren't CI-gated — see AGENTS.md).

ingo-eichhorst:
## Summary
- Move the Usage/Yield/Quota tab picker into `HistoryView`'s header, replacing the static "History" label
- Flatten Range/Chart/Group/Forecast/cross-filters onto one horizontally-scrollable row instead of three stacked rows
- Give the popover a hard fixed height (independent of the width) instead of only capping the content ScrollView, so it no longer inherits whatever height `SessionListView` happened to be at, and stays constant across Usage/Yield/Quota

Closes #785

## Test plan
- [x] `swift build` succeeds
- [x] `swift test --filter HistoryViewSnapshotTests` — all 7 snapshots pass unchanged (they host the inner content views directly, bypassing header/topControls, so no golden drift expected or observed)
- [x] `swift test --filter HistoryPhase2Tests` — all 12 pass unchanged
- [x] Visual verification via a temporary local-only snapshot test hosting the full `HistoryView` against a contrasting backdrop — confirmed the tab picker renders in the header, all Usage-tab controls render on one row, and the panel is a hard-fixed 380×560 box regardless of content (test file was not committed)


ingo-eichhorst:
Addressed the review findings in 511b503a:

- **Hidden filters/forecast** — kept the horizontal scroll indicator visible on the merged Usage row so there's a cue that Forecast + the 3 filters are reachable by scrolling (previously `showsIndicators: false` hid them with no affordance).
- **Quota tab blank gap** — `topControls` (and its divider) is now skipped entirely for the Quota tab instead of rendering empty-but-padded.
- **Custom range stealing chart space** — the panel now grows by a fixed increment when the custom date-range row appears, instead of shrinking the fixed content area to make room for it.
- **Header accessibility-size crowding** — rebuilt as a leading-button/centered-picker/invisible-mirror HStack with guaranteed minimum spacing, instead of a ZStack overlay that could close the gap at large Dynamic Type sizes.
- **Cleanup** — extracted a shared `scrollableControlRow`/`rangePicker` to remove duplication between `usageControls`/`yieldControls`, dropped the pointless `ScrollView` wrapper around Yield's single picker, and removed a redundant `.pickerStyle(.menu)` on the Group picker.

Noted but intentionally not changed: the panel's height is hard-fixed (rather than sizing to content via the `.fixedSize(vertical:)` pattern used elsewhere in the app) so Quota/Yield show some blank space below shorter content — this is the deliberate tradeoff for the literal "fixed height" ask in #785, not an oversight.

All 198 macOS tests pass (`swift build` + `swift test`).

Issue #728, opened 2026-06-22
[feat]: make "click leads to the session" available for terminals inside Obsidian"


gernotstarke:
### What problem are you trying to solve?

for example in my Ghostty terminal sessions, I can "click" into the irrlicht entry of that session and am re-directed to the corresponding window.

I would like that for my in-Obsidian terminal sessions too.

### Proposed solution

No idea how to implement, but you solved the problem for Ghostty and other terminal emulators...

### Who benefits from this?

Everyone, Multi-agent users (3+ concurrent sessions)

ingo-eichhorst:
... investigating 🧐 ...

ingo-eichhorst:
Looked into it 🧐 — short version: doable in principle, with one real unknown.

Click-to-focus dispatches on the *host* app: for editor-embedded terminals (VS Code, Cursor, JetBrains) we detect the host by walking the agent process's parent chain to its `.app`, then activate it and raise the window whose title matches the session's cwd. Adding Obsidian is normally just one line on each side (a host-map entry + reusing the generic AX-title activator with bundle `md.obsidian`).

Two caveats specific to Obsidian:

1. **Detection depends on your terminal plugin.** It only works if the plugin runs the agent as a *child of Obsidian's process tree* (an integrated terminal), so we can trace the ancestry back to `Obsidian.app`. If the plugin launches an external terminal instead, we'd detect that terminal, not Obsidian. → **Which Obsidian terminal plugin are you using?**

2. **Obsidian is a single window where the terminal is a *pane*.** The window title is the active note/vault, not the terminal's cwd — so the best we can realistically do is bring the Obsidian *window* forward, not focus the specific terminal pane (Obsidian exposes no per-pane targeting handle).

Good news: the `processes.txt` from the diagnostic bundle I asked for in #727 shows exactly the process ancestry, so it'll answer caveat 1 directly. So the same capture unblocks both this and the session-list issue.

gernotstarke:
￼

ingo-eichhorst:
Follow-up — rather than hardcode Obsidian (and the next embedded-terminal app after it), there's a **general** version of this worth considering.

[...]

Issue #576, opened 2026-06-04
Consent-grant backfill skips live sessions as orphan transcripts (live process in same sweep ignored)


ingo-eichhorst:
## Summary

When observe consent is granted **after** agent sessions are already running, the watcher-start backfill sweep classifies any transcript idle for more than `orphanTranscriptAge` (2 min) as an orphan and skips it — even when process discovery registered a **live agent process in the same cwd 120 ms earlier in the same sweep**. The session is never created, never tailed, and state transitions (e.g. working→waiting on AskUserQuestion) are silently missed.

Pre-#570 this was rare: the daemon started at boot, so "stale at first sight" reliably meant "exited process". The consent-first gate (#570/#571) makes *grant long after sessions started* the canonical path — including **every first-run onboarding**: a new user installs irrlicht, clicks through the wizard, and their currently-running-but-idle sessions don't appear.

## Observed instance (2026-06-04)

| Time | Event |
|---|---|
| 22:40:21 | Dev daemon (post-#570 build) starts; no `permissions.json` → all pending → zero monitoring (by design) |
| 22:40:27 | `claude --worktree 554` starts (pid 90098); transcript `a39b7ef1` written 22:41→22:44:21 |
| ~22:44+ | Agent raises AskUserQuestion (writes nothing to the JSONL while pending) |
| 22:50:19 | Consent granted via wizard → watchers start, backfill sweep runs |

The sweep, 0.1 ms apart in `events.log`:

```
22:50:19.678  new session detected proc-90098 … worktrees-554   (live pid registered as pre-session)
22:50:19.798  new session detected a39b7ef1 … worktrees-554
22:50:19.799  "skipping orphan transcript"
```

Transcript mtime 22:44:21 was 6 min old > 2 min window → `isStaleTranscript` → skipped. The `proc-90098` pre-session stays a placeholder (state `ready`, no transcript path, no tailing); the upgrade path `cleanupPreSessionsForProject` never runs because the real session was skipped. The user-visible symptom: the session sat on an AskUserQuestion with no working→waiting transition, and the session was missing from the dashboard entirely.

## Code

- Orphan skip: `core/application/services/session_detector_activity.go:68` (`onNewSession`)
- Heuristic: `isStaleTranscript`, `core/application/services/helpers.go:29`; `orphanTranscriptAge = 2 * time.Minute`, `core/application/services/session_detector.go:28`
[...]

PR #791, opened 2026-07-03
fix(antigravity): reject sessions bound to a non-interactive host

### LINKED ISSUE #784: [bug]: irrlicht 0.5.4 shows non-existing entries
### Preflight

- [x] I'm on the latest release from [Releases](https://github.com/ingo-eichhorst/Irrlicht/releases/latest).
- [x] I searched existing issues and didn't find a duplicate.

### Irrlicht version

0.5.4

### macOS version

26.5.1

### Which agent(s) were involved?

Not agent-specific / app or daemon itself

### What happened?

Image

Irrlicht shows THREE circles in menu bar, although just a single Ghossty-based claude-cli session is running.

The screenshot shows that (correct) entry in the first row (from quality.arc42.org), but the other two are strange relicts. I tried to re-start irrlicht, no effect.

### Steps to reproduce

just normal MacOS usage. Starting and stopping terminal sessions (both iterm and ghostty), plus claude and codex 

### Expected vs actual behavior

When just a single agent is active, I expect just a single irrlicht identicator.

### Relevant log excerpts

```shell

```

### Session state file (optional)

```json

```

### Anything else?

_No response_

ingo-eichhorst:
## Summary

CodexBar (a third-party menu-bar app, github.com/steipete/CodexBar) deliberately keeps an Antigravity `agy` CLI process running in the background to poll quota data — confirmed in its own docs/source (PTY-launched, bare argv, cwd forced to `$HOME`, never sent any input). Irrlicht had no way to recognize this as non-interactive infra, so it surfaced a permanent phantom menu-bar circle for it; restarting Irrlicht didn't help since the app just reconnects to the already-running daemon.

A TTL/staleness-based fix was considered and rejected: Antigravity's `NoSubstantiveActivity` signal resets on any parseable line, so a periodically-active CodexBar poll would make a reaped ghost look freshly alive again minutes later — a worse, flickering UX than the original bug.

- Add `Process.RequireKnownHost` (mirrors the existing `ExcludeArgv` opt-in pattern) — antigravity is the only adapter that sets it.
- `PIDManager.AllowsSession` does a synchronous, one-shot ancestry check the moment a candidate PID is discovered, wired into `onNewSession` alongside the existing admission vetoes. A process whose ancestry doesn't resolve to a known terminal/IDE never becomes a session in the first place — no delay for real sessions, no reappearing ghost for CodexBar's.
- Reuses the ancestry-walking machinery built for #728 (Obsidian click-to-focus), extended with an explicit `knownEmbeddedHostBundleIDs` allow-list (starting with Obsidian's bundle ID) — the existing generic "any top-level `.app`" fallback can't be reused as-is since it would also match CodexBar.
- Fails open on non-macOS platforms (ancestry resolution is darwin-only, and CodexBar itself only exists on macOS).

Closes #784

## Test plan

- [x] `go test ./core/... -race -count=1` — full suite green (two unrelated pre-existing `-race` flakes in `core/application/services` reproduced identically on unmodified `main`, confirmed not caused by this change)
- [x] `go vet ./...` and `gofmt -l` clean
- [x] New unit tests: `TestIsKnownInteractiveHostFrom` (deterministic allow-list decisions), `TestAllowsSession` (fail-open on no-PID, reject on unknown host, no-op for non-opted-in adapters), `TestSessionDetector_NewSession_{Skips,Admits}...` (integration-level: a CodexBar-shaped ghost never gets a `SessionState`; a real-host session is admitted immediately)
- [ ] Manual confirmation against the reporter's machine once they reply with an `irrlichd --diagnose` bundle (requested on #784)

### DIFF
--- a/core/adapters/inbound/agents/antigravity/agent.go
+++ b/core/adapters/inbound/agents/antigravity/agent.go
@@ -47,8 +47,9 @@ func Agent() agent.Agent {
 			IconSVGDark:  iconSVGDark,
 		},
 		Process: agent.Process{
-			Match:         agent.ExactName{Name: ProcessName},
-			PIDForSession: DiscoverPID,
+			Match:            agent.ExactName{Name: ProcessName},
+			PIDForSession:    DiscoverPID,
+			RequireKnownHost: true,
 		},
 		Source: agent.FilesUnderRoot{
 			Dir:               cliBrainDir,
--- a/core/adapters/inbound/agents/maps.go
+++ b/core/adapters/inbound/agents/maps.go
@@ -100,6 +100,21 @@ func ArgvExcluders(agents []agent.Agent) map[string]func([]string) bool {
 	return m
 }
 
+// RequireKnownHost produces the adapter-name → Process.RequireKnownHost map
+// consumed by session admission's host-ancestry gate: a candidate PID
+// launched by something other than a known terminal or IDE is rejected before
+// a session is ever created (#784). Only adapters that opt in (today:
+// antigravity) appear; the rest get no entry, so admission is unaffected.
+func RequireKnownHost(agents []agent.Agent) map[string]bool {
+	m := make(map[string]bool)
+	for _, a := range agents {
+		if a.Process.RequireKnownHost {
+			m[a.Identity.Name] = true
+		}
+	}
+	return m
+}
+
 // ControlPresets produces the adapter-name → (preset id → command text) map
 // consumed by the BackchannelEngine to translate a rule's agent-agnostic preset
 // into the concrete command for the session's agent (issue #754). Only adapters
--- a/core/adapters/inbound/agents/processlifecycle/hosts_darwin.go
+++ b/core/adapters/inbound/agents/processlifecycle/hosts_darwin.go
@@ -64,6 +64,18 @@ func termProgramForAppPath(cmdPath string) string {
 	return termProgramByAppName[appName]
 }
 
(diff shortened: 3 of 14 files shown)


ingo-eichhorst:
Pushed fixes for the three correctness bugs found by an xhigh-effort review of the host-ancestry gate:

1. **Gate was a structural no-op for the primary path** — `fswatcher`'s `EventNewSession` never sets `Event.CWD`, so `DiscoverPID`'s cwd-based lookup always short-circuited to `(0, nil)` and `AllowsSession` failed open before ever attempting an ancestry check. Now falls back to the transcript-derived cwd (`GetCWDFromTranscript`, mirroring the existing #576 stale-rescue pattern).
2. **Disambiguator mismatch** — `AllowsSession` passed `nil` (falls back to lowest PID for antigravity), while the real binding path (`TryDiscoverPID`) prefers the highest *unclaimed* PID. A stale PID sharing a cwd with a legitimate new process could get ancestry-checked instead of the real one. Extracted a shared `claimAwareDisambiguate` helper used by both.
3. **Debounce-coalesce bypass** — a rejected session could slip through via the coalesced-activity path, which re-enters admission with an empty adapter identity (`AllowsSession("", ...)` no-ops). Added a small rejection cache checked before the identity-dependent gate.

Also tightened `IsKnownInteractiveHost` to short-circuit once the curated-terminal ancestry walk matches, instead of always running both walks.

Regression tests for all three (each confirmed to fail against the pre-fix code) plus the full suite (`go test ./core/... -race -count=1`) are green.

Issue #750, opened 2026-06-27
[feat]: History tab Phase 2 — branch/worktree + provider/model attribution & drilldowns


ingo-eichhorst:
## Goal

Phase 2 of the History tab (#369). Phase 1 shipped `GET /api/v1/history?chart=cost&group=project` with trailing/calendar/custom ranges + linear forecast, and a History tab in `platforms/web/` (one stacked-area chart + one side panel). Phase 2 extends attribution to **branch/worktree, provider, and model**, wires the remaining chart types and group axes, and adds **drilldowns**.

Follow-up to #369; the Phase 1 PR settled the endpoint shape and the `501 {phase:2}` stubs this issue lights up.

## Design constraint (unchanged)

**One chart + one side panel, ever.** Chart-type and Group are toggles that swap data in place. Drilldowns *replace* the view (with a breadcrumb), never stack a second chart or table.

### Snapshot schema
- Extend the cost snapshot row (`core/adapters/outbound/filesystem/cost_tracker.go`, `snapshotRow`) with `branch` and `model`.
  - **`provider` already exists** on the row and in the per-provider rollup (`CostsInWindows` byProvider) — so `chart=providers` / `group=provider` is mostly a wiring job, not a schema change.
- Bump a schema version; old rows read back as unknown for the new columns (mirrors how empty `provider` is handled today).
- No migration / daemon-restart requirement.

### Attribution
- **Branch/worktree**: worktrees are sub-buckets under their owning project; detached HEAD → `(detached HEAD)` per project. Resolve via the existing git plumbing (`metadata_enricher.go` / git adapter).
- **Cost across mid-flight changes**: a snapshot delta is attributed to the `(branch, model, provider)` active at the **end** of the snapshot interval — never assign a whole session's cost to one branch if it changed mid-session. Extend the per-session baseline/delta model already in `CostSeries`/`scanWindows` to key by the triple, not just project.
- **Unknown/missing**: show an "unknown" bucket only when ≥10% of cells in the window are unknown; otherwise drop the column. No fake zeros.

### Endpoint
- Wire `chart=tokens|models|providers` and `group=branch|provider|model|session` in `handleGetHistory` (currently `501 {phase:2}`).
  - `chart=tokens` returns token series (in/out/cache split) reusing the `cum_*` columns already on the row + `UsageBreakdown` cache 5m/1h semantics (#171).
  - `chart=models` / `chart=providers` reuse the per-model / per-provider ledgers.
[...]

PR #793, opened 2026-07-03
fix(daemon): History Group=Model bucket everything under "unknown"

### LINKED ISSUE #792: History Group=Model shows only "unknown" — cost rows never get a real model name
When grouping the History Usage-tab chart by Model, every cost row falls into a single "unknown" bucket, even when a variety of models were actually used.

## Root cause

`core/adapters/outbound/filesystem/cost_tracker.go:141` and `:219` (`RecordSnapshot` / `RecordBaseline`) record each cost row's `Model` field from `state.Model` (`session.SessionState.Model`). That field is **never assigned anywhere in production code** — only in a test fixture (`core/internal/contracttesting/fixtures.go:63`). It's always `""` on real sessions.

The field that's actually populated live — and what the session list/statusline correctly display — is `state.Metrics.ModelName` (`core/domain/session/session.go:46`, updated live via `session.go:829/890-891`).

Since every persisted cost row ends up with an empty `Model`, `core/cmd/irrlichd/handlers.go`'s `resolveUnknownBucket` (~line 749) relabels the empty group key as `"unknown"` — and because *every* row is empty, that bucket is 100% of the total for every window, so it's never dropped by the 10%-share threshold.

The already-correct fallback pattern exists elsewhere in the same file, `handlers.go:1071-1073` (session-list handler):
```go
model := s.Model
if s.Metrics != nil && s.Metrics.ModelName != "" && s.Metrics.ModelName != "unknown" {
    model = s.Metrics.ModelName
}
```
`cost_tracker.go` never got the equivalent fallback.

## When introduced

Not a regression and not stale historical data — introduced ~2026-06-27 in `2a9fa997` ("feat(history): Phase 2 — branch/worktree + provider/model attribution & drilldowns (#750) (#771)"), which added the `Model` column wired to the wrong source field from day one. Every session's cost has landed under "unknown" for the model axis since that PR shipped.

## Fix

Mirror the `handlers.go:1071-1073` fallback in `cost_tracker.go`'s two `snapshotRow` constructions (`RecordSnapshot` and `RecordBaseline`), reading `state.Metrics.ModelName` (guarding empty/`"unknown"`) instead of the never-set `state.Model`.

ingo-eichhorst:
## Summary
- `RecordSnapshot`/`RecordBaseline` in `cost_tracker.go` stamped each cost row's `Model` field from `SessionState.Model`, a field that's never assigned anywhere in production code (only in a test fixture) — always `""`. Every row landed in the aggregate "unknown" bucket, so grouping History's Usage tab by Model always showed a single "unknown" entry no matter which models were actually used.
- Fall back to `Metrics.ModelName` (guarding empty and the `"unknown"` sentinel), mirroring the pattern already used correctly in the session-list handler (`handlers.go:1071-1073`).
- Introduced ~2026-06-27 in #750/#771 (branch/worktree + provider/model attribution) — the `Model` column was wired to the wrong source field from day one, not a regression and not a data-quality issue with pre-existing history.

Fixes #792

## Test plan
- [x] Added `TestRecordSnapshot_ModelFromMetrics` — confirmed it fails without the fix (`Model` empty) and passes with it
- [x] Added `TestRecordSnapshot_ModelUnknownSentinelFallsBackEmpty` — the `"unknown"` sentinel doesn't get persisted verbatim, keeps bucketing under the aggregate unknown group
- [x] `go test ./core/adapters/outbound/filesystem/... -v` — all pass, including the pre-existing `TestRecordSnapshot_StampsBranchAndModel` (still valid, exercises the `SessionState.Model` fallback path)
- [x] `go test ./core/... -race -count=1` — full suite green

### DIFF
--- a/core/adapters/outbound/filesystem/cost_tracker.go
+++ b/core/adapters/outbound/filesystem/cost_tracker.go
@@ -29,7 +29,10 @@ const (
 	// JSONL with omitempty, so pre-v2 rows simply decode with empty branch and
 	// model (read back as "unknown" for those group axes) — no migration and
 	// no daemon-restart requirement. The constant is documentary; rows do not
-	// carry it.
+	// carry it. Note: rows written between #750 and #792's fix are also
+	// schema-v2 but have an empty model for a different reason (a dead source
+	// field, not a pre-v2 row) — no backfill for those either, they age out
+	// under normal retention.
 	costSchemaVersion = 2
 )
 
@@ -119,10 +122,23 @@ func (t *CostTracker) SetProviderResolver(fn func(*session.SessionState) string)
 	}
 }
 
-// RecordSnapshot appends a row for the session if cost or any cumulative
-// token count has changed since the last stored row, and at least
-// costWriteInterval has elapsed since that row. No-ops on sessions without
-// metrics or a project name — nothing useful to store.
+// modelForRow resolves a session's model name for a cost row: prefer the
+// live Metrics.ModelName (populated by the adapters/tailer as turns stream
+// in), falling back to the SessionState.Model field when ModelName is empty
+// or the "unknown" sentinel. Mirrors the equivalent resolution in the
+// session-list handler (handlers.go's /state endpoint) — shared here between
+// RecordSnapshot and RecordBaseline, the two callers within this file.
+func modelForRow(state *session.SessionState) string {
+	if state.Metrics.ModelName != "" && state.Metrics.ModelName != "unknown" {
+		return state.Metrics.ModelName
+	}
+	return state.Model
+}
+
+// RecordSnapshot appends a row for the session if cost, any cumulative token
+// count, or the resolved model has changed since the last stored row, and at
+// least costWriteInterval has elapsed since that row. No-ops on sessions
+// without metrics or a project name — nothing useful to store.
 func (t *CostTracker) RecordSnapshot(state *session.SessionState) error {
 	if state == nil || state.Metrics == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


ingo-eichhorst:
Ran a review pass and addressed the findings in 149de97e:

- **Model change silently dropped forever** (confirmed, most severe) — `RecordSnapshot`'s dedup check compared cost/token counters but not `Model`, so a session whose model resolved (or changed) with no coincident cost/token delta got that write skipped — and since the skip happens before `lastWrite` updates, the stale model persisted indefinitely for any session that never produces another delta (verified this is common, not rare: several adapters emit model-announcement events with zero usage, on a separate pass from the token-bearing turn). Added `Model` to the comparison.
- **Duplicated fallback logic** — extracted the identical inline snippet in `RecordSnapshot`/`RecordBaseline` into one `modelForRow` helper.
- **Test-coverage gap** — `RecordBaseline` has its own call site but wasn't exercised by the model-fallback tests (they only called `RecordSnapshot`); added a dedicated test, plus a regression test for the dedup fix.
- **Stale doc comment** — noted that schema-v2 rows written before this fix are empty-model for a different reason than genuine pre-v2 rows.

Noted but intentionally not chased further (out of scope for this fix):
- No backfill for existing rows — sessions recorded before this PR keep an empty model forever; they age out under the existing retention window rather than being migrated.
- The same "prefer ModelName, filter the unknown sentinel, fall back to Model" logic now exists in a third place (`handlers.go`'s session-list handler) and a fourth, subtly different place (the macOS Swift client's `effectiveModel`, which is missing the sentinel filter) — worth a follow-up to unify into one shared accessor on `SessionState`, but that's a cross-file/cross-platform refactor beyond this bug fix.

Full suite (`go test ./core/... -race -count=1`) green.

Issue #832, opened 2026-07-04
macOS: SessionRowSnapshotTests' antigravity-icon snapshots are non-hermetic (live daemon dependency + init-time race)


ingo-eichhorst:
## Summary

`swift test --filter SessionRowSnapshotTests` currently fails 2 of 21 tests on a clean `main` checkout:

```
testFixture_AntigravityGhost      — Snapshot does not match reference
testGhostRow_PID0_NilMetrics      — Snapshot does not match reference
```

Confirmed pre-existing and unrelated to #813/PR #826 (same 2 failures reproduce identically on `origin/main` before that PR's changes).

## Root cause

Both fixtures use `adapter: "antigravity"`. Comparing the committed reference PNG to a freshly-recorded one, the only visual difference is the adapter icon on the right edge of the row: the reference shows the antigravity brand icon (a colored arc glyph), the fresh render shows a generic gray "?" placeholder — i.e. `session.adapterIcon` resolves to `nil` now but didn't when the snapshot was captured.

That icon comes from `AgentRegistry`, populated by `SessionManager.hydrateAgents()` (`platforms/macos/Irrlicht/Managers/SessionManager+Hydration.swift:25-32`), which does a **real HTTP GET** to:

```swift
guard let url = URL(string: "\(DaemonEndpoint.httpBase)/api/v1/agents") else { return }
```

`DaemonEndpoint.httpBase` (`platforms/macos/Irrlicht/Managers/DaemonEndpoint.swift:39`) defaults to `http://127.0.0.1:7837` — the **production daemon port** — unless `IRRLICHT_DAEMON_PORT` is set. `SessionRowSnapshotTests.setUp()` constructs a real `SessionManager()`, whose `init` (`SessionManager.swift:257-263`) fires off hydration in an unstructured `Task`, with no way for the synchronous test body / `assertSnapshot` call to await it.

So these two tests' pass/fail status depends on:
1. Whether a real daemon happens to be running on `127.0.0.1:7837` (or whatever `IRRLICHT_DAEMON_PORT` resolves to) at test time.
2. Whether that daemon's `/api/v1/agents` response currently includes `antigravity` branding with icon data (varies by daemon build/state).
3. A timing race between the fire-and-forget hydration `Task` and the synchronous `assertSnapshot(...)` call — the icon can be missing simply because the network round-trip hasn't completed yet when the snapshot is taken.

[...]

PR #845, opened 2026-07-04
fix(macos): recover from a stuck local-daemon reconnect

### LINKED ISSUE #843: macOS app never recovers when the daemon it's connected to dies and a new one starts on the same port
## What happened

During a `/ir:test-mac` (replace mode) session, the daemon backing the running dev app was killed externally by a partial/interrupted restart flow. Starting a fresh, fully healthy `irrlichd` on the same port (7837) did **not** bring the already-running app back: every network call kept failing in a ~1s retry loop indefinitely —

```
💧 Agent branding hydration failed: Could not connect to the server.
💧 Hydration failed: Could not connect to the server.
🔐 Permissions refresh failed: Could not connect to the server.
🔌 WebSocket connected to irrlichd
🔌 WebSocket disconnected: Could not connect to the server.
🔌 Reconnecting in 1.1s
```

— even though `curl` to the exact same host/port/path (`ws://127.0.0.1:7837/api/v1/sessions/stream`) succeeded instantly with a clean `101 Switching Protocols` and live session data. Only killing and relaunching the app process itself could have resolved it (a separate bug prevented even that — see the companion issue on the dev app's resource bundle).

## Where

- `platforms/macos/Irrlicht/Managers/SessionManager+WebSocket.swift` — `connect()`/`scheduleConnect()`, reconnect loop starting `reconnectDelay` at 1.0s
- `platforms/macos/Irrlicht/Managers/DaemonManager.swift` — `scheduleRestart()` only fires when the app itself owns/spawned the daemon process; in `/ir:test-mac` replace mode the app never owns it (it *adopts* an externally-started daemon per SKILL.md step 7), so this restart logic never engages
- `platforms/macos/Irrlicht/Managers/DaemonEndpoint.swift`

## Impact

Any daemon restart/crash while the app is running — including the documented `/ir:test-mac` replace-mode workflow — silently leaves the menu bar app permanently stuck showing stale data, with no visible indication that reconnection has failed for good, until the user manually quits/relaunches the app.

## Suggested fix

Investigate why `URLSession.shared`-based requests keep failing after many minutes of consecutive failures against a now-healthy port — possibly stuck/cached URLSession task state, or a Network.framework path-monitor treating the host as unreachable. Consider:
- using a fresh `URLSession` per reconnect attempt (or recycling it after N consecutive failures), and/or
- surfacing a "disconnected — click to reconnect" UI state instead of retrying silently forever.

May be related to #833 (componentized `/ir:test-mac` restart).

ingo-eichhorst:
## Summary
- Fixes #843: the menu bar app never recovered when the daemon it was connected to died and a new one started on the same port — only killing and relaunching the app itself worked.
- Root cause was two compounding bugs in `SessionManager+WebSocket.swift`'s `connect()`: (1) `reconnectDelay` was reset to 1.0 on every attempt — including failed ones — before the exponential backoff could ever take effect, so a dead daemon got hammered every ~1.1s forever instead of backing off; (2) all local-daemon traffic rode `URLSession.shared`, which can wedge against a host:port that restarted under it and then fail forever even once a healthy daemon is listening again.
- Fix: move all local-daemon HTTP + WebSocket traffic onto a dedicated `URLSession` that gets recreated after 3 consecutive silent-failure cycles, only reset the backoff/failure streak once a message actually confirms the connection (mirrors the existing `relayConnect()` pattern), and surface a "daemon unreachable — retrying" state in the connection tooltip/dot instead of quietly retrying forever with no visible indication.

## Test plan
- [x] `swift build` — compiles clean
- [x] `swift test` — full suite passes, including new `SessionManagerReconnectTests` pinning the backoff-reset and session-recycle state transitions
- [x] `tools/preflight.sh` — full local CI parity (go tests, onboarding-factory, replaydata validate, web suites, ARS gate) — all green
- [ ] Manual: not verified live against a real daemon-restart scenario (no GUI session available in this environment) — the fix is scoped to the state-machine logic described in the issue, unit-tested in isolation since `connect()` itself does real networking (existing tests never dial the real daemon, see #832)

### DIFF
--- a/platforms/macos/Irrlicht/Managers/SessionManager+Hydration.swift
+++ b/platforms/macos/Irrlicht/Managers/SessionManager+Hydration.swift
@@ -25,7 +25,7 @@ extension SessionManager {
     func hydrateAgents() async {
         guard let url = URL(string: "\(DaemonEndpoint.httpBase)/api/v1/agents") else { return }
         do {
-            let (data, response) = try await URLSession.shared.data(from: url)
+            let (data, response) = try await localURLSession.data(from: url)
             guard (response as? HTTPURLResponse)?.statusCode == 200 else { return }
             let entries = try JSONDecoder().decode([AgentBranding].self, from: data)
             AgentRegistry.byName = Dictionary(uniqueKeysWithValues: entries.map { ($0.name, $0) })
@@ -41,7 +41,7 @@ extension SessionManager {
     func refreshPermissions() async {
         guard let url = URL(string: "\(DaemonEndpoint.httpBase)/api/v1/permissions") else { return }
         do {
-            let (data, response) = try await URLSession.shared.data(from: url)
+            let (data, response) = try await localURLSession.data(from: url)
             guard (response as? HTTPURLResponse)?.statusCode == 200 else { return }
             permissionsSnapshot = try JSONDecoder().decode(PermissionsSnapshot.self, from: data)
         } catch {
@@ -65,7 +65,7 @@ extension SessionManager {
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
         do {
             request.httpBody = try JSONEncoder().encode(["answers": answers])
-            let (data, response) = try await URLSession.shared.data(for: request)
+            let (data, response) = try await localURLSession.data(for: request)
             guard (response as? HTTPURLResponse)?.statusCode == 200 else { return false }
             permissionsSnapshot = try JSONDecoder().decode(PermissionsSnapshot.self, from: data)
             return true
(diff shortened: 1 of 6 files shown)


ingo-eichhorst:
Ran an xhigh code-review pass and fixed the two real regressions it found in this diff:

- **Idle daemon never confirms**: gating confirmation on receiving an application message meant a daemon with zero tracked sessions could go an entire reconnect cycle silent, permanently stuck showing "Connecting…" and eventually mis-flagged as stalled. Fixed by also confirming via a WebSocket ping/pong sent right after `resume()`.
- **Local stall masking a healthy relay**: `statusColor`'s new override painted the dot red even when a relay connection was carrying the session list fine. Fixed to only apply when the aggregate connection state isn't otherwise `.connected`.

Two other findings from that pass (a cache-bloat tooltip note and a README dark-mode chart change) turned out to belong to unrelated ambient changes outside this branch's diff — verified via `git diff origin/main...HEAD --name-only` and left untouched.

Filed #846 to track generalizing the URLSession-recycle fix to the relay reconnect path, which has the same structural issue but is out of scope for this PR.

PR #899, opened 2026-07-07
fix(daemon): hold parent working while a background child is active during turn-done waiting

### LINKED ISSUE #897: Session shows "waiting" while a background Agent-tool child is still running
## Summary

A parent session whose turn ends with prose about waiting on a **background subagent** (not the user) gets classified `waiting` instead of `working` — even though a child session is still actively running. The dashboard/menu-bar then reads as "idle, nothing to see" while real work is in progress.

Reproduced live on session `cc24c883-78f7-4657-bfd5-3b5c405b17c3` (worktree `issue-895-sonarqube-cleanup`). Via `/api/v1/sessions`:

```json
"state": "waiting",
"subagents": {"total": 1, "working": 1, "waiting": 0, "ready": 0},
"metrics": {
  "last_assistant_text": "The Go agent is doing significant interface-renaming work across many files — I'll wait for its completion notification before touching any Go/JS files myself, to avoid conflicts."
}
```

## Root cause 1 — waiting-cue regex false positive

`core/domain/session/waiting_cue.go:112` has cue pattern `\bI'?ll wait\b` (category "C. Action gates", added for #381) to catch phrasing like *"I'll wait for your review"* — i.e. waiting on the **human**. It also fires on *"I'll wait for its completion notification"*, which is about a background subagent, not the user.

`IsWaitingForUserInput()` → `true` → `ClassifyState` rule 2a (`core/application/services/state_classifier.go:74-82`) routes to `StateWaiting` instead of `StateReady`/`StateWorking`.

## Root cause 2 — architectural gap (not just the regex)

In `core/application/services/session_detector_activity.go:620-638`, the parent-hold logic is asymmetric:

- Path to `StateReady` → calls `hasActiveChildren()` and holds the parent `working` if any child is still active (see `reevaluateParent` / `hasActiveChildren` in `session_detector_subagent.go`).
- Path to `StateWaiting` → only calls `finishOrphanedChildren()` (fast-forwards *stale* children); it never checks `hasActiveChildren()`. The comment states the assumption outright: *"the parent legitimately waits on the user"*.

So even a textbook, correctly-detected "waiting for the human" classification never reflects "N background children are still running" at the top-level `state` — that only lives in the separate `Subagents` summary field, which is refreshed unconditionally regardless of which branch fired.

## Ground-truth on this specific session

The parent launched 4 background `Agent` calls (~22:50 UTC): shell, JS, Go, Web/CSS. Cross-referencing `events.log` and subagent transcript mtimes at the time of the report:

- Shell — finished, task-notification delivered ~23:19 UTC (correctly woke the parent).
- Web/CSS (`agent-a3ac23539e9810425`) — finished + reaped at 01:08:28 local.
- Go (`agent-aa84cfbb2323264c9`) — finished + reaped at 01:33:06 local (transcript frozen since).
- JS (`agent-ac91a4dd5d0ff53c5`) — still actively growing (confirmed via two mtime/size checks a minute apart), correctly tracked as the one `working` child.

So only 1 background agent was actually still running at report time, not 2 as initially reported — the discrepancy is stale info from the parent's own earlier status update ("4 background agents finishing up...") before Go and Web/CSS had actually finished. The underlying state-classification bug is independent of that count and reproduces with just the one running JS child.

## UX consequence

- `platforms/macos/Irrlicht/Views/SessionStateIcon.swift` renders `.waiting` as a pause glyph — visually indistinguishable from "idle, needs your input".
- `platforms/macos/Irrlicht/Views/MenuBarStatusRenderer.swift` buckets the aggregate menu-bar ring purely by each session's top-level `state`, with no children-awareness.

Both the per-row icon and the aggregate menu-bar badge read as "nothing happening" while a subagent is actively burning tokens and making progress.

## Suggested direction (not yet implemented)

- Extend the `StateWaiting` branch in `processActivity` to also consult `hasActiveChildren()`, mirroring the `StateReady` branch — if active children remain, hold `working` (or introduce a distinct signal) instead of collapsing straight to `waiting`.
- Separately/additionally, tighten the `I'll wait` cue (and similar "action gate" patterns) to require a human-directed object (e.g. "I'll wait for your/you to ...") so it doesn't fire on "I'll wait for the/its/a background process/agent/notification ...".

ingo-eichhorst:
## Summary

- A parent session whose turn ends in a waiting cue (e.g. *"I'll wait for its completion notification..."*) collapsed straight to `waiting` even when a background `Agent`-tool child it had kicked off was still genuinely running — the waiting branch in `processActivity` never checked `hasActiveChildren`, unlike the ready-transition branch just above it. Fixed by sharing a new `holdIfChildrenActive` helper (`session_detector_subagent.go`) between both branches, so the waiting path now holds the parent `working` when a child is genuinely still active, mirroring the existing ready-path behavior.
- The `I'll wait` waiting-cue regex (`waiting_cue.go`) is tightened via a new exclusion list (`its`/`their`/`it` as the wait object) instead of narrowing the inclusion pattern — narrowing it to require a literal `for you`/`for your` would have dropped detection of genuine human-directed waits phrased without that exact object (caught by an xhigh code-review pass on the first draft of this fix).

Closes #897.

## Test plan

- [x] `go test ./core/domain/session/... -race -count=1` — new/updated `TestExtractWaitingCue` / `TestIsWaitingForUserInput_ImperativeCues` cases pass, including the `#897` false-positive and the exclusion-list regression cases.
- [x] `go test ./core/application/services/... -race -count=1` — updated `TestSessionDetector_WaitingParent_HeldWorking_WhenBackgroundChildActive` plus the full `TestSessionDetector_*` suite pass; the two adjacent tests (orphaned children, permission-prompt waiting) are unaffected as expected.
- [x] `go test ./core/... -race -count=1` — full suite green (one pre-existing `-race` timing flake in `TestDebounce_*`, confirmed unrelated: passes reliably in isolation/serially and touches no file in this diff).
- [x] `tools/preflight.sh` (via the pre-push hook) — gofmt, core tests, onboarding-factory tests, replaydata validate, recording-rig smoke test, both web suites, ARS architecture gate, and the security scan all pass.

### DIFF
--- a/core/application/services/session_detector_activity.go
+++ b/core/application/services/session_detector_activity.go
@@ -626,27 +626,24 @@ func (d *SessionDetector) processActivityLocked(id agent.Identity, state *sessio
 		// cleanupChildren either way (finishOrphanedChildren's quiet-window
 		// and open-tool guards already leave a truly running child alone).
 		parentHeldWorking := false
+		holdWorkingIfChildrenActive := func(logMsg string) {
+			if !d.holdIfChildrenActive(state.SessionID) {
+				return
+			}
+			d.log.LogInfo(logComponentSessionDetector, ev.SessionID, logMsg)
+			newState = session.StateWorking
+			reason = ""
+			parentHeldWorking = true
+		}
 		if state.ParentSessionID == "" {
 			switch {
 			case newState == session.StateReady:
-				if d.holdIfChildrenActive(state.SessionID) {
-					d.log.LogInfo(logComponentSessionDetector, ev.SessionID,
-						"holding parent working — active children still running")
-					newState = session.StateWorking
-					reason = ""
-					parentHeldWorking = true
-				}
+				holdWorkingIfChildrenActive("holding parent working — active children still running")
 			case newState == session.StateWaiting && state.Metrics != nil && state.Metrics.IsAgentDone():
 				// Turn-done waiting (question/cue). Permission-prompt
 				// waiting never reaches here: its open tool call makes
 				// IsAgentDone return false.
-				if d.holdIfChildrenActive(state.SessionID) {
-					d.log.LogInfo(logComponentSessionDetector, ev.SessionID,
-						"holding parent working — active children still running (turn ended in waiting cue)")
-					newState = session.StateWorking
-					reason = ""
-					parentHeldWorking = true
-				}
+				holdWorkingIfChildrenActive("holding parent working — active children still running (turn ended in waiting cue)")
 			}
 		}
 
--- a/core/application/services/session_detector_activity.go
+++ b/core/application/services/session_detector_activity.go
@@ -612,17 +612,24 @@ func (d *SessionDetector) processActivityLocked(id agent.Identity, state *sessio
 		// question/cue (#593) — IsAgentDone gates it, so permission-prompt
 		// waiting (open tool call) is excluded. Without this, a parent whose
 		// overnight run ends by asking a question leaves its finished
-		// children stuck in working until the liveness sweep. The waiting
-		// branch only fast-forwards: no hold (the parent legitimately waits
-		// on the user) and no cleanupChildren (background children may still
-		// be running; finishOrphanedChildren's quiet-window and open-tool
-		// guards leave those alone).
+		// children stuck in working until the liveness sweep.
+		//
+		// The waiting branch also holds the parent working when a genuinely
+		// active child remains (#897) — mirroring the ready branch below.
+		// A turn-ending cue like "I'll wait for its completion notification"
+		// can read as a question/cue to the classifier while actually
+		// describing a wait on the parent's own background subagent, not
+		// the user; surfacing "waiting" in that case reads as "nothing
+		// happening" on the dashboard when a subagent is still working.
+		// finishOrphanedChildren runs first so a child that's merely stale
+		// (not genuinely active) doesn't hold the parent forever; no
+		// cleanupChildren either way (finishOrphanedChildren's quiet-window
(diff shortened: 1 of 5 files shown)


ingo-eichhorst:
**CI status**: all required gates green (go-test, Linux build-test — one transient relay-capacity flake on the first run, confirmed unrelated and cleared on re-run — ars-gate, CodeQL, both web-test suites). Not chasing `SonarCloud Code Analysis` further; here's why it's still red and why that's expected:

Per-file new-code duplication breakdown (`new_duplicated_lines` / `new_lines` via the SonarCloud API):

| file | new lines | new duplicated lines |
|---|---|---|
| `session_detector_activity.go` | 26 | 0 |
| `session_detector_subagent.go` | 11 | 0 |
| `waiting_cue.go` | 24 | 0 |
| `session_detector_subagent_test.go` | 11 | 2 |
| `waiting_cue_test.go` | 10 | **10** |

The real duplication the code-review pass flagged (the near-identical ready/waiting hold blocks in `session_detector_activity.go`) is fixed — 0 new duplicated lines there and in the other non-test files.

The remaining 12 lines are Sonar's CPD flagging new rows added to existing table-driven test tables (`cases := []struct{...}{ {...}, {...}, ... }`) as duplicates of each other/neighboring rows — a well-known false-positive pattern for this test style (every `{"name", "text", bool}` literal row normalizes to the same token shape). Restructuring idiomatic table-driven Go tests to dodge CPD would be a worse trade than the metric it's chasing. This check isn't in the repo's required-status-checks ruleset, consistent with how ARS/CodeScene are already treated here — advisory, not a merge gate.

PR #917, opened 2026-07-08
feat(macos): show subscription quota in the menu bar status icon

### LINKED ISSUE #909: [feat]: Show 5h/7d quota usage directly in the menu bar status icon
## What problem are you trying to solve?

Irrlicht already tracks the 5h/7d subscription rate-limit windows (shipped in v0.4.6, #309/#379) and shows them as a quota chip inside the popover — but only after clicking the status item. The icon itself only ever shows session-state dots (or the idle flame / attention badge). Users who want an always-visible sense of "how much of my 5h window is left" currently run a second menu-bar app (e.g. Claude Usage Tracker) just to get that at-a-glance icon, even though Irrlicht's daemon already has the exact same data — cleanly, via Claude Code's own `statusLine` hook, not by probing undocumented API response headers.

## Proposed solution

Image

Add an optional status-item rendering mode that shows the most-imminent rate-limit window's usage directly in the icon (ring, thin progress bar, or percentage text), composed the same way the existing Gas Town rig badge is appended alongside the session dots in `MenuBarImageBuilder`/`MenuBarStatusRenderer`. A small icon-style picker in Settings (percentage / bar / ring / off) would match the mental model users of similar tools already have — Claude Usage Tracker's icon-style picker (Battery, Progress Bar, Percentage, Icon+Bar, Compact) is a useful reference for the interaction, not the visual style. Default should stay as today's dots-only icon (opt-in), since Irrlicht already composes multiple elements (per-project dot groups + Gas Town badge) and quota should not force extra clutter on users who don't want it.

No new data collection needed — this is rendering-only; `RateLimitSnapshot` / `RateLimitInfo` already flow from daemon to macOS app.

## Alternatives considered

Keep using a separate app for the icon-level indicator — works today, but means two menu-bar apps tracking overlapping data, and the separate app's approach (a live Messages API probe to read undocumented `anthropic-ratelimit-*` response headers) is more fragile than Irrlicht's existing statusLine-hook ingestion.

## Who benefits from this?

Single-agent users, multi-agent users who want quota visible without opening the popover.

## How does this fit the philosophy?

Zero-config preserved — the data is already collected automatically via the existing statusLine hook. This only adds an opt-in rendering toggle; default behavior is unchanged.

dernerl:
## Summary

Implements #909: an opt-in Settings picker (**Lights / Usage / Combined**) so the menu-bar status icon can show the 5h/7d subscription-quota windows directly, instead of only inside the popover.

image

- **Lights** — today's behavior, unchanged (default, so existing users see no visual change).
- **Usage** — replaces the session-state dots with stacked 5h/7d quota bars (or a compact ring, see below).
- **Combined** — both side by side: dots left, quota right (closest to the system status icons), matching the mockup posted on the issue.

Additional controls, per your comment on the issue:
- **Quota shape**: Bars (5h/7d stacked) or a single compact ring pinned to the 5h window.
- **Quota provider**: a single-provider picker (not multi-provider) for accounts with more than one subscription (e.g. Claude + Codex) — deliberately not showing every provider at once, to stay within the icon's tight width budget alongside the existing capped-at-5 dot groups.
- **Pace marker**: both shapes carry a thin marker showing how much wall-clock time has elapsed in the window (mirrors `SessionListView.quotaPacePercent`), independent of usage%.

No new data collection — this is rendering-only. The 5h/7d windows were already flowing from the daemon's `statusLine` hook into `session.metrics.rateLimit`; this just adds a second consumer of that existing data.

### Two commits

1. `feat(macos): show subscription quota in the menu bar status icon` — the feature itself.
2. `fix(macos): correct quota-icon rebuild triggers, staleness, and appearance` — a set of real defects an independent second-pass review caught before this reached anyone else: the icon not repainting when the new settings changed, a stale-snapshot handling mismatch with the popover, an absolute-only color ramp that could disagree with the popover's pace-aware one, hardcoded colors invisible in light menu-bar appearance, and a couple of related edge cases. Left as a separate commit so the review trail is visible rather than squashed away.

### Scope note

I deliberately did **not** extract `SessionListView`'s fuller quota-chip bucketing (cost totals, stale-dimming, multi-provider chips) into a shared helper. `QuotaMenuBarRenderer`'s selection logic is intentionally leaner (single provider, no cost aggregation) — pulling popover-only concerns into the status-item icon felt like premature abstraction for what's currently a single consumer. Happy to revisit if you'd rather see that unified.

## Test plan

- 21 new/updated XCTest cases across `QuotaMenuBarRendererTests.swift` (bars/circle rendering, pace-aware color ramp mirroring `QuotaChipBarColorTests`, freshest-renderable snapshot selection) and `MenuBarImageBuilderTests.swift` (image composition geometry, `.usage`-style dots fallback).
- Full local suite (`swift test`, `go test ./core/... -race -count=1`, `tools/preflight.sh`) green, except one pre-existing environment-dependent Swift test failure (`SessionManagerTests.testTitleMatchScoreSkipsGenericTopsAndHomeBasename`) verified to also fail on a clean `main` checkout on this machine — unrelated to this change.
- Manually exercised all three styles + both quota shapes + the provider picker in a locally built dev instance (isolated daemon port, synthetic rate-limit data injected via the statusline endpoint) — confirmed the icon now repaints immediately on a settings change, and that bars/ring/labels are readable in both light and dark menu-bar appearance.
- No screenshots in this PR (local environment couldn't grant screen-recording permission to capture them) — happy to add if useful, or a maintainer/reviewer can pull the branch and check locally.

Closes #909.

### DIFF
--- a/platforms/macos/Irrlicht/App/MenuBarController.swift
+++ b/platforms/macos/Irrlicht/App/MenuBarController.swift
@@ -139,6 +139,13 @@ final class MenuBarController: NSObject {
         // a real change — same shape as SessionManager.init's
         // sourcesSettingsChanged observer.
         let defaultsSignal = NotificationCenter.default.publisher(for: UserDefaults.didChangeNotification)
+            // Unlike NotificationCenter.addObserver(forName:object:queue: .main)
+            // (what SessionManager's sourcesSettingsChanged observer actually
+            // uses), a plain Combine publisher delivers on whatever thread
+            // posted the notification. Hop to main before the compactMap
+            // below mutates lastMenuBarStyle/etc — those are plain, non-atomic
+            // properties read/written elsewhere on the main thread.
+            .receive(on: RunLoop.main)
             .compactMap { [weak self] _ -> Void? in
                 guard let self else { return nil }
                 let style = UserDefaults.standard.string(forKey: MenuBarStyle.storageKey) ?? ""
--- a/platforms/macos/Irrlicht/App/MenuBarController.swift
+++ b/platforms/macos/Irrlicht/App/MenuBarController.swift
@@ -39,6 +39,14 @@ final class MenuBarController: NSObject {
     private var escapeMonitor: Any?
     private var resignObserver: NSObjectProtocol?
 
+    // Last-seen values for the menu-bar quota settings (issue #909), so the
+    // broad UserDefaults.didChangeNotification (fires for *any* default,
+    // mirroring the pattern SessionManager.init uses for sourcesSettingsChanged)
+    // only triggers a rebuild when one of these three actually changed.
+    private var lastMenuBarStyle = UserDefaults.standard.string(forKey: MenuBarStyle.storageKey) ?? ""
+    private var lastMenuBarQuotaProvider = UserDefaults.standard.string(forKey: MenuBarQuotaProvider.storageKey) ?? ""
+    private var lastMenuBarQuotaVisual = UserDefaults.standard.string(forKey: QuotaVisualStyle.storageKey) ?? ""
+
     private static let escapeKeyCode: UInt16 = 53
 
     init(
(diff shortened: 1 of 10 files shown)


ingo-eichhorst:
Very nice!
I'll merge later and release a new Version 0.5.6 this evening

PR #948, opened 2026-07-11
feat(tools): add DORA Change Failure Rate + MTTR (Phase 2)

### LINKED ISSUE #920: macOS: CO2 unit wraps to a second line in the session list row
## Current behavior

In CO2 display mode (session row, list view), the kilogram value wraps onto a second line, splitting the unit from the number — e.g. "16.15k" / "g" on separate lines.

## Root cause

`SessionMetrics.formattedCO2` in `platforms/macos/Irrlicht/Models/SessionState.swift:614-627` formats the kilogram branch with two decimal digits:

```swift
if grams < 1 { return String(format: "%.0fmg", grams * 1000) }
if grams < 1000 { return String(format: "%.1fg", grams) }
return String(format: "%.2fkg", grams / 1000)   // <-- inconsistent, one digit too many
```

The `g` and `mg` branches use 0–1 decimal digits, but the `kg` branch uses 2, producing strings like `16.15kg` (7 chars). The cost/CO2 slot in `SessionRowView.swift` is a fixed-width 36pt column (lines 298/323) sized for short cost strings like `$0.63`, and the `Text` in `costOrCO2Label` (`SessionRowView.swift:100-105`) has no `.lineLimit(1)`, so the extra-wide `kg` string wraps instead of clipping.

## Fix

Use one decimal digit for the `kg` branch too, matching the `g` branch:

```swift
return String(format: "%.1fkg", grams / 1000)
```

This shortens the string to e.g. `16.2kg` (6 chars), which fits the existing 36pt column. Worth also adding `.lineLimit(1)` to the `Text` in `costOrCO2Label` as defense-in-depth against future long values.

## Where

- `platforms/macos/Irrlicht/Models/SessionState.swift:626`
- `platforms/macos/Irrlicht/Views/SessionRowView.swift:100-105`

Follow-up to #829.

### LINKED ISSUE #936: DORA Phase 2: Change Failure Rate + MTTR

ingo-eichhorst:
## Summary
- Fixes #936: extends `tools/dora-metrics.sh` with Change Failure Rate and MTTR, combining the three candidate failure signals from #936's discussion (per the user's decision to combine rather than pick one):
  1. **Hotfix window** — release landed within `--hotfix-window-hours` (default 24) of the prior release.
  2. **Revert commit** — release ships a commit matching `^Revert` whose `This reverts commit <hash>.` trailer resolves to a commit shipped in an earlier release. A non-standard `revert:`-style commit with no trailer is tallied separately as "unresolved" rather than silently dropped or guessed at.
  3. **Bug-labeled issue** — release ships a commit referencing `Fixes/Closes/Resolves #NNN` for a `gh issue list --label bug` issue. Degrades gracefully (signal line shows `n/a`) when `gh` is unavailable/unauthenticated, matching Phase 1's existing convention.
- CFR = deduped union of flagged releases / releases in range. MTTR reported per-signal and combined, one sample per flagged instance (not deduped) since each is a distinct restore event.
- **Also fixes a pre-existing Phase 1 bug** found while extending the file: the Lead Time commit walk silently dropped the *last* commit of every per-tag range, because `git log`'s final line has no trailing newline and `while read` treats that as a failed read (classic missing-final-newline bug). Confirmed via a synthetic single-commit-per-tag repo (previously reported 0 commits; now reports correctly) and cross-checked against the real repo (901 → 935 commits over the same 180-day window, +34 ≈ 1 per tag walked).
- Refactored to dedupe: hoisted `median_of`/`mean_of`/`format_hours` and a shared `idx_in_set` membership helper so Lead Time and the new CFR/MTTR section share logic instead of each reimplementing it (caught in low-effort code review on this PR).

Branches on top of #943 (Phase 1, not yet merged) since this extends that file directly — base branch here is `worktree-issue-919-dora-metrics`, not `main`. Merge #943 first, or merge this into it.

## Test plan
- [x] `bash -n tools/dora-metrics.sh`
- [x] Ran against real repo history at multiple ranges (14/90/180 days), zero-release range, single-release range, `gh`-unavailable (restricted `PATH`)
- [x] Manually cross-checked one instance per signal against real data: a real bug-fix commit (`Fixes #920`, issue labeled "bug", shipped in v0.5.6) correctly flagged; the repo's one non-standard `revert:` commit (no trailer) correctly tallied as "unresolved" rather than silently dropped
- [x] Built a synthetic throwaway git repo to exercise the *resolved*-revert code path end to end (no real example exists in this repo's main history) — confirmed correct MTTR computation
- [x] Code-reviewed at low effort; fixed both findings (duplicate median/mean logic, duplicate membership-check helper) by hoisting shared helpers
- [x] `tools/preflight.sh` (via pre-push hook) — all gates pass

### DIFF
--- a/core/adapters/inbound/agents/processlifecycle/monitor_other.go
+++ b/core/adapters/inbound/agents/processlifecycle/monitor_other.go
@@ -20,7 +20,9 @@ func NewMonitor(handler exitHandler) (*pidMonitor, error) {
 
 func (m *pidMonitor) Watch(pid int, sessionID string) error { return nil }
 
-func (m *pidMonitor) Unwatch(pid int) {}
+func (m *pidMonitor) Unwatch(pid int) {
+	// no-op — this platform has no exit-signal primitive to unregister
+}
 
 // Run blocks until ctx is cancelled, matching the other implementations.
 func (m *pidMonitor) Run(ctx context.Context) error {
--- a/core/application/services/background_process_test.go
+++ b/core/application/services/background_process_test.go
@@ -129,6 +129,78 @@ func TestSessionDetector_BackgroundProcess_HoldsWorkingThenReady(t *testing.T) {
 	<-done
 }
 
+// A session already `ready` (e.g. a prior background job just finished and
+// settled it) that then shows a *fresh* background spawn on the very next
+// activity event must be pulled back to working and held there while that new
+// process is alive — not skip the liveness probe because the pass started
+// from `ready`. Reproduces issue #937: a retried `git push -u
+// run_in_background:true` launched right after the session had settled ready
+// was silently invisible to the liveness gate, and the session bounced
+// ready→working→ready in the same pass despite the push still running.
+func TestSessionDetector_BackgroundProcess_FreshSpawnFromReady_HoldsWorking(t *testing.T) {
+	const sid = "bg3"
+	const path = "/home/.claude/projects/-Users-test/bg3.jsonl"
+
+	metrics := &funcMetrics{fn: func(_, _ string) (*session.SessionMetrics, error) {
+		return &session.SessionMetrics{
+			LastEventType:            "turn_done",
+			BackgroundProcessCount:   1,
+			BackgroundProcessOutputs: []string{"/tmp/x/tasks/retry.output"},
+		}, nil
+	}}
+
+	tw := newMockAgentWatcher()
+	pw := newMockProcessWatcher()
+	repo := newMockRepo()
+	repo.states[sid] = &session.SessionState{
+		SessionID:      sid,
+		State:          session.StateReady,
+		TranscriptPath: path,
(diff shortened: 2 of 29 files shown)


ingo-eichhorst:
Closing along with #943 — same scope misunderstanding (dev-tooling script vs. product feature in the History tab). See the new tracking issue for the real scope.

Issue #951, opened 2026-07-11
Add DORA metrics to the History tab (macOS + web)


ingo-eichhorst:
## Summary

Supersedes #919 and #936, which implemented DORA metrics (deployment frequency, lead time for changes, change failure rate, MTTR) as a standalone `tools/dora-metrics.sh` maintainer script. That's not what was wanted — the actual ask is a **product feature**: DORA metrics shown in the History tab of both the macOS app and the web dashboard, computed by the daemon on request.

## Existing architecture to build on

- **macOS**: `platforms/macos/Irrlicht/Views/HistoryView.swift` — tabbed History panel (`.usage`, `.yield`, `.quota`). Fetches `GET /api/v1/history` via `URLSession`, decoded into `HistoryResponse`/`HistoryYieldResponse` (`platforms/macos/Irrlicht/Models/HistoryData.swift`).
- **Web**: `platforms/web/historyTab.js` — same concern, canvas charts, same `GET /api/v1/history?...` endpoint (`fetchHistory()`/`historyQuery()`).
- **Daemon**: routes registered in `core/cmd/irrlichd/startup.go` (`mux.HandleFunc("GET /api/v1/history", handleGetHistory(...))`), handler bodies in `core/cmd/irrlichd/handlers.go`.
- **Closest precedent**: the Yield feature (`.yield` tab, #373) — `buildYieldResponse` in handlers.go, backed by `core/adapters/outbound/git/adapter.go`'s `RevertedCommits()`. It's exactly the shape a DORA feature should follow: compute-on-request from live git state, zero persistence, no new database.
- `core/adapters/outbound/git/adapter.go` already shells out to git (`GetBranch`, `GetHeadCommit`, `RevertedCommits`, etc.) via a `ports/outbound` interface — extend this (add tag/release listing, commit-range walks) rather than re-implementing git access.
- A new pure-computation package, e.g. `core/domain/dora/`, should hold the actual metric formulas (frequency, lead time, MTTR, CFR ratios) as unit-testable pure functions with no I/O — modeled on `core/domain/stats/percentile.go`.

## Open design question: GitHub API access from the daemon

The closed prototype's Change Failure Rate "bug-labeled issue" signal called `gh issue list` — that's fine for a maintainer's terminal (their own `gh` auth) but not for an unattended background daemon, which has no GitHub credentials by default. Options, to be resolved during planning:
[...]

PR #955, opened 2026-07-11
feat(history): CO2 chart equivalent reference lines + methodology docs

### LINKED ISSUE #952: History tab: CO2 chart — relatable equivalents + methodology docs
## Summary

Two related improvements to the CO2 tracking on the History tab:

### 1. Relatable CO2-equivalent reference lines

Add red dotted horizontal reference lines on the CO2 chart showing what the tracked CO2 corresponds to in everyday terms, similar to how some carbon calculators show "equivalent to driving N miles" or "N% of a flight."

Examples to consider (aim for diversity so the comparisons feel relatable to people in different regions, not just US/UK-centric ones):
- Driving a car for N miles/km
- A short-haul flight (e.g. London → Paris)
- A long-haul flight (e.g. London → New York)
- Charging a smartphone N times
- Streaming video for N hours
- Boiling a kettle N times
- A load of laundry
- N trees' worth of annual CO2 absorption
- A bus/train ride of comparable distance (lower-carbon alternative, useful contrast)

Lines should scale sensibly with the chart's time range/zoom level and pick appropriate equivalents for the magnitude shown (don't show "flights" for a few grams of CO2).

### 2. Link to methodology / sources behind the CO2 numbers

Users should be able to see where the CO2 numbers come from — the calculation approach and the underlying research/papers. Currently there's no visible source or explanation in the UI.

Proposal: add a link (e.g. an info icon next to the CO2 chart) pointing to an Irrlicht docs page that explains:
- How the CO2 figures are calculated (formula/assumptions, e.g. energy-per-token estimates, grid carbon intensity assumptions, etc.)
- Links to the source papers/studies the calculation is based on
- Caveats/uncertainty in the estimate

This may require creating a new docs page if one doesn't exist yet.

## Why

Right now the CO2 numbers in the History tab are just a number — without a relatable frame of reference or a transparent methodology, they're hard to interpret and hard to trust.

ingo-eichhorst:
## Summary
- Adds up to 3 red dotted reference lines to the History tab's CO2 chart, each at the height of a relatable everyday CO2e equivalent (a web search, charging a phone, video streaming, boiling a kettle, driving a km/mile, a load of laundry, a tree's yearly absorption, a short/long-haul flight) — chosen automatically to span the visible range so a session's footprint reads as "about a car mile" instead of an abstract gram count.
- Adds a "ⓘ methodology" link on the chart (visible only for the CO2 chart) to a new **CO2 Methodology** docs page explaining the estimation formula, citing the source for every model coefficient, and citing a source (or explicitly flagging as an unsourced order-of-magnitude estimate) for every equivalent.
- Wires the new docs page into the site nav (sidebar across all doc pages, hub card, prev/next links).

Closes #952.

## Notes
- All equivalent figures were independently verified via web research rather than assumed — where no authoritative primary source exists (phone charge, kettle, laundry), the docs page says so explicitly instead of citing a fake source.
- `pickCO2Equivalents()` is a pure, deterministic function (spread-selection by log-distance to 3 target bands) with its own unit tests; the canvas drawing itself is untested (consistent with how the rest of the chart's drawing is untested — canvas painting isn't observable in jsdom).

## Test plan
- [x] `npm test` in `platforms/web/` — 140/140 pass (7 new tests for `CO2_EQUIVALENTS`/`pickCO2Equivalents`)
- [x] `go vet ./...` (core module) — clean
- [x] `go test ./pkg/capacity/... -race -count=1` — pass (untouched CO2 estimation logic, sanity-checked since the docs page cites it)
- [x] `tidy -e` HTML validation on the new docs page — same warning profile as an existing page (no new issues)
- [x] `/code-review` (low effort) on the diff — no findings
- [ ] Manual: open the History tab, switch to the CO2 chart at a few different ranges, confirm the reference lines and methodology link render correctly in both light and dark theme

### DIFF
--- a/platforms/macos/Irrlicht/Models/CO2Equivalents.swift
+++ b/platforms/macos/Irrlicht/Models/CO2Equivalents.swift
@@ -1,0 +1,84 @@
+import Foundation
+
+/// A relatable everyday CO2e activity used as a reference line on the History
+/// CO2 chart (issue #952) — mirrors the web `CO2_EQUIVALENTS`
+/// (platforms/web/historyTab.js). Kept ascending by `grams`; `id` is a stable
+/// string (never reused across entries) so `pick(maxY:)` can track
+/// already-chosen entries by id instead of by reference identity, which Swift
+/// structs don't have.
+struct CO2Equivalent: Equatable, Identifiable {
+    let id: String
+    let grams: Double
+    let label: String
+}
+
+enum CO2Equivalents {
+    /// Kept in lockstep with the web `CO2_EQUIVALENTS` — same ids, grams, and
+    /// labels, updated together in the same change. Every figure's citation
+    /// lives on the "CO2 Methodology" docs page, not duplicated here.
+    static let all: [CO2Equivalent] = [
+        CO2Equivalent(id: "search", grams: 0.2, label: "a web search"),
+        CO2Equivalent(id: "phone-charge", grams: 10, label: "charging a smartphone"),
+        CO2Equivalent(id: "stream-hour", grams: 36, label: "1 hour of video streaming"),
+        CO2Equivalent(id: "kettle", grams: 60, label: "boiling a kettle"),
+        CO2Equivalent(id: "car-km", grams: 170, label: "driving 1 km by car"),
+        CO2Equivalent(id: "car-mile", grams: 393, label: "driving 1 mile by car"),
+        CO2Equivalent(id: "grid-kwh", grams: 460, label: "1 kWh of average grid electricity"),
+        CO2Equivalent(id: "laundry", grams: 1_500, label: "a load of laundry"),
+        CO2Equivalent(id: "gasoline-gallon", grams: 8_900, label: "burning 1 gallon of gasoline"),
+        CO2Equivalent(id: "flight-short", grams: 43_800, label: "a short-haul flight (London → Paris)"),
+        CO2Equivalent(id: "tree-year", grams: 60_000, label: "a tree's CO2 absorption for a year"),
+        CO2Equivalent(id: "car-commute-month", grams: 118_000, label: "a month of average car commuting"),
+        CO2Equivalent(id: "flight-long", grams: 650_000, label: "a long-haul flight (London → New York)"),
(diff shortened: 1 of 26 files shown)


ingo-eichhorst:
Follow-up fix based on testing feedback:

- **Lines were effectively invisible in practice** — the sparse 10-entry equivalents list + the picking algorithm's target fractions frequently landed on distant, wrong-feeling candidates for realistic CO2 totals. Verified against a concrete report (a 50kg chart should read ~2kg/10kg/40kg): the old code picked a 43.8kg flight for the "low" band instead. Densified the list to 17 entries (0.2g → ~100 tonnes) and retuned the target fractions — now reproduces that example almost exactly (1.5kg/8.9kg/43.8kg).
- **Labels overlapped chart content** — were right-aligned at the far edge of the plot; now left-aligned near where the line starts.
- **macOS app never got the fix at all** — turns out the menu bar app has its own fully native SwiftUI Charts implementation of the History/CO2 chart (`HistoryCostChart` in `HistoryView.swift`), completely independent of the web canvas code. Ported the same equivalents list + picking algorithm to Swift (`CO2Equivalents.swift`) and wired in matching red dotted `RuleMark` reference lines, gated to the CO2 chart only.

All new figures added to the list were verified via WebFetch this session (EPA, Our World in Data, Ember) rather than assumed from memory, and the full updated table is on the methodology docs page. 142 web tests + 265 macOS tests (10 new) all green; full macOS suite unaffected.

ingo-eichhorst:
Rebased onto main to resolve a merge conflict with the DORA metrics PR (#951/#959) and the macOS panel-chrome unification PR (#947), which both touched the same History tab files (`historyTab.js`, `irrlicht.js`, `irrlicht.test.js`, `HistoryView.swift`). Squashed this branch's 3 commits into one first so conflicts only needed resolving once. All conflicts were trivial (independent additions at the same location — DORA's sync call next to CO2's, DORA's exports next to CO2's, separate `describe` blocks) — nothing from either side was dropped. 150 web tests + 254 macOS tests (10 new) pass post-rebase; force-pushed with `--force-with-lease`.

Issue #1064, opened 2026-07-15
kirocli: /rewind fixture re-record + V3 session storage (KIRO_HOME done in #1074)


ingo-eichhorst:
> **⚠️ UPDATED 2026-07-15 after assessment.** All three findings are REAL. One is done (PR #1074); two remain.

## ✅ DONE — `KIRO_HOME` (PR #1074)

kiro-cli v2.3.0 added `KIRO_HOME`, which relocates `~/.kiro` **including sessions**. We hardcoded `defaultRootDir = ".kiro/sessions/cli"`, so a user with it set got **silent zero-session discovery**.

The old comment was factually false — and was already false when written (KIRO_HOME shipped 2026-05-12; the adapter was authored 2026-06-05, PR #590):
```go
// Kiro documents no env var that relocates the session root, so this is
// constant — kept as a function to mirror the sibling adapters.
```

Root cause: commit `1165dcdf` (2026-05-15) swept `codex`, `claudecode` and `pi` to honor their env vars. **kirocli post-dates that sweep and was never folded in.** PR #1074 copies the `codex/adapter.go:31-40` shape (abs-only, `filepath.Clean`, reject-and-log) and adds kirocli's **first-ever `adapter_test.go`**.

**Layout confirmed empirically** (live kiro-cli 2.6.0, driven via tmux) — not by analogy, since kiro.dev's docs don't state it:
```
$KIRO_HOME/sessions/cli/72f34e8f-2514-4384-a015-1ddef1b87f3b.jsonl
$KIRO_HOME/sessions/cli/72f34e8f-2514-4384-a015-1ddef1b87f3b.lock
```
Control: nothing landed in `~/.kiro` during the probe — the override is total. Also incidentally confirmed: **headless (`--no-interactive`) writes no session at all** while still honoring `KIRO_HOME` for `settings/` and `agents/` — corroborating the "headless is invisible" fact from PR #590 on a much newer CLI.

---

## 🔴 OPEN — `/rewind` fixture re-record (v2.4.0)

**No code depends on rewind** (`git grep -in rewind -- core/` yields one substring false positive in `cost_tracker_test.go:187`). The adapter doesn't read `session_created_reason`. **This is a fixture/assessment problem, not a code fix.**

`replaydata/agents/kiro-cli/scenarios/1-6_checkpoint-rewind/expected.jsonl` asserts:
[...]

Issue #349, opened 2026-05-15
[bug]: Pi adapter ignores PI_CODING_AGENT_SESSION_DIR — custom session dirs are invisible


ingo-eichhorst:
### Preflight

- [x] I'm on the latest release from [Releases](https://github.com/ingo-eichhorst/Irrlicht/releases/latest).
- [x] I searched existing issues and didn't find a duplicate.

### Irrlicht version

0.4.2

### macOS version

15 (Sequoia) on Apple Silicon

### Which agent(s) were involved?

Pi

### What happened?

Pi v0.71.0 (released 2026-04-30) added the `PI_CODING_AGENT_SESSION_DIR` environment variable as a runtime override for where Pi writes its session transcripts. The Pi adapter in irrlicht doesn't honour the env var — it always watches the hardcoded `$HOME/.pi/agent/sessions/` path. Any user who sets `PI_CODING_AGENT_SESSION_DIR` will have their Pi sessions silently invisible to the dashboard / menu-bar overlay.

This is a code-level gap I confirmed by reading the adapter source:

- `core/adapters/inbound/agents/pi/adapter.go:14` — `const rootDir = \".pi/agent/sessions\"`
- `core/adapters/inbound/agents/pi/agent.go:34-39` — `Source: agent.FilesUnderRoot{Dir: rootDir, ...}`
- `core/adapters/inbound/agents/fswatcher/watcher.go:50-60` — `New(relDir, ...)` always joins relative to `os.UserHomeDir()`
- `grep -rn PI_CODING_AGENT_SESSION_DIR core/` returns zero matches.

Default installs are unaffected — the gap only manifests when the env var is set.

### Steps to reproduce

1. `export PI_CODING_AGENT_SESSION_DIR=/tmp/pi-sessions`
2. `mkdir -p /tmp/pi-sessions`
3. Start a Pi session from a shell that inherits that env var.
4. Observe Pi writes its transcript under `/tmp/pi-sessions/--<cwd>--/<timestamp>_<uuid>.jsonl` instead of `~/.pi/agent/sessions/...`.
5. Observe the session never appears in Irrlicht's overlay / dashboard.

### Expected vs actual behavior

**Expected:** Irrlicht reads `PI_CODING_AGENT_SESSION_DIR` at daemon startup and uses it as the watch root, falling back to `$HOME/.pi/agent/sessions` when unset.

**Actual:** The env var is ignored. Pi sessions in a custom dir are invisible.

[...]

PR #1074, opened 2026-07-15
fix(kirocli): honor KIRO_HOME for session discovery

ingo-eichhorst:
## What changed

The kirocli adapter hardcoded its session root to `~/.kiro/sessions/cli`. Kiro CLI v2.3.0 (2026-05-12) added `KIRO_HOME`, which upstream docs describe as:

> Overrides the `~/.kiro` directory used for global agents, prompts, skills, steering, settings, and **sessions**.

Because the adapter never read this var, setting `KIRO_HOME` made irrlicht's session discovery **silently find nothing** — no error, no log, just zero sessions.

This follows the pattern from `1165dcdf` ("fix(adapters): honor agent-CLI env vars for relocated session dirs (#349) (#362)"), which wired the same fix into `codex` (`CODEX_HOME`), `claudecode` (`CLAUDE_CONFIG_DIR`), and `pi` (`PI_CODING_AGENT_SESSION_DIR`). kirocli was authored later (PR #590, 2026-06-05) and was missed by that sweep — its `sessionsDir()` comment claiming "Kiro documents no env var that relocates the session root" was already false when written.

### Changes

- `core/adapters/inbound/agents/kirocli/adapter.go` — `sessionsDir()` now reads `KIRO_HOME`, rejects non-absolute values (logging why, same as the sibling adapters), and joins `sessions/cli` onto an absolute override. Default stays `.kiro/sessions/cli` (home-relative). Deleted the false comment.
- `core/adapters/inbound/agents/kirocli/adapter_test.go` — **new file** (kirocli had no adapter tests at all). Table test modeled on `codex/adapter_test.go`: empty→default, absolute→override, trailing slash cleaned, relative rejected, tilde rejected.
- `core/adapters/inbound/agents/kirocli/agent.go` — the `transcripts` permission's `Touches`/`Detail` consent copy hardcoded `~/.kiro/sessions/cli/`. Reworded (matching codex/claudecode/pi's phrasing) to note the path relocates to `$KIRO_HOME/sessions/cli/` when set, so the consent text stays truthful now that the path is overridable.

### The `sessions/cli` append question

**I could not fully confirm this from the docs.** I fetched `https://kiro.dev/docs/cli/reference/settings/`, which states verbatim: "Overrides the `~/.kiro` directory used for global agents, prompts, skills, steering, settings, and sessions" — it does **not** explicitly say whether sessions live directly under `$KIRO_HOME` or under a `sessions/cli` subdirectory.

Given that ambiguity, I implemented the append (`$KIRO_HOME/sessions/cli`) by analogy with `CODEX_HOME` (identical shape: relocates the agent home, sessions subdir appended) rather than guessing a flat layout. This mirrors the existing default-path relationship (`~/.kiro` → `~/.kiro/sessions/cli`), so I'm reasonably confident, but it is **not docs-confirmed** — flagging explicitly per the task instructions.

## Testing

- `go test ./core/... -race -count=1` — all packages pass (includes the architecture test).
- `tools/preflight.sh --only go` — all gates pass (gofmt, core tests, onboarding-factory tests, replaydata validate, recording-rig smoke test, starhistory tests).
- The repo's pre-push hook additionally ran the full `tools/preflight.sh` (web tests, ARS architecture gate, security scan) — all passed.

## Scope

Partially addresses #1064 — this PR only covers the `KIRO_HOME` session-discovery fix. It does **not** do the `/rewind` fixture re-record also mentioned in that issue; that remains open. Out of scope per the task: `core/pkg/tailer/`, the vibe adapter, `replaydata/`, `.claude/skills/`, `core/domain/session/metrics.go` — none of these were touched, and the sidecar/PID paths (`sidecar_metrics.go`, `pid.go`) derive from `transcriptPath` so they follow the new root automatically without changes.

### DIFF
--- a/core/adapters/inbound/agents/kirocli/adapter.go
+++ b/core/adapters/inbound/agents/kirocli/adapter.go
@@ -2,6 +2,12 @@
 // transcript files under ~/.kiro/sessions/cli/<uuid>.jsonl.
 package kirocli
 
+import (
+	"log"
+	"os"
+	"path/filepath"
+)
+
 // AdapterName identifies sessions originating from Kiro CLI.
 const AdapterName = "kiro-cli"
 
@@ -19,9 +25,23 @@ const ProcessName = "kiro-cli"
 // next to it.
 const defaultRootDir = ".kiro/sessions/cli"
 
-// sessionsDir returns the directory the Kiro CLI adapter should watch.
-// Kiro documents no env var that relocates the session root, so this is
-// constant — kept as a function to mirror the sibling adapters.
+// kiroHomeEnvVar is the upstream Kiro CLI env var (added v2.3.0, 2026-05-12)
+// that relocates the agent's home directory (default: ~/.kiro). Kiro's docs
+// state it "[o]verrides the ~/.kiro directory used for global agents,
+// prompts, skills, steering, settings, and sessions" — so when set, sessions
+// move to $KIRO_HOME/sessions/cli.
+const kiroHomeEnvVar = "KIRO_HOME"
+
+// sessionsDir returns the directory the Kiro CLI adapter should watch. Non-
+// absolute env values are rejected so a misconfigured path surfaces in logs
+// instead of silently watching the wrong place.
 func sessionsDir() string {
+	if v := os.Getenv(kiroHomeEnvVar); v != "" {
+		cleaned := filepath.Clean(v)
+		if filepath.IsAbs(cleaned) {
+			return filepath.Join(cleaned, "sessions", "cli")
+		}
+		log.Printf("kirocli: ignoring %s=%q — must be an absolute path (no shell expansion)", kiroHomeEnvVar, v)
+	}
 	return defaultRootDir
 }
--- a/core/adapters/inbound/agents/kirocli/adapter_test.go
+++ b/core/adapters/inbound/agents/kirocli/adapter_test.go
@@ -1,0 +1,25 @@
+package kirocli
+
+import "testing"
+
+func TestSessionsDir(t *testing.T) {
+	tests := []struct {
+		name string
+		env  string
+		want string
+	}{
+		{"empty falls back to default", "", defaultRootDir},
+		{"absolute override produces $KIRO_HOME/sessions/cli", "/tmp/kiro-home", "/tmp/kiro-home/sessions/cli"},
+		{"trailing slash is cleaned", "/tmp/kiro-home/", "/tmp/kiro-home/sessions/cli"},
+		{"relative override is rejected (falls back to default)", "relative/home", defaultRootDir},
+		{"tilde-prefixed override is rejected (no shell expansion)", "~/custom", defaultRootDir},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv(kiroHomeEnvVar, tc.env)
+			if got := sessionsDir(); got != tc.want {
+				t.Errorf("sessionsDir() = %q, want %q", got, tc.want)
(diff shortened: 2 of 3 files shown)


ingo-eichhorst:
## ✅ The `sessions/cli` append is confirmed — empirically, not by analogy

The PR body flags the one real open question: whether `$KIRO_HOME` needs `sessions/cli` appended, implemented by analogy with `CODEX_HOME` because kiro.dev's docs don't say. **I drove a live kiro-cli 2.6.0 to settle it. The implementation is correct.**

**Probe 1 — headless (`--no-interactive`):** `KIRO_HOME` is honored (it created `settings/` and `agents/` under the probe dir), but **no session file is written at all** — consistent with the known "kiro-cli headless is invisible" adapter fact. So headless can't answer the layout question.

**Probe 2 — interactive, driven via tmux** (`KIRO_HOME=/tmp/kh-tmux-probe/home kiro-cli chat`, one real turn):

```
/tmp/kh-tmux-probe/home/settings
/tmp/kh-tmux-probe/home/sessions
/tmp/kh-tmux-probe/home/sessions/cli
/tmp/kh-tmux-probe/home/sessions/cli/72f34e8f-2514-4384-a015-1ddef1b87f3b.jsonl
/tmp/kh-tmux-probe/home/sessions/cli/72f34e8f-2514-4384-a015-1ddef1b87f3b.lock
```

**Control:** nothing new landed in `~/.kiro/sessions/cli` during the probe — so the override is total, not a partial redirect.

`$KIRO_HOME/sessions/cli/<uuid>.jsonl` is exactly what this PR watches. The `.lock` sibling is also present as expected. The docs-uncertainty caveat in the PR body can be downgraded to a confirmed fact.

Verification also green: `go test ./core/... -race -count=1`, `tools/preflight.sh --only go`, and the pre-push hook's full preflight (web + ARS + security scan).

ingo-eichhorst:
## Review: approved. The red SonarCloud check is expected — and is my fault, not the PR's.

**Sonar has zero open issues on this PR** (verified via API: `total: 0`). The gate fails on exactly one condition:

```
new_duplicated_lines_density  actual=7.5  GT  threshold=3
```

With 52 new lines across the three files, that's **≈4 duplicated lines** — the `if v := os.Getenv(...) / filepath.Clean / IsAbs / log-and-fall-through` idiom, which is *character-for-character the point of this PR*. The handover explicitly said "copy the pattern from `codex/adapter.go:31-40`, do NOT invent a new approach," because `codex`, `claudecode`, and `pi` all share it from commit `1165dcdf`. kirocli is the 4th instance.

Per AGENTS.md, SonarCloud/CodeScene are **advisory, not merge gates** — "a red result is a prompt to look closer, not a block — use judgment." Judgment here: a 4-line idiom matching three working siblings is correct-by-consistency. Diverging to satisfy the gate would make kirocli the odd one out. `mergeStateStatus` is `UNSTABLE` (mergeable; only non-required checks red), and all 12 required checks pass.

**Legitimate follow-up, out of scope here:** Sonar is arguably right in aggregate — there are now **four** near-identical `sessionsDir()` env-override implementations. Extracting a shared helper (e.g. `agentEnvDir(envVar, subdir, default string)`) would delete the duplication in all four. That's a real cleanup, but it touches three working adapters and shouldn't ride along with a correctness fix. Worth its own issue.

## What I verified beyond CI

- **The `sessions/cli` append** — the PR's one flagged uncertainty — is now confirmed empirically against live kiro-cli 2.6.0 (see previous comment). Not a guess.
- Required checks: 12/12 pass, including `go-test`, `build-test`, ARS gate, and both web suites.

No changes requested.

Issue #1063, opened 2026-07-15
vibe: reset parser state on rotation; use effective per-model compact threshold


ingo-eichhorst:
> **⚠️ REWRITTEN 2026-07-15 after assessment.** The two HIGH fixes this issue originally proposed were both **rejected** — they were built on premises that don't hold. Two *real* bugs were found underneath and are implemented in **PR #1078**.
> Original title: *"handle in-place messages.jsonl rewrites and the read_file/read tool rename"*.

## ❌ Rejected — do NOT implement these

**"Add inode/truncation detection to the vibe tailer" — ALREADY HANDLED.**
The shared `TranscriptTailer` does `os.Open` + `defer file.Close()` on **every** pass (`core/pkg/tailer/tailer.go:473-477`), so there is no long-lived FD to strand on a dead inode. Shrink is explicitly detected at `tailer.go:491-500`:
```go
case fileSize < t.lastOffset:
    // File rotated/truncated — reset cumulative accumulators...
    startPos = 0
    t.resetAccumulatorsForRotation()
```
Both failure modes the original finding rested on cannot occur.

Reachability was also overstated: `_overwrite_messages_sync` is real (`session_logger.py:273`, `os.replace` at `:290`) but gated on `inplace=True` (`rewind/manager.py:123-167`). **The TUI `/rewind` does NOT pass it** (`cli/textual_ui/app.py:3556-3557` → default `inplace=False` → forks a new session dir). **Only the ACP path passes `inplace=True`** (`acp/acp_agent_loop.py:2003-2004`).

**"Accept both `read_file` and `read` tool names" — NO-OP.**
`vibe/parser.go` keys on exactly **one** tool name: `"todo"` (`parser.go:249`) — which never changed. `parseToolCalls` (`parser.go:297-321`) reads only `id` and `function.name` and passes the name through verbatim; it **never reads args**, so `path`→`file_path` is equally moot. Live 2.19.1 `tools_available`: `ask_user_question, bash, edit, grep, read_file, skill, task, todo, web_fetch, web_search, write_file`.

**"Audit fixtures for the transient `read` name" — CLEAN.**
[...]

PR #1078, opened 2026-07-15
fix(vibe): reset parser state on rotation; use effective per-model compact threshold

ingo-eichhorst:
## Summary

Two well-scoped fixes for real bugs found while assessing issue #1063 — **not** the two fixes the issue text itself proposes (see "About #1063's proposed fixes" below).

### Fix 1 (primary): rotation reset didn't reset parser state → vibe tokens/cost flatline at 0

The tailer's rotation/truncation handler (`resetAccumulatorsForRotation`, `core/pkg/tailer/tailer.go`) zeroes its own cumulative token accumulators when it detects `fileSize < lastOffset`, but had no way to tell a stateful parser to reset its own accumulated state.

Vibe's parser (`core/adapters/inbound/agents/vibe/parser.go`) tracks a token high-water mark (`lastPromptTokens`/`lastCompletionTokens`) derived from `meta.json`'s session-cumulative `session_prompt_tokens`/`session_completion_tokens`, and emits each turn's *delta*. That upstream counter is monotonic within a session. This was previously unreachable because Vibe's `/rewind` always forked a new session directory — but Vibe **2.19.1** made it reachable via the **ACP path only**: `_overwrite_messages_sync` (tempfile + `os.replace`) is gated on `inplace=True`, which the TUI doesn't pass (defaults to `inplace=False` → forks) but ACP does. ACP shares the same session directory, so the rewrite lands under the tailer's watched root as a legitimate rotation.

Without a parser-side reset, the stale (larger) high-water mark computes a negative delta against the freshly-rewound session's smaller counters, both clamp to zero, and `emitContribution`'s early return skips updating the high-water mark too — so every subsequent `turn_done` repeats the same clamp, and session tokens/cost flatline at 0 for the rest of the session.

**Fix**: added an optional `rotationResetter` interface (`ResetForRotation()`) that the tailer type-asserts for in `resetAccumulatorsForRotation` — a no-op for the 6 other adapters sharing this tailer (claudecode, codex, pi, kirocli, aider, antigravity) that don't implement it. Vibe's `Parser.ResetForRotation` resets:
- `lastPromptTokens` / `lastCompletionTokens` — the token high-water mark itself.
- `todos` (the `TodoReconciler`) — its synthetic per-todo task IDs are assigned in lockstep with the tailer's own `taskSeq` counter (both increment once per Create delta, in the same order). The tailer already resets its `taskSeq` to 0 on rotation; leaving the reconciler's `nextID`/`idByKey` stale would desync the two ID spaces.

Deliberately **not** reset: `path` (the transcript file doesn't change across an in-place rewrite) and the `sidecarCache` (already self-invalidates via its own mtime/size check the moment Vibe rewrites `meta.json`).

Added a regression test (`TestTokenAccounting_ResumesAfterRotationReset` in `metrics_test.go`) that drives the production tailer through a pre-rewind turn, simulates the in-place rewrite (smaller transcript + smaller cumulative sidecar counters), and asserts tokens resume correctly instead of flatlining. Also added `core/pkg/tailer/rotation_reset_test.go` pinning the generic seam itself (hook fires on rotation, is a no-op for non-implementing parsers). I verified both tests actually catch the bug by temporarily reverting the tailer-side wiring and confirming they fail as expected, then restored the fix.

### Fix 2 (secondary, latent): `auto_compact_threshold` read the global default, not the effective per-model value

`core/adapters/inbound/agents/vibe/sidecar.go` decoded `config.auto_compact_threshold` (the global default) with no `config.models[]` parsing at all. Upstream's *effective* threshold is per-model (`config.get_active_model().auto_compact_threshold`); `_apply_global_auto_compact_threshold` only pushes the global into models that didn't set their own override, so the two values diverge whenever a per-model override exists. This feeds `SessionMetrics.ContextWindow` → the context-utilization bar, so it is user-visible once a divergence exists.

**Fix**: `resolveAutoCompactThreshold` now resolves `config.active_model` against `config.models[].alias` (not `.name` — a live 2.19.1 `meta.json` has `active_model: "mistral-medium-3.5"` matching `models[].alias`, while `models[].name` is the unrelated `"mistral-vibe-cli-latest"`; matching on name would silently fail to resolve anything), falling back to the global default when there's no match or the override is zero.

**This is a latent fix, not a live bug**: in the verified live `meta.json`, all three models carry the same value (200000) as the global, so nothing is visibly wrong today — it only diverges once a per-model override is set. Added `TestContextWindow_PerModelAutoCompactThresholdOverride`, including a decoy entry to prove the alias-vs-name matching direction, and updated the existing `TestTokenAccounting_EndToEnd` comment to note it only pins the no-override fallback path.

### Drive-by

Fixed a stale comment on `SessionMetrics.PermissionMode` (`core/pkg/tailer/tailer.go`) that listed `"default"`, `"plan"`, `"bypassPermissions"` as the values in use. A census of 320 live Claude Code transcripts shows `auto` (~5000), `plan` (331), `default` (8), `acceptEdits` (3), and `bypassPermissions` (0) — and Claude Code v2.1.200 renamed `"default"` to `"manual"`. Comment-only; the field is pass-through with no classifier branching on its value.

## About #1063's proposed fixes

The issue as filed proposes two HIGH-priority fixes. A deep assessment found both are **already handled or no-ops**, and they are deliberately **not implemented** here so the mismatch with the issue text doesn't confuse reviewers:

- "Add inode/truncation detection to the vibe tailer" — **already handled**. The shared tailer does `os.Open` + `defer file.Close()` every pass (no stale FD), and shrink detection already exists (`fileSize < t.lastOffset` → reset). This PR's Fix 1 is the *actual* gap underneath that report — the tailer's existing detection just didn't propagate to parser-side state.
- "Accept both `read_file` and `read` tool names" — a **no-op**. `vibe/parser.go` keys on exactly one tool name (`"todo"`) for the whole-list-replace decoder, which never changed, and `parseToolCalls` passes tool names through verbatim without ever reading args — there's nothing here that inspects `read_file`/`read` at all.
- A fixture audit found the replaydata fixtures clean (no `read`/`search_replace`/`edit` tool names across 243 files; all 36 recordings postdate the transient window described in the issue).

## Follow-up not implemented here

`isUserBlockingToolName` (`core/pkg/tailer/tailer_config.go`) doesn't match Vibe's `ask_user_question` tool — a real gap, but deliberately out of scope for this PR. It needs a fixture decision first: `replaydata/agents/mistral-vibe/scenarios/2-17_user-blocking-question/metadata.json` currently steers Vibe away from that tool and tests the trailing-`?` path instead.

## Test plan

- [x] `go test ./core/... -race -count=1` — full core suite green, including the architecture test.
- [x] `tools/preflight.sh --only go` — green.
- [x] Full `tools/preflight.sh` ran as the pre-push hook (gofmt, core tests, onboarding-factory tests, replaydata validate, recording-rig smoke test, starhistory tests, both web suites, ARS architecture gate, security scan) — all green.
- [x] `go test ./tools/onboarding-factory/cmd/replay/... -count=1` — all `TestFixtureReplayByteIdentity` cases pass byte-identical across every adapter; no golden regeneration was needed (confirms Fix 1 didn't leak into any adapter's replay output).
- [x] Sanity-checked both new regression tests by temporarily reverting the tailer-side wiring and confirming they fail, then restoring the fix.

Partially addresses #1063

### DIFF
--- a/core/adapters/inbound/agents/vibe/metrics_test.go
+++ b/core/adapters/inbound/agents/vibe/metrics_test.go
@@ -48,8 +48,126 @@ func TestTokenAccounting_EndToEnd(t *testing.T) {
 		t.Errorf("CumOutputTokens = %d, want 320", m.CumOutputTokens)
 	}
 	// The context-utilization bar resolves from the sidecar's auto-compaction
-	// threshold (mistral models aren't in the capacity window map).
+	// threshold (mistral models aren't in the capacity window map). No
+	// config.models[] entries here, so this only pins the no-override
+	// fallback-to-global path; TestContextWindow_PerModelAutoCompactThresholdOverride
+	// below pins the per-model resolution (issue #1063).
 	if m.ContextWindow != 200000 {
 		t.Errorf("ContextWindow = %d, want 200000 (auto_compact_threshold)", m.ContextWindow)
+	}
+}
+
+// TestContextWindow_PerModelAutoCompactThresholdOverride pins issue #1063:
+// the EFFECTIVE auto-compaction threshold is the active model's own
+// config.models[] override, not the top-level global default, whenever the
+// two diverge. It also pins the alias-vs-name matching gotcha found in a
+// live 2.19.1 meta.json: config.active_model ("mistral-medium-3.5") matches
+// models[].alias, while models[].name ("mistral-vibe-cli-latest") is an
+// unrelated CLI/product label that must NOT be used for the match — a decoy
+// entry whose name equals active_model, but whose alias doesn't, is included
+// below to prove the resolution isn't accidentally matching on name.
+func TestContextWindow_PerModelAutoCompactThresholdOverride(t *testing.T) {
+	dir := t.TempDir()
+	transcript := filepath.Join(dir, transcriptFilename)
+	lines := `{"role":"assistant","content":"done","message_id":"a1"}` + "\n"
+	if err := os.WriteFile(transcript, []byte(lines), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	meta := `{"environment":{"working_directory":"/tmp/p"},"config":{` +
+		`"active_model":"mistral-medium-3.5","auto_compact_threshold":200000,` +
+		`"models":[` +
+		`{"name":"mistral-medium-3.5","alias":"decoy-name-match","auto_compact_threshold":1},` +
+		`{"name":"mistral-vibe-cli-latest","alias":"mistral-medium-3.5","auto_compact_threshold":64000}` +
+		`]},"stats":{"context_tokens":9000,"session_prompt_tokens":8500,"session_completion_tokens":320}}`
+	if err := os.WriteFile(filepath.Join(dir, "meta.json"), []byte(meta), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	tl := tailer.NewTranscriptTailer(transcript, &Parser{}, AdapterName)
+	m, err := tl.TailAndProcess()
+	if err != nil {
+		t.Fatalf("TailAndProcess: %v", err)
+	}
+
+	if m.ContextWindow != 64000 {
+		t.Errorf("ContextWindow = %d, want 64000 (active model's own models[].alias-matched override, not the 200000 global default or the 1 name-matched decoy)", m.ContextWindow)
+	}
+}
+
+// TestTokenAccounting_ResumesAfterRotationReset pins issue #1063's primary
+// fix: a rotation/truncation reset must not flatline vibe's token accounting
+// at zero for the rest of the session.
+//
+// Vibe's meta.json session_prompt_tokens/session_completion_tokens are
+// monotonic within a session, so the parser tracks its own high-water mark
+// (lastPromptTokens/lastCompletionTokens) and emits only each turn's DELTA.
+// Vibe 2.19.1's ACP /rewind path rewrites messages.jsonl IN PLACE
+// (_overwrite_messages_sync with inplace=True) instead of forking a new
+// session directory the way the TUI's /rewind does, so a rewind can land as
+// a legitimate rotation under the tailer's watched root: the transcript
+// shrinks (fileSize < lastOffset) and meta.json's cumulative counters drop
+// to reflect the smaller, retained history.
+//
+// Before the fix: the tailer resets its OWN cumulative accumulators to 0 on
+// rotation, but the parser's stale (larger) high-water mark survives. The
+// next turn_done computes dPrompt/dCompletion against that stale mark, both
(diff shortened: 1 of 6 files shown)


ingo-eichhorst:
## Review: approved. The red CodeScene check is on a test, and the test is worth its complexity.

**SonarCloud's quality gate is `OK` on this PR.** The only red is CodeScene, and it's one advisory rule:

```
Enforce advisory code health rules (1 file with Complex Method)
metrics_test.go — TestTokenAccounting_ResumesAfterRotationReset — health 10.00 → 9.69
5 Quality Gates Passed
```

That's the **regression test**, not production code. Simulating this bug genuinely requires a multi-step fixture (write transcript + meta → tail → rewrite smaller in place → tail again → assert tokens resume rather than flatline). I'd rather keep a slightly long test that provably catches the bug than trim it for a 0.31 health point — especially since the author demonstrated it's meaningful by reverting the tailer wiring and confirming it fails (`CumInputTokens = 0, want 3000`) before restoring. That's the discipline I want; I'm not going to penalize it.

Per AGENTS.md, CodeScene is **advisory, not a merge gate**. `mergeStateStatus` is `UNSTABLE` (mergeable; only non-required checks red).

## What I verified independently

The risk here was Fix 1 touching `core/pkg/tailer`, shared by 7 adapters. Checked directly:

- **Only vibe implements the interface** — `git grep ResetForRotation` returns exactly `vibe/parser.go:90` (impl), `tailer/parser.go:460` (decl), `tailer/tailer.go:571` (call).
- **The call site is a guarded type assertion**, so it's a genuine no-op for the other six:
  ```go
  func (t *TranscriptTailer) resetAccumulatorsForRotation() {
      if resetter, ok := t.parser.(rotationResetter); ok {
          resetter.ResetForRotation()
      }
      t.cumInputTokens = 0
      ...
  ```
- **Scope is exactly the 6 intended files** — nothing leaked into other adapters.
- **Replay goldens byte-identical across every adapter** — the strongest available evidence the shared-tailer change didn't leak.

Also credit for two judgment calls I'd have had to make otherwise: resetting `todos` because its synthetic task IDs move in lockstep with the tailer's `taskSeq` (which the tailer already resets, so leaving it would desync the ID spaces), and *not* resetting `sidecarCache` (self-invalidates on mtime/size) or `path` (unchanged by an in-place rewrite).

**Follow-up noted, correctly deferred:** `isUserBlockingToolName` (`tailer_config.go:34-40`) matches `AskUserQuestion`/`ExitPlanMode`/`question` but not vibe's `ask_user_question`. Real gap, but it needs a fixture decision — `2-17_user-blocking-question/metadata.json` deliberately steers Vibe away from that tool to test the trailing-`?` path instead. Right call to leave it.

No changes requested.

Issue #1161, opened 2026-07-16
Phase 1a: wire the Claude Code Stop hook → authoritative ready (#1129)


ingo-eichhorst:
Phase 1a of epic #1129. First, clean slice of the hooks phase — the spike (#1141) confirmed **hooks, not OTel, are the state authority**, so this is the load-bearing direction.

## What

Install and wire the Claude Code **`Stop`** hook so `ready` (and the assistant text the waiting-check needs) come from an authoritative push at true turn end, instead of being inferred from the transcript tail.

Today `ready` is derived from `IsAgentDone()` (transcript `turn_done`), whose legacy fallback matches `assistant`/`assistant_output` and **carves out codex by name in a comment** (`metrics.go:472`) to stop ready↔working flicker — a landmine for every new adapter. The `Stop` hook gives claudecode a clean, per-adapter turn-done signal that doesn't depend on that heuristic, and it carries `last_assistant_message` — exactly the text `IsWaitingForUserInput` reads — so the same signal improves the waiting-vs-ready decision.

Spike-measured behavior: the `Stop` hook fires at real turn end (~4s after the answer in the driven run, i.e. at genuine completion, not mid-turn).

## Scope

- Add `HookStop` and register `"Stop"` in `installedHookEvents` (`claudecode/hookinstaller.go`) with an empty matcher (Stop takes no tool matcher). Cover install idempotency + clean uninstall, matching the existing five-event machinery.
- Handle the `Stop` payload in the hooks handler (`claudecode/hooks.go`): surface an authoritative turn-done signal and capture `last_assistant_message` into the same field the classifier reads for `LastAssistantText`.
- Wire the signal through the classifier (`state_classifier.go`): a hook-delivered `Stop` routes through the existing `classifyAgentDone` path (so a turn that ended on a question still goes `waiting`, not `ready`), taking precedence over the transcript-derived `IsAgentDone` for claudecode.
- Move the installed hook command to native `"type": "http"` delivery instead of the `curl` wrapper while here (this is the change that likely removes the reason `OpenToolStalled`/#488 exists — call that out but don't necessarily remove the rule in this issue).

## Acceptance criteria (user-observable)

- A claudecode session transitions to `ready` on turn completion driven by the `Stop` hook, not only by transcript tailing.
- A turn that ends on a question/cue still lands in `waiting` (the `Stop`-carried `last_assistant_message` feeds `IsWaitingForUserInput`).
[...]

PR #1174, opened 2026-07-18
feat(codex): hook-authoritative live-state tier (PermissionRequest→waiting, Stop→ready)

### LINKED ISSUE #1171: Codex hooks: PermissionRequest → waiting, Stop → ready (hook-authoritative state tier for Codex)
## Goal

Give the Codex adapter a **hook-authoritative live state tier**, mirroring what Phase 1 does for Claude Code. Codex CLI shipped a Claude-Code-shaped hooks system (experimental in `rust-v0.114.0`, ~March 2026; still iterating through v0.144, July 2026), so we can observe Codex's UI state from a structured push channel instead of inferring it from transcript prose.

Part of epic #1129 (observe UI state, stop inferring it). This is the fix for the gap the OTel spike (#1141) opened: OTel is retrospective and not the waiting authority for **any** of the three OTel adapters. Claude Code has its own hooks (Phase 1); geminicli is unmaintained (#1068); **Codex was left as the orphan** — no live-waiting plan, and it is not in the Phase 5 network-probe list. Codex hooks close that.

## Where Codex sits today

- **Turn-end is already structured and prospective** — `task_complete` / `turn_aborted` `event_msg` payloads map to `turn_done` → `ready` (`core/adapters/inbound/agents/codex/parser.go:170-225`). So a `Stop → ready` hook is *marginal* for Codex; the transcript already covers turn-end.
- **Waiting is the gap.** Codex approval prompts (shell escalation, network access) are interactive TUI overlays. The adapter only handles one waiting case explicitly — it maps a `<proposed_plan>` block to `ExitPlanMode` (`parser.go:295`). Ordinary approval prompts are not handled, so they fall through to the shared `waiting_cue` prose regex — exactly the inference this epic wants to delete.

## What the hooks give us (ranked by value)

1. **`PermissionRequest` → `waiting` (the real win).** Fires *while* Codex is blocked on an approval — prospective, live, JSON on stdin. Direct analog of Claude Code's #108 permission hook. Retires `waiting_cue` reliance for Codex approvals.
2. **`Stop` → `ready`, carrying `last_assistant_message`.** Marginal for turn-end (transcript already covers it), but the `last_assistant_message` payload feeds `IsWaitingForUserInput` directly — the same double-duty noted for Claude Code's Stop hook in Phase 1a.
3. **`PreCompact` / `PostCompact`** — optional; only if Codex grows a transcript-goes-silent-during-compaction problem like Claude Code's #657.

## Reuse — this is a small, bounded slice

Codex's hook system is deliberately Claude-Code-shaped: same event names (`PreToolUse`, `PermissionRequest`, `PostToolUse`, `PreCompact`, `Stop`), same `matcher` + `hooks: [{ "type": "command", "command": ... }]` schema, JSON-on-stdin, `Stop` carrying `last_assistant_message` (snake_case). The existing Claude Code hook stack is ~80% reusable:

- **Receiver** — the `serveHookRequest` / `hookPayload` / `HandlePermissionHook` / `HandleCompactHook` shapes in `core/adapters/inbound/agents/claudecode/hooks.go` transfer directly. Add a `POST /api/v1/hooks/codex` route (mirrors `startup.go:911`).
- **Installer** — a Codex hook-installer mirroring `claudecode/hookinstaller.go`, writing to **`~/.codex/hooks.json`** (a dedicated JSON file) rather than `~/.codex/config.toml`. `hooks.json` lets us reuse the settings.json merge/idempotency logic almost verbatim, and keeps blast radius off the user's main config (malformed TOML could break Codex entirely).
- **Consent** — the adapter declares only `transcripts` today (`codex/agent.go:47`). Add a `hooks` permission (`KindModify`) with `Apply`/`Remove` closures that install/uninstall the hook, and wire `contracttesting.AssertPermissionGated` for it.
- **Delivery command** — the same `curl -fsS --max-time 1 -X POST --data-binary @- http://localhost:7837/api/v1/hooks/codex || true` shape works, since Codex execs a command with the payload on stdin.

## Caveats / constraints

- **Port-7837 fixture problem applies identically** (see [[project_claudecode_hook_port_hardcoded]] / epic Phase 1 note): hook-delivered observations can't be re-recorded on a dev daemon at an alt port. Fix the port or plan the fixture story before relying on recorded coverage.
- **Version-gate the install.** Hooks are experimental and the event set has changed across versions (`UserPromptSubmit` was added later, plugin hooks, etc.). The adapter already captures `cli_version` from `session_meta` — only install when the running Codex supports the events we use. Let `ir:agent-releases` catch schema drift.
- **Config blast radius** — prefer `~/.codex/hooks.json` over `~/.codex/config.toml` for the reasons above; consent-gate the write.

## Dependencies / sequencing

Depends on the epic's reconcile work: this is a *second* hook consumer, so `PermissionRequest`/`Stop` inherit the "late but authoritative" concern (the `Notification`/`idle_prompt` family fires ~6s late for Claude Code; Codex's turn-scoped hooks need the same reconcile-and-correct handling). **Land Phase 3 (tier/provenance + reconcile, and the minimal slice pulled into Phase 1b) first**, then have Codex reuse that machinery rather than land ahead of it.

## Done when

- [ ] `PermissionRequest` from Codex drives a live `waiting` transition (approval prompt shows → session is `waiting` before the transcript would reveal it).
- [ ] `Stop` drives `ready` and carries `last_assistant_message` into `IsWaitingForUserInput`.
- [ ] Hook install/uninstall is gated behind a new `hooks` permission on the Codex adapter, with `AssertPermissionGated` coverage; install is version-gated on `cli_version`.
- [ ] Codex no longer depends on `waiting_cue` prose matching for approval prompts.
- [ ] Fixture/recording story for the hardcoded port is resolved (or explicitly deferred with a note).

## References

- Codex hooks (official): https://learn.chatgpt.com/docs/hooks
- Codex `notify` (older, distinct — argv JSON, `agent-turn-complete` only): https://learn.chatgpt.com/docs/config-file/config-advanced
- Epic: #1129 · OTel spike verdict: #1141 · Claude Code permission hook precedent: #108 / #307

ingo-eichhorst:
## What

Gives the Codex adapter a **hook-authoritative live-state tier**, mirroring what Claude Code already has (#1161/#108). Codex shipped a Claude-Code-shaped hooks system (experimental in `rust-v0.114.0`), so the daemon can now observe Codex's UI state from a structured push channel instead of inferring it from transcript prose. Part of epic #1129.

Closes #1171.

## How

The daemon installs three hooks into **`~/.codex/hooks.json`** (a dedicated file — never `config.toml`, so a malformed write can't corrupt the user's main Codex config) that `curl`-POST their payload to `POST /api/v1/hooks/codex`:

| Event | Effect |
|---|---|
| `PermissionRequest` | live `waiting` the moment an approval overlay appears — **the real win**, retiring `waiting_cue` prose matching for approvals |
| `PostToolUse` | clears the pending overlay once an approved tool runs |
| `Stop` | `ready`, carrying `last_assistant_message` into the classifier; also clears any lingering pending |

### Key design points
- **Reuse, no services rewrite.** `SessionDetector.HandlePermissionHook`/`HandleStopHook` are already agent-agnostic (plain strings), so the Codex receiver dispatches straight into them. Only one small additive method was added: `ClearPermissionPending`.
- **Session-ID correctness.** Codex session IDs are the `session_meta` `payload.id`, *not* the filename stem. The receiver resolves the ID via `sessionIDFromPath(transcript_path)` — the same function fswatcher uses — so the overlay maps key on the exact detector ID. (Both `PermissionRequest` and `Stop` payloads carry `transcript_path`.)
- **Denial edge.** Codex has **no `PostToolUseFailure`** event, so a denied-then-aborted approval could otherwise leave `permissionPending` stuck. A `Stop` authoritatively means no approval is pending (a turn can't end while blocked), so the receiver clears pending on every `Stop`.
- **Consent + version gate.** New `hooks` (`KindModify`) permission with `Apply`/`Remove` install closures; install is version-gated on the running Codex's captured `cli_version` (floor `0.114.0`, fail-open on unknown). `AssertPermissionGated` covers **both** the live receiver gate and the install gate.
- **No prose change needed.** Approval prompts are TUI overlays that never appeared reliably as assistant prose, so the hook simply becomes authoritative for approvals; `PendingWaitingCue` stays for its legitimate turn-end-question job.

## Testing
- `go test ./core/... -race -count=1` — full suite green (46 pkgs).
- New: `codex/hooks_test.go` (dispatch, session-ID resolution, method/JSON/identity guards, live-gate `AssertPermissionGated`), `codex/hookinstaller_test.go` (install/uninstall idempotency, Stop-has-no-matcher, preserves unrelated config, version-gate table, install-gate `AssertPermissionGated`).

## Deferred / notes
- **Fixture/recording deferred** (Done-when allows it): hook delivery hardcodes port `7837`, so hook-driven observations can't be re-recorded on a dev daemon at an alt port — same constraint as claudecode hooks.
- **DRY:** the `WaitingScanWindow` + `ExtractQuestionSnippet‖ExtractWaitingCue` tail rule now has a third copy (parser.go, claudecode/hooks.go, codex/hooks.go) — a candidate for a shared helper in a follow-up.


ingo-eichhorst:
## ✅ Live-verified against a real Codex (0.144.4) + factory regression added

Drove **codex-cli 0.144.4** under tmux against a worktree-built `irrlichd` on `:7837` (grant-all), `-a untrusted -s read-only` on a write task:

| Signal | Result |
|---|---|
| Installer wrote `~/.codex/hooks.json` (version-gated 0.144.4 ≥ 0.114) | ✅ exact schema (`.*` matchers, Stop no-matcher, curl→:7837) |
| **`PermissionRequest` → `waiting`** at the approval overlay (a TUI state never in the rollout JSONL) | ✅ session flipped to `waiting` |
| `PostToolUse` → clear, second prompt re-triggered `waiting` | ✅ |
| **`Stop` → `ready`** at turn end | ✅ |
| Doc-based assumptions (`transcript_path` delivered, `.*` accepted, Stop no-matcher) | ✅ all held |

**New deployment detail:** Codex gates hook *execution* behind a one-time interactive **"Trust hooks"** prompt — after irrlicht installs them, the user must trust them in Codex before they fire (a "Review hooks" option lets them inspect the installed curl commands first). Worth a docs mention.

### Factory regression (`replaydata/agents/codex/scenarios/2-19_tool-gate-permission-prompt`)
Re-assessed **`incapable` → `daemon_capability: full`** (`of validate` green). The prior verdict rested entirely on *"codex Source is FilesUnderRoot … there is NO HTTP hook channel … PermissionPending is always false for codex"* — which this PR invalidates. The cell is `pending-record`: the `events.jsonl` recording is **deferred** because the hook POSTs to hardcoded `:7837` while the factory coexist recorder runs on `:7838` (same limitation as claudecode hooks). A recording needs a `:7837 --record` daemon via `of record run --attach`, or a configurable hook port — filed as follow-up work.

The `test-first` recommendation is now satisfied — the integration is confirmed, so this is safe to merge.

PR #1219, opened 2026-07-31
test(factory): floor $HOME so the snapshot test can never reach a real config

### LINKED ISSUE #1212: Harden hook-config-snapshot_test.sh so it can never write to the real $HOME
## What happened

On 2026-07-31 at 16:41:51 a draft of `tools/onboarding-factory/scripts/lib/hook-config-snapshot_test.sh` ran without HOME isolation and wrote its fixture straight over the real `~/.claude/settings.json`:

```
{"hooks":{"Stop":"localhost:7837"}}
```

That is the literal fixture from the test's first case (line 42). Byte-for-byte identical to what landed in the user's config.

Blast radius: it destroyed the `statusLine` entry **and** all seven irrlicht hook entries. The daemon kept detecting sessions via transcripts, so the only visible symptom was the Claude Code status line vanishing — statusline ingestion stopped dead at 16:41:51 and stayed dead for 18 minutes. No backup existed (no Time Machine destination; newest `settings.json.backup_*` was from Nov 2025), so recovery meant hand-reconstructing the file and re-running `EnsureHooksInstalled()` / `EnsureStatuslineInstalled()`.

The draft was fixed minutes later (`fresh_env` added) and the committed test is correctly isolated. This issue is about the residual hazard, not that draft.

## The residual hazard

The committed test's safety rests entirely on every case calling `fresh_env` **before** it writes. `HOME` is only redirected inside that function:

```bash
fresh_env() {
  ROOT="$TMP/$1"
  mkdir -p "$ROOT/home/.claude" "$ROOT/codex"
  HOME="$ROOT/home"
  ...
}
```

A new case added above the first `fresh_env` call — or one that forgets it — writes to the developer's real `~/.claude/settings.json` again. Nothing structurally prevents it.

This area is inherently risky by design: per the lib's own header comment, `~/.claude/settings.json` and `$CODEX_HOME/hooks.json` follow `$HOME`, not `IRRLICHT_HOME`, so `IRRLICHT_ONBOARD_HOME` does not isolate them.

## Fix

Redirect `HOME` (and `CODEX_HOME`) to a temp dir once, immediately after `TMP` is created, so no case can reach the real `$HOME` even before `fresh_env` runs:

```bash
TMP="$(mktemp -d)"
trap 'rm -rf "$TMP"' EXIT

HOME="$TMP/nohome"        # no case can reach the real $HOME, even before fresh_env
CODEX_HOME="$TMP/nocodex"
```

`fresh_env` keeps overriding both per-case; this is purely the floor.

Worth a quick sweep of the sibling `lib/*_test.sh` files for the same pattern while in there.

## Base branch

`tools/onboarding-factory/scripts/lib/hook-config-snapshot_test.sh` **does not exist on `main`** — it is introduced by PR #1207 (branch `feat/1178-configurable-hook-port`, issue #1178), which is still open.

Land this on `feat/1178-configurable-hook-port` rather than branching off `main`, or wait until #1207 merges. Do not branch off `main`: the target file will not be there.

## Verification

`bash tools/onboarding-factory/scripts/smoke-test.sh` (runs this test), with the real `~/.claude/settings.json` left untouched.

ingo-eichhorst:
Closes #1212.

## Problem

`hook-config-snapshot_test.sh` only redirects `$HOME` **inside** `fresh_env()`. Any case that runs before it — or forgets to call it — writes to the developer's real `~/.claude/settings.json`.

That is not hypothetical. A draft of this very file did exactly that on 2026-07-31 at 16:41:51, writing its own fixture over a live config:

```
{"hooks":{"Stop":"localhost:7837"}}
```

It destroyed the `statusLine` entry **and** all seven irrlicht hook entries. The daemon kept detecting sessions via transcripts, so the only visible symptom was the Claude Code status line vanishing — statusline ingestion stopped dead for 18 minutes. No backup existed (no Time Machine destination; newest `settings.json.backup_*` was from Nov 2025), so recovery meant hand-reconstructing the file and re-running `EnsureHooksInstalled()` / `EnsureStatuslineInstalled()`.

The draft was fixed minutes later and the committed test is correctly isolated. This PR removes the residual hazard, which is structural: nothing stops the next case from being added above the first `fresh_env` call.

The risk is inherent to this area — per the lib's own header, `~/.claude/settings.json` and `$CODEX_HOME/hooks.json` follow `$HOME`, not `IRRLICHT_HOME`, so `IRRLICHT_ONBOARD_HOME` does not isolate them.

## Change

Point `$HOME` and `$CODEX_HOME` at non-existent paths under `$TMP`, immediately after `TMP` is created and before the first case:

```bash
HOME="$TMP/nohome"
CODEX_HOME="$TMP/nocodex"
```

`fresh_env` still overrides both per-case, so what the tests exercise is unchanged — this is purely a floor. The paths are **deliberately not created**, so an omitted `fresh_env` fails loudly on a missing directory instead of silently writing somewhere plausible.

Swept the four sibling `lib/*_test.sh` files: none of them touch `$HOME`, so the fix stays scoped to this one file.

## Verification

- `bash tools/onboarding-factory/scripts/lib/hook-config-snapshot_test.sh` — all 5 assertions pass.
- `bash tools/onboarding-factory/scripts/smoke-test.sh` — `smoke-test: PASS`.
- Real `~/.claude/settings.json` verified byte-identical (sha1) before and after both runs.
- Negative proof: injected a case that writes to `$HOME/.claude/settings.json` *before* any `fresh_env`. It targeted `$TMP/nohome/.claude/settings.json` and failed loudly with `No such file or directory`; the real config was untouched. (The destructive path was **not** re-run against a real `$HOME`.)

## Base branch

Targets `feat/1178-configurable-hook-port` (PR #1207), not `main` — the file it hardens does not exist on `main` yet; #1207 introduces it. Safe under either merge order: if #1207 lands first, this retargets to `main` and still applies; if this lands first, #1207 carries it up.

### DIFF
--- a/tools/onboarding-factory/scripts/lib/hook-config-snapshot_test.sh
+++ b/tools/onboarding-factory/scripts/lib/hook-config-snapshot_test.sh
@@ -17,6 +17,18 @@ source "$DIR/hook-config-snapshot.sh"
 TMP="$(mktemp -d)"
 trap 'rm -rf "$TMP"' EXIT
 
+# Floor under every case: point $HOME and $CODEX_HOME somewhere harmless before
+# the first one runs. fresh_env overrides both per-case, so this changes nothing
+# about what the tests exercise — it exists so a case that is added above the
+# first fresh_env call, or forgets it, cannot reach the developer's real
+# ~/.claude/settings.json. That is not hypothetical: a draft of this file did
+# exactly that and wiped a live config's statusLine and all seven irrlicht hook
+# entries (#1212). These paths are deliberately NOT created — an omitted
+# fresh_env then fails loudly on a missing directory instead of silently
+# writing somewhere plausible.
+HOME="$TMP/nohome"
+CODEX_HOME="$TMP/nocodex"
+
 fails=0
 pass() { echo "  PASS: $1"; return 0; }
 fail() { echo "  FAIL: $1 — expected [$2] got [$3]"; fails=$((fails + 1)); return 0; }


ingo-eichhorst:
## Seen red, then green

Per AGENTS.md ("a defect test proves nothing until it has been seen red"), here is the hazard reproducing on the base branch and being blocked by this PR.

Method: take each version of `hook-config-snapshot_test.sh`, inject one case that writes to `$HOME/.claude/settings.json` **before** any `fresh_env` call, and run it with `$HOME` pointed at a stand-in home pre-seeded with `PRECIOUS-USER-CONFIG`. The real `~/.claude/settings.json` is never in the path — the destructive run is aimed at a sandbox.

```
RED   (base, pre-fix):
  pre.sh    stand-in ~/.claude/settings.json = CLOBBERED
GREEN (this PR):
  post.sh   stand-in ~/.claude/settings.json = PRECIOUS-USER-CONFIG
```

Sanity-checked that both variants were real and comparable: `pre.sh` 80 lines, `post.sh` 92 lines, exactly one injected case in each, lib present alongside both.

So the failure mode is not theoretical — an omitted `fresh_env` writes to the inherited `$HOME` on the base branch, and cannot on this branch.

Also worth noting: the local pre-push gate `tools/security-scan.sh --local` currently fails on a **pre-existing, unrelated** `postcss` (high) advisory in `tools/onboarding-factory/internal/viewer/web`. It is transitive via vitest, not touched by this one-file bash diff, and there is no security workflow in `.github/workflows/`, so it gates nothing on the PR. The recent dependabot bump covered `platforms/web` but not this tree. Filed separately rather than folded in here.

ingo-eichhorst:
Follow-up: the pre-existing postcss advisory mentioned above is now tracked as #1225 (local pre-push gate only; no CI impact).

Issue #1255, opened 2026-07-31
fswatcher: addExistingDirs swallows watcher.Add errors, so Ready() can promise a watch that isn't armed


ingo-eichhorst:
## Problem

`fswatcher.addExistingDirs` discards the result of every watch registration:

```go
if d.IsDir() {
    _ = watcher.Add(path)      // core/adapters/inbound/agents/fswatcher/watcher.go:498
    w.emitExistingFiles(path)
}
```

`Watch` then calls `signalReady()` unconditionally once the scan returns. So if `Add` fails for a project directory — FD exhaustion (`EMFILE`), a permissions change, an inotify watch-limit on Linux — the watcher reports itself ready over a tree it is **not** fully watching.

## Why it matters

`Ready()`'s own doc comment states the guarantee this breaks:

> the watch is live for the root and every pre-existing subdir … file mutations under the tree are guaranteed to be observed — tests and wiring can wait on it instead of sleeping a guessed interval

In production the consequence is a **silent monitoring blind spot**: every session under that directory is simply never observed. Nothing logs, nothing degrades visibly, and the daemon looks healthy. That is the failure class hardest to notice from the outside, and the reason `Ready()` was given an explicit contract in the first place (#998).

The root watch at `watcher.go:217` already treats this correctly — `if err := watcher.Add(w.root); err != nil { return err }`. Only the per-subdirectory adds inside the scan swallow it.

## Not the cause of #1248

Worth stating up front so nobody re-runs it: I instrumented this exact call to report every failure and ran the full `./core/...` race suite twice while hunting #1248 — **0 add-failures**. This is a latent defect found while investigating that flake, not its explanation. See https://github.com/ingo-eichhorst/Irrlicht/issues/1248#issuecomment-5145458876.

## Proposed

Surface the failure without making it fatal — an unreadable directory must not kill the watcher, which is presumably why it was discarded in the first place. Options, roughly in increasing order of ambition:

1. Log it via the adapter's `Logger` (per the Key Conventions error rule — logged, not propagated).
2. Count failures and expose the count, so a blind spot is at least queryable.
3. Narrow `Ready()`'s doc comment to what it actually guarantees if the decision is to keep swallowing.

Whatever the shape, the outcome to avoid is the current one: a documented guarantee that can lapse with no trace.

## References
[...]

Issue #998, opened 2026-07-13
daemon: fswatcher cold-start backlog scan delays new-session discovery (O(N) with session-directory count)


ingo-eichhorst:
## Follow-up from #996

While fixing #996 (mistral-vibe's `1-3_long-idle-live-session` onboarding scenario going `known_failing` because a brand-new session's first turn completed before the daemon discovered its transcript), the root cause of the ~17.3s discovery delay itself was identified but deliberately **not** fixed there, to keep that PR scoped to the state-machine synthesis fix.

### The mechanism

`core/adapters/inbound/agents/fswatcher/watcher.go`'s `Watch()` does a **synchronous** cold-start scan (`addExistingDirs`, `watcher.go:388-399`) over every existing session directory under an adapter's transcript root — for each directory it does a `watcher.Add()` (fsnotify/kqueue syscall) plus `emitExistingFiles()` (an `os.ReadDir` + a per-file `os.Stat`) — **before** it ever arms the live fsnotify watch on the root (`watcher.Add(w.root)`, `watcher.go:167`) or signals `Ready()` (`watcher.go:173`).

`filepath.WalkDir` (used internally) visits directory entries in **lexical order**. Mistral Vibe (and several other adapters) name session directories with a timestamp prefix (`session_<YYYYMMDD>_<HHMMSS>_<hash>`), which is lexically sortable by creation time — so a **brand-new** directory always sorts **last**, and every older directory's scan cost is paid first. On a dev machine with a large pre-existing session backlog (e.g. an isolated `IRRLICHT_ONBOARD_HOME` recording daemon that has no persisted state and must treat the entire real `$HOME` backlog as "new" on every cold start), this pushed discovery of one new directory out to ~17.3 seconds — long enough for an agent's entire first turn to complete before the daemon ever looked.

No debounce/poll constant in the fswatcher or session-detector layers explains this delay (confirmed by an audit of every timing constant in that code) — it's a genuine O(N)-with-backlog-size synchronous scan cost, not a timer.

### Why this wasn't fixed in #996

- It's environment/backlog-size-dependent (this dev machine's cumulative onboarding-factory recording history), not something every production install would see the same way.
- Reducing the latency would **not fully eliminate** the underlying race #996 fixes — a busy machine, a daemon restart mid-turn, or a burst of concurrent sessions can always reproduce a nonzero discovery delay. The state-synthesis fix in #996 is the actual defect; the latency here is just what made it observable.
[...]

PR #1286, opened 2026-07-31
fix(fswatcher): reconcile sweep so a dropped kqueue event can't hide a session

### LINKED ISSUE #1248: TestWatch_MetadataParentIsPresentOnFirstNewEvent (fswatcher) flakes ~1-in-20 under load
## Problem

`TestWatch_MetadataParentIsPresentOnFirstNewEvent` in `core/adapters/inbound/agents/fswatcher` fails intermittently — roughly 1 run in 20 under load, green otherwise.

It is **pre-existing**, not introduced by any recent change: it was reproduced on `main` without a diff applied, under `-count=20`, and it fails in a package with no dependency on the code the reporting branch touched.

## Why it matters

Every agent following AGENTS.md runs the full suite (`go test ./core/... -race -count=1`) before marking a ticket done. A ~5% failure rate means a meaningful fraction of those runs go red for a reason unrelated to the work in hand, and each one costs an investigate-and-dismiss cycle. That is a tax on every ticket, and it also trains agents to dismiss red suites — the habit AGENTS.md's "dismissals carry evidence" rule exists to prevent.

## Reproduce

```
go test ./core/adapters/inbound/agents/fswatcher/... -race -count=20 -run TestWatch_MetadataParentIsPresentOnFirstNewEvent
```

Load helps: running it concurrently with another full `./core/...` race build reproduces it faster than an idle machine.

## Proposed

Diagnose the race rather than papering over it — the assertion is about a parent link being present on the *first* new event, so the likely shapes are a watcher-registration/write ordering window or a fixture written before the watcher is fully armed. If it turns out to be a genuine product race in parent-link population rather than test timing, that is a more valuable find than the flake itself.

Please avoid "fixing" it with a bare sleep or a retry loop; if the settle is genuinely asynchronous, gate on the observable condition.

## References
- Surfaced during `/ir:exec full 1215` (PR #1241), 1 of 6 full-suite runs

ingo-eichhorst:
Closes #1248.

## The failure mode

#1248's investigation pinned the symptom precisely: the flake is always
`watcher_test.go` *"timed out waiting for linked new-session event"* — an event
that **never arrives**, not one that arrives late. Across 660 instrumented runs
with every core saturated the event arrived every time, worst case **9.9 ms**.
Two causes were ruled out with evidence (CPU-load latency; a failed
`watcher.Add`). The surviving lead was a genuinely dropped event inside
fsnotify's macOS kqueue backend, which discovers new files by diffing a
directory listing rather than receiving a per-file notification.

Scope for this PR was confirmed with the maintainer: **fswatcher only** (not the
wider `./core/...` flake family), fixed by **closing the production hole** rather
than instrumenting and soaking. That matters because the same drop in production
means a session that is *simply never observed* for the daemon's whole lifetime —
the same silent-blind-spot class as #1255, which this builds directly on.

## What changed

`Watch` now runs a periodic **reconcile sweep** (`defaultReconcileInterval`, 15s;
`WithReconcileInterval` overrides, negative disables) that re-walks the tree and
reports what fsnotify didn't:

- a transcript with no emission recorded → `EventNewSession`
- one whose size moved since the last emission → `EventActivity`

Every emission now goes through a new `emit()` that records the reported size, so
**in the healthy case the sweep finds nothing to say and stays silent** —
notifications remain the fast path, this is only the safety net. It runs on
`Watch`'s goroutine (so `pendingNew`/`emitted`/`watched` stay lock-free) and
drains fsnotify's unbuffered channel between top-level directories, the same way
the existing backlog scan does.

**Deletions are deliberately not reconciled.** Inferring `EventRemoved` from a
file's absence would turn a transient `ReadDir` failure into a spurious teardown
of every session under that directory; a lost `Remove` leaves a session stale
rather than invisible, which is the cheaper failure.

### Interaction with #1255 (merged mid-branch)

#1255 landed while this was in flight and touches the same function, so this
rebase folds the two together rather than duplicating them. `addWatch` keeps
#1255's logging and gains this PR's semantics:

- **already-armed dirs are skipped**, so the sweep isn't re-registering the whole tree every interval;
- **a rejected registration is not recorded**, so the next sweep *retries* it — a watch lost to FD exhaustion or a permissions blip now recovers on its own instead of staying blind until restart;
- **the rejection is logged once per directory**, not once per retry — without this, #1255's log line would repeat every 15s forever for a genuinely unreadable directory.

## Why the tests are trustworthy

A real dropped kqueue notification can't be forced (that's what 660 runs
established). The stand-in is a directory whose watch registration **fails**:
fsnotify then structurally cannot report anything created inside it, which is the
same observable situation — and unlike a real drop, it can be produced on demand.

`TestWatch_ReconcileRecoversFileUnderAnUnwatchableDir` carries its own control:
the `sweep disabled` subtest pins that the file really is invisible to fsnotify,
so the recovered case is evidence about the sweep rather than about fsnotify
quietly delivering the event anyway.

### Seen red (per AGENTS.md)

Every new assertion was run against a deliberate mutation and observed failing:

| Mutation | Result |
|---|---|
| `reconcile` is a no-op | 6 tests fail, incl. `sweep enabled recovers it` — *timed out waiting for "new_session"* |
| dedup guard never fires | `DoesNotReEmitWhatWasAlreadyReported`, `SilentAfterNormalEventPathEmitted` — *unexpected "activity" event* |
| `Remove` no longer forgets the emitted size | `RecreatedFileIsANewSessionAgain` — *event type = "removed", want "new_session"* |
| negative interval no longer disables the sweep | `EffectiveReconcileInterval/negative` — *= -1ns, want 0s* |
| `reconcileFile` ignores the staleness cutoff | `SkipsStaleFile` — *unexpected "new_session"* |
| `reconcileFile` ignores an empty session ID | `SkipsFileWithoutSessionID` — *unexpected "new_session"* |
| every retry re-logs | `AddWatch_Retries…` — *logged 3 errors …, want 1* |
| a rejected dir is recorded as watched | `AddWatch_Retries…` — *must not be recorded as watched, or it is never retried* |

Two earlier mutation attempts (deleting the guards outright) were discarded as
invalid — they orphan `prev`/`mtime` and fail to compile, so they proved nothing.

## Verification

- `go test ./core/... -race -count=1` — **2 consecutive green runs**, 49 packages, 0 failures (plus 3 green before the #1255 rebase).
- `TestWatch_MetadataParentIsPresentOnFirstNewEvent` — `-race -count=50`, green.
- `tools/preflight.sh --changed` via the pre-push hook: gofmt, core module tests, ARS architecture gate, security scan all **PASS**.

### One unrelated flake seen, reported not dismissed

During the first full-suite run, `TestAnyLiveOutputWriter_RealProcess`
(`core/application/services`) failed once: *"probe reported no live writer while
background process is running"*. Evidence that it is not from this change:
this branch touches only `core/adapters/inbound/agents/fswatcher/` (3 files), and
that probe's code path is byte-identical to `origin/main`. It polls `lsof` — which
scans every process — against a **2-second** deadline, so it is load-sensitive by
construction. It did **not** reproduce in 30 `-race` runs under full CPU
saturation, nor in 4 subsequent full-suite runs. Marked as *not reproduced*: this
looks like a third member of the load-dependent flake family #1248's follow-up
comment describes, and is worth its own issue rather than a fix here.

### DIFF
--- a/core/adapters/inbound/agents/fswatcher/reconcile_test.go
+++ b/core/adapters/inbound/agents/fswatcher/reconcile_test.go
@@ -1,0 +1,465 @@
+package fswatcher
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+
+	"irrlicht/core/domain/agent"
+)
+
+// newFsnotify returns a real fsnotify watcher for the direct reconcile tests,
+// closed when the test ends. reconcile only uses it to arm directories.
+func newFsnotify(t *testing.T) *fsnotify.Watcher {
+	t.Helper()
+	fsw, err := fsnotify.NewWatcher()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = fsw.Close() })
+	return fsw
+}
+
+// writeTranscript writes content to path, failing the test on error.
+func writeTranscript(t *testing.T, path, content string) {
+	t.Helper()
+	if err := os.WriteFile(path, []byte(content), 0644); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// mkdir creates dir and any parents, failing the test on error, and returns it.
+func mkdir(t *testing.T, dir string) string {
+	t.Helper()
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		t.Fatal(err)
+	}
+	return dir
+}
+
+// expectEvent asserts the next event on ch has the wanted type and session ID.
+func expectEvent(t *testing.T, ch <-chan agent.Event, want agent.EventType, sessionID string) agent.Event {
+	t.Helper()
+	select {
+	case ev := <-ch:
+		if ev.Type != want {
+			t.Errorf("event type = %q, want %q", ev.Type, want)
+		}
+		if ev.SessionID != sessionID {
+			t.Errorf("session ID = %q, want %q", ev.SessionID, sessionID)
+		}
+		return ev
+	case <-time.After(2 * time.Second):
+		t.Fatalf("timed out waiting for %q event for %q", want, sessionID)
+		return agent.Event{}
+	}
+}
+
+// expectNoEvent asserts nothing arrives on ch within d. A negative assertion
+// needs some bound; d is kept short because every path that would emit here is
+// synchronous with the call under test.
+func expectNoEvent(t *testing.T, ch <-chan agent.Event, d time.Duration) {
+	t.Helper()
+	select {
+	case ev := <-ch:
+		t.Fatalf("unexpected %q event for %q", ev.Type, ev.TranscriptPath)
+	case <-time.After(d):
+	}
+}
+
+func TestEffectiveReconcileInterval(t *testing.T) {
+	tests := []struct {
+		name string
+		set  time.Duration
+		want time.Duration
+	}{
+		{"zero uses the default", 0, defaultReconcileInterval},
+		{"negative disables the sweep", -1, 0},
+		{"positive is honored", 250 * time.Millisecond, 250 * time.Millisecond},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			w := NewWithRoot("/tmp/fake", testAdapter, 0).WithReconcileInterval(tc.set)
+			if got := w.effectiveReconcileInterval(); got != tc.want {
+				t.Errorf("effectiveReconcileInterval() = %v, want %v", got, tc.want)
+			}
+		})
+	}
+}
+
+// A transcript file that fsnotify never reported is surfaced by the sweep —
+// the dropped-notification failure mode behind issue #1248, where the symptom
+// is an event that never arrives rather than one that arrives late.
+func TestReconcile_EmitsNewSessionForFileFsnotifyMissed(t *testing.T) {
+	root := setupFakeProjects(t)
+	w := NewWithRoot(root, testAdapter, 0)
+	ch := w.Subscribe()
+
+	path := filepath.Join(root, "-Users-test-myproject", "ghost.jsonl")
+	writeTranscript(t, path, `{"type":"init"}`+"\n")
+
+	// No fsnotify event is ever delivered for ghost.jsonl: the sweep is the
+	// only thing that can find it.
+	w.reconcile(context.Background(), newFsnotify(t))
+
+	ev := expectEvent(t, ch, agent.EventNewSession, "ghost")
+	if ev.ProjectDir != "-Users-test-myproject" {
+		t.Errorf("project dir = %q, want %q", ev.ProjectDir, "-Users-test-myproject")
+	}
+	if ev.Size == 0 {
+		t.Error("size = 0, want the file's actual size")
+	}
+}
+
+// The sweep must be silent when fsnotify is doing its job, or every interval
+// would replay the whole tree as new sessions.
+func TestReconcile_DoesNotReEmitWhatWasAlreadyReported(t *testing.T) {
+	root := setupFakeProjects(t)
+	w := NewWithRoot(root, testAdapter, 0)
+	ch := w.Subscribe()
+	fsw := newFsnotify(t)
+
+	path := filepath.Join(root, "-Users-test-myproject", "abc-123.jsonl")
+	writeTranscript(t, path, `{"type":"init"}`+"\n")
+
+	w.reconcile(context.Background(), fsw)
+	expectEvent(t, ch, agent.EventNewSession, "abc-123")
+
+	// Nothing changed on disk, so two further sweeps must emit nothing.
+	w.reconcile(context.Background(), fsw)
+	w.reconcile(context.Background(), fsw)
+	expectNoEvent(t, ch, 100*time.Millisecond)
+}
+
+// A write fsnotify missed shows up as activity, not as a second new session.
+func TestReconcile_EmitsActivityWhenFileGrewUnobserved(t *testing.T) {
+	root := setupFakeProjects(t)
+	w := NewWithRoot(root, testAdapter, 0)
+	ch := w.Subscribe()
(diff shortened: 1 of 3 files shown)


ingo-eichhorst:
## Review + simplify round — 4 real defects found and fixed

Ran `/ir:code-review high` and the 4-agent `/simplify` fan-out against `origin/main...HEAD`. Both surfaced genuine problems; everything below is fixed and pushed.

### From code review

**1. The sweep was inert for flat-layout adapters.** `newestFirst` skips non-directory entries, so root's own `.jsonl` files were never reconciled. kiro-cli's layout is exactly that — verified on this machine: `~/.kiro/sessions/cli/` holds **79 `.jsonl` at depth 1 and 1 directory**. The fix would have shipped doing nothing for that adapter. `reconcile` now walks root's own entries too.

**2. The dedup was one-directional.** `reconcile` skipped what fsnotify had reported, but not the reverse. fsnotify's event channel is unbuffered, so a `Create` can sit blocked in the producer while a sweep walks the same file and reports it, then get delivered and reported *again*. The guard now lives in `emit` and is symmetric (new-session only — activity duplicates are absorbed by the detector's debounce).

**3. `watched` was never pruned on removal,** so a directory deleted and recreated at the same path was short-circuited as already-watched and never re-armed. `handleEvent` now clears it.

**4.** Cost comment corrected — it understated the real per-sweep work.

### From simplify

**5. Ordering bug worth its own line: `reconcileFile` called `idFor` before the cheap filters.** For codex, `idFor` opens the transcript and JSON-decodes its first record — so every sweep re-read every transcript in the tree to produce nothing. Measured on a real 301-file codex tree: **~251ms → ~2.7ms per sweep (93×)**, on the goroutine that also dispatches fsnotify events. The sweep was delaying the fast path it exists to back up. Filters now run stat → staleness → size-compare → `idFor`.

**6. Dropped `newestFirst` from the sweep.** Its ordering bounds *first*-discovery latency over a one-shot backlog (#998); the sweep visits everything unconditionally and repeats, so the sort bought nothing and cost an lstat per directory per sweep.

**7. Pre-existing bug found and fixed: the startup scan skipped flat layouts too.** Not just the sweep — `addExistingDirs` never made root itself a walk root, so `Ready()`'s promise that *"every pre-existing transcript file has already been emitted"* was **false for kiro-cli**: its sessions were invisible after a daemon restart until something wrote to them. Verified empirically before fixing — a pre-existing flat transcript produced no event at all. One line (`emitExistingFiles(w.root)`), plus a regression test.

Also: `emitExistingFiles` now uses the `isStale` helper instead of inlining its body.

### A flake I introduced, caught and fixed

`TestHandleEvent_RemovedDirIsRearmedWhenRecreated` failed 1 run in ~150. Root-caused rather than retried: re-adding a path through the *same* fsnotify watcher races with that watcher's own teardown of the deleted directory's fd and transiently returns **`bad file descriptor`**. Production already copes — a rejected registration isn't recorded, so the next sweep retries it — but a single-shot assertion on it is nondeterministic. The retry now goes through a fresh watcher. **300 consecutive `-race` runs green.** (Incidentally this is the first confirmed real-world trigger for that retry path.)

### Seen red

Every new assertion was run against a targeted mutation and observed failing — 8 mutations before this round, 3 more for the review fixes (root-file walk removed; dedup guard disabled; `delete(w.watched)` removed), plus the startup fix. Two mutation attempts were discarded as invalid because they orphaned a variable and didn't compile.

### Verification

- `go test ./core/... -race -count=1` — **3 consecutive green runs**, 49 packages, 0 failures.
- `tools/preflight.sh --changed` via pre-push: gofmt, core tests, ARS gate, security scan all PASS.

### Deliberately not done in this PR

- **Extracting the shared walk skeleton.** `reconcile` and `addExistingDirs` are now the same loop twice. All three reviewers flagged it; the altitude review argued against doing it here, and I agree — it sits on top of #998's newest-first and drain-between-directories guarantees and deserves its own red-first pass. Follow-up issue.
- **Pruning the sweep by directory mtime.** ~93% of its stats re-confirm that ancient files are still ancient. A blanket dir-mtime skip is *unsafe* (verified: on APFS an append does not bump the parent's mtime, so it would break activity reconciliation); the safe split is dir-mtime for discovery + stat-only-known-paths for growth. Real design change, prerequisite for shortening the interval.
- Collapsing `watched`/`watchFailed` into one map, and the test-boilerplate helpers — cosmetic.

### Unrelated flake, now seen twice

`TestAnyLiveOutputWriter_RealProcess` (`core/application/services`) failed on two separate full-suite runs. It polls `lsof` — which scans every process — against a 2s deadline, so it is load-sensitive by construction, and this branch touches only `fswatcher`. It did **not** reproduce in 30 `-race` runs under full CPU saturation. Third member of the load-dependent flake family #1248's follow-up comment describes; worth its own issue.

ingo-eichhorst:
## CI status — all gating checks green

**12 pass / 1 fail.** The only red is **CodeScene**, which AGENTS.md defines as advisory and explicitly not a merge gate. Reporting it rather than burying it:

- `reconcile_test.go` **Code Duplication — fixed.** Collapsed 23 boilerplate blocks into `writeTranscript`/`mkdir` helpers. That violation is gone. (Both the simplify fan-out and CodeScene independently flagged it; I'd deferred it as cosmetic and was wrong to.)
- `watcher.go` **Complex Method: `Watch`** — still flagged, health 8.35 → 8.04. I extracted `reconcileTicker()` and `resetRunState()` out of it, which is everything this PR added to that function beyond one `select` case. Getting further means the shared-walk extraction (`reconcile` and `addExistingDirs` are now the same loop twice) — which the altitude review argued belongs in its own PR on top of #998's ordering and drain guarantees, and I agree.

Net effect on `Watch` from this PR is one call, two lines of ticker setup, and one `select` case; it was already the file's most complex function beforehand.

**Verification after the last two rounds:** `go test ./core/... -race -count=1` green (49 packages, 0 failures) on every run since; `TestHandleEvent_RemovedDirIsRearmedWhenRecreated` 300 consecutive `-race` runs green; the reconcile no-op mutation still fails 8 tests, confirming the test-helper refactor didn't neuter any assertion.

PR #1295, opened 2026-07-31
feat(state): explicit signal tiers + shared reconcile mechanism (#1288)

### LINKED ISSUE #1288: Phase 3: explicit signal tiers + reconcile in the state classifier (#1129)
## Goal

Replace the classifier's **implicit, hand-ordered** authority model — and the three **bespoke, per-signal** overlay lifecycles that grew alongside it — with **one explicit tier + reconcile mechanism**. This is Phase 3 of #1129, and the epic now calls it "the central design problem of the whole epic": every phase after it (4, 5, 6, 7) adds another signal source, and each one currently costs a new hand-rolled overlay with its own lifecycle policy.

## Why now

Phase 1 shipped both hook consumers, and they proved the epic's central risk is real:

- **claudecode** — `Stop` → ready (#1161) and `Notification`/`idle_prompt` → waiting (#1173)
- **codex** — `PermissionRequest` → waiting, `Stop` → ready (#1171)

Both are **late but authoritative** (`idle_prompt` measured at ~6s in the #1141 spike), so both had to defend against a lower-tier guess that already fired. Each one solved that separately, by hand:

| Overlay | Lifecycle policy | Why it differs |
|---|---|---|
| `overlayPermissionPending` | persistent; cleared by *transcript evidence* (`LastWasToolDenial`) | no PostToolUseFailure on denial |
| `overlayHookTurnDone` | **consume-once** — deleted as it is read | must not bleed into the next turn |
| `overlayIdlePrompt` | **persistent** — re-applied every pass, gated on `IsAgentDone` | a stray reclassify would flip waiting back to ready via rule 2 |

Three signals, three different answers to the same question ("how long does an authoritative signal outrank the transcript?"), plus a special-case bypass in `forceReadyToWorkingIfActive` (`hasPendingIdlePrompt`) that exists only to stop one of them being clobbered. Phases 4–7 add OTel, network-probe, screen, and ACP signals. Without a shared mechanism that is **six more** bespoke overlays and a combinatorial ordering problem in `ClassifyState`.

## Seam correction — the epic's stated seam is wrong

The epic body says *"Tag each `ParsedEvent` with its tier."* That does not fit the code as built:

- `ParsedEvent` (`core/pkg/tailer/parser.go:107`) is **transcript-only by construction** — it is what a `LineParser` returns from a transcript line. Every tier above transcript is, by definition, not a transcript line.
- Every hook signal arrives **out-of-band**, via the HTTP hook endpoint into a detector-held per-session map, and is folded onto `SessionMetrics` as an overlay immediately before `ClassifyState` (`session_detector_activity.go:868-952`). It never passes through a parser.

**So the tier model belongs where the signals converge: `SessionMetrics` + the classifier — not `ParsedEvent`.** Tagging `ParsedEvent` would tier only the tier-5 source and miss every higher tier. Adopt the corrected seam; the epic body should be updated to match.

## Current state — the tier ladder already exists, undeclared

`ClassifyState` (`core/application/services/state_classifier.go:26`) is already a de-facto tier ladder, encoded as *comment-numbered rule order*:

```
0.  PermissionPending    → waiting   (hook, tier 1)
0b. CompactInProgress    → working   (hook, tier 1)
1.  NeedsUserAttention() → waiting   (transcript, tier 5)
1b. OpenToolStalled      → waiting   (5s timer fallback, tier 5)
1c. IdlePromptPending    → waiting   (hook, tier 1)   ← Phase 1b's reconcile slice
2.  IsAgentDone()        → ready     (hook OR transcript — HookTurnDone folded inside)
3.  ESC / tool denial    → ready     (transcript)
4.  default              → working   (transcript)
```

Two problems this issue fixes:

1. **Authority is positional, not declared.** Rule 1c sits above rule 2 purely so the hook can correct the turn-done verdict. Nothing in a type says so; the invariant lives in a comment and is enforced by nothing. Reordering the function silently changes arbitration.
2. **Tier is invisible downstream.** `IsAgentDone()` (`metrics.go:503`) consults `HookTurnDone` *inside* itself, so a caller cannot tell an authoritative hook-delivered turn end from a transcript-tail guess. Transition reasons say "agent finished turn → ready" for both.

## Scope — build

1. **Declare the tier ladder** as a domain type in `core/domain/session` (`SignalTier` + constants), matching the epic's five tiers: hook → structured (OTel/ACP) → process/PTY → screen → transcript inference. Tier 5 is `waiting_cue`.
2. **Carry provenance onto the state signal.** Each state-bearing input records the tier that produced it, so the classifier's verdict and the emitted `lifecycle.Event` reason both name the tier (`"idle prompt hook (tier 1) → waiting"`). This is the observability payoff — a transition log that says *which channel decided* is what makes Phases 4–7 debuggable.
3. **One reconcile mechanism replacing the three bespoke overlays.** A single per-session hold with an explicit policy — persistent vs consume-once, and its clearing condition — expressed as data per signal rather than as three hand-written functions. Preserve today's behaviour exactly (see AC1); this is a refactor to a shared mechanism, not a behaviour change.
4. **Grace-timer hysteresis**, as the epic specifies, so an authoritative signal landing after a lower-tier guess corrects the badge without flapping. `forceReadyToWorkingIfActive`'s `hasPendingIdlePrompt` special case should fall out as an instance of the general rule.
5. **`waiting_cue` becomes an explicit, bypassable tier 5.** Keep it — it is still the only signal for the seven adapters with no hook, and still the instant first-guess for the two that have one. But make "a higher tier is present, so skip the cue" a declared rule rather than an emergent property of rule ordering. **Do not delete `waiting_cue.go` in this issue** (it is 346 lines and load-bearing; deletion is gated on every adapter having a higher tier, which is Phases 4-6).

## Acceptance criteria

- **AC1 — no observable behaviour change.** The full existing suite passes unchanged, and `tools/replay-fixtures.sh` + the replay goldens are byte-identical. Any golden diff is a regression, not an update, and must be explained before it is committed. This is the primary safety property: the mechanism is new, the verdicts are not.
- **AC2 — arbitration is declared, not positional.** A test asserts the tier ordering directly (a higher-tier signal wins over a conflicting lower-tier one) without depending on the source order of `ClassifyState`'s rules.
- **AC3 — provenance is visible.** Every state transition driven by a tiered signal records its tier in the transition reason; a test asserts a hook-delivered `ready` is distinguishable from a transcript-tail-inferred `ready`.
- **AC4 — the three overlays converge.** `overlayPermissionPending`, `overlayHookTurnDone`, and `overlayIdlePrompt` are expressed through the one mechanism, each with its lifecycle policy stated as data. Adding a Phase 4-7 signal must not require a new overlay function.
- **AC5 — no flap under a late signal.** A test drives the real Phase 1b sequence — turn ends → lower tier guesses `ready` → hook lands ~6s later saying `waiting` — and asserts the correction emits a single clean transition, not a ready→waiting→ready oscillation.
- **AC6 — architecture holds.** `core/architecture_test.go` passes: `SignalTier` lives in `domain/`, and the classifier does not import any adapter (`session_detector_lifecycle.go:361` already notes the application layer must not import claudecode for its `HookStop` constant — the tier type is the proper home for that).
- Full gate green: `tools/preflight.sh`.

## Observability

No new gap. Hook-delivered observations became recordable when #1178 (configurable hook port) landed, so hook-driven cells can be captured by the onboarding factory — this issue's signals are all either already-recorded transcript paths or hook paths that #1178 unblocked. Replay goldens (AC1) are the regression net.

## Out of scope

- Deleting `waiting_cue.go` (gated on Phases 4-6 giving every adapter a higher tier).
- Any new signal *source* — no OTel ingest (Phase 4), no network probe (Phase 5), no screen manifests (Phase 6), no ACP (Phase 7). This issue builds the arbitration those phases plug into.
- `compactHoldTimeout` — the epic explicitly keeps the safety-net timer.
- `OpenToolStalled` removal — it is a tier-5 fallback that stays until its native-`http`-delivery replacement is confirmed (#488).

Part of #1129 (Phase 3). Unblocks Phases 4-7, all of which the epic gates on this landing first.

ingo-eichhorst:
Closes #1288. Phase 3 of #1129.

## What this is

The classifier's authority ladder existed only as the comment-numbered order of the if-statements inside `ClassifyState`. Nothing in a type said a hook outranks a transcript guess — the invariant survived exactly as long as nobody reordered the function. Meanwhile each of the three hook signals had grown its own hand-written overlay with its own lifecycle rule, plus a fourth copy in the replay harness:

| Overlay | Policy |
|---|---|
| `overlayPermissionPending` | persistent; cleared by transcript evidence (`LastWasToolDenial`) |
| `overlayHookTurnDone` | consume-once — deleted as read |
| `overlayIdlePrompt` | persistent — re-applied every pass, gated on `IsAgentDone` |
| `sidecarReplayer.overlayPermissionPending` | a fourth, "mirrors SessionDetector" by hand |

Phases 4–7 add four more signal sources. This replaces all of that with one declared mechanism.

## Changes

- **`session.SignalTier`** — the five-tier ladder (hook → structured → process → screen → transcript) as a domain type. `Outranks` handles the `TierNone` zero value explicitly, which a bare `<` gets backwards.
- **`session.SignalHolds` + `signalPolicies`** — each signal's tier, persistence and staleness rule declared as *data*. Adding a Phase 4–7 signal is a row in the table, not a fourth map and a fourth overlay.
- **Table-driven ladder** — each `stateRule` carries the tier its evidence arrives on. `agent_done` is deliberately dynamic: `TierHook` when a Stop hook delivered it, `TierTranscript` when the transcript tail was guessed at.
- **Structured provenance** — `ClassifierInputs` gains `DecidedByTier` / `DecidedByRule`, plus the `HookTurnDone` / `IdlePromptPending` fields that were missing from the recorded snapshot entirely.
- **The replay harness now shares the mechanism**, killing the fourth copy.

## A deliberate deviation from AC3

The issue's AC3 says the tier should go "in the transition reason". It can't: **338 committed replay goldens pin `Reason` strings byte-for-byte**, so prose-encoding the tier would rewrite every one of them — destroying the regression net that protects this very refactor (AC1). Provenance therefore lands as **structured fields** on the lifecycle event, which is not in the goldens' curated `transition` struct. This satisfies AC3's stated intent ("a hook-delivered ready is distinguishable from a transcript-tail-inferred ready") and is queryable in a way prose never was.

## Verification

`tools/preflight.sh`: **all 12 gates PASS**, including the ARS architecture gate (AC6).

**AC1 — behaviour is unchanged.** All 338 replay goldens byte-identical, zero `replaydata/` churn; full `go test ./core/... -race` green. Reason strings are untouched by design.

**These are locks, not defect tests** — they pin behaviour that must not change, so they pass on `main` by construction. Per AGENTS.md, their value is that they *can* fail, so each new invariant was proven red by deliberate mutation:

| # | Mutation | Result |
|---|---|---|
| 1 | Reverse `signalOrder` (idle_prompt before turn_done) | 🔴 `ApplicationOrderIsLoadBearing`: *"idle_prompt must survive: turn-done ran first…"* |
| 2 | Move the hook-tier `idle_prompt` rule below `agent_done` | 🔴 3 tests, incl. *"rule `agent_done` (tier transcript) decided, but lower rule `idle_prompt` (tier hook) outranks it"* |
| 3 | Drop a policy from `signalOrder` | 🔴 *"policy `idle_prompt` is missing from signalOrder — it would never be applied"* |
| 4 | Remove `idle_prompt`'s staleness rule | 🔴 *"user-blocking-tool-open/idle-prompt: rule `user_blocking_tool` (tier transcript) decided, but lower rule `idle_prompt` (tier hook) outranks it"* |

Mutation 4 is the load-bearing one. Two transcript-tier rules sit *above* the hook-tier `idle_prompt` rule, which is only sound because an open tool makes `IsAgentDone` false, which makes the idle hold stale, so the pair can never contend. `TestStateRules_LadderIsTierConsistent` builds its corpus by running real holds through `Overlay` rather than setting metrics fields directly — so it **tests** that reachability argument instead of assuming it, and names the contending pair the moment it stops holding.

Mutation 1 is worth flagging too: `signalOrder` replaced a map, and a map range would have randomised it. `idle_prompt`'s staleness check reads the `HookTurnDone` that `turn_done` sets one step earlier — so on any adapter whose transcript tail isn't literally `turn_done`, the wrong order silently discards the correction the hook exists to deliver.

## Not in scope

`waiting_cue.go` is untouched and stays that way — it is the instant first guess for the two adapters with hooks and the only signal for the seven without. Deleting it is gated on Phases 4–6. No new signal *source* here; this is the arbitration they plug into.

### DIFF
--- a/core/adapters/inbound/agents/claudecode/hookinstaller.go
+++ b/core/adapters/inbound/agents/claudecode/hookinstaller.go
@@ -66,7 +66,7 @@ const hookMatcher = "Bash|Write|Edit|MultiEdit|NotebookEdit|WebFetch|mcp__.*|Ask
 
 // hookMatcherPreToolUse is the narrow matcher for the PreToolUse event. We
 // only want to flip working→waiting on the user-input tools — matching every
-// Bash/Write/… would set permissionPending on every tool call. (Issue #307.)
+// Bash/Write/… would hold SignalPermissionPrompt on every tool call. (Issue #307.)
 const hookMatcherPreToolUse = "AskUserQuestion|ExitPlanMode"
 
 // hookMatcherPreCompact is the matcher for the PreCompact event. Unlike the
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -193,12 +193,25 @@ type SessionDetector struct {
 	recorder    outbound.EventRecorder
 	recorderSeq int64
 
-	// permMu guards permissionPending. The map tracks sessions with an active
-	// PermissionRequest hook that hasn't been cleared by PostToolUse/
-	// PostToolUseFailure. Written by HandlePermissionHook (HTTP handler
-	// goroutine), read by processActivity (event-loop goroutine).
-	permMu            sync.Mutex
-	permissionPending map[string]bool // sessionID → true
+	// signals holds the out-of-band state signals (hook-delivered today,
+	// OTel/process/screen in the later phases of #1129) between arrival on
+	// their own goroutine and the classify pass that consumes them. It
+	// replaces the three hand-rolled per-signal overlay maps this struct
+	// carried before #1288 — permissionPending, hookTurnDone and
+	// idlePromptPending — each of which had its own lifecycle rule spelled
+	// out in its own overlay method. The per-signal policy now lives in
+	// session.signalPolicies; adding a Phase 4-7 signal is a row there, not a
+	// fourth map and a fourth overlay here.
+	//
+	// Carries its own lock (hooks arrive on HTTP handler goroutines, the
+	// event loop classifies), so it is deliberately NOT guarded by permMu.
+	signals *session.SignalHolds
+
+	// permMu guards the remaining per-session hook/bookkeeping maps below
+	// (compactPending, editToolOpenSince, idleProjectRetryAttempts). Written
+	// from HTTP handler goroutines, read by processActivity (event-loop
+	// goroutine).
+	permMu sync.Mutex
 
 	// compactPending tracks sessions in a manual /compact: sessionID → the Unix
 	// time the PreCompact hook fired. Set by HandleCompactHook; cleared when the
@@ -209,25 +222,6 @@ type SessionDetector struct {
 	// window (#657). Guarded by permMu — same goroutine-crossing story as
 	// permissionPending.
 	compactPending map[string]int64 // sessionID → unix seconds (hook fire time)
-
-	// hookTurnDone tracks sessions whose Claude Code Stop hook (#1161) has fired
-	// but whose resulting classify pass hasn't consumed it yet: sessionID → the
-	// turn's final assistant text (display-truncated) and whether it carried a
-	// waiting cue. Set by HandleStopHook; consumed and cleared by
-	// overlayHookTurnDone on the next classify pass (consume-once, so the
-	// authoritative turn-done signal never bleeds into the following turn).
-	// Guarded by permMu — same goroutine-crossing story as permissionPending.
-	hookTurnDone map[string]hookStopSignal // sessionID → Stop-hook payload
-
-	// idlePromptPending tracks sessions whose Claude Code Notification/idle_prompt
-	// hook reported the agent is idle at the prompt waiting for the user (#1173):
-	// sessionID → true. Set by HandleIdlePromptHook; overlaid onto metrics by
-	// overlayIdlePrompt every classify pass while the finished turn stays idle,
-	// and cleared there the moment new activity arrives (IsAgentDone flips false)
-	// — so it holds the session in waiting without pinning it into the next turn.
-	// Persistent (not consume-once, unlike hookTurnDone): a lower-tier reclassify
-	// must never revert the corrected waiting back to ready. Guarded by permMu.
-	idlePromptPending map[string]bool // sessionID → true
 
 	// editToolOpenSince tracks, per session, the Unix time a permission-gated
 	// file-edit tool first appeared open. Guarded by permMu. Drives the
@@ -332,10 +326,8 @@ func NewSessionDetector(watchers []inbound.Watcher, deps SessionDetectorDeps) *S
 		debounce:                 make(map[string]*debounceEntry),
 		debouncedEvents:          make(chan agent.Event, 64),
 		deletedCooldown:          10 * time.Second,
-		permissionPending:        make(map[string]bool),
+		signals:                  session.NewSignalHolds(),
 		compactPending:           make(map[string]int64),
-		hookTurnDone:             make(map[string]hookStopSignal),
-		idlePromptPending:        make(map[string]bool),
 		editToolOpenSince:        make(map[string]int64),
 		idleProjectRetryAttempts: make(map[string]int),
 		bgLiveProbe:              anyLiveOutputWriter,
@@ -534,7 +526,23 @@ func classifierInputs(m *session.SessionMetrics) *lifecycle.ClassifierInputs {
 		LastEventType:                     m.LastEventType,
 		LastWasUserInterrupt:              m.LastWasUserInterrupt,
 		LastWasToolDenial:                 m.LastWasToolDenial,
+		HookTurnDone:                      m.HookTurnDone,
+		IdlePromptPending:                 m.IdlePromptPending,
 	}
+}
+
+// withProvenance stamps the deciding rule and tier onto a classifier-inputs
+// snapshot. Separate from classifierInputs because provenance is a property of
+// the *verdict*, not of the metrics: the synthesizers emit transitions that no
+// rule decided, and those must record no tier rather than borrow whichever one
+// the ladder would have reached on its own.
+func withProvenance(in *lifecycle.ClassifierInputs, v StateVerdict) *lifecycle.ClassifierInputs {
+	if in == nil || !v.Tier.Known() {
+		return in
+	}
+	in.DecidedByTier = v.Tier.String()
+	in.DecidedByRule = v.Rule
+	return in
 }
 
 // AddWatcher registers a watcher with the running (or not-yet-running)
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -45,7 +45,7 @@ const staleWorkingRefreshInterval = 5 * time.Second
 // (Edit/Write/MultiEdit/NotebookEdit) may stay open before the detector treats
 // it as a held permission prompt and sets OpenToolStalled — the transcript
 // -based fallback for when the PermissionRequest hook can't reach the daemon
-// (#488, ClassifyState rule 1b).
+// (#488, ClassifyState's open_tool_stalled rule).
 //
 // It is deliberately NOT staleWorkingRefreshInterval. That constant is a
 // polling cadence (how often a lingering working session is re-read); reusing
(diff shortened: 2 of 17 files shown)


ingo-eichhorst:
## `/simplify` pass (4 agents) — applied

Four-way convergence on one finding, which drove the main change.

**Merged `signalPolicies` (map) + `signalOrder` (slice) into one ordered table.** All four angles reached this independently: *simplification* and *altitude* as "two sources of one truth", *efficiency* with numbers — the two string-keyed map lookups per held pass were **~400ns of `Overlay`'s ~380ns held path**, roughly half its cost. The win that matters most isn't the ns though: "added a policy, forgot to order it" is now **unrepresentable** rather than caught by a test, so that test is gone with it.

Also applied:
- **`session.TierOf`** — a signal's tier was declared twice (its policy row *and* every classifier rule reading it), with nothing checking they agreed. Setting a policy's tier used to change no behaviour and fail no test. Rules now resolve through `TierOf`, giving `signalPolicy.tier` a real consumer and one source of truth.
- **`stateRule.eval` → `when`/`decide`**, dropping the `("", "", false)` sentinel protocol and the `tierAlways` closure factory (7 of 8 uses just wrapped a constant). ~45 lines.
- **Provenance no longer keys off prose.** The old guard compared `reason` strings to decide whether the verdict still applied — coupling trace correctness to string wording, in a PR whose whole premise is that reason strings are load-bearing golden data. It now keys off a structural signal (did a synthesizer re-base `state.State`?), and when the collapsed-waiting synthesizer re-classifies, provenance is attributed to that re-run rather than dropped.
- **Test consolidation.** The migrated overlay tests were driving a domain type through an inert `SessionDetector` shell, duplicating the new domain suite ~1:1. The three genuinely unique policy assertions moved to `signal_hold_test.go`; `TestHasPendingIdlePrompt` stays, being the one case that exercises detector code.
- **Finished the rule-renumbering sweep** — it had stopped at the `_test.go` boundary, leaving ~25 `Rule N` comments as the last holders of the vocabulary the sweep retired.
- Inlined a one-line pass-through in the replay harness; fixed a doubled `(consume-once)` parenthetical.

## Deferred, filed as #1297 — and it is a real gap in this PR's scope

Two agents independently found that **a fourth hook-delivered signal was left hand-rolled**: `PreCompact` (`compactPending` + `applyCompactHold`) is structurally identical to the three that migrated — per-session hold, persistent, one staleness rule, one flag applied — and the classifier *already* declares it `compact_in_progress` at `TierHook`. It didn't migrate for exactly one reason: `signalPolicy.stale` has no clock, and its rule is half wall-clock.

That same gap means **#1288's scope item 4, "adopt grace-timer hysteresis", is not delivered here.** The reconcile mechanism landed without needing a timer, but Phases 5 (network probe) and 6 (screen stability) are inherently time-based and will otherwise force a signature change touching every row. #1297 covers both; it should land before Phase 5. Flagging it rather than quietly rescoping — a code comment on the `compact_in_progress` rule now points at it so the exception is visible in place.

## Verification

`tools/preflight.sh`: **12/12 PASS**. All 338 replay goldens still byte-identical, zero `replaydata/` churn.

One caveat worth stating plainly: on the first post-simplify run the **ARS architecture gate reported FAIL**, and its diagnostic was not captured in the log. On three isolated re-runs it passes identically (composite **7.9634 → 7.9646**, *improved*, no category regressions). I'm treating it as a transient — the gate fetches `origin/main` to build its baseline and ~20 sibling worktrees share one `.git` dir — but I could not diagnose it from evidence, so it's reported as observed, not explained. ARS is advisory rather than a merge gate in this repo either way.

Issue #445, opened 2026-05-23
daemon: track Bash background processes (run_in_background) so the session stays working


ingo-eichhorst:
## Context

Re-assessment of the `background-process` cell (2026-05-23) confirms
`agent_supports: yes`, `irrlicht_observes: **no**` (confidence 0.9).
claude supports `Bash` with `run_in_background: true` (+ `BashOutput` /
`KillShell`), but irrlicht has no transport for it. Source of record:
`replaydata/agents/claudecode/scenarios/background-process/assessment.json`.

A core-wide grep for any background-process tracking is empty; the
claudecode parser introspects only `TaskCreate` / `TaskUpdate` /
`AskUserQuestion` tool inputs, never `Bash`.

## Spec shape (unobservable today)

- Agent stays `working` while a background process is alive, even after
  the turn's normal `end_turn`.
- `BackgroundProcessCount > 0` while any background process is alive.
- When the last one exits → agent `ready`.

## Proposed work

- Parse `Bash` `tool_use` entries whose input has
  `run_in_background: true`; snapshot the `bash_id`; increment a new
  `BackgroundProcessCount` on `SessionMetrics`.
- Decrement on a `BashOutput` `tool_result` reporting a terminated
  status, or a `KillShell` tool_use.
- Hold the session in `working` while the count > 0 — a
  `BackgroundProcessCounter` hook analogous to the existing
  `SubagentCounter`, so the `IsAgentDone` path considers it.

## Caveat

claude reports background-process exit only **on demand** via
`BashOutput`, not as an async event — so liveness is approximate unless
the daemon walks the process tree from the `claude` PID. Worth deciding
the fidelity bar before implementing.

Discovered while re-evaluating the unfinished claudecode onboarding
cells. Sibling: background-subagent observability.

PR #1307, opened 2026-07-31
fix(daemon): don't read a timed-out lsof probe as a dead background process

### LINKED ISSUE #1299: TestAnyLiveOutputWriter_RealProcess flakes under full-suite load — lsof's own 2s timeout is swallowed as "not alive"
## Summary

`TestAnyLiveOutputWriter_RealProcess` (`core/application/services/background_probe_internal_test.go:40`) failed on **two separate** full `go test ./core/... -race -count=1` runs, on a branch that touches only `core/adapters/inbound/agents/fswatcher/`. Same failure both times:

```
--- FAIL: TestAnyLiveOutputWriter_RealProcess (4.25s)
    background_probe_internal_test.go:40: probe reported no live writer while background process is running
```

Found while verifying #1286 (which closes #1248). This is a **third** member of the load-dependent flake family [#1248's follow-up comment](https://github.com/ingo-eichhorst/Irrlicht/issues/1248#issuecomment-5145458876) proposed tracking, alongside `TestWatch_MetadataParentIsPresentOnFirstNewEvent` and `TestSessionDetector_NewSession_CreatesState`.

## The duration is the tell — 4.25s

`anyLiveOutputWriter` (`background_probe.go:53-56`) runs `lsof` under its **own** 2-second `context.WithTimeout`:

```go
ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
defer cancel()
args := append([]string{"-t", "--"}, outputPaths...)
out, _ := exec.CommandContext(ctx, lsofPath, args...).Output()
```

The test's structure is a 2s polling loop followed by one final probe:

```go
deadline := time.Now().Add(2 * time.Second)
for time.Now().Before(deadline) && !anyLiveOutputWriter([]string{out}) {
    time.Sleep(20 * time.Millisecond)
}
if !anyLiveOutputWriter([]string{out}) { t.Fatalf(...) }
```

If `lsof` hits its timeout, the first probe alone consumes the loop's entire 2s budget — the loop then exits after exactly one iteration, and the final probe burns another 2s. **2 + 2 ≈ the observed 4.25s.** A healthy run of this test is milliseconds.

*Marked as inference, not instrumented:* the 4.25s duration matches two consecutive `lsof` timeouts closely enough that I'm confident, but I did not instrument the probe to confirm `ctx.Err() == DeadlineExceeded` at the moment of failure. Confirming it is one line (see below), and that instrumentation is worth having regardless.

## Why the test can't tell you what went wrong

`Output()`'s error is discarded, deliberately. The reasoning at `background_probe.go:38-48` is sound and should not be reverted — `lsof` exits 1 when *any* named path has no holder while still printing the other paths' PIDs, so branching on the error would report a live process as dead (#445 review).

But the same line makes **three different outcomes indistinguishable**:

| Reality | `anyLiveOutputWriter` returns |
|---|---|
| no process holds the file | `false` ✅ correct |
| `lsof` timed out under load | `false` ❌ wrong |
| `lsof` missing / failed to exec | `false` ❌ wrong |

The doc comment names this and accepts it — *"A timeout or a missing lsof yields empty stdout → 'no live writer', which is the safe ('don't pin forever') degradation."* So this is a **documented trade-off, not an oversight**, and that framing should survive into whatever gets done here.

## The part worth a second look

The accepted degradation produces exactly the failure the surrounding code was written to prevent. `background_probe.go:41-44` explains that branching on lsof's exit code "would ... wrongly report a still-running process as dead, flipping the session to `ready` while it's alive (issue #445 review)" — and a timeout does precisely that, just via a different route.

In production `anyLiveOutputWriter` is the default `bgLiveProbe` (`session_detector.go:341`), so a timeout means the daemon concludes a live backgrounded Bash command is finished. And the correlation runs the wrong way: `lsof` scans every process on the machine, so it is *slowest exactly when the machine is busiest* — which is also when long-running background agent work is most likely to be in flight. "Don't pin forever" and "don't call a live session dead" are both real; the current code picks one silently.

## Suggested direction (not prescriptive)

1. **Make the timeout observable.** Keep reading stdout regardless of exit code, but check `ctx.Err()` separately and log/count deadline-exceeded. That alone turns this flake from "no writer found" into a diagnosis, and gives production a trace where today there is none.
2. **Decide the degradation deliberately.** If a timeout should mean "unknown" rather than "dead", the caller needs a tri-state; if "dead" is genuinely preferred, the doc comment should say so with the #445 tension spelled out.
3. **Fix the test's budget regardless.** Its 2s deadline is meaningless while a single probe call can consume all of it — under timeout conditions the loop runs exactly once. The outer deadline should exceed the probe's own, or the probe timeout should be injectable for tests.

## Reproduction

Both occurrences were inside a full `go test ./core/... -race -count=1`. It did **not** reproduce in:

- 30 `-race` runs of the test alone under full CPU saturation (all 10 cores hogged);
- 10 `-race` runs of the package in isolation;
- ~7 further full-suite runs.

So budget for a long soak rather than a quick loop — the same shape as #1248. Suggestion 1 above is the cheap way in: land the instrumentation and let a natural occurrence explain itself.

## Environment

macOS (Darwin 25.5.0), Apple silicon, 10 cores. `go test ./core/... -race -count=1`.

ingo-eichhorst:
Closes #1299.

## What the flaky test was actually reporting

`TestAnyLiveOutputWriter_RealProcess` failed on two separate full-suite runs, at **4.25s** — and that duration was the clue. `anyLiveOutputWriter` runs `lsof` under its own 2s timeout, and the test's polling budget was *also* 2s, so one slow `lsof` consumed the entire window and the loop got exactly one attempt. 2s + 2s ≈ the observed failure time; a healthy run is milliseconds.

But chasing the flake surfaced the real problem underneath it.

## The bug

`Output()`'s error is discarded — correctly, for the #445 reason that `lsof` exits 1 when *any* named path has no holder while still printing the other paths' PIDs. The side effect is that **three outcomes collapse into one**: no holder, timed out, and lsof-not-found all return `false`.

The old doc comment called this "the safe (don't pin forever) degradation". It isn't safe, because of what the caller does with a dead verdict:

```
runBackgroundLivenessProbe → purgeDeadBackgroundProcesses
  → metrics.PurgeDeadBackgroundProcs → tailer.PurgeBackgroundProcs + saveLedger
```

That purge is **written through to the persisted ledger**. So one slow `lsof` permanently drops a still-running background process and flips a live session to `ready` with no way back — the exact false-negative the exit-code handling at `background_probe.go:38-48` was written to prevent, reached by a different route.

The correlation runs the wrong way, too: `lsof` walks every process on the machine, so it is slowest precisely when the machine is busiest — which is when long-running background work is most likely to be in flight.

## The fix

`backgroundProbe` becomes `func([]string) (alive, conclusive bool)`.

- A **timeout** is inconclusive. The detector then skips the purge *and* holds the previous liveness verdict instead of writing "dead", so the next probe decides.
- **Any other** exec failure still degrades to a conclusive dead. Retrying can't fix a missing `lsof` (`MustResolve` falls back to the bare name, so this is reachable), and treating it as inconclusive would pin every affected session `working` forever — the failure the old degradation was guarding against. Only the transient case gets the retry.
- The `DeadlineExceeded` check runs **after** stdout has yielded no PID: `lsof` can be killed by the deadline having already printed a holder, and that PID is still proof of life worth keeping.
- The inconclusive path logs through `d.log`, so production leaves a trace where it had none.

Both halves matter and are tested separately — skipping the purge alone would still flip the session to `ready` via the cached verdict.

## Seen red

| Mutation | Result |
|---|---|
| timeout collapses back into conclusive-dead | `TimeoutIsInconclusive` — *a timed-out probe reported a conclusive verdict; the caller would purge on it* |
| inconclusive verdict still purges | `InconclusiveVerdictDoesNotPurge` — *an inconclusive probe purged [...] — a timeout is not evidence of death* |
| retained-verdict hold removed | `InconclusiveVerdictHoldsWorking` — *session flipped to ready 3 time(s) on an inconclusive probe* |

The third mutation initially produced **no failure** — I had written the hold without covering it, so the session would still have flipped to `ready` on a timeout. `InconclusiveVerdictHoldsWorking` was added to close that gap, and the mutation is red against it now.

## The test fix, and a mistake worth flagging

My first rework asserted that the post-kill probe was conclusive — and that assertion promptly flaked 2-in-25 under `-race`, because under load `lsof` legitimately times out and the probe correctly says "I don't know". Same mistake class as the original test: asserting something load can invalidate.

`awaitConclusiveVerdict` now treats an inconclusive verdict as a **retry, not a result**, which is both the behavior under test and what keeps the test honest. If every probe in the budget times out it **skips** rather than fails — the probe did its job; the machine was simply too loaded to observe the transition, which is not a defect in the code under test.

## Verification

- `go test ./core/... -race -count=1` — **4 green runs** across the change (49 packages, 0 failures).
- `TestAnyLiveOutputWriter*` — **40 runs `-race`** green, plus **15 runs `-race` under full CPU saturation** (all 10 cores hogged) green. The original failed twice in ordinary full-suite runs.

## Not done here

The `backgroundPIDProbe` (`anyLivePID`) has no timeout and no exec, so it has no inconclusive state to model — left alone deliberately rather than changed for symmetry.

### DIFF
--- a/core/application/services/background_fixture_test.go
+++ b/core/application/services/background_fixture_test.go
@@ -1,0 +1,98 @@
+package services_test
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"irrlicht/core/application/services"
+	"irrlicht/core/domain/agent"
+	"irrlicht/core/domain/session"
+)
+
+// bgOutputFixture is the setup every background-output liveness test shares: a
+// `working` session whose turn has ended and whose only open background process
+// is an output file, wired to a detector with an injectable probe.
+type bgOutputFixture struct {
+	det     *services.SessionDetector
+	tw      *mockAgentWatcher
+	metrics *funcMetrics
+	rec     *mockRecorder
+	probes  atomic.Int32
+	sid     string
+	path    string
+}
+
+// newBGOutputFixture builds the fixture with probe as the injected liveness
+// verdict; every call is counted so a test can require positive evidence the
+// probe actually ran rather than passing on the mere absence of an effect.
+func newBGOutputFixture(sid, path, outPath string, probe func() (alive, conclusive bool)) *bgOutputFixture {
+	return newBGOutputFixtureInState(sid, path, outPath, session.StateWorking, probe)
+}
+
+// newBGOutputFixtureInState is newBGOutputFixture with the session's starting
+// state chosen explicitly — the ready-start case covers a background process
+// spawned right after the session had already settled (#937).
+func newBGOutputFixtureInState(sid, path, outPath, state string, probe func() (alive, conclusive bool)) *bgOutputFixture {
+	f := &bgOutputFixture{sid: sid, path: path}
+	f.metrics = &funcMetrics{fn: func(_, _ string) (*session.SessionMetrics, error) {
+		return &session.SessionMetrics{
+			LastEventType:            "turn_done",
+			BackgroundProcessCount:   1,
+			BackgroundProcessOutputs: []string{outPath},
+		}, nil
+	}}
+	f.tw = newMockAgentWatcher()
+	repo := newMockRepo()
+	repo.states[sid] = &session.SessionState{
+		SessionID:      sid,
+		State:          state,
+		TranscriptPath: path,
+		FirstSeen:      time.Now().Unix(),
+		UpdatedAt:      time.Now().Unix(),
+	}
+	f.det = newDetectorWithMetrics(f.tw, newMockProcessWatcher(), repo, f.metrics)
+	f.rec = &mockRecorder{}
+	f.det.SetRecorder(f.rec)
+	f.det.SetBackgroundProbeForTest(func([]string) (bool, bool) {
+		f.probes.Add(1)
+		return probe()
+	})
+	return f
+}
+
+// start runs the detector and stops it when the test ends.
+func (f *bgOutputFixture) start(t *testing.T) {
+	t.Helper()
+	ctx, cancel := context.WithCancel(context.Background())
+	done := make(chan error, 1)
+	go func() { done <- f.det.Run(ctx) }()
+	t.Cleanup(func() { cancel(); <-done })
+}
+
+// activity drives one re-evaluation. terminal short-circuits the 2s debounce,
+// the way production's periodic re-probe does.
+func (f *bgOutputFixture) activity(terminal bool) {
+	f.tw.ch <- agent.Event{
+		Type:           agent.EventActivity,
+		SessionID:      f.sid,
+		ProjectDir:     "-Users-test",
+		TranscriptPath: f.path,
+		Terminal:       terminal,
+	}
+}
+
+// awaitProbe blocks until the injected probe has run at least once, failing if
+// it never does — without it a test asserting only that nothing happened would
+// pass just as happily when the probe was never launched.
+func (f *bgOutputFixture) awaitProbe(t *testing.T) {
+	t.Helper()
+	deadline := time.Now().Add(2 * time.Second)
+	for f.probes.Load() == 0 {
+		if !time.Now().Before(deadline) {
+			t.Fatal("background liveness probe was never invoked; the test would prove nothing")
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
--- a/core/application/services/background_probe.go
+++ b/core/application/services/background_probe.go
@@ -15,6 +15,10 @@ import (
 // than trusted PATH, per go:S4036.
 var lsofPath = pathutil.MustResolve("lsof")
 
+// lsofTimeout bounds a single lsof invocation. A package var rather than a
+// const so tests can drive the timeout path without waiting seconds.
+var lsofTimeout = 2 * time.Second
+
 // backgroundProbe answers "does any of these output files still have a live
 // writer?" — the daemon's authoritative liveness check for Claude Code Bash
 // background processes (run_in_background). Claude Code reports a background
@@ -22,7 +26,12 @@ var lsofPath = pathutil.MustResolve("lsof")
 // transcript alone can't tell us when one dies; this probe walks the live
 // process state instead. Modeled as a field so tests can inject a fake.
 // See issue #445.
-type backgroundProbe func(outputPaths []string) bool
+//
+// The second return is "was this conclusive?", not a second liveness bit. It
+// is false only when the probe could not find out — see anyLiveOutputWriter —
+// and a false there must never be read as "dead", because the caller's
+// response to death is a durable purge (issue #1299).
+type backgroundProbe func(outputPaths []string) (alive, conclusive bool)
 
 // anyLiveOutputWriter reports whether any of the given background output files
 // is currently held open by a live process. Each backgrounded command's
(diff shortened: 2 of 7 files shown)


ingo-eichhorst:
## Review round — two real findings fixed, plus two tests that had regrown the flake

Ran `/ir:code-review medium`. It confirmed the core fix (the `ctx.Err()` placement after the stdout scan, the deliberate conclusive-on-non-timeout carve-out, and the purge suppression all check out, and found no race in the new `bgMu` usage). Two findings were worth acting on before merge, and both are now fixed.

### 1. The hold was unbounded — I'd traded one failure for another

Nothing limited consecutive inconclusive verdicts. On a machine where `lsof` *always* exceeds 2s — a permanently oversized process table, or a hung NFS/SMB mount that `lsof` blocks on — every probe returns inconclusive, `live` pins to `true` forever, and the session never leaves `working`. Because the purge is skipped too, the ledger entry survives a daemon restart and replays the same loop.

That is precisely the failure the old always-believe-the-silence behavior avoided. My doc comment reasoned carefully about non-timeout errors and then assumed "timed out" implies transient — which it doesn't.

`maxInconclusiveBackgroundProbes` now releases the hold after ~5 consecutive inconclusive verdicts (~30s at the re-probe cadence), degrading to the pre-#1299 semantics as a floor. New test `InconclusiveHoldIsBounded`, mutation-verified.

### 2. Two of my own tests had regrown the exact flake shape

`UnheldPathIsConclusivelyDead` asserted `conclusive` off a **single un-retried** `lsof` call — the identical full process-table scan as the live case, which the reviewer measured at **0.51–1.21s against a 2s timeout**, under 2× headroom. Under full-suite load that fails on a correct implementation. It now goes through `awaitConclusiveVerdict`; `EmptyAndMissing` drops its `conclusive` assertion for the same reason.

Worth noting as a pattern: the un-retried shape regrew the moment a new test was added next to the fixed one. A shared retry/skip helper for anything shelling out to a full-process-table tool is the durable answer.

### Also fixed

- **Guard bug:** the hold tested `!livePID` but overwrote `live` (= `liveOut || livePID`), so a probe returning `(alive, inconclusive)` would have had its liveness discarded in favour of a stale cache. Not reachable from `anyLiveOutputWriter` today, but the `backgroundProbe` seam permits it. Now guards on `live`.
- **Log level/volume:** the inconclusive path used `LogError` every cycle — on a chronically slow machine ~1800 lines/hour into a 10MB rotating log, pushing real errors out of the retained window. Now `LogInfo`, edge-triggered on the first hold, plus one `LogError` when the bound releases and a possibly-live process is about to be purged.
- **False-green risk:** both detector tests asserted only *absence* (no purge / no ready transition), which stays true if the probe is never launched at all. Both now require positive evidence the probe ran. The no-purge test waits on that rather than unconditionally burning 2.01s.

### Not taken

The reviewer's finding 5 (the purge decision assumes alive while the cached verdict can hold `prev == false`) — narrow and self-correcting, and a `prev == false` normally implies an earlier purge already emptied `BackgroundProcessOutputs`, so the probe wouldn't run. Left as-is rather than restructuring on a path I can't demonstrate.

### Seen red

| Mutation | Result |
|---|---|
| hold unbounded again | `InconclusiveHoldIsBounded` — *never released across 5s of back-to-back probes* |
| purge not suppressed on hold | `InconclusiveVerdictDoesNotPurge` — *an inconclusive probe purged [...]* |
| retained-verdict hold removed | `InconclusiveVerdictHoldsWorking` — *flipped to ready 3 time(s)* |
| timeout collapses to conclusive-dead | `TimeoutIsInconclusive` — *reported a conclusive verdict; the caller would purge on it* |

### Verification

- `go test ./core/... -race -count=1` — **6 green runs** across the change (49 packages, 0 failures).
- Probe + inconclusive tests: **25 runs `-race`** green after the review fixes; earlier, **40 runs `-race`** plus **15 under full CPU saturation** green.

### Follow-up worth its own issue

`lsofTimeout` is 2s against a measured 0.5–1.2s baseline on this machine — under 2× headroom. The reviewer's point stands that the constant itself, not just its error handling, is a root cause of #1299. Passing `lsof -b` (non-blocking) would also stop a hung mount from consuming the budget at all. Both are behavior changes beyond this PR's scope.

ingo-eichhorst:
## Quality-gate status — stopping here deliberately

**11 pass / 2 fail.** Both failures are advisory: I verified that neither branch protection nor the "Protect Main" ruleset configures *any* required status check, so nothing here blocks merge. AGENTS.md says the same about CodeScene, and I'm treating SonarCloud the same way on that evidence.

**SonarCloud — Duplication on New Code: 15.2% → 3.9%** (gate ≤3%). Fixed by extracting `bgOutputFixture`, which owns the setup the three new inconclusive tests were each repeating, and migrating `DeadVerdictPurgesLedger` onto it as their natural counterpart. The residual 0.9% sits in `HoldsWorkingThenReady` and `FreshSpawnFromReady` — pre-existing tests that would need the fixture to grow a toggleable probe and a configurable initial state. That's a refactor of code this PR doesn't otherwise touch, for a gate that doesn't block, so I stopped.

**CodeScene** — `background_probe_internal_test.go` **improved** (9.69 → 9.92). `session_detector_activity.go` flagged Complex Method on an existing hotspot; I extracted `commitBackgroundVerdict`, so `runBackgroundLivenessProbe` now reads as probe → hold decision → purge → commit. `background_process_test.go` declined (9.10 → 8.55) from the fixture type plus three tests landing in an already-large file.

### Final verification

- `go test ./core/... -race -count=1` — **8 green runs** across this PR (49 packages, 0 failures).
- All four mutations re-confirmed red after both refactors, so neither the fixture extraction nor `commitBackgroundVerdict` neutered an assertion.
- Probe tests: 40 runs `-race`, plus 15 under full CPU saturation, plus 25 after the review fixes — all green. The test this PR started from failed twice in ordinary full-suite runs.

Issue #1320, opened 2026-08-02
Hook-name → SignalKind mapping is written twice, and has already drifted: replay ignores PreToolUse


ingo-eichhorst:
Found while reviewing #1297 / PR #1318. Independently surfaced by two reviewers on that PR.

## The drift

Which hook name produces which `session.SignalKind` is written in two places:

- **Daemon**: `SessionDetector.HandlePermissionHook` (`core/application/services/session_detector_lifecycle.go`) treats **`PermissionRequest` *and* `PreToolUse`** as `SignalPermissionPrompt`.
- **Replay harness**: `applyHookEvent` (`tools/onboarding-factory/cmd/replay/replay_sidecar.go`) handles only `HookPermissionRequest`, and falls through `default: return` on `PreToolUse`.

`PreToolUse` is not hypothetical — it is in the catalog today. `hook_name` values across all of `replaydata/`:

| hook_name | count |
|---|---|
| `PostToolUse` | 40 |
| `PermissionRequest` | 6 |
| `PostToolUseFailure` | 2 |
| **`PreToolUse`** | **2** |

Both `PreToolUse` events are `hook_received` on the primary session, in `2-18_user-blocking-plan-mode-approval` and `5-4_architect-editor-pair`.

## Is anything wrong right now?

**No — and that is the problem.** Verified: in both recordings a `PermissionRequest` follows the `PreToolUse` by 19 ms and 135 ms respectively and produces the same hold, so no golden is currently wrong. The two paths agree by luck, not by construction.

This is exactly the class of drift #1288 set out to make unrepresentable when it moved both paths onto the shared `SignalHolds` — the *store* is now shared, but the *hook-name mapping* into it is not.

## Proposed change

Move the mapping into shared data — a `map[string]session.SignalKind` or a small `HookSignal(name) (SignalKind, bool)` next to the hook-name constants in `core/adapters/inbound/agents/claudecode/hooks.go` — consumed by both the detector's handlers and `applyHookEvent`.

Then "add a signal" is "add a policy row plus one table entry" instead of "a row + a detector method + a replay switch case", and a hook the daemon honours can't be silently dropped by the harness whose job is catching classifier regressions.

## Related

[...]

PR #1341, opened 2026-08-05
fix(replay): pair recordings with their sibling events.jsonl sidecar

### LINKED ISSUE #1326: Hook-carrying recordings never reach applyHookEvent: the replay hook path has no fixture coverage
Found while reviewing #1320 / PR #1323, where "no goldens moved" was offered as evidence the fix was safe. It was vacuous evidence: no committed gate replays a hook-carrying recording through `applyHookEvent`, so goldens *cannot* move on a hook-mapping change.

## Two gaps that compose

**1. The two sidecar fixture tests use recordings with no hooks.** `TestReplayWithSidecar_GoldenFixture` and `TestReplayWithSidecar_ContinueFixture` do drive `replayWithSidecar` over real recordings, passing `events.jsonl` explicitly. But:

```
10-full-lifecycle-839f0678:        0 hook_received events
13-full-lifecycle-continue-8a525d27: 0 hook_received events
```

So `applyHookEvent` is never entered on either.

**2. Every recording that *does* carry hooks is replayed transcript-only.** `resolveInputPaths` (`cmd/replay/main.go:214`) auto-detects a sidecar only at `<transcript-basename>.events.jsonl`:

```go
if candidate := strings.TrimSuffix(src, ".jsonl") + eventsSidecarExt; candidate != src {
	if _, err := os.Stat(candidate); err == nil { ... }
}
```

Every sidecar in the catalog is named plain `events.jsonl` beside `transcript.jsonl`:

```
$ find replaydata -name '*.events.jsonl' | wc -l
0
$ find replaydata -name 'events.jsonl' | wc -l
335
```

`os.Stat("transcript.events.jsonl")` therefore always fails and `useSidecar` stays false for both fixture drivers:

- `tools/replay-fixtures.sh:193` — `find ... \( -name 'transcript.jsonl' -o -name 'transcript.md' \)`
- `discoverReplayFixtures` (`cmd/replay/fixtures_test.go:83`) explicitly skips `events.jsonl`

Measurable: replaying `claudecode/scenarios/2-18_user-blocking-plan-mode-approval` (which carries a `PreToolUse`) through `replay-fixtures.sh` yields transition causes `['debounce_coalesce', 'init']` — no `hook` cause anywhere, despite 50 `hook_received` events across `replaydata/`.

## Consequence

The hook path of the replay harness — the harness whose job is catching classifier regressions — is exercised only by hand-written synthetic unit tests in `main_test.go`. This is what let #1320's drift (replay silently ignoring `PreToolUse`, which the daemon holds on) sit undetected.

## Verified impact of closing the gap

Building `cmd/replay` before and after PR #1323 and running both in **explicit** sidecar mode over the two `PreToolUse` recordings does show output differences:

- `5-4_architect-editor-pair`: `cause=hook working→waiting @23:34:37.42887` → `cause=debounce_coalesce working→waiting @23:34:37.398345`; `state_durations.waiting` 10412487000 → 10443012000.
- `2-18_user-blocking-plan-mode-approval`: transition `virtual_time` 00:30:55.233678 → 00:30:55.214489.

The post-#1323 output is the daemon-faithful one.

## Proposed change

Teach `resolveInputPaths` (or the fixture drivers) to also pair `transcript.jsonl` with a sibling plain `events.jsonl`, then regenerate goldens. That converts 335 recordings from transcript-only into full sidecar coverage.

Expect one large, deliberate golden churn on the first run — that churn *is* the finding, and it should be reviewed rather than rubber-stamped, since it is the first time these recordings will have been replayed the way the daemon actually sees them. Worth checking during that review whether any currently-green cell was only green because its hook events were being dropped.

ingo-eichhorst:
Closes #1326

⚠️ **Read the "Cost" section before merging.** This change does what #1326 asked, and in doing so exposed a replay-engine defect that costs 22 fixtures their state-machine assertions. That tradeoff is a maintainer call, not mine.

## The gap

`resolveInputPaths` auto-detected a sidecar only at the legacy `<transcript-stem>.events.jsonl`. Every sidecar in `replaydata/` is a plain `events.jsonl` beside its transcript (396 of them; zero of the legacy spelling). So `os.Stat` always failed, `useSidecar` stayed false for the entire catalog, and **both** fixture drivers replayed all ~380 recordings transcript-only. `applyHookEvent` was never reached over a real recording — the hook→signal mapping could drift without moving a single golden, which is what let #1320's drift sit undetected and made "no goldens moved" vacuous evidence on PR #1323.

## The fix

Probe the sibling `events.jsonl` too. The probe stays in the transcript's own directory rather than walking up: a subagent transcript sits one level from its parent recording's sidecar, which describes a *different* session.

Three consequences had to be handled:

**1. 56 recordings hard-errored.** opencode (30) and hermes (24) are `ProcessOwnedStore` and record no fswatcher fires; aider is `FilesUnderCWD` with only synthetic `proc-<pid>` session ids. These degrade to transcript-only via an `errSidecarNotDrivable` sentinel and **record why** in the new `sidecar_fallback` field. The sentinel is keyed on the *symptom*, not the adapter — two mistral-vibe recordings also fall back despite mistral-vibe being `FilesUnderRoot`; the comment says so rather than claiming otherwise. A sidecar with **malformed lines is not eligible** and stays fatal.

**2. The metric vector regressed.** Taking metrics from the sidecar timeline cost 112 goldens their totals and flipped `of verify` to FAIL for five cells, because `observations.go` reads that block as the catalog's only source of truth for token/cost/model. Fixed by splitting concerns: transitions stay sidecar-driven; the metric vector always comes from a full parse. **0 of 365 changed goldens now differ from `main` in their metric vector.**

**3. `replay-fixtures.sh` aborted** on the first extended-check divergence under `set -e`. Now informational (the idiom `run-cell.sh:516` already used), with divergences listed at the end; a missing report is still fatal and the report is `rm -f`'d first so a stale file can't read as success.

## Cost — 22 fixtures stop asserting anything

This is the part that needs a decision. Sidecar mode is *not* uniformly more faithful. For 22 recordings it reproduces **fewer** transitions than the daemon actually recorded — in most cases zero — so those goldens now pin only the synthetic `init` row:

| cell | golden transitions (main → branch) | daemon recorded | replay reproduced |
|---|---|---|---|
| `codex/regressions/agent-question-pending` | 3 → 1 | 1 | 0 |
| `codex/regressions/baseline-hello` | 3 → 1 | 1 | 0 |
| `codex/scenarios/1-1_session-start` | 3 → 1 | 0 | 0 |
| `codex/scenarios/1-2_session-end` | 6 → 1 | 2 | 0 |
| `codex/scenarios/2-1_basic-turn` | 3 → 1 | 0 | 0 |
| `codex/scenarios/2-1_basic-turn` | 3 → 1 | 2 | 0 |
| `codex/scenarios/2-1_basic-turn` | 3 → 1 | 2 | 0 |
| `codex/scenarios/2-5_synchronous-slash-command` | 3 → 1 | 2 | 0 |
| `codex/scenarios/5-2_model-identification` | 3 → 1 | 2 | 0 |
| `pi/regressions/baseline-hello` | 3 → 1 | 2 | 0 |
| `pi/regressions/full-lifecycle-toolcall` | 3 → 1 | 1 | 0 |
| `pi/scenarios/1-1_session-start` | 3 → 1 | 2 | 0 |
| `pi/scenarios/1-2_session-end` | 3 → 1 | 2 | 0 |
| `pi/scenarios/1-2_session-end` | 3 → 1 | 4 | 0 |
| `pi/scenarios/1-8_model-context-display` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-13_turn-end-terminal-text` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-16_oversized-transcript-line` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-17_agent-question-pending` | 3 → 1 | 2 | 0 |
| `pi/scenarios/2-1_basic-turn` | 3 → 1 | 2 | 0 |
| `pi/scenarios/2-2_auto-executed-tool-call` | 3 → 1 | 2 | 0 |
| `pi/scenarios/2-5_synchronous-slash-command` | 3 → 1 | 4 | 0 |
| `pi/scenarios/5-2_model-identification` | 3 → 1 | 4 | 0 |

129 goldens lose transitions overall. The extended check is what proves this is a defect rather than a fidelity gain: it compares replay against *what the daemon itself logged*, and reports 0 reproduced against 1–4 recorded.

**Cause: not established.** Two candidates were tested and one ruled out — relaxing `flushPendingDebounce`'s `!d.coalesced` early-return changes nothing (measured), and incomplete fs-event coverage explains only 1 of these 22 (the other 21 have 100% coverage). I have deliberately not asserted a cause in the code comments.

**Options:** (a) merge and fix the sidecar replayer under a follow-up issue, accepting a weaker net for these 22 in the interim; (b) hold this PR until the replayer reproduces at least what the daemon recorded. I lean (b) — the whole point of #1326 is that a harness which silently asserts nothing is worse than one that fails.

## Red-first evidence

Six defect tests, each seen red before its fix: `PairsSiblingEventsJSONL`, `HookReachesApplyHookEvent`, `EveryHookCarryingRecordingReplaysWithSidecar` (17 recordings), `StoreAdapterFallsBackToTranscriptOnly`, `CorruptSidecarIsFatal`, and `SidecarArgumentIsNotItsOwnTranscript` (after strengthening — the original passed while `./events.jsonl` still self-paired, because `t.TempDir()` hands back an already-cleaned path).

Locks (pass by construction): `LegacySpellingWins`, `RelativeTranscriptStillPairs`, `BlankLinesAreNotMalformed`.

## Golden churn

| | count |
|---|---|
| goldens changed | 365 |
| gained a hook-caused transition | 5 |
| gained an `extended_check` | 309 |
| transcript-only fallback recorded | 56 |
| **lost transitions** | **129** |
| **reduced to `init` only** | **22** |
| metric vector changed | 0 |

The 5 cells gaining hook-caused transitions are the direct #1326 win, all claudecode: `15-permission-hooks-a845d6c4` (+4), `2-7_autonomous-loop` (+2), `2-18_user-blocking-plan-mode-approval`, `2-19_tool-gate-permission-prompt`, `5-4_architect-editor-pair`.

## Also not addressed

198 extended-check divergences are now visible for the first time. The dominant kind is a missing terminal `working→ready`. They are listed by name on every sweep rather than dismissed.

## Verification

`tools/preflight.sh` — all 13 gates PASS. `of validate` OK. CI green.

## Review

One `/ir:code-review` subagent at **high** effort (8 findings, all addressed) plus a 4-agent `/simplify` pass (reuse / simplification / efficiency / altitude). Between them they caught two things the local gates structurally could not: `extended_check.sidecar_path` baking an absolute path into 309 goldens (invisible locally — only the machine that wrote the goldens runs the test before CI), and the 22-fixture degradation above.

### DIFF
--- a/replaydata/agents/antigravity/scenarios/1-1_session-start/recordings/2026-06-20-12-33-40_irrlichd-0.5.2+be45695/transcript.jsonl.replay.json.golden
+++ b/replaydata/agents/antigravity/scenarios/1-1_session-start/recordings/2026-06-20-12-33-40_irrlichd-0.5.2+be45695/transcript.jsonl.replay.json.golden
@@ -8,25 +8,30 @@
     "flicker_max_duration": 10000000000
   },
   "summary": {
-    "total_events": 5,
-    "consumed_events": 5,
+    "total_events": 4,
+    "consumed_events": 4,
     "total_transitions": 3,
-    "first_event_time": "2026-06-20T10:33:40Z",
-    "last_event_time": "2026-06-20T10:33:46Z",
-    "wall_clock_session_duration": 6000000000,
+    "first_event_time": "2026-06-20T12:33:40.074291+02:00",
+    "last_event_time": "2026-06-20T12:33:47.762677+02:00",
+    "wall_clock_session_duration": 7688386000,
     "state_durations": {
-      "ready": 6000000000
+      "ready": 1080640000,
+      "working": 6607746000
     },
-    "flicker_count": 0,
-    "flickers_by_category": {},
-    "flickers_by_reason": {},
+    "flicker_count": 1,
+    "flickers_by_category": {
+      "working_between_ready": 1
+    },
+    "flickers_by_reason": {
+      "force ready→working on first activity": 1
+    },
     "model_name": "Gemini 3.1 Pro"
   },
   "transitions": [
     {
       "index": 0,
       "event_index": -1,
-      "virtual_time": "2026-06-20T10:33:40Z",
+      "virtual_time": "2026-06-20T12:33:40.074291+02:00",
       "cause": "init",
       "prev_state": "",
       "new_state": "ready",
@@ -39,24 +44,23 @@
     },
     {
       "index": 1,
-      "event_index": 3,
-      "virtual_time": "2026-06-20T10:33:40Z",
-      "cause": "debounce_coalesce",
+      "event_index": 0,
+      "virtual_time": "2026-06-20T12:33:40.074291+02:00",
+      "cause": "event",
       "prev_state": "ready",
       "new_state": "working",
       "reason": "force ready→working on first activity",
-      "last_event_type": "turn_done",
+      "last_event_type": "user_message",
       "has_open_tool": false,
-      "is_agent_done": true,
+      "is_agent_done": false,
       "needs_user_attention": false,
-      "waiting_for_user_input": false,
-      "last_assistant_text_head": "ok"
+      "waiting_for_user_input": false
     },
     {
       "index": 2,
-      "event_index": 3,
-      "virtual_time": "2026-06-20T10:33:40Z",
-      "cause": "debounce_coalesce",
+      "event_index": 2,
+      "virtual_time": "2026-06-20T12:33:46.682037+02:00",
+      "cause": "event",
       "prev_state": "working",
       "new_state": "ready",
       "reason": "agent finished turn → ready",
@@ -67,5 +71,36 @@
       "waiting_for_user_input": false,
       "last_assistant_text_head": "ok"
     }
-  ]
+  ],
+  "sessions": [
+    {
+      "session_id": "36842803-601e-4679-860a-8bf09b3798c8",
+      "adapter": "antigravity",
+      "first_seen": "2026-06-20T12:33:40.072384+02:00",
+      "last_seen": "2026-06-20T12:33:47.762799+02:00",
+      "duration_ms": 7690,
+      "event_count": 10,
+      "state_changes": 3,
+      "final_state": "ready",
+      "debounce_coalesced_events": 2,
+      "state_durations_ms": {
+        "ready": 7688,
+        "working": 0
+      }
+    }
+  ],
+  "extended_check": {
+    "sidecar_path": "",
+    "recorded_transition_count": 2,
+    "replayed_transition_count": 2,
+    "ordered_matches": 2,
+    "recorded_unique_kinds": [
+      "ready→working",
+      "working→ready"
+    ],
+    "replayed_unique_kinds": [
+      "ready→working",
+      "working→ready"
+    ]
+  }
 }
--- a/replaydata/agents/antigravity/scenarios/1-2_session-end/recordings/2026-06-20-12-36-01_irrlichd-0.5.2+a00b826/transcript.jsonl.replay.json.golden
+++ b/replaydata/agents/antigravity/scenarios/1-2_session-end/recordings/2026-06-20-12-36-01_irrlichd-0.5.2+a00b826/transcript.jsonl.replay.json.golden
@@ -8,25 +8,30 @@
     "flicker_max_duration": 10000000000
   },
   "summary": {
-    "total_events": 13,
-    "consumed_events": 13,
-    "total_transitions": 6,
-    "first_event_time": "2026-06-20T10:36:01Z",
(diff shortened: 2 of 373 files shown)


ingo-eichhorst:
Landing per maintainer decision. The 22-fixture degradation documented in the Cost section above is tracked in #1342 — cause not established, two candidates ruled out by measurement, reproduction steps included.

Issue #1360, opened 2026-08-09
fix(state): hook-asserted persistent holds have no wall-clock ceiling — a missed release pins a session at waiting forever


ingo-eichhorst:
Parent: #1355 — Phase B, work item 1. **Blocks the whole hook rollout.**

## The defect

`core/domain/session/signal_hold.go:258-275` — the `SignalPermissionPrompt` row is `TierHook`, persistent, and its *only* staleness rule is:

```go
stale: func(c holdContext) bool { return c.Metrics.LastWasToolDenial },
```

No clock. `LastWasToolDenial` is set from one claudecode-shaped transcript marker (`[Request interrupted by user for tool use]`). Its normal release is `PostToolUse` / `PostToolUseFailure` (`core/domain/session/hook_signal.go:82-83`). It sits at the top of the tier ladder, so it resolves to `waiting`.

Therefore: if the agent CLI renames or stops firing the release event, or the daemon misses the POST (crash, restart, port change, uninstalled hook), **the hold never releases and the session reads `waiting` indefinitely.** No lower-tier transcript signal can overrule it — that is by design, and it is exactly what makes this unrecoverable.

`compactHoldTimeout` (5 min, `signal_hold.go:67`) is the only hold with a ceiling today, and it is not the one with the highest authority.

## Why this blocks the epic

The transcript path is *total*: the classifier ladder ends in an always-firing `transcript_activity` default, so losing parsing fidelity falls **down** the ladder and self-corrects. The hook path does the opposite — it **pins**. Going from 2 hook adapters to 10 multiplies a failure mode that does not degrade gracefully.

## Scope

1. Give every **hook-asserted persistent** hold a wall-clock ceiling. `holdContext.Now` and `HeldSince` already exist (#1318), so this is a `stale` predicate, not new machinery — copy the shape of the `SignalCompactInProgress` row (`signal_hold.go:334`).
2. Enumerate the rows in scope from the code (`TierHook` and not `consumeOnce`) — do not assume the list is just `SignalPermissionPrompt` and `SignalIdlePrompt`. Say in the PR what you found.
3. Ceilings are **per-row constants with a stated rationale**, like `compactHoldTimeout` — not one global number. A human genuinely sitting on a permission prompt can take a long time; state in the comment what real-world duration each ceiling is calibrated against and why cutting a real wait short is the lesser error.
[...]

Issue #988, opened 2026-07-12
mistral-vibe: 5 scenarios flagged daemon_capability=bug in the onboarding matrix


ingo-eichhorst:
## mistral-vibe: 5 scenarios flagged daemon_capability=bug in the onboarding matrix

Roundup of every mistral-vibe cell currently marked `daemon_capability: "bug"` in the scenario × agent coverage matrix (`tools/onboarding-factory`'s viewer), filed together since three of the five share one root cause.

### 1. `2-5_synchronous-slash-command` — tracked separately as **#905**

A content-less `messages.jsonl` touch (mtime bump, no new line) during a synchronous slash command (`/help`, `/status`) force-works the ready session. Root cause: `core/pkg/tailer/tailer.go:494` + `core/application/services/session_detector_activity.go`'s `forceReadyToWorkingIfActive`. **A fix has since landed** (this same PR branch) gating the force-bounce on real transcript growth or a hook-synthetic event — see the `fix(services): stop force-working ready sessions on content-less transcript touches (#905)` commit. The *recording* is frozen evidence from before the fix and still validates as `known_failing`; only a live re-record against the fixed daemon can flip it to observed.

### 2. `1-3_long-idle-live-session` — likely the same mechanism as #905, unverified

Session survives idle past `readyTTL` correctly (core claim holds), but a spurious `ready→working→ready` flicker fires ~2s after each `turn_done`, with **no `transcript_activity` event of its own**, immediately following a `debounce_coalesced` marker. That "no activity event, ~2s after the fact, right after a debounce coalesce" shape is exactly the content-less-touch force-bounce #905 fixed — the recorded evidence:

```
22:13:28.530  force ready→working (no preceding transcript_activity)
22:13:50.589  same pattern, second turn
```

**Ask:** re-record `1-3_long-idle-live-session` against the daemon build containing the #905 fix. If the flicker is gone, drop `known_failing` and flip `daemon_capability` back to `full`. If it persists, it's a distinct mechanism worth its own investigation.

- `core/application/services/pid_manager.go:942-948` (`sweepStaleSnapshot` — the adapter-agnostic PID-liveness skip that correctly keeps the session alive)
- `core/adapters/inbound/agents/vibe/sidecar.go` (meta.json decode — no `session_cost` field read, unrelated to this finding but co-documented in the same assessment)

### 3. `2-10_mid-turn-message-queued` — daemon under-observes a queued turn (distinct root cause)

[...]

PR #1404, opened 2026-08-09
feat(state): grace-timer hysteresis on the one edge that withdraws the cue

### LINKED ISSUE #1366: feat(state): classifier grace-timer hysteresis — the residual of #1129 Phase 3
Parent: #1355 — Phase I. Depends on #1360 (adjacent code).

## Context

#1129's Phase 3 shipped in full — `a6e6aba3` (#1288/#1295), `b43cb4e9` (#1306), `14337b77` (#1318), `12994e4c` (#1323), `84571212` (#1324). Signal tiers, `SignalHolds` with per-signal policies, `ClassifyStateTiered`, `stateRule.tierFor` and `TestStateRules_LadderIsTierConsistent` all exist today.

The one thing it did not deliver is **grace-timer hysteresis**: grepping `core/` for `hysteresis`/`grace` hits only backchannel and pid_manager, never the classifier.

## Why it matters now

#1355 adds hook-tier signals to up to 8 more adapters. `Notification`/`idle_prompt` is ~6s late (#1141), so a late-but-authoritative signal routinely lands against a lower-tier guess the user has already been shown. Eight new sources is eight new ways to flap the badge.

## Scope

1. A minimum dwell time before a state change is published, so a transition that is immediately reversed never reaches the UI.
2. **Asymmetric where the states are asymmetric.** Pinning `waiting` late is a worse error for the user than showing `working` a moment longer. Decide which direction gets the grace, and state the reasoning in the code rather than picking a symmetric number.
3. Must **not** delay the first classification of a session.
4. Clock-injected on the same timeline as `holdContext.Now`, so replay stays deterministic. A `time.Now()` in the classifier breaks the offline harness.
5. Replay goldens: adding hysteresis can legitimately change recorded transitions. If goldens move, the diff is the reviewed deliverable — explain each changed transition in the PR rather than regenerating silently.

**Complexity:** 6

ingo-eichhorst:
Closes #1366.

A classified state change was published the instant the classifier decided it, so a transition reversed one pass later still reached the UI. #1355 adds hook-tier signals to up to eight more adapters and `Notification`/`idle_prompt` is ~6s late (#1141) — eight new ways to flap the badge.

## What this adds

`session.StateDwell` (`core/domain/session/state_dwell.go`): a per-session record of state changes decided but not yet published. Shaped after `SignalHolds` — own mutex, and **the clock passed in on every call**, never read inside. `SessionDetector.admitTransition` replaces the bare `newState != state.State` at the single publication point in `classifyAndTransition`.

No `time.Now()` enters the classifier. The pass clock now comes from `SessionDetector.nowFn()` — the same instant `holdContext.Now` already receives — so the dwell runs on the transcript's virtual timeline under replay. `TestStateDwell_MeasuresTheInjectedClockNotTheWallClock` anchors its virtual timeline in the year 2000, so a `time.Now()` anywhere inside would read decades of elapsed time and publish immediately.

## Which direction gets the grace

**Exactly one edge is debounced: `waiting→working`.** Entering `waiting` is never delayed, from any state.

The reasoning is in `graceFor`'s doc comment. A phantom `waiting` is self-limiting — the session claims it needs a human, the human looks, and looking is what corrects it. A dropped or late `waiting` is silent: nothing prompts the look that would reveal the error. So the grace is spent entirely on staying in `waiting` a moment too long, never on getting there a moment too late.

It is narrower than "leaving waiting", and that narrowing was **forced by evidence, not chosen**. What the grace protects is not the string `waiting` but *the advertisement that a human is needed* — and `ready` carries that advertisement too. `working` is the only state that withdraws it. A first version that debounced every waiting exit broke two existing behaviours:

- `TestSessionDetector_Activity_CancellationFromWaiting_TransitionsToReady` — `state: got "waiting", want ready (ESC from permission prompt)`
- `TestSessionDetector_IdlePromptHook_ReconcilesReadyToWaiting` — `next turn completed: state = "waiting", want ready (idle signal must not leak into the next turn)`

Both are a user action or an authoritative signal, not a flap. `waiting→ready` now publishes immediately and both pass.

Dwell value: **6s**, calibrated to #1141's measured `Notification` lag rather than picked symmetric — see the constant's comment.

## Can a dwell swallow a `hold_expired`-driven transition?

**Yes — and it did.** A #1360 ceiling expiry unpins a session, which is a `waiting→working` edge: exactly the debounced one. It is not merely delayed, it is lost, and the mechanism is invisible from either feature's own tests: `refreshStaleSessions` selects non-working sessions on `SignalHolds.HasAny`, and *dropping the hold is what makes `HasAny` false*. The same pass that starts the dwell is the pass that stops the ticker ever coming back. The session stays pinned at `waiting` for the life of the process — the defect #1360 was opened to remove.

Seen red against a naive dwell (gate on, exemption off):

```
--- FAIL: TestClassify_CeilingExpiryIsNotSwallowedByTheDwell (0.00s)
    state_dwell_test.go:246: the ceiling expiry must unpin the session on the
    pass it fires, got "waiting" — a dwell that defers this can never be
    satisfied, because dropping the hold is what stops refreshStaleSessions
    from revisiting the session at all
--- FAIL: TestClassify_ADwellOutstandingKeepsTheTickerComing/a_dwell_alone_selects_the_session
    state_dwell_test.go:308: a deferred waiting exit must keep the ticker
    coming, or the grace timer is a dropper and not a delay
```

Closed twice over, deliberately:

1. `admitTransition` **bypasses the dwell entirely** when `Overlay` reported an expiry this pass. An expiry is not a flap and cannot become one — the hold is deleted, nothing re-asserts it, and it only fires after hours of wall clock.
2. `shouldRevisitIdleSession` also selects on `StateDwell.Pending`, bounding *every* dwell whatever started it, so the timer is always a delay and never a drop.

A synthesizer that re-based the current state this pass is exempt for the same class of reason: it authors a *pair* of transitions, and holding the second half back would strand the session in a synthetic `waiting` the pass itself knows has ended.

## Red-first evidence

Both captures taken against the final code by reverting only the gate (`git commit -m wip` checkpoint, never `git stash`).

**No hysteresis (pre-#1366 publication rule):**
```
state_dwell_test.go:94: a waiting exit reversed inside the dwell must never be
published, got 2 transition(s): [... PrevState:waiting NewState:working ...
PrevState:working NewState:waiting ...]
--- FAIL: TestClassify_AReversedWaitingExitNeverReachesTheUI (0.00s)
state_dwell_test.go:121: precondition: the first pass must not publish, got "working"
--- FAIL: TestClassify_AWaitingExitThatHoldsIsPublished (0.00s)
```

**LOCKS** — pass on `main` by construction, and say so: `TestClassify_EnteringWaitingIsNotDelayed`, `TestClassify_FirstClassificationIsNotDelayed`, `TestClassify_ADwellOutstandingKeepsTheTickerComing/{nothing_outstanding_is_still_left_alone, a_hold_alone_selects_the_session}`, `TestStateDwell_OnlyWaitingToWorkingIsDebounced`'s undebounced arm.

## Replay goldens: none moved

`tools/preflight.sh --only go`'s `replay fixtures` gate is **PASS**, unmodified. The harness deliberately does not mirror the dwell, documented at the call site in `replay_sidecar.go` with the measurement rather than a hunch:

- The dwell is a *publication* policy; this replayer has no UI, and `runExtendedCheck` diffs `state_transition` events only. The goldens are a regression net for the parsers, tailer and classifier — debouncing that net lowers its resolution, since a correctly derived `waiting→working` with no later event behind it would stop appearing and a future regression that stopped deriving it would be invisible.
- Measured blast radius had it been mirrored: **41 transitions across 15 goldens in 4 adapters** (claudecode 9, codex 3, hermes 2, pi 1).
- Precedent for deliberate daemon/harness divergence on timing-derived behaviour: #988 keeps `ShouldSynthesizeCollapsedTurnBoundary` out of the replay engine ("only the live path has real timing to gate this on"), and #1360 discards `Overlay`'s `[]SignalExpiry` at all three sidecar call sites.
- `ClassifyStateTiered` is byte-for-byte unchanged, so the harness measures exactly what it measured before.

## Verification

Chunked foreground `tools/preflight.sh`, all green: `go` (gofmt, core module tests, onboarding-factory, replaydata validate, recording-rig smoke, **replay fixtures**, starhistory), `web` (both trees), `arch` (ARS), `tools`, `skills`, `security`.

### DIFF
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -384,9 +384,22 @@ func (d *SessionDetector) nowFn() time.Time {
 }
 
 // SetClockForTest replaces the detector's clock. Exported for tests outside
-// this package that need to advance the #1366 dwell without sleeping.
+// this package that need to drive the #1360 ceilings and the #1366 dwell
+// without sleeping.
+//
+// MUST be called before Run. The write is unsynchronised, while nowFn is read
+// from the event-loop goroutine — setting it on a started detector is a data
+// race, not a supported reconfiguration.
 func (d *SessionDetector) SetClockForTest(now func() time.Time) {
 	d.now = now
+}
+
+// StartDwellForTest records a decided-but-unpublished state change (#1366)
+// with an explicit start time. Exported only so tests outside this package can
+// set up a dwell the ticker then has to notice; production starts one solely
+// from inside the classify pass, via admitTransition.
+func (d *SessionDetector) StartDwellForTest(sessionID, current, candidate string, at time.Time) {
+	d.dwell.Admit(sessionID, current, candidate, at)
 }
 
 // HoldSignalForTest places an out-of-band signal hold with an explicit
@@ -715,7 +728,7 @@ func (d *SessionDetector) handleTerminalUISignal(sig terminalUISignal) {
 			Adapter: state.Adapter, UIKind: string(sig.ui), Reason: uiReason,
 		})
 
-		now := time.Now().Unix()
+		now := d.nowFn().Unix()
 		d.record(lifecycle.Event{
 			Kind: lifecycle.KindStateTransition, SessionID: sig.sessionID,
 			PrevState: state.State, NewState: newState, Reason: reason,
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -372,26 +372,29 @@ func (d *SessionDetector) RunStaleSessionRefreshForTest() {
 }
 
 // nowFn reads this detector's clock. The indirection is what lets a test drive
-// the ceiling and dwell timers on a virtual timeline; production leaves the
-// field nil and gets time.Now. Every elapsed-time decision in the classify
-// pipeline must come through here — a bare time.Now() alongside it would make
-// two mechanisms that are supposed to share one instant disagree.
+// the #1360 ceilings and the #1366 dwell on a virtual timeline; production
+// leaves the field nil and gets time.Now.
+//
+// WHAT MUST COME THROUGH HERE: every instant that is later compared against
+// another instant by the classify pipeline. Concretely that is the pass clock
+// (holdContext.Now and the dwell), every SignalHolds placement (HeldSince is
+// measured against the pass clock by both ceiling and ripe), and every
+// SessionState.UpdatedAt write (reclassifyFromTranscript measures the refresh
+// interval against it). Mixing a bare time.Now() into any of those makes two
+// stamps that are supposed to share one timeline disagree, and the symptom is
+// a guard silently short-circuiting rather than anything failing loudly.
+//
+// DELIBERATELY EXEMPT, because they are self-consistent wall-clock pairs that
+// no pipeline guard reads: the deletedSessions tombstone and its prune
+// threshold (session_detector_helpers.go / session_detector_lifecycle.go), and
+// historyTracker timestamps, which pair with record()'s own event stamping.
+// Naming them is the point — otherwise a reader cannot tell "deliberately wall
+// clock" from "missed in the sweep".
 func (d *SessionDetector) nowFn() time.Time {
 	if d.now != nil {
 		return d.now()
 	}
 	return time.Now()
-}
-
-// SetClockForTest replaces the detector's clock. Exported for tests outside
-// this package that need to drive the #1360 ceilings and the #1366 dwell
-// without sleeping.
-//
-// MUST be called before Run. The write is unsynchronised, while nowFn is read
-// from the event-loop goroutine — setting it on a started detector is a data
-// race, not a supported reconfiguration.
-func (d *SessionDetector) SetClockForTest(now func() time.Time) {
-	d.now = now
 }
 
 // StartDwellForTest records a decided-but-unpublished state change (#1366)
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -182,6 +182,19 @@ type SessionDetector struct {
 	// event loop classifies), so it is deliberately NOT guarded by idleMu.
 	signals *session.SignalHolds
 
+	// dwell is the grace timer state changes serve before they are published
+	// (#1366). Like signals it carries its own lock and takes the pass clock
+	// as an argument, so it is deliberately NOT guarded by mu. A nil value
+	// disables hysteresis; see session.StateDwell.
+	dwell *session.StateDwell
+
+	// now is this detector's clock, read once per classify pass and threaded
+	// into everything that reasons about elapsed time — the signal holds'
+	// ceilings and ripeness rules (holdContext.Now) and the #1366 dwell. nil
+	// means time.Now; tests replace it to drive both mechanisms on a virtual
+	// timeline instead of sleeping. Read through nowFn, never directly.
+	now func() time.Time
+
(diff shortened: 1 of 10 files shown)


ingo-eichhorst:
## Review + simplify rounds (commits 2 and 3)

A `high`-effort review subagent returned **6 findings, all applied**; a 4-agent simplify pass returned a further set, applied or explicitly skipped. Two of them changed behaviour, so they belong in the PR description rather than only the commit log.

### Behaviour changes since the first commit

**A hook-tier verdict is never debounced** (found on the altitude axis). `compact_in_progress` is a `TierHook` rule that decides *working*. A manual `/compact` issued while the session read `waiting` was serving the full 6s dwell plus a ticker interval — directly contradicting that hook handler's own reason for existing ("there is no transcript flush coming during the compaction window"). Now exempt, covered by `TestClassify_AHookTierVerdictIsNotDebounced`.

It does **not** subsume the other two bypasses, which I checked rather than assumed: after a ceiling expiry, and after the collapsed-waiting synthesizer re-bases, the deciding rule is `transcript_activity` at `TierTranscript`.

**The clock seam had a trap in it.** The injected clock was compared against wall-clock `UpdatedAt`, so `reclassifyFromTranscript` silently no-opped under any test clock behind wall time — which would have made the missing liveness test green while exercising nothing. The reviewer demonstrated this with a throwaway probe, not an argument. Every `UpdatedAt` write and every hold placement in the detector now goes through `nowFn`, and the interval guard is direction-safe.

### Two guards were shipping untested, and are not any more

Mutation testing found that deleting the synthesizer bypass, or the reap-path eviction, left the entire package green. Both now have a test that was **seen red**:

```
--- FAIL: TestClassify_ASynthesizedWaitingIsNotLeftStranded
    expected the synthesized pair (working→waiting, waiting→working), got 1
--- FAIL: TestSessionDetector_ReapingASessionEvictsItsDwellAndHolds
    a reaped session must not leave a pending transition behind
    a reaped session must not leave a signal hold behind
```

The reap finding also closes a **pre-existing `SignalHolds` leak**: `onRemoved` fires on the transcript file disappearing, but a dying process does not delete its transcript, so most sessions tear down through the PIDManager path — which was clearing neither store. Teardown is now one seam, `forgetSessionScopedState`, called from both paths.

And the end-to-end liveness test the review found missing (everything pinned `shouldRevisitIdleSession`'s *return value*; nothing pinned that the ticker then runs a real pass), also seen red:

```
--- FAIL: TestSessionDetector_StaleRefreshFinishesAnOutstandingDwell
    the ticker never ran a classify pass for a session with a dwell outstanding
```

### Deliberately skipped, with reasons

- **Making `SignalHolds` nil-safe** to delete two `!= nil` guards — belongs to #1360's type, not this diff.
- **An `IsStaleAt` domain helper** — one caller, and the direction guard is deliberately local to its long rationale.
- **Removing `pendingExit`'s `from`/`to`** — a mutation proved the shape check is unreachable while `graceFor` debounces a single edge. Kept as defence-in-depth for a future second edge; the comment and the two test docs that implied a live invariant are corrected to say so plainly.
- **A dirty-check in `finalizeActivityPass`** so a no-op refresh pass stops doing `Save` + `broadcast` + cache invalidation. Real, pre-existing, and the thing that makes any widening of the ticker's reach expensive — worth its own issue rather than riding along here.

Goldens still unchanged: `replay fixtures` PASS on every run.

Issue #1377, opened 2026-08-09
fix(claudecode): instructions consent copy names the retired task-summary block, never names task-question (#570 violation)


ingo-eichhorst:
## What

The Claude Code `instructions` permission's consent copy names a managed block that is no longer written, and never names one that is. Same class as #1356 (the hooks permission's undisclosed Notification entry), different permission.

`core/adapters/inbound/agents/claudecode/agent.go:114-123`:
```
FeatureUnlocked: "Task-completion ETA chip + task summary from agent-reported markers",
Touches:         "Maintains two managed blocks in ~/.claude/CLAUDE.md",
Detail:          "... instructing the agent to emit hidden markers: a task-progress
                  marker (which irrlicht reads to project a completion ETA) and a
                  one-line task summary (so a human can tell what a session is about)."
```

`core/adapters/inbound/agents/claudecode/instructioninstaller.go:256-265` — what `Apply` actually does:
```go
func applyInstructionBlocks() error {
	if _, err := EnsureTaskEtaBlockInstalled(); err != nil { return err }
	if _, err := UninstallTaskSummaryBlock(); err != nil { return err }
	_, err := EnsureTaskQuestionBlockInstalled()
	return err
}
```

So the two blocks installed are **task-eta** and **task-question** (#759). **task-summary** was retired in #1186 and is *removed* on apply, not written — yet it is the block the copy names. The task-question block, which instructs the agent to emit a pending-question marker at the end of every turn, is disclosed nowhere.

## Why it matters

The count ("two managed blocks") is right by luck; the contents are wrong in both directions — one named block is not written, one written block is not named. Per #570 the `Touches`/`Detail` text *is* the consent contract for a `modify`-kind permission, so this is the same violation shape as #1356, in the adjacent permission of the same `Agent()` literal.

## Fix

1. Correct the copy: name task-eta and task-question, drop the task-summary mention (including in `FeatureUnlocked`).
2. Consider generalizing #1356's contract. `contracttesting.AssertHookDisclosureMatchesInstalled` is keyed on hook events, so it structurally cannot see this instance. The deeper seam suggested during #1356's review: an `Installs []string` inventory on `agent.Permission` that both the `Apply` closure and the copy read, plus one contract over *every* `KindModify` permission asserting copy ⊇ `Installs` — that subsumes hooks, instructions and statusline at once.

## Verification

[...]

Issue #1357, opened 2026-08-09
fix(hooks): --uninstall-hooks and the recording rig hardcode two adapters — guaranteed to be missed on adapter #3


ingo-eichhorst:
## What

Two independent places enumerate "the agent config files irrlicht writes hooks into" as a hardcoded two-element list. Both are correct today and both are guaranteed to be forgotten the first time a third hook adapter lands.

**1. Uninstall** — `core/cmd/irrlichd/main.go:95-97`:
```go
func uninstallHooks() {
	reportUninstall("~/.claude/settings.json", claudecode.UninstallHooks)
	reportUninstall("~/.codex/hooks.json", codex.UninstallHooks)
```
(The consent-revoke block below it hardcodes the same two adapter/key pairs.)

**2. The recording rig** — `tools/onboarding-factory/scripts/lib/hook-config-snapshot.sh:32`:
```bash
HOOK_CONFIG_FILES=("$HOME/.claude/settings.json" "${CODEX_HOME:-$HOME/.codex}/hooks.json")
```

## Why each is harmful

**Uninstall.** An adapter not listed leaves its hook entries in the user's real config **permanently** after irrlicht is uninstalled — pointing at a dead port, firing on every turn. The user has no way to know where they came from.

**The recording rig is the worse of the two.** Its own header explains the hazard it exists to prevent: a recording daemon runs `IRRLICHT_PERMISSION_MODE=grant-all` and installs hooks into the user's **real** config, because those paths follow $HOME and `IRRLICHT_ONBOARD_HOME` does not isolate them. Since #1178 the endpoint carries the recorder's own bind port. So recording a hook-enabled adapter that is not in this list repoints the user's real config at a dead recorder port and leaves it there — exactly #1214, one adapter later.

There is a second, sharper failure in `restore_hook_configs`: a listed file with no backup is treated as "the daemon created it" and `rm -f`'d. Adding a path that is a **multi-purpose** config the daemon merely happened to create therefore **deletes the user's file**. `~/.gemini/settings.json` is exactly that shape — it holds far more than hooks. So the naive fix (append gemini's path to the array) is itself dangerous; the restore logic needs to distinguish "file we created" from "file we edited" before the list grows.

## Fix

- Drive both off the agent registry rather than a literal list. The adapters already declare their hooks permission and their uninstall closure; the paths should come from the same declaration (this is the `HookSpec` seam #1355 Phase F introduces, but neither fix needs to wait for it).
[...]

Issue #1214, opened 2026-07-31
run-cell-multi.sh duplicates run-cell.sh's record-daemon spawn/teardown, so recorder fixes reach only one path


ingo-eichhorst:
## Problem

Two scripts spawn an `irrlichd --record` daemon, and each carries its own copy of the spawn + teardown block:

| | `run-cell.sh` | `run-cell-multi.sh` |
|-|-|-|
| env assembly | `:255` (`DAEMON_ENV=(…)`) | `:200` (`env … \`) |
| `SHUTDOWN_REASON` + INT→TERM→KILL escalation | `:271-292` | `:208-231` |

Same three env vars (`IRRLICHT_RECORDINGS_DIR`, `IRRLICHT_BIND_ADDR`, `IRRLICHT_PERMISSION_MODE=grant-all`), same 6s-grace escalation ladder, same `daemon.shutdown` write — written twice. `run-cell-multi.sh` does **not** delegate to `run-cell.sh`; it sources `lib/shard-lib.sh`, `lib/reconcile.sh` and `lib/recipe-lint.sh`, then spawns its own daemon.

## Why it bit

#1178 / PR #1207 had to make the recorder snapshot and restore the shared agent config it rewrites (`~/.claude/settings.json`, `$CODEX_HOME/hooks.json` follow `$HOME`, not `IRRLICHT_HOME`, so `IRRLICHT_ONBOARD_HOME` never isolated them — and since #1178 the endpoint written there carries the recorder's own port).

The first cut added that inline in `run-cell.sh` only. `run-cell-multi.sh` kept spawning its own daemon with no snapshot, so the cross-adapter recorder would still have exited leaving the user's config repointed at a dead `:7838` — and a comment added to `run-cell-multi.sh` in the same change *claimed* it was covered. Review caught it; the fix was to hoist into `lib/hook-config-snapshot.sh` and source it from both.

That is the failure mode this duplication produces: a recorder-lifecycle fix silently covers half the surface, and the divergence is invisible because both scripts read as complete.

## Proposed

[...]

PR #1409, opened 2026-08-09
fix(claudecode): derive the instructions consent copy from the blocks Apply installs

ingo-eichhorst:
Closes #1377.

## What was wrong

The Claude Code `instructions` permission's consent copy described an install
that never happened and hid one that did:

- it named **task-summary**, retired in #1186 — `applyInstructionBlocks` *removes*
  that block, it does not write it;
- it never named **task-question** (#759), which it has been writing into the
  user's `~/.claude/CLAUDE.md` on every grant since;
- `FeatureUnlocked` advertised "task summary from agent-reported markers", a
  capability that no longer exists.

The count ("two managed blocks") was right by luck. Per #570 the `Touches`/`Detail`
text *is* the consent contract for a `modify`-kind permission, so this is the same
violation shape as #1356, one permission over in the same `Agent()` literal.

## What changed

`installedInstructionBlocks` (in `instructioninstaller.go`) is now the single
declaration of what the grant effect writes — name, sentinels, content, and the
clause the copy uses to disclose it. `applyInstructionBlocks` iterates it, and
`instructionsTouched()` / `instructionsDetail()` render the copy from it, so the
count and the list cannot be restated by hand and cannot drift. Retired blocks
live in their own `retiredInstructionBlocks` list: `Apply` sweeps them, and
because the copy is rendered from the *installed* list only, a retired block can
never be presented as a write. The cleanup is still disclosed, as a removal,
without naming a historical marker the user has no reason to care about.

Rendered copy today (captured by *executing* `Agent()` and printing the
fields, not transcribed from the format strings):

> **Touches** — Maintains 2 managed blocks (task-eta and task-question) in ~/.claude/CLAUDE.md
>
> **Detail** — Writes 2 irrlicht-managed blocks (each between BEGIN/END sentinels)
> to ~/.claude/CLAUDE.md, instructing the agent to emit hidden markers: task-eta —
> a task-progress marker, which irrlicht reads to project a completion ETA;
> task-question — a one-line version of the question it is waiting on, so a human
> can tell what a session is blocked on. All surrounding file content is preserved
> byte-for-byte, and a managed block an earlier irrlicht version wrote but this one
> no longer uses is removed on the same pass. Toggling off removes these blocks and
> any such retired block, and nothing else (also available via the macOS Settings
> toggle).
>
> **FeatureUnlocked** — Task-completion ETA chip + waiting-question headline from
> agent-reported markers

**Deliberately not in scope.** The issue's fix step 2 — an `Installs []string`
[...]

PR #1411, opened 2026-08-09
fix(recording): protect every managed user file, not just hook configs

### LINKED ISSUE #1383: fix(recording): grant-all recordings leave CLAUDE.md and kitty.conf modified — only hooks configs are snapshotted
## What

#1357 made the recording rig's protected-file set a projection of the adapter registry — but it projects only permissions that declare `agent.HookInstall`, i.e. **hooks**. Other `modify`-kind permissions also write shared, user-owned files that follow `$HOME` (not `IRRLICHT_HOME`), and a recording daemon runs `IRRLICHT_PERMISSION_MODE=grant-all`, so it exercises *every* `Apply` closure — including theirs. Those files are still neither snapshotted nor restored by `tools/onboarding-factory/scripts/lib/hook-config-snapshot.sh`.

This was surfaced and deliberately deferred while implementing #1357 (PR #1375); filing it so the deferral is a tracked decision rather than a note in a merged PR description.

## The unprotected set

| Permission | File written | Isolated by `IRRLICHT_ONBOARD_HOME`? |
|---|---|---|
| `claudecode` / `instructions` | `~/.claude/CLAUDE.md` | No — `claudeMemoryPath()` is `filepath.Join(os.UserHomeDir(), ".claude", "CLAUDE.md")` (`core/adapters/inbound/agents/claudecode/instructioninstaller.go:173-179`) |
| `kitty` / `remote-control` | `$XDG_CONFIG_HOME/kitty/kitty.conf`, else `~/.config/kitty/kitty.conf` | No — `kittyConfPath()` (`core/adapters/inbound/agents/processlifecycle/kitty_permission.go:150`) |

Both are declared on entries the rig could not have seen anyway: `kitty` is one of three non-agent declarations appended to the consent catalog in `core/cmd/irrlichd/startup.go` (with `gastown` and `launcher`), which `agents.HookConfigs` does not project at all. `core/cmd/irrlichd/hookconfigs_test.go`'s `TestNonAgentPermissionDeclarationsDeclareNoHooks` is the tripwire holding that gap visible.

A recording that grants these leaves the user's real `CLAUDE.md` carrying irrlicht's managed instruction blocks, and their real `kitty.conf` with `allow_remote_control yes` + `listen_on unix:/tmp/kitty`, after the recording daemon is gone.

## Why each is lower risk than the hooks case

The hooks hazard (#1214, #1178, #1357) is that the written content carries **the recording daemon's own bind port**, so a leftover entry points at a dead port and fires on every turn — broken state, not merely unwanted state. These two are different on all three counts:

- **Sentinel-delimited — VERIFIED by reading the code.** `instructioninstaller.go:28-29, 101-102, 109-110` define `<!-- BEGIN/END IRRLICHT MANAGED BLOCK (task-eta|task-summary|task-question) -->`; `kitty_permission.go:38-40` defines `# >>> irrlicht remote control >>>` / `# <<< ... <<<`. Uninstall removes exactly the delimited block (`uninstallBlock`, `UninstallKittyConfig`), and kitty refuses outright on an unbalanced marker pair (`errKittyBlockMalformed`) rather than risk swallowing user lines.
- **Idempotent — VERIFIED by reading the code.** `ensureBlockInstalled` (`instructioninstaller.go:185-200`) writes only when `patchManagedBlock` reports `changed`; `EnsureKittyConfigPatched` (`kitty_permission.go:103-121`) compares the rebuilt file to the original and returns `false, nil` when identical. Repeated recordings therefore do not accumulate content.
- **Port-free — VERIFIED by reading the code.** `grep -n 'daemonaddr\|hookEndpointURL\|localhost\|127.0.0.1\|:7837\|BindAddr\|Port'` over `instructioninstaller.go`, `kitty_permission.go` and `kitty_config.go` returns **no matches**. The instruction blocks are static prose; the kitty block is the static pair `allow_remote_control yes` / `listen_on unix:/tmp/kitty` (a fixed unix socket path, not the daemon's bind address).

**Assumed, not verified:** that no *user-visible misbehaviour* follows from a leftover instruction block (an agent would keep emitting `irrlicht-eta` markers into transcripts with no daemon reading them — cosmetic as far as I can tell, but I did not test it). Also assumed: that `~/.gemini/settings.json` remains irrelevant here — `geminicli` declares no `modify` permission today (`core/adapters/inbound/agents/geminicli/agent.go`), so nothing in the daemon writes it; the only writes in the repo are workspace-local, in `replaydata/agents/gemini-cli/driver-interactive.sh`.

So the residue is a **consent violation** — content installed under `grant-all` that the user may have explicitly denied — rather than the broken-state hazard hooks pose. Real, but not the same severity.

## What would change that

Any of the following flips one of these into the hooks severity class and should force this issue, not a fresh discovery:

1. **A managed block gains port-bearing or otherwise environment-specific content** — e.g. the kitty block's `listen_on` becoming per-daemon, or an instruction block embedding a daemon URL. Then a leftover points at a dead recorder exactly like #1214.
2. **A managed block stops being idempotent** — anything that appends rather than patches in place makes repeated recordings accumulate.
3. **`claudecode/statusline` moves out of `settings.json`.** This is the sharpest one. `statusline` writes `statusLine.command` containing `daemonaddr.LocalURL(StatuslineEndpointPath)` (`core/adapters/inbound/agents/claudecode/statusline_installer.go:59-62`) — it **is** port-bearing, and it is protected today **only incidentally**, because it happens to share `~/.claude/settings.json` with the `hooks` permission that declares the file. Nothing asserts that coincidence. If statusline ever moved to its own file, it would be both unprotected and port-bearing on day one.
4. **A hooks declaration lands on a non-agent catalog entry** (gastown/launcher/kitty), which `agents.HookConfigs` does not project. `TestNonAgentPermissionDeclarationsDeclareNoHooks` fails loudly if so.

## Suggested direction (not a decision)

Generalize the seam rather than adding a second list: broaden `agent.HookInstall` (or add a sibling) to "shared user-owned file this permission writes", have `agents.HookConfigs` become a `ManagedUserFiles`-shaped projection over the **full consent catalog** — `All()` plus the three declarations `startup.go` appends — and let `--uninstall-hooks` filter on the `hooks` key while the recording rig protects everything. Point 4 above is a prerequisite either way. Note `--print-hook-configs` is a shipped flag name with exactly one consumer (the rig), so it can be renamed cheaply if the concept widens.

## References

- #1357 — the hooks half of this, fixed in PR #1375
- #1214, #1178 — the recording-rig hazards that motivated the snapshot

ingo-eichhorst:
Closes #1383.

## What

#1357 (PR #1375) made the recording rig's protected-file set a projection of the adapter registry — but it projected only permissions declaring `agent.HookInstall`, i.e. **hooks**. A recording daemon runs `IRRLICHT_PERMISSION_MODE=grant-all`, which grants every consent-gated capability and therefore runs **every** modify permission's `Apply` closure against the user's real `$HOME`. `~/.claude/CLAUDE.md` and `~/.config/kitty/kitty.conf` were written and never handed back.

Per the issue's "Suggested direction", the seam is generalized rather than a second list added:

- **`agent.HookInstall` → `agent.ManagedUserFile`** (`Path` + `Uninstall` + `Version`), and **`Permission.Hooks` → `Permission.Writes`** — "the shared user-owned file this permission writes". `agent.HooksPermissionKey` is now the single spelling of `"hooks"`; `claudecode.PermissionKeyHooks` and `codex.PermissionKeyHooks` alias it instead of restating it.
- **`agents.HookConfigs` → `agents.ManagedUserFiles`**, projecting anything that declares `Writes`. `HookConfigs` survives as **the hooks slice of it**, so `--uninstall-hooks` behaviour is unchanged: it must not revoke the CLAUDE.md instruction blocks or the kitty patch nobody asked it to touch. The two consumers deliberately read different slices of one declaration.
- **Issue point 4 (the prerequisite) is handled.** `consentCatalog` (`core/cmd/irrlichd/consentcatalog.go`) is now the single composition of `agents.All()` plus the three daemon-wide declarations `startup.go` used to append inline (gastown, launcher, kitty). `setupPermissionService` and both flag paths read it. Projecting only the adapter registry is exactly how the kitty config patch was offered by the wizard while being invisible to both lists.
- **New declarations:** `claudecode/instructions` (`~/.claude/CLAUDE.md`, via a new exported `UninstallInstructionBlocks`), `claudecode/statusline`, and `kitty/remote-control`.
- **`--print-hook-configs` → `--print-managed-files`** (renamed with the concept, per the issue). The rig's `lib/hook-config-snapshot.sh` is renamed to `lib/managed-file-snapshot.sh` with its functions (`snapshot_managed_files` / `restore_managed_files` / `managed_file_paths`) and state var; `site/docs/cli-tools.html` and `AGENTS.md` updated in the same PR.

`statusline` is declared even though it resolves to the same `settings.json` the hooks permission declares — that is issue point 3. It writes a command carrying the daemon's bind port and was protected **only** by the coincidence of sharing a file with `hooks`, with nothing asserting the coincidence. It stays covered if it ever moves.

## The safety bar PR #1375 set is unchanged

The `saved`/`absent` manifest and move-aside-not-delete are untouched; widening the set adds cases on top of them rather than relaxing them. Every test that exercises snapshot/restore still runs against a fake `$HOME`/`$XDG_CONFIG_HOME`/`$CODEX_HOME` under `mktemp -d`. Nothing in this PR reads or writes a real `~/.claude`, `~/.config/kitty` or `~/.codex`.

## Defect tests, seen RED first

Written and run against **unmodified `origin/main`** (as a main-compatible probe — the final tests name the new API and could not compile there; the assertions and messages are the same):

```
=== RUN   TestEveryModifyPermissionDeclaresTheFileItWrites
    redfirst_1383_test.go:31: claude-code/statusline is a modify permission with an Apply closure but declares no managed user file: a grant-all recording daemon runs that closure against the user's real $HOME and the recording rig has no way to learn which file to hand back (#1383)
    redfirst_1383_test.go:31: claude-code/instructions is a modify permission with an Apply closure but declares no managed user file: ...
    redfirst_1383_test.go:31: kitty/remote-control is a modify permission with an Apply closure but declares no managed user file: ...
--- FAIL: TestEveryModifyPermissionDeclaresTheFileItWrites (0.00s)
=== RUN   TestPrintManagedFilesNamesTheFilesTheIssueFound
    redfirst_1383_test.go:56: the flag never names a /.claude/CLAUDE.md file, so a grant-all recording leaves claude-code/instructions's managed block in the user's real file (#1383):
        /Users/ingo/.claude/settings.json
        /Users/ingo/.codex/hooks.json
    redfirst_1383_test.go:56: the flag never names a /kitty/kitty.conf file, so a grant-all recording leaves kitty/remote-control's managed block in the user's real file (#1383):
        /Users/ingo/.claude/settings.json
        /Users/ingo/.codex/hooks.json
--- FAIL: TestPrintManagedFilesNamesTheFilesTheIssueFound (0.00s)
FAIL	irrlicht/core/cmd/irrlichd	0.399s
```

Both are the bug: three modify permissions write user-owned files nothing declares, and the flag the rig reads names neither of the two files the issue was filed about. Green after the fix.

## Deliberate mutations, one per contract obligation

Contract assertions pass by construction, so each was made to fail on purpose and reverted (AGENTS.md).

| # | Mutation | Observed failure |
|---|---|---|
| M1 | the flag path builds its own `agents.All()`-only list instead of `consentCatalog` | `the consent catalog omits the daemon-wide "gastown"/"launcher"/"kitty" declaration…` + `--print-managed-files never names a /kitty/kitty.conf file` |
| M2 | `HookConfigs` stops narrowing on `HooksPermissionKey` | `HookConfigs returned claude-code/statusline, which is not a hooks permission — --uninstall-hooks would revoke something the user never asked it to` (×3), `ManagedUserFiles returned 5 entries and HookConfigs 5`, plus `HookConfigs() returned 4 entries for 2 declared hooks permissions` and the synthetic-catalog arm |
| M3 | `printManagedFiles` emits only the first entry | `claude-code/instructions declares "…/CLAUDE.md", which --print-managed-files does not list` (×3) |
| M4 | `writeManagedFilePaths` drops the dedup | `printed "/Users/ingo/.claude/settings.json" twice` |

M4 is worth a note: the dedup was previously testable **only** synthetically (`TestWriteManagedFilePathsDeduplicates` said so in its comment). Now that `hooks` and `statusline` both declare `~/.claude/settings.json`, the **shipped** registry produces the duplicate, so `TestPrintManagedFilesCoversEveryDeclaredFile` catches it too.

## Locks (pass by construction — not red-first evidence)

- **All three new `managed-file-snapshot_test.sh` cases.** The lib is path-agnostic, so it handles a `CLAUDE.md` exactly as it handles a `settings.json` and these pass on `origin/main` too. The defect is upstream of the lib — the *daemon* never named these files. They are here to make the widened set's behaviour explicit:
  - `a user's own instruction file is restored byte-for-byte (#1383)` — pre-existing `CLAUDE.md`, managed block appended mid-run, restored to a byte-identical `cksum`.
  - `a host config the daemon created is moved aside, not destroyed (#1383)` — no `kitty.conf` before the run, created by the Apply closure, path absent afterwards **and** contents recoverable under `created/`.
  - `a multi-purpose file the daemon created mid-run keeps the user's own content (#1383)` — the case `--print-managed-files`'s move-aside exists for: no `CLAUDE.md` before the run, the daemon creates one, the **agent** then writes the user's own rule into the same file. Pre-run state (absent) restored; the user's mid-run content recoverable.
- Every pre-existing `managed-file-snapshot_test.sh` and `spawn-record-daemon_test.sh` case, carried forward unchanged apart from the rename — including `restore acts only on paths the manifest records`, `a directory at a declared config path is left alone`, and the `#1214` single-owner assertions.
- `TestConsentCatalogMatchesThePermissionWizard` — a lock as long as both callers go through `consentCatalog`; M1 is what proves it can fail.
- `TestUninstallHooksReadsOnlyTheHooksSlice` — a lock relative to `main` (where the projection was hooks-only by construction), a regression test relative to this branch (where the narrowing is a removable filter); M2 proves it.
- `TestManagedUserFilesIgnoresPermissionsThatDeclareNoFile`, `TestUninstallHookConfigsLeavesUngrantedPermissionsAlone`, `TestUninstallHookConfigsSurvivesOneFailingAdapter`, `TestManagedUserFilesRejectsUnusableDeclarations`.

## Tests that changed meaning

- **`TestNonAgentPermissionDeclarationsDeclareNoHooks` is deleted, not weakened.** Its premise — "non-agent declarations are NOT projected, so they must declare nothing" — is exactly what this PR dissolves; kitty now declares a managed file on purpose. It is replaced by `TestConsentCatalogMatchesThePermissionWizard`, which pins the stronger property the tripwire was standing in for: the catalog the flags project **is** the catalog the wizard offers.
- **`TestEveryHookInstallDeclaresAVersionFloor` now narrows on `agent.HooksPermissionKey`.** A version floor is an obligation of installing into an agent's *own* config format, where an older CLI can reject what irrlicht writes. The CLAUDE.md and kitty blocks are files irrlicht owns end-to-end and depend on no upstream feature; a blanket rule would have failed them for no reason. Recorded in the test's comment and in AGENTS.md.

## Manual check on a real build

```
$ irrlichd --print-managed-files
/Users/ingo/.claude/settings.json
/Users/ingo/.claude/CLAUDE.md
/Users/ingo/.codex/hooks.json
/Users/ingo/.config/kitty/kitty.conf

$ XDG_CONFIG_HOME=/tmp/altxdg CODEX_HOME=/tmp/altcodex irrlichd --print-managed-files
/Users/ingo/.claude/settings.json
/Users/ingo/.claude/CLAUDE.md
/tmp/altcodex/hooks.json
/tmp/altxdg/kitty/kitty.conf
```

`settings.json` appears once despite two permissions declaring it, and both env overrides are honoured.

## Gates

- `tools/preflight.sh` (unscoped) — **13/13 PASS**: gofmt, core module tests, onboarding-factory tests, replaydata validate, recording-rig smoke test, replay fixtures, starhistory tests, both web suites, ARS architecture gate, tools/lib shell-lib tests, skill-file lint, security scan.
- `bash tools/onboarding-factory/scripts/smoke-test.sh` — **PASS**.
- Rebased onto `origin/main` after #1402/#1403/#1404 landed; `go test ./core/... -race -count=1` and the smoke test re-run green afterwards. #1403 touched `AGENTS.md` and `core/cmd/irrlichd/startup.go`; both reconciliations were read on both sides (their startup.go line is `buildDiagnostics`, unrelated to the catalog; their AGENTS.md edit bumped "five contract families" to "six", and the new Managed-user-files bullet is placed after that paragraph as a separate seam, not a contract family).

## Coordination note

**#1377 had not landed when this was pushed.** It corrects the consent copy of the `instructions` permission literal in `core/adapters/inbound/agents/claudecode/agent.go` — the same literal this PR adds a `Writes:` declaration to, so a rebase conflict there is expected and is a straight both-sides merge. This PR does not touch that copy.

**Follow-up this seam makes possible (not done here):** `Writes.Path` is now the machine-readable answer to "which file does this permission write", while every permission's `Touches`/`Detail` still names that path as hand-written prose — including `~/.claude/CLAUDE.md` in the instructions copy #1377 is fixing. The same derivation `hookjson.EntriesTouched` did for the hook count in #1370 could be done for the path, so the consent copy and the declared file cannot disagree. Deliberately out of scope: it would rewrite the copy #1377 is mid-flight on.


ingo-eichhorst:
## Rebased onto `c2bb439a` (was `CONFLICTING`, now `MERGEABLE`)

Handed back at Phase 6 by the previous session; rebased twice (main moved mid-run) and re-verified from scratch. New head **`a3855479`**, 4 commits on `origin/main@c2bb439a`.

### Conflicts and how each was resolved

| File | Conflict | Resolution |
|---|---|---|
| `claudecode/agent.go`, `codex/agent.go` | #1422 added `Verify` to `HookInstall`; this PR renames it to `ManagedUserFile`/`Writes`/`Path` | Kept both — renamed fields **and** carried `Verify: VerifyHooksInstalled` |
| `cmd/irrlichd/main.go` | #1412/#1413 refactored into `selectAction`/`runCLIAction`; this PR had an inline flag chain | Took upstream's structure, re-applied the rename **into** it (`actionPrintManagedFiles`) |
| `claudecode/agent.go` (2nd rebase) | #1409 replaced hand-written instructions consent copy with derived `instructionsTouched()`/`instructionsDetail()` | Kept #1409's derived copy, appended only the `Writes` declaration. **Taking this branch's side would have silently reverted #1409.** |

### Textually clean but semantically wrong — caught, not merged

Four auto-merges compiled or read fine and were wrong:

1. **`instructioninstaller.go` — duplicate `UninstallInstructionBlocks`.** Both sides added one; git merged both. This branch's called `UninstallTaskEtaBlock`/`TaskSummary`/`TaskQuestion`, which #1409 deleted. Kept upstream's inventory-driven version (that's #1377's fix), folded in the `ManagedUserFile.Uninstall` doc note.
2. **`HookChannelReady`** — a blind `p.Hooks`→`p.Writes` rename widens a hooks-only filter to `statusline`, `instructions` and kitty, arming the liveness watchdog for adapters whose hooks were *denied*. Narrowed to `p.Key != agent.HooksPermissionKey`.
3. **`hookPermission`** (feeds `RepairGrantedHookInstall`) — same widening would let the re-verification loop re-run arbitrary effects by name, which is exactly what that filter exists to prevent. Narrowed.
4. **`hookAdapterNames`** — would earn a hook-liveness row for an adapter that only patches CLAUDE.md, i.e. a permanently silent channel reported as broken. Narrowed.

`declaration.go` auto-merged **correctly** (verified, not assumed): `ManagedUserFile` kept `Verify`, `HookEntryStatus` and the `strings` import.

Rename sites the branch never saw: `core/cmd/irrlichd/dispatch_test.go` (new in #1412) still asserted `--print-hook-configs`. Swept repo-wide — no stale references remain. `site/install.sh`'s #1421 sweep calls `--uninstall-hooks`, not the renamed flag, so it is unaffected.

`AGENTS.md`: upstream's "All seven contract families…" closer had this branch's new bullet appended *after* it. Count re-derived (`grep -c "^func Assert"` = **7**, correct); paragraph moved back to the end of the run.

### Re-verified on the rebased tree

Red-first and per-obligation mutations were all re-run and all reddened:

- **#1383 defect test** — removing both new `Writes` declarations: `TestPrintManagedFilesNamesTheFilesTheIssueFound` and `TestEveryModifyPermissionDeclaresTheFileItWrites` fail, output showing the pre-fix list (`settings.json` + `hooks.json` only).
- **`HookChannelReady` narrowing** — reverting it fails `TestHookChannelReady_IgnoresNonHookPermissions`. Its fixture was strengthened to match post-#1383 reality (statusline now declares a `ManagedUserFile`); without that the row could not catch this.
- **Verifier tripwire scoping** — unscoping it fails `TestEveryHookInstallDeclaresAVerifier` for `claude-code/statusline` and `claude-code/instructions`.

**End-to-end snapshot/restore, against a temp `HOME`** (both files, both directions):

- CLAUDE.md pre-existing + kitty.conf created → CLAUDE.md restored byte-for-byte, kitty.conf removed, `.config/kitty` removed.
- Mirror (kitty.conf pre-existing + CLAUDE.md created) → kitty.conf restored byte-for-byte, CLAUDE.md removed.
- The `absent` record moves the file aside into `backup/created/<n>` rather than `rm -f`-ing it, and a pre-existing file is never deleted.

`--print-managed-files` under a temp HOME now lists `.claude/settings.json`, `.claude/CLAUDE.md`, `.codex/hooks.json`, `.config/kitty/kitty.conf`; `XDG_CONFIG_HOME` override honored.

### Non-agent declarations

`kitty` **is** now projected (it appears in the list above) — the #1383 gap is closed. `gastown/state` is allowlisted in `applyWritesNoUserFile` ("Apply starts/stops the watcher, writes nothing"); `launcher` declares no `Apply`-bearing modify permission. `TestNonAgentPermissionDeclarationsDeclareNoHooks` is **gone**, deliberately: it held the gap *visible* by asserting the non-agent entries declared no hooks precisely because they weren't projected. It is superseded by catalog-wide tripwires that assert the stronger property — `TestEveryModifyPermissionDeclaresTheFileItWrites`, `TestConsentCatalogMatchesThePermissionWizard`, `TestPermissionWizardIsBuiltFromConsentCatalog`.

### Gates

`tools/preflight.sh` run chunked in the foreground, all PASS: `go` (7/7), `web` (2/2), `arch` (ARS composite 7.9495 → 7.9506, no regression), `tools`, `skills`, `security`.

Required checks confirmed present and green **on `a3855479`**: `go-test` pass, `build-test` pass, `ars-gate` pass.

CodeScene is red (advisory, not a merge gate). Findings are mostly in this PR's own new table-driven tripwire tests (`managedfiles_test.go` ×2, Bumpy Road / Complex Method). Two are attributable to the correctness narrowings above: `hookPermission` Complex Conditional (8.00 → 7.88, now a 4-clause filter) and `claudecode` `Agent()` Large Method (9.49 → 9.45). Not chased to green — flagging for reviewer judgment.

No test ever pointed at the real `~/.claude`, `~/.config/kitty` or `~/.codex`; all runs used a temp `HOME`/`CODEX_HOME`/`XDG_CONFIG_HOME`. Verified afterwards: `CLAUDE.md`, `kitty.conf`, `codex/config.toml`, `settings.json` and `codex/hooks.json` all byte-identical to their session-start baseline.

Issue #1372, opened 2026-08-09
feat(hooks): re-verify installed hook entries — an agent's own settings UI silently deletes them


ingo-eichhorst:
Parent: #1355 — Phase B. Discovered by the Phase C audit; see the report on #1355.

## The defect, live-verified on gemini-cli

gemini-cli's settings writer implements **sync-by-omission**. `applyKeyDiff` deletes any key present on disk but absent from the in-memory snapshot the CLI took at startup — recursively, at every nesting level.

Live-verified: with gemini running, an externally added top-level key **and** an externally added `hooks.PreCompress` entry were both silently deleted by a single `/theme` change.

The write paths are numerous — `/theme`, `/auth`, `/settings`, `/hooks`, `/editor`, IDE setup — and one of them, `saveModelChange`, is **not user-initiated** and can fire on its own.

## Why this is a Phase B item and not a gemini detail

irrlicht's whole hook install model is *install once on consent, then assume it is there*. That assumption is false for any agent whose own UI rewrites its config, and gemini is simply the one where it was caught. claudecode and codex have the same exposure to a user hand-editing or a future in-app settings surface; we have never checked.

The user-visible symptom is identical to every other silent hook failure: "irrlicht stopped detecting waiting", with a permission that still reads `granted` and no error anywhere.

## Scope

1. Periodically **re-verify** that the installed hook entries are still present and still point at the resolved daemon port, for every adapter with a granted hooks permission. `hookjson.IsCanonical` already expresses "is this entry ours and current" — this is about calling it on a schedule, not about new matching logic.
2. Re-install when they are missing. Silently and idempotently — a user who never noticed the clobber should never notice the repair either.
3. **Rate-limit and log.** A repair loop that fights a live CLI writing the same file every few seconds must back off and say so, not spin. Repeated repairs are themselves a signal worth surfacing.
4. Re-verification is a **write**, so it stays behind the same `KindModify` consent as the original install. A revoked permission must stop the loop. Do not let a background repair become a way to write to a config the user withdrew consent for — that would break #570 more thoroughly than the bug it fixes.
5. Interacts with #1368 (hook-liveness watchdog): "entries present but nothing arriving" and "entries went missing" are different diagnoses and should be distinguishable in `--diagnose`, not collapsed into one warning.

**Complexity:** 6

Issue #1364, opened 2026-08-09
fix(hooks): an unrecognized hook event is an Info log and an HTTP 200 — silently uncounted


ingo-eichhorst:
Parent: #1355 — Phase B, work item 4. Depends on #1361 (same files — land that first).

## The defect

An event name the hook receiver does not recognize is logged at **Info** and answered **HTTP 200**. There is no counter, nothing in `--diagnose`, and nothing anywhere that would tell you an upstream rename had happened.

Verify against `claudecode/hooks.go` and `codex/hooks.go` and quote the lines.

The failure this hides: upstream renames the *assert* event. The daemon receives nothing meaningful. The permission still reads `granted`, the config still contains entries, and state detection just quietly gets worse. Hook schemas are young — mistral-vibe had breaking changes at 2.15 — so this is a when, not an if.

## Scope

1. Count unrecognized events per `(adapter, event name)`.
2. Surface in `--diagnose` and wherever the daemon already reports health.
3. Log at a level that is actually noticed, **once per distinct name** — an upstream rename would otherwise flood the log at every tool call.
4. **Keep answering 200.** A hook that receives a non-2xx can block or slow the user's turn; failing open is deliberate (see codex's `|| true`) and must not change here.

**Complexity:** 4

ingo-eichhorst:
## Confirmed by evidence gathered in #1361 / PR #1380

This issue's instruction — *"Keep answering 200. A hook that receives a non-2xx can block or slow the user's turn; failing open is deliberate and must not change"* — is the correct one, and it now has evidence behind it rather than just caution.

I had written the opposite in #1361 item 4 ("reject with a 400"). That has been corrected there; the two tickets no longer disagree.

What was established while resolving it:

- Claude Code documents that for `type: http` hooks, **HTTP status codes alone cannot block actions** — blocking requires 2xx plus a JSON decision body. The `claude 2.1.226` binary confirms: `validateStatus: () => true`, and a non-2xx becomes `ok:false` rather than an exception.
- **Unknown, and stated as such:** whether `ok:false` surfaces to the user, is retried, or is only logged.
- The fail-closed hazard is real but lives elsewhere: gemini-cli's `BeforeTool` and copilot's `preToolUse` **deny the tool call** on a hook error. Those are command-hook exit codes, not HTTP status, but they are why "what does the CLI do when our endpoint refuses" must never be answered by assumption.

[...]

Issue #1362, opened 2026-08-09
fix(permissions): a failed Apply is recorded as "granted" and the error is swallowed


ingo-eichhorst:
Parent: #1355 — Phase B, work item 5. **Blocks the hook rollout.**

## The defect

In `core/application/services/permission_service.go`:

- the grant is recorded as `granted` and `store.Save` is called **unconditionally** — before/regardless of whether `runEffects` succeeded;
- `runClosureEffect` **swallows** the returned error;
- `PermissionView` has no error field.

**Verify each of those three claims against the code and quote the lines in the PR body.**

Observable result today: the consent wizard says "granted", no hooks are installed, and the only symptom the user ever sees is "irrlicht doesn't detect waiting". Note that `hookjson.ReadSettings` *deliberately* errors on malformed JSON — correct behaviour, and it routes straight into this swallow.

## Scope

1. Persist the `Apply` failure alongside the grant. The state is not `granted`; it is *granted but not applied, because `<reason>`*. Those are different facts.
2. Surface it on `PermissionView` and through to everything that renders permissions — **the web dashboard and the macOS UI**. Check both; fixing one and leaving the other is the failure mode here.
3. Make it retryable: a grant whose `Apply` failed must be re-appliable without the user revoking and re-granting.
4. **Do not roll the grant back on failure.** The user did consent; the effect failed. Conflating them means a transient filesystem error silently revokes consent.
5. The #570 guarantee is unchanged: nothing is exercised while pending or denied.

## Red-first

Make an `Apply` closure fail, assert that today's code reports a plain `granted` with no error anywhere. Seen red on `main`, failure pasted in the PR.

**Complexity:** 6 — services + ports + two UIs.

Issue #1361, opened 2026-08-09
fix(hooks): claudecode's hook receiver forwards a caller-supplied transcript_path unconfined


ingo-eichhorst:
Parent: #1355 — Phase B, work item 9. **Blocks the hook rollout.**

## The defect

`core/adapters/inbound/agents/codex/hooks.go` resolves symlinks and confines the incoming `transcript_path` to its own sessions dir. `core/adapters/inbound/agents/claudecode/hooks.go` does **not** — it derives the session id from the path basename but forwards the raw, caller-supplied path downstream to the detector, which reads it.

The hook endpoint is local and unauthenticated: any local process can POST to it.

**Verify both statements against the code before writing anything, and quote the lines in the PR body.** If the claudecode receiver turns out to confine after all, say so and close this — a wrong dismissal here is the expensive kind.

## Why now

This is a 2-receiver problem today and a 10-receiver problem after #1355's rollout. Confinement has to become a mandatory property of the receiver path, not a thing each adapter remembers.

## Scope

1. One **shared** confinement helper on the receiver path: given the adapter's declared transcript root(s) and a caller-supplied path, return either a resolved confined path or a rejection.
2. Roots come from the adapter's existing `agent.Source` declaration (`FilesUnderRoot.AllRootsFor(goos)`) — **not** a second hardcoded list that can drift from the first.
3. Resolve symlinks **before** the containment check. That ordering is the actual bypass; getting it backwards passes a naive test and fixes nothing.
4. Reject with a 400 and a logged, counted reason. Never silently drop.
5. A `contracttesting` assertion so a new hook-receiving adapter cannot skip confinement. Same shape as `AssertHookEndpointFollowsBindAddr`: it passes by construction against a correct adapter, so its whole value is that it *can* fail — land it with the deliberate mutation seen red, per AGENTS.md, and record that in the PR.

**Complexity:** 5

ingo-eichhorst:
## Correction to scope item 4 — resolved in PR #1380

Item 4 of this issue says *"Reject with a 400 and a logged, counted reason."* **That was wrong**, and it contradicted #1364, which I also wrote and which says a non-2xx must not happen because it can block or slow the user's turn.

The implementing agent established the evidence I did not have:

- Claude Code's hooks guide, on `type: http`: **"HTTP status codes alone can't block actions."** Blocking or denying requires 2xx *plus* a JSON decision body.
[...]

PR #1413, opened 2026-08-09
feat(hooks): per-adapter hook-liveness watchdog (#1368)

### LINKED ISSUE #1368: feat(hooks): per-adapter hook-liveness watchdog — detect "granted, installed, but nothing arriving"
Parent: #1355 — Phase B, work item 2. Depends on #1364 (needs the counting).

## The failure this catches

Permission granted **+** hook entries present in the agent's config **+** zero `hook_received` in N turns. Today that state is indistinguishable from "the user just hasn't hit a permission prompt lately", and its only symptom is silently degraded detection. After #1355's rollout there are up to 10 channels that can be in it independently.

## Scope

1. Per adapter, track turns observed against hook events received.
2. On crossing the threshold: log loudly, **drop that adapter to tier-5**, and surface it in `--diagnose`. Continuing to trust a channel that is demonstrably dead is the current behaviour and the thing being fixed.
3. **Recover automatically** when events resume. This is a health signal, not a latch — a user who simply had a quiet hour must not have to restart the daemon.
4. "N turns" tunable, with a defaulted value and a stated rationale for it.
5. Interaction with #1360's wall-clock ceilings: a dead channel plus a pinned hold is the compound failure. Make sure the watchdog dropping to tier-5 also releases holds that channel asserted, or say explicitly why it must not.

**Complexity:** 6

ingo-eichhorst:
Closes #1368

## The failure

`permission granted` **+** `install effect succeeded` **+** zero hook requests arriving across N completed turns. Today that is indistinguishable from "the user just hasn't hit a permission prompt lately", and its only symptom is silently degraded detection.

Three diagnoses are deliberately kept apart, because they have different first moves:

| | what it means | where it shows |
|---|---|---|
| **this PR (#1368)** | entries were written, nothing is arriving | `hooks.json` `channels[].silent` + a loud log line |
| #1372 (not built) | entries were written and later removed | reports *the same way* as above — the note says so explicitly rather than claiming otherwise |
| #1362 (landed) | the install effect failed, entries never written | `effect_error` on the permission; the watchdog **stays disarmed** so it cannot re-report this in weaker words |
| #1364 (landed) | events arrive under names we don't know | still counts as a *receipt* — the channel is alive; the unknown-event counters say the names are wrong |

## Design

- **`hookjson.ObserveHookReceipt`** (`receipt.go`) counts, per adapter, that the channel delivered. Called immediately after the consent gate and *before* any rejection: a malformed body, an unrecognized name (#1364) or a confinement refusal (#1361) all still prove delivery. A consent-*denied* request is not counted — noting that a POST arrived is itself an observation, and every observation is gated.
- **`services.HookLivenessWatchdog`** (`hook_liveness.go`) tracks turns against receipts per adapter. Turns, not wall time: a minutes-based window cannot tell a dead channel from a lunch break, which is the whole ambiguity. The turn signal is transcript-derived `IsAgentDone`, read *before* the hook overlay, so a dead channel cannot suppress the measurement that convicts it.
- **Recovery** is checked on *every* pass; demotion only on a turn boundary. Convicting needs evidence that turns went by; acquitting needs only that a hook arrived. Waiting for the next boundary to believe a recovery would be a latch wearing a different hat.
- **`PermissionService.HookChannelReady`** is the arming predicate: a hook-install permission that is granted *and* whose Apply did not fail. Built from what the daemon did, not from re-reading the config — that would cost a JSON parse per turn and answers #1372's question instead.
- **Diagnostics** extend #1364's *existing* `HookHealthSnapshot` seam rather than adding a second one, and inherit its nil-means-CLI honesty rule.

## The hard question: must the demotion release the channel's holds?

**Yes.** A hold is not a record that a hook fired — it is a live assertion of authority: while a `TierHook` hold stands, nothing below it may correct the session. So the holds *are* the tier the demotion claims to be lowering, and a demotion that left them standing would demote nothing: the session stays pinned at `waiting` by the channel just declared dead, and the only observable change is a row in a bundle.

It acts on evidence already collected, not a guess: conviction requires N turns with zero receipts, so any hold that channel placed necessarily predates all N of them, and the release POST that would retire it is exactly what has been shown not to arrive.

It cannot invent a state — `ReleaseTier` removes a *suppression*; the classifier then decides from the transcript, which is what `TierTranscript` means and what the seven hookless adapters do all the time.

#1360's 12-hour ceilings already drop these holds, so this changes **when**, not **whether** — and that is the value: five turns is minutes, twelve hours is a working day at the wrong state. The ceiling remains the backstop for the case the watchdog cannot see (an adapter producing no further turns, where there is no evidence to convict on). Two mechanisms, different evidence, deliberately not merged — and a distinct `hook_hold_released` kind so a trace never conflates them.

Pinned by `TestSessionDetector_SilentHookChannelReleasesItsPinnedHolds`.

## Tunable

`IRRLICHT_HOOK_SILENT_TURNS`, default **5**, `0` disables. Rationale in `config.go`: both hook adapters install a Stop hook, so a healthy channel owes ≥1 receipt per turn; 1 would be far too twitchy given hook/transcript interleaving, a mid-turn install, and a session already in flight when consent was granted (each costs 1–2 turns); and the upper bound is set by the point of the demotion — a threshold in the tens would be slower than the 12h backstop it exists to beat.

## Red-first evidence

Every behaviour was mutated and seen fail. Verbatim failures are in the thread below.

## Not done deliberately

- No macOS/web UI surface. This is a daemon-internal health signal whose consumer is the diagnostics bundle and events.log; the issue scopes it to `--diagnose`.
- No detection of *why* a channel is silent. That is #1372, and the note says so instead of guessing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -206,7 +206,20 @@ Before marking a ticket done, run the full suite — every layer must pass:
   the diagnostics bundle's `hooks.json`; note that `--diagnose` runs in a process
   that never served a hook, so that form omits the counts and says so rather than
   publishing zeros (the live counts come from `GET /debug/bundle`).
-  All six contract families pass by construction against a correct adapter, so
+- Hook receipts: `contracttesting.AssertHookReceiptObserved`
+  (`core/internal/contracttesting/hook_receipt.go`) is the #1368 counterpart —
+  the event that never *arrives*, rather than the one that arrives unrecognized.
+  It is the only receiving-side contract whose failure is a false **accusation**
+  instead of a silence: the liveness watchdog demotes a channel that produces no
+  receipts across N turns and releases the holds it placed, so a receiver that
+  forgets `hookjson.ObserveHookReceipt` is reported dead and stops being trusted
+  while working perfectly — and the watchdog cannot tell that apart from a real
+  outage. Four obligations: a recognized event counts exactly one receipt; an
+  *unrecognized* one counts too (alive-but-misunderstood is #1364's diagnosis);
+  a path-confinement rejection counts too (alive-but-misrouted is #1361's); and a
+  consent-denied request counts nothing, because noting that a POST arrived is
+  itself an observation.
+  All seven contract families pass by construction against a correct adapter, so
   their whole value is that they *can* fail: a new or reworked contract
   assertion lands with the deliberate mutation that was seen red for each
   obligation recorded in its PR — the same bar the red-first rule above sets
--- a/core/adapters/inbound/agents/claudecode/hookreceipt_test.go
+++ b/core/adapters/inbound/agents/claudecode/hookreceipt_test.go
@@ -1,0 +1,36 @@
+package claudecode
+
+import (
+	"net/http"
+	"testing"
+
+	"irrlicht/core/internal/contracttesting"
+	"irrlicht/core/ports/outbound"
+)
+
+// TestHookReceiptObserved wires this adapter's hook receiver into the shared
+// issue #1368 contract: a consent-passed request is counted as a receipt for
+// this adapter whatever the receiver then does with it, and a consent-denied
+// one is not.
+//
+// The counter is what tells the liveness watchdog this channel is alive, so a
+// receiver that stops calling it is not merely unmonitored — it is actively
+// reported as dead and demoted to the transcript tier while working.
+func TestHookReceiptObserved(t *testing.T) {
+	contracttesting.AssertHookReceiptObserved(t, contracttesting.HookReceiptReceiver{
+		Adapter:         AdapterName,
+		Root:            claudeProjectsRoot,
+		WriteTranscript: writeContractTranscript,
+		New: func(t *testing.T, log outbound.Logger) http.Handler {
+			t.Helper()
+			return NewHookHandlerWithConfiner(&mockTarget{}, nil, fakeGate(true), log, TranscriptConfiner())
+		},
+		NewDenied: func(t *testing.T, log outbound.Logger) http.Handler {
+			t.Helper()
+			return NewHookHandlerWithConfiner(&mockTarget{}, nil, fakeGate(false), log, TranscriptConfiner())
+		},
+		PayloadFor:   contractPayload,
+		KnownEvent:   HookPermissionRequest,
+		EndpointPath: HookEndpointPath,
+	})
+}
--- a/core/adapters/inbound/agents/claudecode/hooks.go
+++ b/core/adapters/inbound/agents/claudecode/hooks.go
@@ -274,6 +274,13 @@ func serveHookRequest(target HookTarget, markers MarkerTarget, gate ConsentGrant
 		return
 	}
 
+	// The channel delivered (issue #1368). Counted here — after the consent
+	// gate, before anything can reject the payload — because the question this
+	// answers is "are hooks arriving at all", and a body we then refuse still
+	// proves they are. See receipt.go for why each rejection below still
+	// counts and why a consent-denied request must not.
+	hookjson.ObserveHookReceipt(AdapterName)
+
 	var payload hookPayload
 	if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
 		http.Error(w, "bad request: invalid JSON", http.StatusBadRequest)
--- a/core/adapters/inbound/agents/codex/hookreceipt_test.go
+++ b/core/adapters/inbound/agents/codex/hookreceipt_test.go
@@ -1,0 +1,39 @@
+package codex
+
+import (
+	"net/http"
+	"testing"
+
+	"irrlicht/core/internal/contracttesting"
+	"irrlicht/core/ports/outbound"
+)
+
+// TestHookReceiptObserved wires this adapter's hook receiver into the shared
+// issue #1368 contract: a consent-passed request is counted as a receipt for
+// this adapter whatever the receiver then does with it, and a consent-denied
+// one is not.
+//
+// Codex's receiver has two gates, and the receipt is counted against the first.
+// That is deliberate: "transcripts" authorizes READING the transcript, while
+// noting that a POST arrived needs only the "hooks" consent — and a
+// hooks-granted / transcripts-denied install has a channel that plainly works
+// and must not be convicted of silence.
+func TestHookReceiptObserved(t *testing.T) {
+	contracttesting.AssertHookReceiptObserved(t, contracttesting.HookReceiptReceiver{
+		Adapter:         AdapterName,
+		Root:            codexSessionsRoot,
+		WriteTranscript: writeContractTranscript,
+		New: func(t *testing.T, log outbound.Logger) http.Handler {
+			t.Helper()
+			return NewHookHandlerWithConfiner(&mockTarget{},
+				keyedGate{PermissionKeyHooks: true, PermissionKeyTranscripts: true}, log, TranscriptConfiner())
+		},
+		NewDenied: func(t *testing.T, log outbound.Logger) http.Handler {
+			t.Helper()
+			return NewHookHandlerWithConfiner(&mockTarget{}, keyedGate{}, log, TranscriptConfiner())
+		},
+		PayloadFor:   contractPayload,
+		KnownEvent:   HookPermissionRequest,
+		EndpointPath: HookEndpointPath,
+	})
+}
--- a/core/adapters/inbound/agents/codex/hooks.go
+++ b/core/adapters/inbound/agents/codex/hooks.go
@@ -136,6 +136,12 @@ func serveHookRequest(target HookTarget, gate ConsentGranter, log outbound.Logge
 		return
 	}
 
+	// The channel delivered (issue #1368). Counted against the "hooks" consent
+	// only — the transcripts gate below authorizes READING the transcript, not
+	// observing that a POST arrived, and a hooks-granted / transcripts-denied
+	// install has a working channel that this counter must not call dead.
+	hookjson.ObserveHookReceipt(AdapterName)
+
 	var payload codexHookPayload
 	if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
 		http.Error(w, "bad request: invalid JSON", http.StatusBadRequest)
--- a/core/adapters/inbound/agents/hookjson/receipt.go
+++ b/core/adapters/inbound/agents/hookjson/receipt.go
@@ -1,0 +1,130 @@
+// receipt.go counts, per adapter, that the hook channel delivered anything at
+// all (issue #1368).
+//
+// It is the missing half of #1364. That issue made an event we do not
+// UNDERSTAND visible; this one makes an event that never ARRIVES visible. The
+// two failures look identical from every surface a user can see — permission
+// reads `granted`, the config still holds our entries, no error is logged
+// anywhere — and they have opposite fixes, so a single "hooks are unhealthy"
+// signal would be worse than none.
+//
+// What this counter is NOT: a count of hook events successfully handled. It is
+// deliberately incremented at the earliest point the receiver is allowed to
+// observe anything, which is immediately after the consent gate. Three
+// consequences, each chosen:
+//
+//   - A malformed body still counts. A decode failure is a different diagnosis
+//     with a different fix, and it is already answered 400. For the question
+//     this counter answers — "is the channel delivering?" — a 400 is a yes.
+//   - A path-confinement rejection still counts (#1361). Same argument, and
+//     confine.go's own counters say what is actually wrong. A channel that is
+//     alive but 100% rejected must not be reported as silent, or the watchdog
+//     sends its reader to the wrong file.
+//   - A consent-denied request does NOT count. Everything an adapter observes
+//     must be gated behind a declared permission (AGENTS.md), and "a POST
+//     arrived" is an observation. It is also the honest answer: while consent
+//     is withheld the daemon has no business claiming to know whether the
+//     channel works.
+//
+// Unlike unknown.go's table this map needs no bound. Its key is the adapter's
+// own compile-time AdapterName constant, never caller-supplied input, so the key
+// space is closed at the size of the agent registry — the same reason
+// unknown.go's per-adapter DROP counter is unbounded while its per-name table is
+// not.
+package hookjson
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+var (
+	// receiptMu guards slot allocation only. Once an adapter owns a slot every
+	// later receipt goes through its atomic under the read lock — the same
+	// argument unknown.go makes, and it applies harder here: this increments on
+	// EVERY hook, not only on the unrecognized ones, so it sits squarely on the
+	// receiver hot path.
+	receiptMu sync.RWMutex
+	receipts  = make(map[string]*atomic.Uint64)
+)
+
+// ObserveHookReceipt records that adapter's hook receiver was handed a
+// consent-passed request.
+//
+// It takes no logger and writes no response on purpose. There is nothing to
+// report here: a hook arriving is the healthy case, and the interesting event
+// is its ABSENCE, which by definition no call site can observe. The watchdog in
+// the application layer is what notices the absence, by reading HookReceipts
+// against turns it counted independently.
+func ObserveHookReceipt(adapter string) {
+	receiptMu.RLock()
+	counter := receipts[adapter]
+	receiptMu.RUnlock()
+	if counter != nil {
+		counter.Add(1)
+		return
+	}
+
+	receiptMu.Lock()
+	defer receiptMu.Unlock()
+	if counter := receipts[adapter]; counter != nil {
+		// Lost the race to insert; the winner's slot is the live one.
+		counter.Add(1)
+		return
+	}
+	fresh := &atomic.Uint64{}
+	fresh.Add(1)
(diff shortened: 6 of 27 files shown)


ingo-eichhorst:
## Red-first evidence (AGENTS.md / `ir:exec` step 11a)

Each behaviour was neutered in the worktree with the implementation otherwise intact, so every failure below is a real assertion failure, not a compile error. Verbatim output.

### 1. Demotion without releasing holds — i.e. answering "no" to the hard question

Mutation: `checkHookLiveness` logs and records the verdict but never calls `ReleaseTier`.

```
--- FAIL: TestSessionDetector_SilentHookChannelReleasesItsPinnedHolds (0.00s)
    session_detector_hook_liveness_test.go:110: the hold placed by a channel now proven dead is still pinning the session — the demotion demoted nothing
```

### 2. Threshold never crosses

Mutation: `if st.silent || st.silentTurns < w.cfg.SilentTurns {` → `if true {`.

```
--- FAIL: TestHookLiveness_DemotesAfterThresholdTurnsWithNoReceipts (0.00s)
    hook_liveness_test.go:86: transition on the threshold turn = 0, want HookLivenessSilent
--- FAIL: TestHookLiveness_SnapshotPublishesEveryConfiguredChannel (0.00s)
    hook_liveness_test.go:346: dead channel row = {Adapter:claude-code Armed:true Receipts:0 TurnsSinceReceipt:2 Silent:false}, want armed+silent with 0 receipts and 2 silent turns
--- FAIL: TestSessionDetector_SilentHookChannelReleasesItsPinnedHolds (0.00s)
    session_detector_hook_liveness_test.go:110: the hold placed by a channel now proven dead is still pinning the session — the demotion demoted nothing
--- FAIL: TestSessionDetector_SilentVerdictIsLoggedLoudlyAndRecorded (0.00s)
    session_detector_hook_liveness_test.go:157: expected exactly one error-level line for the verdict, got 0: []
```

### 3. Silence is a latch (scope item 3)

Mutation: the recovery branch in `OnActivity` never fires.

```
--- FAIL: TestHookLiveness_RecoversWhenEventsResume (0.00s)
    hook_liveness_test.go:137: transition = 0, want HookLivenessRecovered
--- FAIL: TestHookLiveness_RecoveryNeedsNoTurnBoundary (0.00s)
    hook_liveness_test.go:164: transition on a non-boundary pass = 0, want HookLivenessRecovered
--- FAIL: TestSessionDetector_RecoveredChannelStopsReleasingHolds (0.00s)
    session_detector_hook_liveness_test.go:223: a hold placed after recovery was released — the demotion behaved as a latch
```

### 4. `HookChannelReady` ignores a failed install (the #1362 boundary)

Mutation: drop the `!s.effectFailedLocked(...)` term.

```
--- FAIL: TestHookChannelReady_FalseWhenTheInstallFailed (0.00s)
    permission_hook_channel_test.go:92: a granted permission whose Apply failed reads as an installed channel — the watchdog would report 'installed but silent' for entries that were never written
```

### 5. Liveness rows never reach the bundle

Mutation: `Channels: channels` → `Channels: nil` in `hooksView`'s daemon shape.

```
--- FAIL: TestHooksBundleReportsChannelLiveness (0.00s)
    diagnostics_service_test.go:394: channels has 0 entries, want 3: <nil>
```

### 6–9. The new contract family, one mutation per obligation

AGENTS.md requires a new contract assertion to land with the deliberate mutation seen red **for each obligation**.

**Obligations 1–3** — mutation: both receivers stop calling `hookjson.ObserveHookReceipt`.

```
--- FAIL: TestHookReceiptObserved (0.00s)
    --- FAIL: TestHookReceiptObserved/recognized_event_counts_one (0.00s)
        hook_receipt.go:103: event "PermissionRequest" added 0 receipts for adapter "claude-code", want exactly 1 — a receiver that does not count its own traffic is reported as a DEAD channel by the liveness watchdog and demoted while working perfectly
    --- FAIL: TestHookReceiptObserved/unrecognized_event_counts_too (0.00s)
        hook_receipt.go:104: event "IrrReceiptUnknownEvent" added 0 receipts for adapter "claude-code", want exactly 1 — a receiver that does not count its own traffic is reported as a DEAD channel by the liveness watchdog and demoted while working perfectly
    --- FAIL: TestHookReceiptObserved/confinement_rejected_request_counts_too (0.00s)
        hook_receipt.go:105: a path-refused request added 0 receipts for adapter "claude-code", want 1 — a misconfigured transcript root (#1361) must not be reported as a channel that stopped delivering (#1368)
FAIL	irrlicht/core/adapters/inbound/agents/claudecode	3.414s
--- FAIL: TestHookReceiptObserved (0.01s)
    --- FAIL: TestHookReceiptObserved/recognized_event_counts_one (0.00s)
        hook_receipt.go:103: event "PermissionRequest" added 0 receipts for adapter "codex", want exactly 1 — a receiver that does not count its own traffic is reported as a DEAD channel by the liveness watchdog and demoted while working perfectly
    --- FAIL: TestHookReceiptObserved/unrecognized_event_counts_too (0.00s)
        hook_receipt.go:104: event "IrrReceiptUnknownEvent" added 0 receipts for adapter "codex", want exactly 1 — a receiver that does not count its own traffic is reported as a DEAD channel by the liveness watchdog and demoted while working perfectly
    --- FAIL: TestHookReceiptObserved/confinement_rejected_request_counts_too (0.00s)
        hook_receipt.go:105: a path-refused request added 0 receipts for adapter "codex", want 1 — a misconfigured transcript root (#1361) must not be reported as a channel that stopped delivering (#1368)
FAIL	irrlicht/core/adapters/inbound/agents/codex	3.754s
```

**Obligation 4** — mutation: count moved *above* the consent gate.

```
--- FAIL: TestHookReceiptObserved (0.00s)
    --- FAIL: TestHookReceiptObserved/consent_denied_request_is_not_counted (0.00s)
        hook_receipt.go:106: a consent-denied request added 1 receipts for adapter "claude-code", want 0 — noting that a POST arrived is an observation, and every observation an adapter makes is gated
```

## Locks (pass by construction — NOT red-first evidence)

Stated explicitly so their green is not read as proof:

- `TestSessionDetector_LiveHookChannelKeepsItsHolds` — nothing released holds on `main`, so this passes there by construction. Its job is to fail if the release ever escapes the silent-channel condition and starts dropping holds on a **healthy** channel, which would break every hook-driven state transition in the product.
- `TestHooksViewDoesNotAliasTheInjectedChannels` — `hooksView` copies but does not sort the channel slice, so there is no in-place mutation to catch today. It locks against a future reorder-in-place (#965/#967/#975).
- `TestHooksBundleOmitsCountsWhenNotCollectedInDaemon` (pre-existing, #1364) — still passes; the new `liveness_note` is additive and does not disturb the CLI shape's existing assertions.

## What `hooks.json` now prints, verbatim

**Daemon, one channel silent:**

```json
{
  "channels": [
    { "adapter": "claude-code", "armed": true, "receipts": 0, "silent": true, "turns_since_receipt": 7 },
    { "adapter": "codex", "armed": true, "receipts": 118, "silent": false, "turns_since_receipt": 0 }
  ],
  "collected_from": "daemon",
  "hook_requests_received": 0,
  "silent_channel_note": "Hook channel(s) claude-code delivered nothing across the watchdog's turn threshold while consent was granted and the install reported success, so they were demoted to the transcript tier and any hook-tier holds they had placed were released (#1368). This says entries were WRITTEN and nothing is arriving through them. It does NOT mean the entries are still present — if something removed them since, this is how that looks too (#1372). And it is a different fault from an install that FAILED, which never wrote entries at all and shows up as effect_error on the permission rather than here (#1362). Check the agent's own hook config next, then the daemon's bind port.",
  "unknown_events_total": 0
}
```

**`irrlichd --diagnose` (a process that served no hooks):** counts omitted, not zeroed, matching #1364's honesty rule — and the liveness note states the *stronger* reason.

```json
{
  "collected_from": "cli",
  "liveness_note": "The per-adapter hook-liveness watchdog (#1368) is omitted here for the same reason and one more: this process served no hooks AND observed no turns, so both sides of the ratio it reports are structurally absent — a channel would read as silent with zero turns behind it, which is an accusation, not a measurement. Its verdicts are logged, so events.log in this bundle still carries any `hook channel for <adapter> looks dead` line. For the live view, fetch GET /debug/bundle from the running daemon.",
  "note": "Collected by `irrlichd --diagnose`, which runs in a process that never served a hook, so the receivers' unrecognized-event counters are not readable here and are omitted rather than reported as zero. ..."
}
```

## Local CI parity

`tools/preflight.sh` run chunked in the foreground, all gates PASS on the rebased head:

| gate | result |
|---|---|
| `--only go` (gofmt, core `-race`, factory, replaydata validate, rig smoke, **replay fixtures**, starhistory) | PASS — no golden drift |
| `--only web` (dashboard + onboarding viewer) | PASS (85 tests) |
| `--only arch` (ARS) | PASS — composite 7.9503 → 7.9447, no category regressions |
| `--only tools` | PASS |
| `--only skills` | PASS — 23 files, 0 errors, 0 warnings |
| `--only security` (govulncheck + gosec + npm audit ×2) | PASS |

## Rebase note

`origin/main` advanced from `b0fca88c` to `285fed30` mid-run, landing #1404 (grace-timer hysteresis), which touches `session_detector_activity.go` and `session_detector_lifecycle.go` — both rewritten here. Two conflicts, both resolved semantically rather than by side:

- **`session_detector_lifecycle.go`**: #1404 replaced the inline `signals.DropSession` + idle-retry cleanup with a shared `forgetSessionScopedState` seam. The watchdog's `Forget` moved *into* that seam rather than sitting beside the code #1404 deleted — which also gets it the PIDManager reap path for free.
- **`session_detector_activity.go`**: #1404 hoisted `Overlay`'s result into an `expiries` local and switched to `d.nowFn()`. Kept both, with `checkHookLiveness` before the overlay for the reason in its doc comment.

The contract-family count in `AGENTS.md` ("All **six**" → "All **seven**") was verified against the code rather than assumed:

```
$ grep -h "^func Assert" core/internal/contracttesting/*.go | wc -l
7
```

ingo-eichhorst:
## Round 2 — review findings applied, then a `/simplify` pass

### Review (one subagent, `high` effort, base `origin/main...HEAD`)

Five findings, all fixed. Each fix was mutated and seen red; verbatim failures below.

**1. A session's first observed pass counted as a completed turn — CONFIRMED false positive, the worst of the five.**
`lastDone[sid]` returns `false` for an unseen session, so a session already at a turn boundary the first time it is observed registered a rising edge for a turn that finished before the watchdog existed. `refreshStaleSessions` re-reads *every* persisted working session at startup, so **a daemon restart alone could convict a healthy channel** — with no agent having run and no hook having had a chance to arrive. Fixed by seeding on first observation (`wasDone, seen := w.lastDone[...]`), which is the same argument the receipts baseline already made, applied to the other side of the ratio.

```
--- FAIL: TestHookLiveness_FirstObservationOfASessionIsNotATurn (0.00s)
    hook_liveness_test.go:416: session restored-3: first observation produced 1 — a turn that finished before the watchdog was watching is not evidence of silence
--- FAIL: TestHookLiveness_ForgottenSessionDoesNotDonateAPhantomTurn (0.00s)
    hook_liveness_test.go:444: a revived session's first pass produced 1
```

**2. The un-pinning was being handed straight back to #1366's grace timer.**
`checkHookLiveness` deletes the hold *before* `Overlay`, so a watchdog release can never appear in `expiries` — the pass was therefore not `authoritative` and `admitTransition` debounced it, while its own doc argues the identical ceiling-expiry case must bypass the dwell. `checkHookLiveness` now returns the release count and `processActivity` ORs it into `rebased`.

```
--- FAIL: TestSessionDetector_CheckHookLivenessReportsWhatItReleased (0.00s)
    session_detector_hook_liveness_test.go:286: reported 0 releases, want 2 — processActivity reads this to decide the pass is authoritative, so a zero here silently re-debounces the un-pinning
```

**3. A hook arriving mid-pass had its fresh hold swept.** `ReleaseTier` now skips any hold whose `heldAt` is after the pass clock — a hold asserted after the verdict was taken is not covered by evidence gathered before it.

```
--- FAIL: TestReleaseTier_KeepsAHoldPlacedAfterTheVerdictClock (0.00s)
    signal_hold_release_tier_test.go:160: released [{Kind:permission_prompt Tier:hook HeldFor:1h0m0s} {Kind:idle_prompt Tier:hook HeldFor:-1s}], want only the hold that predates the pass clock
```

**4. The CLI-honesty loop didn't cover the three new fields.** Added `hook_requests_received`, `channels`, `silent_channel_note` to `TestHooksBundleOmitsCountsWhenNotCollectedInDaemon`.

```
--- FAIL: TestHooksBundleOmitsCountsWhenNotCollectedInDaemon (0.00s)
    diagnostics_service_test.go:371: hook_requests_received is present without a daemon to read it from: 0
```

**5. `d.signals` dereferenced without the nil guard its sibling seam documents as required.** Added; the mutation panics.

```
--- FAIL: TestSessionDetector_HookLivenessToleratesAMissingHoldStore (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference
```

The reviewer additionally verified and found nothing on: lock ordering (`w.mu → PermissionService.mu` has no reverse edge; no `s.mu` holder does blocking work), replay goldens (`tools/onboarding-factory` never constructs a `SessionDetector`, so no replay path reaches the watchdog; all new event fields are `omitempty` and stay zero), the two rebase conflict resolutions, and the three-way diagnosis separation.

### `/simplify` (4 agents: reuse, simplification, efficiency, altitude)

Applied:

- **Whole-map copy per classify pass.** `receipts(adapter)` called `hookjson.HookReceipts()`, allocating and copying the entire table to index one key — measured ~19x slower and 256 B of garbage per pass, on a path that scales with session count, not hook traffic. Added `hookjson.HookReceiptsFor(adapter)`; the seam is now `func(adapter string) uint64`. Also fixes `Snapshot()` doing one full copy *per adapter*.
- **`hookChannelState.silent` deleted** — it was exactly `silentTurns >= threshold`, i.e. a second field that had to agree with the first forever. Out-of-step would have produced `"silent": true, "turns_since_receipt": 0`, which `hooksView`'s own note calls an accusation rather than a measurement. One `silentLocked` definition now.
- **`SetReceiptSource` removed** — it was called five lines after the constructor with no ordering constraint at all. Now a `HookLivenessConfig.Receipts` field. `SetChannelReady` stays: it is genuinely deferred until the permission service exists.
- **`HookLivenessChange.Silent`** — removes a second `w.mu` acquisition and a second adapter resolution per pass that could have disagreed with the first.
- **`resolve()`** — one definition of the empty-adapter fallback instead of two.
- **`Enabled()` → `enabled()`** (no external consumer).
- **Redundant `channels` slice copy dropped** — it copied an already-fresh slice from `Snapshot()`. Its accompanying "lock" test went with it: it passed with *or without* the copy, so it was pinning nothing.
- **~25 lines of rationale de-duplicated** — `checkHookLiveness`'s doc restated three paragraphs near-verbatim from `ReleaseTier`'s. The argument now lives once, next to the mechanism.
- **`playbackTimeline.js`** — all three new event kinds fell to the grey `default` (size 7, opacity 0.5, i.e. *bookkeeping* styling) while the neighbouring `hold_expired` is amber/full-size specifically so it "reads as a fault". `hook_channel_silent` is now red, `hook_hold_released` amber, `hook_channel_recovered` green.
- **Two doc gaps named rather than left implicit**: `ReleaseTier` now states the caller invariant it cannot check and the per-adapter/per-session equivalence it relies on; `config.go` names the premise a future #1355 adapter could violate (one that installs hooks but no per-turn event would owe no receipt on a quiet turn).
- Test harness de-duplication (`passAs`/`turnAs`, `pass` returning the release count).

Skipped, deliberately, with reasons:

- **Extract a shared rising-edge tracker** with `CacheBloatDetector` (raised by two agents). Real — and it would incidentally fix a *pre-existing* leak, since CacheBloat's three per-session maps are swept by nothing. But it edits landed code outside this diff. **Better as its own issue.**
- **Share a fixture struct between the two contracttesting receiver types** (6 identical fields). Test-only ergonomics; would edit a landed contract family and 4 adapter wirings right at hand-back.
- **Fold the receivers' 12-line prefix into a `hookjson.Admit`**, making the receipt unforgeable and deleting the new contract family. The most interesting suggestion of the four, but it restructures landed adapter code, and it would *not* subsume obligations 2–3 (confinement happens after the point `Admit` would sit) nor codex's second gate. **Worth an issue.**
- **`PermissionService.EffectInForce` generalization** and **moving `hookAdapterNames` into the `agents` package** — both reasonable, both reshape landed APIs for no behavioural gain here.
- **Replacing `receipt.go`'s double-checked locking with a plain mutex.** One agent flagged it as over-built; kept, because the file's two siblings in the same package use exactly this shape and consistency is worth more than 15 lines.

### Gates, re-run chunked in the foreground after every change

`go` (incl. replay fixtures — no golden drift), `web` (both suites), `arch` (ARS, no regression), `tools`, `skills`, `security` — **all PASS**.

### Advisory check

`CodeScene Code Health Review` is red and is **not** addressed. Per AGENTS.md it is advisory (not required by branch protection). Looking closer rather than waving it off: the plausible driver is file/function growth — `session_detector_activity.go` 1669 → 1791 lines and `classifyAndTransition` 98 → 116, both dominated by comments rather than code; the new `OnActivity` is 92 lines of which **54** are non-comment. The high-effort reviewer and the simplification agent both examined these and neither flagged them. I have not opened the CodeScene delta itself, so *which* rule fired is **unverified** — flagging it rather than claiming it is benign.

Issue #1416, opened 2026-08-09
fix(install): site/install.sh --uninstall never removes hook entries — uninstalling irrlicht leaves it in the user's agent configs


ingo-eichhorst:
Found while building the beacon (#1373 / PR #1412). This is the concrete, shipping form of the risk #1355 records as *"Uninstall-on-crash — `Remove` closures run only on explicit revoke. Nothing sweeps hooks left by a daemon that was deleted rather than revoked."*

## The defect, verified

```
$ git show origin/main:site/install.sh | grep -c "uninstall-hooks"
0
```

`site/install.sh`'s uninstall path (lines ~118-146) removes, in order: the LaunchAgent, the Linux systemd user unit, the app bundle, and the `~/.local/bin/irrlichd` binary. It never asks the daemon to remove what it wrote into the user's *agent* configs.

So after a clean uninstall, `~/.claude/settings.json` and `~/.codex/hooks.json` still carry irrlicht's hook entries, pointing at a binary that no longer exists. The user has uninstalled the product and it is still in their agent's configuration.

## Why it matters more now than it did last week

`--uninstall-hooks` was **registry-driven** by #1357 (PR #1375), so it now covers every hook-installing adapter automatically rather than the two that were hardcoded. The sweep exists and is correct; nothing calls it.

And the blast radius grows with #1355's rollout. Two adapters today, up to ten planned. Every one of them is a config file we write into and currently never clean up on uninstall.

## Severity, honestly bounded

Today the residue is close to harmless: claudecode's entries use native `type: http`, so a dead endpoint is an unanswered POST, and codex's entry ends in `|| true`. The beacon added in #1412 is also `|| true`-guarded, deliberately.

But that is a property of the three installers we happen to have, not of the design — and it is exactly the kind of "harmless today" that stops being true quietly. A future `command`-type installer without a fail-open suffix turns leftover entries into a broken agent for a user who has already uninstalled us, with no irrlicht left on the machine to explain why.

## Scope

1. Call `irrlichd --uninstall-hooks` from the uninstall path **before removing the binary** — the ordering matters, the binary is deleted at line ~141 and the sweep needs it.
2. Fail soft. An uninstall must complete even if the sweep errors: the user asked for removal, and a hook-cleanup failure must not leave a half-uninstalled product. Log it, continue.
3. Handle the binary already being gone (a partial earlier uninstall, a manually deleted binary). Skip with a message rather than failing.
[...]

Issue #1209, opened 2026-07-31
No CI gate reads .claude/skills/**/*.md — add a skill-file linter to preflight


ingo-eichhorst:
## Problem

`.claude/skills/**/*.md` has **zero mechanical coverage**. These files are
executable instructions — they decide how agents triage, plan, implement and
review — but no gate reads them.

Confirmed on PR #1204, a change touching only two skill files:
`tools/preflight.sh --changed` skipped all ten gates and `gofmt` was the single
check that ran. All 13 PR checks went green while proving nothing about the
files actually changed.

```
gofmt                                                      PASS
core module tests                                          SKIP
onboarding-factory tests                                   SKIP
replaydata validate                                        SKIP
recording-rig smoke test                                   SKIP
starhistory tests                                          SKIP
web: platforms/web                                         SKIP
web: onboarding-factory viewer                             SKIP
ARS architecture gate                                      SKIP
security scan                                              SKIP
```

## Why it matters

Skill files are merged on a human read alone. That is fine for judgement calls
about wording, and bad for the mechanical defects a linter catches for free —
the ones a reviewer's eye slides over precisely because they look like prose.

Two such defects were found by hand during recent work and would each have been
caught by a trivial check:

- `ir:triage` had a section reachable only as an orphaned paragraph under an
  unrelated `### Instrument menu` heading, and a numbered list that said "Two
  flavors:" and then listed three (the third orphaned outside the list).
- A section producing the Specification axis verdict sat 60 lines *below* the
  axis table that consumes it, so the table forward-referenced a section the
  reader had not met yet.

Both survived review. Neither is a judgement call.

## Desired

A linter over `.claude/skills/**/*.md`, wired into `tools/preflight.sh` (and
therefore the pre-push hook and CI) as its own gate, scoped to skill files so it
skips when none changed — the same shape every existing gate already has.

Checks worth having, roughly in order of value per line of implementation:

1. **Unresolved conflict markers** — `<<<<<<<` / `=======` / `>>>>>>>` at line
   start. Cheap, absolute, and the highest-consequence miss: a merged conflict
   marker in a skill file corrupts agent behaviour silently.
2. **Unfilled template tokens** — `{{TOKEN}}` surviving in a rendered artifact,
   and `REPEAT:` / `OPTIONAL:` scaffolding comments left in output. `ir:exec`'s
   `templates/plan.html` already documents both as must-strip and relies on a
   runtime banner to notice.
[...]

PR #1428, opened 2026-08-09
fix(ci): gate #!/bin/sh scripts against a real POSIX shell and a bashism linter

### LINKED ISSUE #1423: fix(ci): the shell-lib gate runs on macOS, where /bin/sh is bash — POSIX violations in site/install.sh cannot be caught anywhere
Found while adding installer coverage in #1416 (PR #1421).

## The gate is green by construction

`site/install.sh` is the script that installs and uninstalls irrlicht on users' machines. It is a POSIX `sh` script — the shebang and the whole idiom set say so.

`test.yml`'s shell-lib step runs on **`macos-latest`**, where `/bin/sh` is **bash 3.2 in POSIX mode**. Bash-in-POSIX-mode accepts a large amount of bash that a real POSIX shell rejects: `[[ ]]`, arrays, `local` in some forms, `${var,,}`, `+=`, process substitution. So a bashism in `site/install.sh` passes CI and fails on a user's machine running `dash` — which is `/bin/sh` on Debian and Ubuntu.

There is currently **no runner in this repo where such a violation can be caught.** The gate reports success without having tested the property it exists to test.

This is the same shape as several findings from the #1355 fleet run, and it is the most expensive one of the family, because the artifact is the *installer*: the failure lands on a new user's first command, before they have anything installed that could report it.

## Why it has not bitten yet

Presumably because the script has stayed disciplined and because macOS is the primary platform. That is not a mechanism — it is a run of luck that no gate is protecting.

Note also that #1416 just widened `site/install.sh`'s exposure: the uninstall path now invokes the daemon binary, redirects stdin, and applies a bounded wait. That is more shell surface, and all of it is currently unverified against a real POSIX shell.

## Scope — pick one, both are defensible

1. **Move the shell-lib step to `ubuntu-latest`**, where `/bin/sh` is `dash`. Smallest change and it tests the actual property. Check first whether anything else in that step depends on macOS; if it does, split the step rather than moving the job.
2. **Add `checkbashisms`** (from Debian's `devscripts`) as a lint over `site/install.sh` and any other `#!/bin/sh` script in the repo. Catches violations statically and names them, rather than failing at runtime on whichever line happens to execute.

Doing both is better than either: `checkbashisms` names the problem, `dash` proves the script runs.

Whichever is chosen, **run the existing `site/install.sh` through it before merging** and report what it finds. If it finds nothing, say so — that is the useful result, and it is the one that tells us the gate was merely redundant rather than hiding real breakage. If it finds violations, they are pre-existing user-facing bugs and should be split out rather than fixed silently in the same PR.

## Verification

The gate must be shown capable of failing: introduce a deliberate bashism into a scratch `#!/bin/sh` fixture, confirm the new gate goes red, remove it. Per AGENTS.md's rule for gates that pass by construction, that mutation is the evidence, and the current step would not fail it.

**Complexity:** 3

ingo-eichhorst:
Closes #1423

## The defect

`site/install.sh` reaches users as `curl -fsSL https://irrlicht.io/install.sh | sh`. On Debian and Ubuntu that `sh` is **dash**. A bashism there lands on a new user's first command, before anything is installed that could report it. #1416 (PR #1421) widened the surface — the uninstall path now invokes the daemon binary, redirects stdin and applies a bounded wait.

The issue's framing was that no runner can catch such a violation. That turned out to be **half right, and the other half is worse**. `tools/lib/install-uninstall_test.sh` check 9 (added by #1416) already prefers a real POSIX shell, and macOS ships one — `/bin/dash` is an Apple system binary (`com.apple.dash`) — so `dash -n` *was* running on `macos-latest`. The real gap is that `dash -n` is a **parser** check, and a parser accepts most bashisms as ordinary commands.

Measured, one bashism per file:

| bashism | `dash -n` | `checkbashisms` | `shellcheck --shell=sh` |
|---|---|---|---|
| `[[ "$x" == 1 ]]` | **pass** | CATCH | CATCH (SC3010/SC3014) |
| `arr=(a b c)` | CATCH | CATCH | CATCH (SC3030/SC3054) |
| `${v,,}` | **pass** | CATCH | CATCH (SC3059) |
| `s+=b` | **pass** | CATCH | CATCH (SC3024) |
| `<(cmd)` | CATCH | CATCH | CATCH (SC3001) |
| `echo -e` | **pass** | CATCH | CATCH (SC3037) |
| `function greet() {}` | CATCH | CATCH | CATCH (SC2112) |
| `source ./other.sh` | **pass** | CATCH | CATCH (SC3046) |
| | **3 of 8** | **8 of 8** | **8 of 8** |

So the gate missed five of the most common bashisms, including the one the issue named first.

## What the gate finds today: nothing

Run over both tracked `#!/bin/sh` scripts — `site/install.sh` and `tools/linux-replay-entrypoint.sh` — `checkbashisms`, `shellcheck --shell=sh` and `dash -n` all come back **clean**. Reported as the issue asked: the gate was **redundant, not hiding breakage**. No pre-existing user-facing bugs to split out.

## Route taken: both, neither by moving the job

- **Static** (`tools/posix-lint.sh`): every tracked file whose **first line** is a `#!/bin/sh` shebang gets a real POSIX parser *and* a static bashism linter. Line 1 only — `install-uninstall_test.sh` is a bash file that writes `#!/bin/sh` stubs inside a heredoc, and a content grep would lint it as POSIX sh.
- **Runtime**: `install-uninstall_test.sh` now *executes* the installer under `dash` instead of `sh`. That is the issue's "dash proves the script runs", obtained without a new runner.

**Placement — the decision, not an accident.** The gate is hosted in `linux.yml`'s `build-test`, not `test.yml`. Checked first, as the issue instructed: `test.yml`'s `go-test` is a single job pinned to `macos-latest` for the macOS runtime paths in `go test ./core/...` (kqueue exit-watch, pgrep/lsof discovery), so the whole job cannot move — and the `macos-15` image ships **no shellcheck**, so hosting it there means `brew install`ing a linter on every PR to check two files. `ubuntu-latest` is the only runner where `/bin/sh` is genuinely dash **and** the image ships shellcheck (`Ubuntu2404-Readme.md` lists 0.9.0). The step is split out rather than the job moved.

`checkbashisms` is preferred when present (it reports bashisms and nothing else, so the tool fixes the scope); `shellcheck` is the fallback because it is free on the CI runner, filtered to `SC3xxx`/`SC2039`/`SC2112` so the gate cannot drag in unrelated style debt from scripts nobody has linted before.

## The gate is shown capable of failing

Deliberate `[[ ]]` injected into the real `site/install.sh`:

```
=== OLD gate (dash -n) on the mutated installer ===
>>> dash -n: PASS — the pre-#1423 gate does NOT catch it <<<

=== NEW gate on the mutated installer ===
posix-lint: 2 file(s); parser=dash, bashisms=checkbashisms
FAIL [checkbashisms] site/install.sh
possible bashism in site/install.sh line 299 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ -z "${VERSION}" ]]; then :; fi
  ok  tools/linux-replay-entrypoint.sh

posix-lint: FAILED — the above are bashisms in a #!/bin/sh script.
  These ship to users running dash (Debian/Ubuntu /bin/sh).
>>> new gate exit: 1 <<<
```

The mutation was then reverted (`git diff --stat site/install.sh` empty). Rather than leave that as a one-off, the corpus is **committed** under `tools/lib/testdata/posix-lint/` — one deliberately-broken fixture per bashism class, driven by `tools/lib/posix-lint_test.sh`, so the evidence outlives this PR. `good-clean.sh` is the vacuity guard: without a clean fixture that must pass, a linter that rejected everything would satisfy all eight.

## A real defect this found — in the gate itself

The first draft of `lint_bashisms` reproduced #1423 **inside the gate built to fix it**: it piped the linter into `grep` and tested the captured string for emptiness, so when the filter failed to execute the capture came back empty, empty read as clean, and it printed `ALL PASS` over the installer *while the deliberate `[[ ]]` was still in it*. Every branch now checks the tool's own exit status and filters in bash rather than through a second process. Test case 9c pins it, seen red against the naive version:

```
FAIL: broken linter: silent exit-2 is a FAILURE, not a clean file — expected [1] got [0]
FAIL: broken linter: says it could not check — expected [contains 'could not check'] ...
posix-lint_test: 7 FAILED
```

More generally: **no POSIX shell, no static linter, and an empty file set are hard failures, never skips.** A gate whose absence reads as a pass is the defect being fixed.

## Red-first / locks

- **Seen red:** case 9c above (naive pipeline → `expected [1] got [0]`), and all eight `bad-*` fixtures, which the pre-#1423 gate passes on five of.
- **Locks** (pass by construction, pinning behaviour that must not change): `good-clean.sh` vacuity, the discovery cases, and the existing `install-uninstall_test.sh` suite, which stays green now that the installer executes under `dash`.

## Verification

`tools/preflight.sh` chunked in the foreground — `go`, `web`, `arch`, `tools`, `skills`, `security`, `posix`: **all PASS**.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -28,6 +28,23 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v7
 
+      # POSIX-sh gate (#1423). This job is the host because ubuntu is the only
+      # runner where the property can actually be tested: `/bin/sh` here is
+      # dash, and the image ships shellcheck (0.9.0). test.yml's `go-test` job
+      # cannot host it — that job is pinned to macos-latest for the macOS
+      # runtime paths in `go test ./core/...` (kqueue exit-watch, pgrep/lsof
+      # discovery), and the macos image ships no shellcheck at all, so the
+      # gate would have to `brew install` a linter on every PR to check two
+      # files. Runs before setup-go: it needs no toolchain, takes well under a
+      # second, and behind the Go steps a compile error would abort the job
+      # and leave an installer-only PR with no installer feedback.
+      #
+      # site/install.sh reaches users as `curl … | sh`, which on Debian and
+      # Ubuntu is dash — so a bashism here lands on a new user's first
+      # command, before anything is installed that could report it.
+      - name: Lint POSIX sh scripts
+        run: tools/posix-lint.sh
+
       - name: Set up Go
         uses: actions/setup-go@v7
         with:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -45,6 +45,15 @@ jobs:
       - name: Lint POSIX sh scripts
         run: tools/posix-lint.sh
 
+      # The gate's own unit tests, over the deliberately-broken corpus under
+      # tools/lib/testdata/posix-lint/. They live here rather than in test.yml's
+      # `tools/lib/*_test.sh` loop for the same reason the gate does: they need
+      # a static bashism linter, and the macos image ships none. A gate that
+      # passes by construction is what #1423 is about, so the mutation corpus
+      # has to run somewhere — this is where it can.
+      - name: Test the POSIX-sh gate
+        run: bash tools/lib/posix-lint_test.sh
+
       - name: Set up Go
         uses: actions/setup-go@v7
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,6 +97,15 @@ jobs:
       - name: Test the shared shell libs
         run: |
           for t in tools/lib/*_test.sh; do
+            # posix-lint_test.sh drives tools/posix-lint.sh, which REQUIRES a
+            # static bashism linter and refuses to pass without one (#1423 is
+            # about a gate that passed having checked nothing, so it fails
+            # rather than skipping). This runner is macos-latest, whose image
+            # ships neither shellcheck nor checkbashisms — which is the same
+            # reason the gate itself lives in linux.yml. Both run there.
+            # Installing a linter here would cost every PR 20-40s to re-test
+            # what linux.yml already tested.
+            case "$t" in */posix-lint_test.sh) continue ;; esac
             echo "== $t =="
             bash "$t"
           done
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -271,14 +271,21 @@ Before marking a ticket done, run the full suite — every layer must pass:
   `tools/lib/install-uninstall_test.sh` is a bash file that writes `#!/bin/sh`
   stubs inside a heredoc, and a content grep would try to lint it as POSIX sh.
   It runs two different kinds of check on each file: a real POSIX shell's
-  parser (`dash -n`) and a static bashism linter (`checkbashisms`, else
-  `shellcheck --shell=sh` filtered to its POSIX-compatibility codes —
-  `SC3xxx`, plus `SC2039` and `SC2112`, so general style debt stays out of
-  scope). Both, because the parser alone is far weaker than it looks:
-  measured one bashism per file, `dash -n` catches **3 of 8** — it flags
-  arrays, process substitution and the `function` keyword, and accepts
-  `[[ ]]`, `${v,,}`, `+=`, `echo -e` and `source`. Either static linter
-  catches 8 of 8. That gap is #1423: `site/install.sh` reaches users as
+  parser (`dash -n`) and **every** static bashism linter installed —
+  `shellcheck --shell=sh` filtered to its POSIX-compatibility codes
+  (`SC3xxx`, plus `SC2039` and `SC2112`/`SC2113`, so general style debt stays
+  out of scope) and `checkbashisms`. Both kinds, because the parser alone is
+  far weaker than it looks: measured one bashism per file, `dash -n` catches
+  **3 of 8** — it flags arrays, process substitution and the `function`
+  keyword, and accepts `[[ ]]`, `${v,,}`, `+=`, `echo -e` and `source`, where
+  either static linter catches all eight of those. *Every* installed linter
+  rather than the first one found, because the two disagree beyond that
+  sample: `checkbashisms` accepts `local`, `set -o pipefail` and `echo -n`,
+  which `shellcheck` rejects (SC3043/SC3040/SC3037) and which an installer
+  accretes. CI has shellcheck and not checkbashisms, so preferring the other
+  one locally would let a developer's preflight pass a diff CI rejects —
+  running both is monotone, and a run without shellcheck says out loud that
+  it is weaker than CI. That gap is #1423: `site/install.sh` reaches users as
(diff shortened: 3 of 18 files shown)


ingo-eichhorst:
## Review round applied (6 findings) + a live-CI correction

The review subagent (medium effort) found **6 findings, one a confirmed blocker**, and CI proved it: `go-test` was **red** on the first push. All are fixed in `23f4071`/`7cb9ab5`.

**The blocker is worth recording, because it is this issue's own defect reappearing.** `tools/lib/posix-lint_test.sh` landed in test.yml's `tools/lib/*_test.sh` glob, which runs on `macos-latest` — the runner this PR argues cannot host the gate, because it ships neither linter. The diff contradicted itself. My local preflight passed because this machine has both tools installed; only CI could see it. Collateral: that step aborts on first failure and `posix-lint_test` sorts *before* `skill-lint_test`, so #1209's coverage had silently stopped running in CI too. Both now fixed — the loop skips that one file by name, `linux.yml` runs it beside the gate.

Second-most important, finding 2: the gate *preferred* checkbashisms and fell back to shellcheck, making its strength depend on the host. The two disagree — checkbashisms accepts `local`, `set -o pipefail` and `echo -n`, which shellcheck rejects (SC3043/SC3040/SC3037) and which an installer accretes. CI has shellcheck only, so a developer with checkbashisms would have had preflight pass a diff CI rejects. It now runs **every installed linter** (monotone: local can be stricter, never weaker), and a run without shellcheck says so out loud.

Also: discovery silently dropped C-quoted non-ASCII paths (4), the `--changed` trigger under-fired on extensionless scripts (3), SC1xxx parse-aborts could be filtered into a false clean (5), and case 9b claimed to pin the severity filter but did not (6) — now pinned by `noisy-but-posix.sh`, seen red against the unfiltered version:

```
FAIL: filter: shellcheck-noisy but POSIX-clean still passes — expected [0] got [1]
```

A `/simplify` inline glance then caught one more, in the same class the gate guards: shellcheck reads any comment whose **first token** is `shellcheck` as a *directive*, so six prose lines naming the tool raised SC1072/SC1073 — the parse-abort family — leaving both new scripts unanalysable by the linter this PR wires into CI. Reworded; both are now shellcheck-clean.

## Verified on the runner, not assumed

```
posix-lint: 2 file(s); parser=dash, bashisms=shellcheck
  ok  site/install.sh
  ok  tools/linux-replay-entrypoint.sh
posix-lint: ALL PASS
```

All 8 bashism classes are detected on ubuntu's shellcheck 0.9 (a version concern I had; unfounded). `/bin/dash` is confirmed present on the **macOS** runner too — `syntax: site/install.sh parses under /bin/dash (real POSIX shell; also ran it)`.

`go-test`, `build-test`, `ars-gate` all pass on head `7cb9ab5`, and the two new steps are confirmed executed inside `build-test` (not merely dispatched).

**Not fixed, deliberately:** `tools/preflight.sh` carries two pre-existing `SC2164` warnings, both present on `origin/main` and unrelated to this change.

PR #1470, opened 2026-08-10
fix(focus): select the right Ghostty tab by working directory

MBoegers:
## Summary

Clicking a Ghostty session in the menu bar brought Ghostty to the front but left whatever tab was last focused. Ghostty was registered as a generic `AXTitleMatchActivator`, which picks the best **title** match among the app's Window-menu entries — and that key doesn't identify an agent's tab. For a cwd of `…/IdeaProjects/Irrlicht`, the tab title `✳ Debug Irrlicht navigation` scores 4 and lands correctly while `✳ Fix the flaky test` scores 0 and lands nowhere. Same session, same directory; the result turns on the agent's phrasing, which is why it read as intermittent. Ghostty also abbreviates its own titles (`~/…`, `…/…`), so the exact-cwd tier that would otherwise dominate can never fire.

Ghostty 1.3.1 ships a scripting dictionary exposing `working directory` on each terminal surface, plus `focus`. Matching on that is decidable where a title match is a guess.

## Changes

- New `GhosttyActivator`: enumerate surfaces via AppleScript, match `working directory` against the session cwd, `focus` a **unique** match. Replaces the `AXTitleMatchActivator` registry line for `ghostty`.
- Not an identity match, and the design follows from that. Ghostty exports no per-surface variable to child processes, its `terminal` class carries only `id`/`name`/`working directory`, and it has no remote-control CLI — so two surfaces in one directory are indistinguishable. Zero matches and several both raise the window and **leave the tab selection alone** rather than guessing into another agent's terminal.
- Neither non-unique path falls back to the title match. `kAXPressAction` on a Window-menu item makes AppKit open that menu and run an `NSMenuTrackingSession` on the target's main thread; observed against Ghostty 1.3.1 that session does not always end — the main thread parks in `-[NSMenuTrackingSession startRunningMenuEventLoop:]` and the app stops answering Apple Events, activating, or repainting (captured in a `sample(1)` trace). Freezing the terminal you're jumping to is worse than not moving the tab.
- Path comparison resolves symlinks then strips a leading `/private` for `/tmp`, `/var`, `/etc`. `resolvingSymlinksInPath()` alone is existence-dependent in both directions, so `/tmp/x` and `/private/tmp/x` normalise differently once `x` is gone — exactly when someone clicks the row to go look at it.
- `NSAppleEventsUsageDescription` named only iTerm2 and Terminal.app; it's the Automation consent dialog text.
- `site/docs/macos-setup.html`: a Ghostty section covering the permission prompt, the shell-integration dependency, and the same-directory limitation.

## Test plan

- [ ] ~~Full test suite passes locally~~ — **the Swift tests in this PR have never been executed.** Only CommandLineTools is installed here, so `swift test` cannot find XCTest, and **no CI workflow runs `swift build` or `swift test`** at all. Flagging rather than hiding it.
- [x] Added tests: `GhosttyActivatorTests` (matching, normalisation, parsing) and two `LauncherHarnessTests` (unique match selects the tab; ambiguous cwd does not move the selection).
- [x] Manually exercised against live Ghostty, driving the production activator from a binary compiled from the real source: unique cwd → selects the target tab; unknown cwd → selection unchanged; ambiguous cwd → selection unchanged. All pass.
- [x] 27 assertions on the matching layer, compiled against the real source; the two `/private` normalisation cases were **seen red first**.
- [x] `swift build` clean; `tools/preflight.sh --changed` all PASS/SKIP.

**Not verified:** the harness test has not been seen red — `titleMatchScore` returning 0 for these titles is calculated, not recorded from a run. Marked as such in the file.

## Related issues

None filed — diagnosed and fixed in one session.

## Checklist

- [x] Follows AGENTS.md / CONTRIBUTING.md conventions
- [x] Conventional Commits
- [x] Documentation updated (`site/docs/macos-setup.html`)
- [x] No `cancelled` session state introduced — three states only: `working`, `waiting`, `ready`


ingo-eichhorst:
## The Swift tests in this PR have now been executed — 21/21 pass

Your test plan flagged this as the PR's one unverified item:

> ~~Full test suite passes locally~~ — **the Swift tests in this PR have never been executed.** Only CommandLineTools is installed here, so `swift test` cannot find XCTest

That is no longer true on this machine — `xcode-select -p` now resolves to `/Applications/Xcode.app/Contents/Developer` (Apple Swift 6.2.4, target `arm64-apple-macosx26.0`), so XCTest is available. Ran against your branch at `bbd8166d`:

```
Test Suite 'GhosttyActivatorTests' passed
	 Executed 21 tests, with 0 failures (0 unexpected) in 0.007 seconds
```

Every case passes, including the two `/private` normalisation ones you noted were seen red first.

**I deliberately did not run `LauncherHarnessTests`.** Those guard on `NSRunningApplication.runningApplications(withBundleIdentifier:)` and drive live Ghostty, and there is a live Ghostty on this machine hosting real work — moving its tab selection to satisfy a test is not something to do unasked. So the harness half of your test plan remains verified only by you.

## The rest of the suite: two failures, and they are not yours

A full `swift test --skip LauncherHarnessTests` reports two failures in `SessionRowSnapshotTests` (`testFixtureAntigravityGhost`, `testGhostRowPID0NilMetrics`).

**Both reproduce identically on `main` with your branch absent**, so this PR does not cause them. I measured rather than assumed: the two differ in the same 415 of 67 200 pixels, in the same 26×26 box at (655, 35), by the same maximum channel delta of 72, with the alpha channel identical — one shared drawn element, which magnifies to the Irrlicht logo glyph rendering with different stroke antialiasing. The references were last regenerated four weeks ago by #1044 for the same reason.

Filed separately as #1509 so it is not attached to your PR.

## On the substance

The diagnosis in your description is the good part — that `AXTitleMatchActivator` scores a tab title against a cwd, so `✳ Debug Irrlicht navigation` lands and `✳ Fix the flaky test` does not, *for the same session in the same directory*. That explains the intermittency in a way "focus is flaky" never would.

Two decisions I'd have had to argue for and you already made correctly: refusing to fall back to the title match on the non-unique paths (given the `NSMenuTrackingSession` hang you captured in a `sample(1)` trace, freezing the terminal you are jumping to is strictly worse than not moving the tab), and treating zero matches and several the same way rather than guessing into another agent's terminal.

Merging is the maintainer's call, not mine — this is an external contribution and I am not going to land it unilaterally. Flagging that from my side the code review is clean, the unit tests now run and pass, and the only red in the suite is pre-existing and filed.

ingo-eichhorst:
## Correction to my previous comment — one claim of mine was weaker than I stated

A follow-up investigation filed **#1523**: `swift test` intermittently aborts (SIGABRT) inside `SoundPlayerTests` via XCTest's stall detector, and when it does it **silently drops 8 suites and the aggregate total** while every suite that did run still reports `0 failures`. Measured at roughly 1 run in 5 on this machine.

Two things follow for what I told you.

**Still stands — your 21/21.** That came from `swift test --filter GhosttyActivatorTests`, which does not include `SoundPlayerTests` and which printed its own aggregate (`Executed 21 tests, with 0 failures`). A filtered run that prints its total is exactly the shape #1523 says is trustworthy.

**Does not stand — "the only red in the suite is pre-existing and filed".** That came from my `swift test --skip LauncherHarnessTests` run, and that run ended with `Exited with unexpected signal code 6` — i.e. it aborted. So I saw 32 of 40 suites and told you something about all of them. The two `SessionRowSnapshotTests` failures are real (independently reproduced on `main` with a narrow filter, and they are #1509), but I cannot claim nothing else is red, and I should not have.

Nothing here changes the assessment of your PR: the diff is 6 files confined to `platforms/macos/` and `site/docs/`, none of them anywhere near the dropped suites, and your unit tests pass on their own terms. I am flagging it because I stated it with more confidence than the measurement supported, and because #1523 means **anyone** re-running the macOS suite on this PR should check for `exit=0` and a printed `Executed N tests` line rather than the absence of a `FAIL`.

Issue #1447, opened 2026-08-10
codex sessions are now born working, not ready — six cells' specs disagree with today's recordings, blocking #1388


ingo-eichhorst:
Blocks #1388. Found while recording the missing codex hook cells (PR #1443, merged) — that PR fixed three rig blockers and proved `hook_received` arrives for both `PermissionRequest` and `Stop`, but the recordings still cannot be promoted because of this.

## The observation

**Codex sessions are now born `working`, not `ready`.** Every codex cell's `expected.jsonl` asserts born-`ready`; recordings from May 2026 show that, today's do not.

Concretely, cell 2-13 scores **5/6**, failing only this assertion. 2-19 cascades to **1/6**. `promote-recording.sh` is validate-then-write, so it refuses and nothing is half-written — which is why `of coverage --hooks` still reports:

```
codex   yes   44   36   0   GAP — declares hooks, zero hook-bearing recordings
```

The recordings exist and are correct about hooks. They cannot land while they disagree with the specs about the opening state.

## Two candidate causes, not yet separated

1. **Upstream drift.** Codex 0.147 may materialize its rollout file later than it used to, so at the moment irrlicht first sees the session the agent genuinely *is* working. If so the specs are stale and the fix is a re-assessment.
2. **A daemon regression.** Something in irrlicht's own classification changed the opening verdict. If so the fix is in `core/`, and six cells' worth of specs are correct as written.

**This was explicitly not isolated** by the recording work — it is recorded here as a hypothesis with two branches, not a diagnosis.

## Why it needs a decision rather than an edit

The recording agent deliberately **did not weaken any `expected.jsonl`** to make a cell pass, and that restraint is the right default: editing six specs to match today's observation would erase the evidence of whichever cause is real, and if it is cause 2 it would bless a regression permanently.

So the first deliverable is the **discrimination**, not the fix:

- Check whether a codex session's rollout file exists at the moment `pid_discovered` fires, and what the transcript's first lines say. That distinguishes "the agent is working" from "we think it is".
- Compare against a May 2026 recording of the same cell — the fixtures are on disk and are exactly the controlled comparison this needs.
- If it is upstream drift, say what codex changed and when, then re-assess the affected cells as a group rather than editing them one at a time.
- If it is a daemon regression, it is a `core/` bug with its own red-first test and this issue becomes its parent.

## Scope of the blast radius — measured, not assumed

[...]

Issue #329, opened 2026-05-14
Session stuck in working after Claude Code writes away_summary post-turn


ingo-eichhorst:
## Problem

A Claude Code session that has correctly transitioned `working → ready` gets bounced back to `working` and stays there indefinitely as soon as Claude Code writes a post-turn `system / away_summary` record (≈3 minutes after the turn ends). The session never returns to `ready`.

Live repro: session `29e274cf-f35f-4d79-9492-35c3da6a6bcb` (transcript: `~/.claude/projects/-Users-ingo-projects-irrlicht/29e274cf-f35f-4d79-9492-35c3da6a6bcb.jsonl`).

## Evidence

Transcript tail (relevant timestamps):

| Line | Time (UTC) | Event |
|---|---|---|
| 242 | 11:22:37.710 | `type=assistant`, `stop_reason=end_turn` |
| 243 | 11:22:37.779 | `type=system`, `subtype=turn_duration` → parser emits `turn_done` |
| 244 | — (no ts) | `type=last-prompt` (metadata) |
| 245 | — (no ts) | `type=ai-title` (metadata) |
| 246 | — (no ts) | `type=permission-mode` (metadata) |
| 247 | 11:25:41.324 | `type=system`, `subtype=away_summary` — Claude Code's "user has been away" recap |

Daemon log for this session:

- 13:22:37 (local) — `agent finished turn → ready` ✓
- 13:22:39 — `agent finished turn → ready` ✓ (duplicate check, idempotent)
- 13:25:41 — **`transcript activity (ready → working)`** ← stuck here
- After that: only `transcript_activity` events, no further state transitions

Current daemon state via `/api/v1/sessions`: `state=working`, `has_open_tool_call=false`, `last_event=transcript_activity`, ~330s elapsed since the bounce.

## Root cause (with one open question)

The `away_summary` line is correctly skipped by the claudecode parser (`core/adapters/inbound/agents/claudecode/parser.go:150-156` — `handleSystemEvent` sets `ev.Skip = true` for any subtype other than `turn_duration` / `stop_hook_summary`). The tailer honors this at `core/pkg/tailer/tailer.go:414-431`: `LastEventType` is not updated, `addMessageEvent` is not called.

But the fswatcher still fires a transcript-activity event (it only watches mtime). `SessionDetector.processActivity` (`core/application/services/session_detector_activity.go:244`) runs the full pipeline anyway:

1. `RefreshOnActivity` recomputes metrics (line 299).
[...]

Issue #144, opened 2026-04-12
replay: extra transitions for /continue sessions with multiple process lifetimes


ingo-eichhorst:
## Problem

Fixture 13 (`13-full-lifecycle-continue-8a525d27`) produces 6 extra `ready→working→ready` transitions that the daemon never emitted. All 10 recorded transitions match perfectly, but the replay emits 16 total.

## Root cause

The session uses `/continue` — the same session ID (`8a525d27`) spans two separate process lifetimes:

1. **1st launch** (20:52–20:56): 6 recorded transitions, process exits at seq 673
2. **2nd launch** (21:11–21:11): 4 recorded transitions, process exits at seq 728

When the daemon's process exits, it tears down the session and stops processing fswatcher events. But the transcript file persists on disk and the `/continue` session appends to the same `.jsonl`.

`ReplayWithSidecar()` collects all `transcript_activity` events for the session across both lifetimes into a single fswatcher stream. It only uses the *first* `process_exited` timestamp to cancel the final debounce timer — it doesn't split the stream at process boundaries. This causes the debounce state machine to fire on fswatcher events that arrived between the two process lifetimes, producing ghost transitions the daemon never emitted.

## Fix

Split the fswatcher event stream at `process_exited` boundaries and reset the debounce state machine between process lifetimes. Each segment between `transcript_new` and `process_exited` for the same session should be replayed independently, with the debounce state cleared at each process exit.

## Affected files

- `core/cmd/replay/main.go` — `ReplayWithSidecar()`, specifically the merged timeline / debounce loop (~lines 830–900)
- `testdata/replay/claudecode/13-full-lifecycle-continue-8a525d27.events.jsonl` — fixture that exercises this path

## Expected outcome

Fixture 13 extended check should report 0 mismatches: 10 recorded, 10 replayed, all ordered matches.

PR #1476, opened 2026-08-11
fix(replay): reproduce the daemon's transitions in sidecar-driven replay

### LINKED ISSUE #1342: Sidecar replay reproduces fewer transitions than the daemon recorded: 22 fixtures now assert nothing
Follow-up to #1326 / PR #1341. That PR paired every recording with its `events.jsonl` sidecar, which turned sidecar-driven replay on catalog-wide for the first time. Doing so exposed a pre-existing defect in the sidecar replay path that had never been observable.

## The defect

For 22 committed recordings, sidecar-driven replay produces **zero** state transitions where the daemon that made the recording logged 1–4. The extended check is the witness — it diffs replay against the daemon's own `state_transition` events:

| cell | golden transitions (pre-#1341 → now) | daemon recorded | replay reproduced |
|---|---|---|---|
| `codex/scenarios/1-2_session-end` | 6 → 1 | 2 | 0 |
| `codex/scenarios/2-1_basic-turn` (×2 recordings) | 3 → 1 | 2 | 0 |
| `codex/scenarios/2-5_synchronous-slash-command` | 3 → 1 | 2 | 0 |
| `codex/scenarios/5-2_model-identification` | 3 → 1 | 2 | 0 |
| `codex/regressions/agent-question-pending` | 3 → 1 | 1 | 0 |
| `codex/regressions/baseline-hello` | 3 → 1 | 1 | 0 |
| `pi/scenarios/1-2_session-end` | 3 → 1 | 2 / 4 | 0 |
| `pi/scenarios/1-8_model-context-display` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-13_turn-end-terminal-text` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-16_oversized-transcript-line` | 3 → 1 | 4 | 0 |
| `pi/scenarios/2-5_synchronous-slash-command` | 3 → 1 | 4 | 0 |
| `pi/scenarios/5-2_model-identification` | 3 → 1 | 4 | 0 |

(full list of 22 in PR #1341's body; all `codex` and `pi`)

Those goldens now contain only the synthetic `init` row, so they pin **nothing** about the state machine. 129 goldens lose transitions in total; these 22 are the ones reduced to asserting nothing.

## Why this matters more than the count suggests

The replay harness's job is catching classifier regressions. A fixture whose golden is `[init]` cannot catch one. This is the same failure shape as #1326 itself — a gate that looks green because it stopped looking — which is why it is worth fixing rather than accepting.

## Cause: NOT established

Stating this explicitly so nobody treats the guesses below as findings. Two candidates were tested during #1341:

- **Ruled out — skipped trailing debounce flush.** Relaxing `flushPendingDebounce`'s `if !d.pending || !d.coalesced` early-return (`cmd/replay/replay_sidecar.go`) and rebuilding changes nothing: `codex/scenarios/2-1_basic-turn` still replays 1 transition. Measured, not reasoned.
- **Ruled out as the general cause — incomplete fs-event coverage.** 30 of 311 sidecar-drivable recordings have a last observed `file_size` below 95% of the transcript, but only **1 of these 22** is among them; the other 21 have 100% coverage.

What *is* established: for `codex/scenarios/2-1_basic-turn` the daemon's two transitions (seq 5651/5652) occur **after** its last recorded `transcript_activity` (seq 5648), and the replay consumes all 12 fs events without emitting either. A plausible next lead — untested — is that the classifier's staleness overlay uses the sidecar's wall-clock `virtTime` while transcript-only replay derives virtual time from the transcript's own timestamps, so content that reads as fresh in one mode reads as stale in the other. Marked as a lead, not a diagnosis.

## Reproduce

```
go build -o /tmp/replay ./tools/onboarding-factory/cmd/replay
/tmp/replay --out /dev/null \
  replaydata/agents/codex/scenarios/2-1_basic-turn/recordings/2026-05-23-20-36-20_irrlichd-0.4.7+723609a/transcript.jsonl
# replay: 12 events → 1 transitions, 0 flickers [extended-check: kinds FAIL ordered 0/2 FAIL]
```

## Acceptance

Sidecar-driven replay reproduces at least the transitions the daemon recorded for these 22 recordings (extended check's `replayed_transition_count` ≥ `recorded_transition_count`, `missing_kinds` empty), goldens regenerated, and a test that fails on `main` for at least one of them.

## Related

- 198 of 309 sidecar-driven recordings report *some* extended-check divergence (listed at the end of every `tools/replay-fixtures.sh` run). The dominant kind is a missing terminal `working→ready`. These 22 are the extreme tail of that distribution; fixing the tail may or may not address the rest.

ingo-eichhorst:
Closes #1342.

Sidecar-driven replay emitted **zero** state transitions for 20 committed recordings whose daemon logged 1–4, leaving those goldens holding nothing but the synthetic `init` row. **Two independent causes**, plus a third consistency defect that fix 1 would have amplified 10×.

## Relationship to #1447 / PR #1464 — established, not assumed

**Unrelated.** #1464 changes only `core/application/services/session_detector_activity.go`, the live `SessionDetector`. The sidecar replayer never calls it — `replay_sidecar.go` mirrors `processActivity`'s pipeline itself and reaches into `services` only for `ClassifyState` and the shared reason constants.

Measured rather than reasoned: I built the replay binary at `2350b97c` (the #1464 merge) and at `2350b97c^`, swept **all 309** sidecar-driven recordings with each, and diffed — **byte-identical**. The issue's reproduction also still prints exactly the line the issue recorded, on a tree containing #1464:

```
replay: 12 events → 1 transitions, 0 flickers (ww=0 wr=0 rw=0) [extended-check: kinds FAIL ordered 0/2 FAIL]
```

## Cause 1 — `process_exited` discarded an already-expired debounce window

A recording's fs events routinely all land inside one debounce window, so the leading-edge pass is the only classify that runs and every later event merely extends the deadline. The daemon's real timer then fires that window and emits the turn's transitions. The replayer reached the `process_exited` branch first, which reset the debounce state wholesale.

For `codex/scenarios/2-1_basic-turn` the arithmetic is unambiguous:

| moment | virtual time |
|---|---|
| last fs event (`seq 5648`, `file_size=59472` = 100% of the transcript) | `20:36:39.129` |
| window deadline (last event + 2 s) | `20:36:41.129` |
| **daemon's two `state_transition` events** (`seq 5651`, `5652`) | `20:36:41.1317`, `20:36:41.1325` |
| `process_exited` (`seq 5653`) | `20:36:44.492` |

The daemon's timer fired 2 ms after the computed deadline; `process_exited` did not arrive until **3.4 s later**. Cancelling a still-**live** window on teardown is correct and is preserved — the fix calls the *existing* `flushDebounceIfExpired`, which no-ops unless the deadline has passed, exactly as the sibling `timelineHook` and `timelineChildOrphan` branches already do.

**This explains #1342's own "ruled out" measurement.** Relaxing `flushPendingDebounce`'s `!d.coalesced` early-return genuinely changes nothing — by the time that trailing flush is reached, the process-exit reset has already zeroed `d.pending`, `d.coalesced` *and* `d.pendingSize`. That measurement was correct; the real cause was one step upstream.

## Cause 2 — a manufactured header-only first pass suppressed classification

Fix 1 took the population from 20 to 7. The remaining 7 share one shape: one classify pass, on the first fs event, `NoSubstantiveActivity=true`, `LastEventType=""`.

The sidecar records `file_size` from the fswatcher's stat at fire time — `d.record(lifecycle.Event{…, FileSize: ev.Size})` runs *before* `d.onActivity(id, ev)`, whose `RefreshOnActivity` then tails to EOF. `classifyAt` clamps to that stat, so for codex and pi — both of which open a transcript with a multi-kilobyte `session_meta` — it can land exactly on the end of that header and manufacture a "header line only" pass with no counterpart in the daemon's history.

`codex/regressions/baseline-hello`: first fs event reports `file_size=15401`, *exactly* the byte offset of the end of line 1. The daemon read to EOF ~57 ms later, at which point the file was 28252 bytes and carried a user message, and it logged `ready→working` at `11:29:00.354`.

`readBoundaryFor` widens that one shape to the next recorded stat and re-tails. **It widens the boundary, never the verdict** — #329's guard is re-applied unchanged, so a pass still non-substantive at the wider boundary still skips.

### Review caught a real defect in the first draft

The first version narrowed #329's guard itself (skip only when `LastEventType != ""`). That **classified on no evidence** and fabricated a `ready→working` in two goldens whose daemon recorded **zero** transitions — including `codex/scenarios/1-1_session-start`, the sole recording for a core-twelve scenario, which went from *agreeing* with its recording to *disagreeing*. Rewritten to fix the read boundary instead.

### Models tried, measured not argued

All 309 sidecar-drivable recordings:

| model | zero-transition | fabricated | divergent |
|---|---|---|---|
| `main` | 20 | 1 | 196 |
| debounce fix only | 7 | 1 | 146 |
| \+ global one-event lookahead (rejected) | 4 | 1 | **153 — worse** |
| \+ narrowing #329's guard (rejected) | **0** | **3 — fabricates** | 145 |
| \+ `readBoundaryFor` (**shipped**) | 4 | 1 | 145 |

The single "fabricated" case (`copilot/2-19`) **pre-dates this change** and neither fix moves it.

**The acceptance criterion is met for 16 of the 20**, not all 20 — stating that plainly rather than forcing it. The remaining 4 need a *time-aware* read boundary: their transcripts are written essentially instantaneously (`pi/2-1_basic-turn` writes 2556 bytes across 4 fs events spanning **3 ms**), so the daemon's single read absorbed the whole file while later fswatcher fires were still queued carrying stale sizes. Reaching them by widening more aggressively is exactly what fabricates. They are pinned in `knownZeroTransition`, which fails if the set **grows** *or* if an entry is fixed and left to rot, and are tracked by **#1478** with the full evidence and both rejected models.

## Cause 3 — `state_durations` summed past `wall_clock_session_duration`

A window fires at its **deadline**, past the last recorded fs event, so those transitions charged duration outside the reported window while the trailing remainder was silently dropped as negative. Fix 1 would have grown this from **10** committed goldens to **100**. The transition timestamp is the honest one and is *not* clamped — the daemon's timer really did fire there — so `extendWindowToLastTransition` widens the window instead. Now **0 catalog-wide**, clearing the 10 pre-existing cases too.

## Red-first evidence

All three tests were written before their fixes and confirmed failing. With `replay_sidecar.go` restored from `origin/main`:

```
--- FAIL: TestReplayWithSidecar_Issue1342_ExpiredWindowSurvivesProcessExit (0.00s)
    issue1342_debounce_test.go:115: replayed 0 transitions, daemon recorded 2
    issue1342_debounce_test.go:115: missing_kinds = [ready→working working→ready], want empty (daemon kinds [ready→working working→ready] vs replayed [])
--- FAIL: TestReplayWithSidecar_Issue1342_HeaderOnlyFirstPassIsWidened (0.00s)
    issue1342_debounce_test.go:132: replayed 0 transitions, daemon recorded 1
    issue1342_debounce_test.go:132: missing_kinds = [ready→working], want empty (daemon kinds [ready→working] vs replayed [])
--- FAIL: TestSidecarReplayAgreesWithTheDaemonsOwnLog (0.59s)
    issue1342_debounce_test.go:243: 16 recordings reproduce zero transitions where the daemon recorded at least one, so their goldens assert nothing …
    issue1342_debounce_test.go:263: checked 309 sidecar-driven recordings
FAIL
```

Cause 2's test is **independently** red — proved with cause 1's fix already applied, so it cannot ride on it:

```
--- FAIL: TestReplayWithSidecar_Issue1342_HeaderOnlyFirstPassIsWidened (0.00s)
    issue1342_debounce_test.go:132: replayed 0 transitions, daemon recorded 1
--- FAIL: TestSidecarReplayAgreesWithTheDaemonsOwnLog (0.58s)
    issue1342_debounce_test.go:243: 3 recordings reproduce zero transitions …
```

The catalog test now asserts **both directions** — zero-where-recorded *and* invented-where-none — because the first draft's regression passed a deficit-only assertion. The two fixture tests pin a **named recording** rather than using `fixturePath`'s newest-wins resolution, so the per-event timestamps they quote as evidence cannot silently stop describing the file being read.

## Measured blast radius

**172 goldens across 8 adapters** — 7.8× the 22 cells the issue names, measured not estimated:

| adapter | goldens | | adapter | goldens |
|---|---|---|---|---|
| gemini-cli | 37 | | claudecode | 9 |
| codex | 35 | | antigravity | 9 |
| mistral-vibe | 27 | | copilot | 8 |
| kiro-cli | 26 | | | |
| pi | 21 | | | |

Net **+119** transitions restored (275 → 394). **No golden lost a transition** — checked per file, not inferred from the net. 174 files total: those 172 goldens plus `replay_sidecar.go` and the new test file. (32 of the 172 are cause 3's window correction only.)

The issue's related figure moves: **198 → 145** of 309 recordings reporting some extended-check divergence. My baseline measurement of that predicate on `main` is 196, not 198; the difference is a counting detail in the predicate, not a behaviour change — 196 is what the committed tree produces today.

## Follow-up filed

**#1478** — the 4 remaining recordings, with the measurement table above, the reason one-step widening cannot reach them, and the open question of whether re-recording them (all four are old: two `irrlichd-unknown`, two `0.4.7`) beats adding a heuristic. Two are `regressions/` fixtures and so are deletion-guarded (#268).

## Verification

- `go test ./tools/onboarding-factory/... -race -count=1` — all 10 packages green
- `go run ./tools/onboarding-factory/cmd/of validate` — OK
- `tools/preflight.sh --only go` — all 7 gates PASS, including **replay fixtures** and **replaydata validate**
- `tools/preflight.sh --only arch` — PASS (ARS composite unchanged)
- `tools/preflight.sh --only tools`, `--only skills` — PASS
- pre-push security scan — PASS
- CI on the head SHA: `go-test`, `build-test`, `ars-gate`, `guard`, both `test (platforms/web)` and both `test (tools/onboarding-factory/internal/viewer/web)`, `Analyze (go/actions/javascript-typescript)`, `CodeQL`, `SonarCloud` — all pass

**CodeScene** (advisory, not a merge gate per AGENTS.md) flagged `classifyAt` for *Bumpy Road Ahead* on the first push. Addressed rather than suppressed: both of that function's reads now go through one `readTo` helper, which flattens it and puts the write cursor in a single place. The advisory test-file findings were addressed too — `pinnedRecording` now takes one catalog-relative path, the same notation `knownZeroTransition` is keyed by.


ingo-eichhorst:
---

## Post-review addendum (orchestrator)

An independent `/ir:code-review high` pass on the source diff confirmed the PR's claims and found **one real failure mode the PR did not mention**. Recording it here because the PR would otherwise read as cost-free, and it is not.

### Confirmed by the reviewer, with evidence

- **Fix 2's test does not ride on fix 1.** Reverting only the `readBoundaryFor` call, keeping fix 1: `TestReplayWithSidecar_Issue1342_HeaderOnlyFirstPassIsWidened` fails (`replayed 0 transitions, daemon recorded 1`) while fix 1's test still passes, and the catalog test reports exactly the 3 new zero-transition recordings the PR's table predicts.
- **Fix 1 is independently red** — reverting only the process-exit flush fails its own test and adds 9 zero-transition recordings.
- **`flushDebounceIfExpired` really does no-op unless expired** (`replay_sidecar.go:865` returns before any mutation; the fire condition matches `advanceFSEvent`'s exactly, `==` boundary included). **There is no path where a still-live window is now flushed** — cancellation-on-teardown is preserved.
- **`readBoundaryFor` cannot mask a mid-session classifier regression.** `LastEventType` persists across passes and survives rotation, and the replay scratch file is append-only, so the guard can only fire before the session's first event that sets it. Empirically all 114 firings are at `eventIdx=0`, one per recording.
- **`knownZeroTransition` fails in both directions** — removing a live entry, and adding an already-fixed one.
- **`extendWindowToLastTransition` is sound**, not papering over an off-by-one: after extension the telescoping sum equals `WallClockDuration` exactly, and `of validate`/`of verify` read neither field.

### The finding: the widening is bounded by SIZE, never by TIME

`readBoundaryFor` widens to the next recorded stat regardless of how far away in time that stat is, so across a long idle gap it classifies against bytes that did not exist at the pass being classified. Over 114 firings the gap distribution is `p50=42ms, p75=13.5s, p90=15.9s, max=31.0s` — **42 of 114 exceed 1s**, where the doc comment's "tens of milliseconds the read takes" justification does not hold.

Result: **20 goldens now pin a first transition more than 1s ahead of their own `events.jsonl`**, worst `mistral-vibe/2-12_context-compaction` at **30.976s** early. The three recordings the fix rescues have gaps of 1ms, 5ms and 60ms — the win comes entirely from the justified regime, the drift entirely from the other.

To be precise about what this is and is not: it is **not** the fabrication disease of the rejected guard-narrowing variant. #329's guard is genuinely re-applied to the re-tailed metrics, which is why fabrication stays at 1. It is a fourth failure mode — *right transition, wrong time*.

**The obvious remedy was measured and refuted.** A 200ms gap bound keeps all three #1342 tests green with an unchanged `knownZeroTransition` set and collapses the six worst drifts (−30.976s → −0.022s), but breaks **35 gemini-cli recordings** that reproduce exactly today (−0.000s → +10…+28s), because there the long gap is idle time *after* a write the daemon did absorb. The sidecar cannot distinguish those two shapes — the same limitation the four `knownZeroTransition` entries concede — and the bounded variant is worse on aggregate (175/118 vs 146/88 recordings drifting >1s/>5s).

**Net, the widening still moves times toward the daemon**: 46 recordings closer, 20 further; >1s drift 171 → 146, >5s 117 → 88. So this is a documented limitation, not a blocker.

### Why it was invisible, which is the more important half

**Nothing asserts transition timestamps.** `compareOrdered` walks `prev_state`/`new_state` index-by-index and never reads the time, and the extended check's headline divergence figure is counts and kinds only. A transition moved 31s is a full pass on every gate in this package.

That is the same shape as #1342 itself — a net that stopped being able to fail — one axis over, and it now applies to all 399 goldens. Filed as **#1480**, which also notes that **#1478 cannot be evaluated without it**: you cannot judge a time-aware read boundary against a harness that does not measure time.

### Also corrected

`extendWindowToLastTransition`'s doc claimed "10 committed goldens to 100". Measured: **42 of 399** on `origin/main`, **132 of 365** with both fixes and the extension disabled, **0 of 399** with it. Direction and mechanism right, magnitudes understated ~4×.

Both doc comments corrected in `dadc41fc` (comments only, no behaviour change).

### Two latent items, neither reachable today

- The whole-file EOF widening branch (no next fs event → boundary is the entire transcript) is taken by **0 of 114** firings, but 5 sidecar-drivable recordings have exactly one fswatch event, so a future recording of that shape could take it. Unbounded and untested.
- `readBoundaryFor` indexes `r.fswatches[next]` raw and does not consult `d.alive`, so it could in principle widen across a daemon-deaf gap (#144). **0 of 114** firings cross a `process_exited`.

### Golden verification (orchestrator, independent of the reviewer)

Across all 172 changed goldens, compared against each recording's own `events.jsonl`:

| | |
|---|---|
| lost a transition | **0** |
| fabricate vs daemon (branch) | **0** |
| fabricate vs daemon (`main`) | **0** |
| gained | 97 |
| unchanged count | 75 |

The goldens are strictly strengthened in the count-and-kind dimension. #1480 is about the dimension this table cannot see.

Issue #1350, opened 2026-08-07
herdr: click-to-focus for herdr-hosted sessions (Phase 3 of #1348)


ingo-eichhorst:
Follow-up to #1348, which shipped Phases 1–2 (daemon-side launcher capture + backchannel control/read-back) in #1349. This is Phase 3: making a click on a herdr-hosted session land in the right window.

## Why it was split out

#1349 removed the *misroute*: a herdr pane no longer inherits a stranger's `term_program`, so the macOS app stops raising an unrelated application. What it does not yet do is raise the *correct* one — `resolveActivator` gets no `term_program` and no `host_bundle_id` for such a session, so `SessionLauncher.jump` logs "no activator" and does nothing. That is honest failure rather than a wrong jump, but it is not the feature.

It was split because the central design question has no obvious answer, and guessing would reintroduce exactly what #1349 fixed.

## The hard part, with the evidence gathered so far

**The agent's own ancestry is useless here.** Verified live (herdr 0.8.0): the pane's process tree is `claude → zsh → herdr server`, and a detached server is reparented to init with no controlling tty:

```
  PID  PPID TTY      COMM
67531     1 ??       herdr        # server
67992 67531 ttys018  -zsh         # pane shell
```

So `resolveHostFromAncestry` terminates without ever reaching a GUI application. This is why #1349 skips the ancestry fallbacks for herdr panes entirely.

**The window belongs to the *client*, not the server.** An attached client is a separate process with a real tty, and *its* ancestry leads to the hosting terminal:

```
30453 11253 ttys012  herdr --remote factory --session factory
30520 30453 ttys012  /Users/ingo/.local/bin/herdr client
```

So focus has to (a) find the live client for this pane's session, (b) resolve *its* host application, (c) tell herdr to select the pane, then (d) raise that host window — the `TmuxActivator` ordering, one indirection deeper.

## Open questions

1. **Matching client → session.** A named session's client carries `--session <name>` in argv, but the default session's client is bare `herdr`. The daemon captures `HERDR_SOCKET_PATH`, not the session name — is the socket recoverable from the client process, or should capture also record `$HERDR_SESSION`? (It is injected; #1349 deliberately did not capture it, since it is not the addressing key.)
2. **No attached client.** A detached session has no window at all. Report "no window" honestly, or attach one? Probably the former, but it needs a decision.
3. **Multiple clients.** herdr supports attaching from more than one place, including over SSH. Which window wins?
[...]

Issue #1348, opened 2026-08-07
[feat]: Support herdr (agent multiplexer) as a terminal backend — focus, backchannel, read-back


ingo-eichhorst:
## Summary

[herdr](https://herdr.dev/) is a terminal multiplexer purpose-built for coding agents ("tmux for coding agents"): a headless Rust server owns a pty per pane, agents survive detach/reattach, and it exposes a first-class local socket API + CLI for reading pane output, sending input, and focusing a pane. It ships a curated list of agent kinds that overlaps heavily with our adapter set.

Irrlicht **discovers** agents running under herdr today (process scan works — they are ordinary processes), but every capability that depends on knowing *who owns the pty* is broken or unavailable: click-to-focus targets the wrong application, and the backchannel (send input / interrupt / read-back) reports the session as not-controllable and not-readable. herdr is a third `tmux`/`kitty`-class terminal backend, and it is the one backend that also hands us the pane identity for free, in the environment.

This is not hypothetical for this repo: herdr is already installed here (`0.8.0`) and in daily use, including a remote session (`herdr --remote factory --session factory`).

## Evidence — what happens today

Probe: isolated herdr server (`herdr --session irrprobe server`), one workspace, `herdr agent start probe --kind claude --pane w1:p1`, against the running production `irrlichd`.

**1. The session is discovered, but its `Launcher` is wrong.** `GET /api/v1/sessions`:

```json
{
 "session_id": "proc-69461",
 "state": "ready",
 "adapter": "claude-code",
 "pid": 69461,
 "cwd": "/Users/ingo/projects/irrlicht",
 "launcher": { "term_program": "vscode", "tty": "/dev/ttys018" }
}
```

`term_program: "vscode"` is a **lie**. The pane is not a VS Code terminal — the value was inherited from the environment the herdr *server* was started in and then handed to every pane the server ever launches. Because the server is long-lived and detached, that stale value never expires. Clicking this session in the macOS app routes to the VS Code activator and raises the wrong window.

**2. Nothing else identifies the host.** In-pane probe:

```
/dev/ttys018
PID=67992 PPID=67531
  PID  PPID TTY      COMM
67531     1 ??       herdr        # server: reparented to init, no controlling tty
67992 67531 ttys018  -zsh         # pane shell
```

[...]

Issue #744, opened 2026-06-23
[bug]: Agent View background agents (kind: bg) keep running detached in the daemon pool and show as active sessions after their window is closed


ingo-eichhorst:
## Summary

A **background agent started via Claude Code Agent View** keeps running **detached** in the persistent `claude daemon run` pool after its agent-view window/terminal is closed. Irrlicht shows it as a live `ready` session — correctly, because it genuinely *is* a live `kind: bg` claude process — but from the user's seat it's a "ghost": there's no terminal/window open for it, yet it sits in the list (and menu bar) as an active session.

This is the same **user-facing symptom** as #727 ("shows more sessions than active"), but a **distinct sub-cause** from the `--bg-spare` infra-ghost fix in #734 / #644 — see "Why #734/#644 don't cover this" below.

## Evidence (captured live, Claude Code 2.1.187, 2026-06-24)

Reporter played with Agent View, then flagged session `eb8dcbdb-36b3-4d7e-9ce7-f75b93a13220` (worktree `eta-marker-v4`, branch `fix/eta-marker-v4-first-marker-carrier`).

**The process is alive and has no controlling terminal:**

```
PID    PPID  STAT TTY  ELAPSED COMMAND
87636     1  Ss   ??   39:25   claude daemon run --origin transient --spawned-by {…}   ← persistent CC daemon
 └ 87705 87636 SNs ??  39:21   claude --bg-pty-host …/79a91bc0.pty.sock … -- … --resume eb8dcbdb … ← pool worker
    └ 87742 87705 SN  ?? 39:20  claude … --resume eb8dcbdb-… --agent claude --model opus  ← the agent (idle)
```

**Session registry** `~/.claude/sessions/87742.json`:
```json
{ "pid": 87742, "sessionId": "eb8dcbdb-36b3-4d7e-9ce7-f75b93a13220", "kind": "bg", "status": "idle" }
```

**Daemon `/state`** shows it as `state: ready`. `TTY=??` on the entire chain confirms no controlling terminal — the agent is fully detached and parented (transitively) to `claude daemon run` (ppid 1), so it survives the terminal/Agent-View window being closed.

The session's transcript (`…/eta-marker-v4/eb8dcbdb-….jsonl`) is marked `"sessionKind":"bg"` and carries a `worktree-state` block pointing back at the original interactive session (`e3f20cdc-…`) that spawned the worktree.

## Why #734 / #644 don't cover this

[...]

Issue #727, opened 2026-06-22
[bug]: irrlicht shows more sessions than active


gernotstarke:
### Preflight

- [x] I'm on the latest release from [Releases](https://github.com/ingo-eichhorst/Irrlicht/releases/latest).
- [x] I searched existing issues and didn't find a duplicate.

### Irrlicht version

0.5.3

### macOS version

26.5.1

### Which agent(s) were involved?

Not agent-specific / app or daemon itself

### What happened?

<img width="380" height="283" alt="Image" src="https://github.com/user-attachments/assets/913aec29-8824-4186-adab-b1e82f289a28" />

See the screenshot: Irrlicht shows three active sessions (Aquarius, Swabook and gernotstarke), but only the two Aquarius sessions are really existing/running right now.

The swabook session has been terminated >24h ago, was running in a terminal inside Obsidian...

No antigravity session is currently running on my machine...

### Steps to reproduce

I usually use different terminals (ghostty, iterm) plus Obsidian-terminal-plugins to start agent cli sessions, plus claude and codex native.

This time it was obsidian and ghostty. No idea how you can reproduce...

### Expected vs actual behavior

I want only the REALLY active / open sessions to be shown in Irrlicht.

### Relevant log excerpts

```shell

```

### Session state file (optional)

```json

```

### Anything else?

Ingo, your Irrlicht is absolutely awesome!! I've already shown and recommended it to dozends of colleagues.

Regards from a very happy user!!

ingo-eichhorst:
Thanks so much for the kind words, Gernot — and for the report! 🙏

A scripted repro would be tricky, but there's a better path: a one-shot diagnostic bundle captured **while the ghost session is still showing in the menu bar**. That tells us exactly what the daemon thinks is running vs. what's actually alive.

**1. Quick check (paste the output here):**
```bash
curl -s http://127.0.0.1:7837/state
```
If the stale `swabook`/`gernotstarke` entries show up in there, that's the smoking gun.

**2. Full bundle.** Open **Terminal** and paste this to *collect* (it won't upload anything — it just gathers files and opens the folder for you to review):
```bash
DIR="$HOME/Desktop/irrlicht-diag-$(date +%Y%m%d-%H%M%S)"; mkdir -p "$DIR"
APP="$HOME/Library/Application Support/Irrlicht"
curl -s http://127.0.0.1:7837/state          > "$DIR/state.json"
[...]

PR #1483, opened 2026-08-11
fix(launcher): refresh a herdr session's host on the liveness sweep

### LINKED ISSUE #1405: herdr: refresh the resolved client host on the liveness sweep, not only at seed
Follow-up to #1350 (PR #1399), which made click-to-focus work for a session
running in a herdr pane by resolving the host window from the attached herdr
*client*.

## The gap

Every other `session.Launcher` field is static for a process's lifetime. Herdr
client attachment is not — detach-here-reattach-there is the entire point of a
persistent multiplexer. #1350 stores that dynamic property as a capture-time
snapshot:

- `captureLauncher` resolves it once, when the PID is first bound.
- `backfillLauncher` re-resolves it only at daemon seed, and only when the need
  fires — and `launcherBackfillNeedsFor` gates `herdrHost` on
  `TermProgram == "" && HostBundleID == ""`, i.e. only when no host was *ever*
  resolved.

So two things go unnoticed:

1. A session captured while detached heals only at the next daemon restart, not
   at the moment a client attaches.
2. A session whose host *did* resolve is never re-checked. Attach in iTerm →
   start the agent → detach → re-attach in Ghostty, and the session points at
   iTerm for the rest of its life, across restarts. Clicking it then raises the
   wrong window — the misroute class #1348 was opened to remove, arriving by a
   slower route.

Both limits are stated in `backfillLauncher`'s doc comment today; this issue is
to fix them rather than let "documented" stand in for "handled".

## Shape of the change

Re-run the herdr host resolution from the existing periodic liveness sweep
(`PIDManager.CheckPIDLiveness`) rather than only at seed, and drop the "only if
never resolved" precondition for herdr launchers specifically.

Two things already in place make this cheaper than it sounds:

- `herdrClientCache` (5s TTL, keyed by socket path) collapses every pane of one
  herdr server into a single `lsof` scan, so the cost is one scan per herdr
  *server* per interval, not per session.
- `applyLauncherBackfill` returns false when nothing changed, so a steady state
  produces no `UpdatedAt` churn and no websocket push churn.

## An approach that is *not* needed

Resolving at focus time instead is the wrong shape: the macOS row-tap path
(`SessionRowView` → `SessionLauncher.jump`) reads the session state the app
already holds and does not round-trip to the daemon, so focus-time resolution
would require routing the most-used interaction in the app through the daemon.
Refreshing on the sweep needs no click-path change at all — the daemon already
pushes launcher updates, so the row tap picks up a corrected host for free.

## Verification

A real detach/re-attach cycle against a live herdr (0.8.0) — attach in one
terminal, bind a session, detach, re-attach in a different terminal, confirm
the click follows — plus a unit test that a herdr launcher with a stale host is
refreshed while a non-herdr one is still left alone.

ingo-eichhorst:
Closes #1405. Follow-up to #1350 (PR #1399).

## Root cause

Every other `session.Launcher` field describes something fixed for the life of
the agent process, so capturing it once when the PID is bound is correct. A
herdr pane's host is not — it belongs to whichever client is attached *right
now*, and detach-here-reattach-there is the entire point of a persistent
multiplexer.

#1350 stored that dynamic property as a capture-time snapshot. Two independent
gates each froze it:

| Gate | Where | Effect |
|---|---|---|
| `captureLauncher` is set-once (`state.Launcher != nil` → no-op) | `pid_manager.go:280` | a session bound while detached never picks up a client that attached later, except at seed |
| `launcherBackfillNeedsFor` required `TermProgram == "" && HostBundleID == ""` | `pid_manager.go:1294` | a host that *did* resolve was never re-checked, ever |

and `backfillLauncher` — the only thing that re-reads a launcher at all — runs
only from `SeedPIDs`, i.e. at daemon startup.

So: attach in iTerm → start the agent → detach → re-attach in Ghostty, and the
session names iTerm for the rest of its life, **across daemon restarts** (the
seed-time backfill's own precondition rejects it, because a host *is*
recorded). Clicking it raises the wrong window — the misroute class #1348 was
opened to remove, arriving by a slower route.

Both limits were already written down in `backfillLauncher`'s doc comment as
"deliberate for now". This replaces "documented" with "handled".

## Red first

Written before the fix existed, run against `main`:

```
$ go test ./core/application/services/ -run 'TestCheckPIDLiveness_Herdr|TestCheckPIDLiveness_NonHerdr' -count=1
--- FAIL: TestCheckPIDLiveness_HerdrHostRefreshedOnReattach (0.00s)
    pid_manager_herdr_refresh_test.go:65: host not refreshed on the sweep: TermProgram = "iTerm.app", want ghostty
    pid_manager_herdr_refresh_test.go:68: the previous client's iTerm selector survived the re-attach: &{TermProgram:iTerm.app ITermSessionID:w0t0p0-OLD TermSessionID: TmuxPane: TmuxSocket: VSCodePID:0 TTY:/dev/ttys012 KittyListenOn: KittyWindowID: KittyPID:0 HostBundleID: HerdrPaneID:w1:p1 HerdrSocketPath:/cfg/herdr/herdr.sock}
    pid_manager_herdr_refresh_test.go:71: TTY still names the old client's tab: "/dev/ttys012"
--- FAIL: TestCheckPIDLiveness_HerdrHostClearedOnDetach (0.00s)
    pid_manager_herdr_refresh_test.go:109: a detached session must report no host, not the last one: &{TermProgram:iTerm.app ITermSessionID:w0t0p0-OLD TermSessionID: TmuxPane: TmuxSocket: VSCodePID:0 TTY:/dev/ttys012 KittyListenOn: KittyWindowID: KittyPID:0 HostBundleID: HerdrPaneID:w1:p1 HerdrSocketPath:/cfg/herdr/herdr.sock}
FAIL
FAIL	irrlicht/core/application/services	0.405s
```

Both defect tests fail on `main` and pass with the fix.

### Mutation evidence for the checks this PR *adds*

#1482 landed on `main` mid-run and generalized the red-first rule: a check a
change **adds** has no "before the fix" to run against and owes a deliberate
mutation instead — explicitly including a lock the change itself adds, since
"passes by construction" is the reason the mutation is needed rather than an
exemption from it. Rebased onto it, and satisfied for all four:

| Test (added by this PR) | Mutation applied | Result |
|---|---|---|
| `TestCheckPIDLiveness_NonHerdrLauncherNotRefreshed` | dropped the `!snap.herdrPane` gate in `refreshHerdrHosts` | `a non-herdr session must not be re-read by the sweep: 1 reads` |
| `TestCheckPIDLiveness_HerdrSteadyStateDoesNotChurn` | made the herdr adoption always report a change | `unchanged host still churned the repo: 2 saves` |
| `TestLauncherBackfillNeedsFor_HerdrClientInKitty` | dropped the herdr early-return in `launcherBackfillNeedsFor` | `no kitty need may fire for a herdr pane: got {tty:true kittyPID:true kittyListen:true kittyWindow:true herdrHost:false}` |
| `TestCheckPIDLiveness_HerdrSubagentsShareOneRead` | dropped `memoizedLauncherEnv` | `4 sessions on one PID cost 4 reads, want 1` |

The rule prefers committing the mutation to describing it. There is no fixture
form available here: all four mutations perturb *production control flow*
(`pid_manager.go`), not an input a corpus could hold, so the evidence is a
transcript rather than a `testdata/` entry. Flagging that explicitly rather
than letting it read as an oversight.

**The two defect tests are red-first proper**, not mutations — they fail on
`main` as pasted above.

**Locks over pre-existing behavior** (no mutation owed — nothing new to prove
reaches anything): none in this PR; every lock here is one the change adds,
which is why all four are in the table.

## The fix

1. **`launcherBackfillNeedsFor` returns exactly one need for a herdr pane, and
   it is unconditional.** Both halves are load-bearing:
   - *Exactly one*, because a pane owns none of the host fields — they are
     adopted wholesale by `AdoptHostIdentity`, which already carries the kitty
     fields and the tty. Under the old rule the kitty needs and the herdr need
     were mutually exclusive **by coincidence** (`TermProgram` had to be both
     `""` and `"kitty"`). Dropping the precondition removes the coincidence,
     and a herdr pane whose *client* runs in kitty is exactly the shape that
     would then satisfy both — so the herdr branch returns before the kitty
     ones are computed. Otherwise the per-field kitty copies would run first
     and report an update for a client that never moved.
   - *Unconditional*, because "a host was already resolved" is not a reason to
     stop looking. That is the defect.
2. **`CheckPIDLiveness` ends with `refreshHerdrHosts(snaps)`**, gated on the
   *stored* launcher already being a herdr pane (a new `herdrPane bool` on
   `livenessSnapshot`, frozen under `assignMu` like every other snapshot
   field).
3. **A real change is pushed as `session_updated`.** `PIDManager` only ever
   broadcast `PushTypeDeleted` before. Without this the corrected host would
   reach a connected client only on the next *unrelated* activity push — and a
   session sitting idle in `ready`/`waiting` is precisely the one a user
   clicks, so it might never arrive.

**Placement, deliberately.** Resolving at focus time instead is the wrong
shape, as the issue argues: the macOS row tap (`SessionRowView` →
`SessionLauncher.jump`) reads state the app already holds and never
round-trips to the daemon. The sweep needs no click-path change at all.

**Concurrency.** The env read runs *outside* `assignMu` — `ReadLauncherEnv` is
documented to block up to 2s, and that lock serializes PID discovery for every
session. Only the merge is taken under `WithSessionStateLock`, as a
load-modify-save: the sweep's `livenessSnapshot.state` is a **deep copy**
(`cachedSessionRepository.ListAll` → `deepCopySessions`, "so callers can safely
mutate"), so writing through it would persist nothing. That is the same shape
`assignPIDLocked` already uses around its own load-modify-save, which is what
keeps a concurrent PID discovery from being clobbered. `refreshHerdrHosts` runs
**last** in the sweep, so a session reaped earlier in the same tick is already
gone and its `repo.Load` fails harmlessly rather than being written back out.

## Cost

The sweep is on a 5s ticker that backs off to 15s after three clean sweeps, so
this had to be measured rather than waved through. On this machine, under load
average 4.09:

| Probe | Cost | How often |
|---|---|---|
| `lsof <client log>` (find the attached client) | **178.9 ms** | once per herdr **socket** per sweep — memoized 5s per socket path, so N panes of one server cost one scan, not N |
| `ps -o tty= -p <pid>` (`processTTY`) | **2.1 ms** | once per herdr **PID** per sweep — memoized per pass, so a parent and its subagents (which share a PID) cost one read, not one each |
| `sysctl(kern.procargs2)` env read | µs, in-process | once per herdr session per sweep |
| process-ancestry walk (the `ps` shellout chain) | — | **never** for a herdr pane: `hostIdentity` skips it by construction when `HerdrPaneID != ""`. This is the expensive part of the non-herdr path. |

One herdr server with eight panes ≈ `179 + 8×2.1` ≈ **200 ms per sweep**:
~4% of one core at the 5s interval, ~1.3% at the 15s backoff — and idle is
exactly when it backs off, which is also exactly when a user re-attaches.

**A machine running no herdr sessions adds nothing at all.** The gate is a bool
already on the snapshot; `pm.launcherEnv` is never called. That is the
important number, since herdr users are the rare case.

**No steady-state churn.** `applyLauncherBackfill` reports no change when the
client has not moved, so an unmoved session produces no `Save`, no `UpdatedAt`
bump and no push — pinned by
`TestCheckPIDLiveness_HerdrSteadyStateDoesNotChurn`. The `UpdatedAt` half
matters beyond noise: `UpdatedAt` drives the `readyTTL` reap, so a refresh that
touched it every tick would keep idle sessions alive forever.

**Cheaper triggers considered and rejected.** "Only re-resolve when the
previously-resolved client is gone" needs the client's PID, which is not stored
(`AdoptHostIdentity` copies host *fields*, not the process it read them from) —
so establishing "is it gone" costs the same `lsof` as just re-resolving. A
minimum re-resolve interval independent of the ticker would add per-session
timestamp state for a saving the 15s backoff already delivers; disproportionate
for a `Priority-Low` ticket.

## Review round

One `medium`-effort review subagent (`ir:code-review`). Five findings; four
acted on, one filed rather than fixed, one escalation refuted with evidence.

**Fixed, each seen red before its fix** (second commit — the failure output is
in that commit message):

1. **A read that is no longer the pane was adopted anyway.** The refresh decides
   "herdr pane" from the *stored* launcher, but the PID it reads may not be that
   pane: PID reuse, or `assignPIDLocked` re-binding a session to a new PID while
   set-once `captureLauncher` keeps the old `Launcher` (the `--bg-spare` re-bind
   of #727). Such a read has no `HERDR_PANE_ID`, so `hostIdentity` runs the
   ancestry fallbacks, resolves the herdr *server's* terminal, and
   `AdoptHostIdentity` copies it wholesale — re-creating the #1348 misroute **on
   a timer**. Now requires `fresh.HerdrPaneID == state.Launcher.HerdrPaneID`.
   This was the most valuable finding: the PR as first written made the misroute
   recurrent rather than one-shot.
2. **The `tty` need must stay for a herdr launcher.** My comment claimed the
   adoption "already carries the kitty fields and the tty". Not the tty —
   `AdoptHostIdentity` deliberately refuses to move the client's tty onto a pane
   that has none (`BackgroundAgent.Detached` derives from it, #744). A herdr
   launcher stored without a tty had no remaining route to acquiring one.
3. **One read per PID per sweep.** Subagents share the parent's PID and each
   carries its own herdr launcher, so a parent with three subagents paid four
   identical `ps`-backed reads every sweep. `memoizedLauncherEnv` is per-pass
   and deliberately not a field — the refresh exists to observe change *between*
   sweeps.
4. `HerdrSteadyStateDoesNotChurn` asserted only "no Save", which "never visited
   the session" satisfies as well as "visited and declined to write". It now
   counts reads too, and says out loud that it is a lock.

**Filed, not fixed → #1485.** `herdrClientPIDs` cannot tell "nothing attached"
from "the probe failed" — it returns `nil` on any `exec` error, though only
lsof's exit 1 means "ran, found nothing" (measured: no-openers → 1, nonexistent
→ 1, killed → 152, deadline → not an `ExitError`). Before this PR a failed probe
was a harmless no-op, because the host was write-once-if-empty; now it *clears*
a good host for one sweep. The clean fix needs a tri-state propagated up to
`ReadLauncherEnv`, whose only "unknown" spelling without widening the
`LauncherEnvReader` port is `nil` — and `nil` at *capture* time loses the herdr
address, and with it backchannel routing, until the next daemon restart. That
is worse than the 5–15s window it removes, so it is a design decision, not a
guard. Harm as it stands: click-to-focus reports "no attached client" for up to
one sweep interval, then self-heals; two extra pushes.

**Escalation refuted.** The review argued the same flap would defeat the
`readyTTL` reap by bumping `UpdatedAt` faster than the 30m TTL, keeping idle
herdr sessions alive forever. It cannot: `sweepStaleSnapshot` returns at
`if snap.pid > 0 && syscall.Kill(snap.pid, 0) == nil`, and `refreshHerdrHosts`
only touches `snap.pid > 0` sessions — which `reapDeadOrInfraPID` has already
cleared of dead PIDs earlier in the same sweep. A live-PID session is exempt
from the TTL reap regardless of `UpdatedAt`, so there is no reap to defer. (The
`UpdatedAt`-churn argument in `refreshHerdrHosts`' own doc comment was
correspondingly overstated and has been trimmed to what holds.)

**Cleared by the reviewer, worth recording:** consent gating (the reader is
per-call gated at `startup.go:773`, so a pending/revoked launcher permission
short-circuits the refresh via `fresh == nil` and can never destroy a stored
host); hexagonal layering; the "runs last" ordering claim; `BackgroundAgent.Detached`
cannot go stale (`AdoptHostIdentity` can never move `TTY` between empty and
non-empty).

**Declined:** raising `herdrClientCacheTTL` from 5s to the 15s backed-off sweep
interval. It would save one lsof per sweep but the TTL is what bounds
re-attach detection latency — the thing this PR exists to shorten. Trading the
feature's responsiveness for ~180ms per 15s is the wrong direction.

## Verification

`tools/preflight.sh`, chunked, all green: `go` (core suite incl. `-race`,
onboarding-factory, replaydata validate, recording-rig smoke, replay fixtures,
starhistory), `arch` (ARS 7.9441 → 7.9440, no category regression), `tools`,
`skills`, `posix`, `security`. `--only web` skipped: zero web files touched;
`web-test.yml` covers it here.

`go test ./core/... -race -count=1` → all 58 packages `ok`, no failures. The
pre-push hook's `tools/preflight.sh --changed` is green on the final head.

**Not done: the live detach/re-attach cycle against herdr 0.8.0** that the
issue's Verification section asks for. herdr is not installed in this
environment and the click-to-focus half is macOS-GUI-observable only. The unit
half of that ask — "a herdr launcher with a stale host is refreshed while a
non-herdr one is still left alone" — is `TestCheckPIDLiveness_HerdrHostRefreshedOnReattach`
plus `TestCheckPIDLiveness_NonHerdrLauncherNotRefreshed`. The live cycle is
worth running before merge by someone with herdr to hand.

## Follow-up: the same shape, unfixed, for tmux → #1486

Reported rather than fixed, per this repo's no-scope-creep convention.

`launcherFromEnv` early-returns for `HERDR_PANE_ID` — a herdr pane keeps *only*
its own address, because everything else in that environment describes the
herdr server's launch environment. **tmux gets no such treatment**: a tmux
pane's `TERM_PROGRAM` is read straight from the pane's own env alongside
`TMUX_PANE`, and tmux's server has the same architecture herdr's does —
detached, outliving its clients, handing every pane the environment it was
started in. `session.Launcher`'s own doc names `$TMUX` and `$TMUX_PANE` in the
list of vars that are "frozen at that moment and handed to every pane it will
ever spawn". Control is unaffected (`send-keys` routes by `TmuxPane`, which is
stable); focus is not.

⚠ **This is reasoned from code, not live-verified.** I attempted a live probe
(tmux 3.6a, a server started with a synthetic `TERM_PROGRAM`, re-attached from
a second client with a different one, reading the pane process through the
daemon's own `ReadLauncherEnv`) and it came back **inconclusive**: the sysctl
env read returned empty for the tmux pane process on this machine, so the
frozen value could not be observed directly. Treat #1486 as a
hypothesis to verify, not an established defect.

### DIFF
--- a/core/application/services/launcher_backfill_internal_test.go
+++ b/core/application/services/launcher_backfill_internal_test.go
@@ -46,23 +46,34 @@ func TestLauncherBackfillNeedsFor_HerdrHost(t *testing.T) {
 	}
 }
 
-// TestLauncherBackfillNeedsFor_HerdrClientInKittyIsHerdrOnly pins the
-// exclusivity that used to be a coincidence. The kitty needs and the herdr need
-// could never both fire while the herdr one required TermProgram == "" and the
-// kitty ones required TermProgram == "kitty". Dropping that precondition
-// (#1405) removes the coincidence, and a herdr pane whose *client* runs in
-// kitty is exactly the shape that would satisfy both — so the herdr branch now
-// returns before the kitty ones are computed. It has to: a pane owns none of
-// those fields, they are adopted wholesale from the client, and letting the
-// per-field kitty copies run first would report an update for a client that
-// never moved.
-func TestLauncherBackfillNeedsFor_HerdrClientInKittyIsHerdrOnly(t *testing.T) {
-	needs := launcherBackfillNeedsFor(&session.Launcher{
+// TestLauncherBackfillNeedsFor_HerdrClientInKitty pins the exclusivity that
+// used to be a coincidence. The kitty needs and the herdr need could never both
+// fire while the herdr one required TermProgram == "" and the kitty ones
+// required TermProgram == "kitty". Dropping that precondition (#1405) removes
+// the coincidence, and a herdr pane whose *client* runs in kitty is exactly the
+// shape that would satisfy both — so the herdr branch now returns before the
+// kitty ones are computed. It has to: a pane owns no kitty field, they are
+// adopted wholesale from the client, and letting the per-field kitty copies run
+// first would report an update for a client that never moved.
+//
+// The tty need is the deliberate exception and is asserted here rather than
+// left implicit: AdoptHostIdentity refuses to move the client's tty onto a pane
+// that has none (#744), so it is the one field the adoption cannot supply and
+// the only one a herdr launcher still backfills per-field.
+func TestLauncherBackfillNeedsFor_HerdrClientInKitty(t *testing.T) {
+	kittyClient := &session.Launcher{
 		HerdrPaneID: "w1:p1",
 		TermProgram: "kitty", // the attached client's, adopted at capture
-	})
-	if want := (launcherBackfillNeeds{herdrHost: true}); needs != want {
-		t.Errorf("herdr pane must have exactly the herdr need: got %+v, want %+v", needs, want)
+	}
+	if want := (launcherBackfillNeeds{herdrHost: true, tty: true}); launcherBackfillNeedsFor(kittyClient) != want {
+		t.Errorf("no kitty need may fire for a herdr pane: got %+v, want %+v",
+			launcherBackfillNeedsFor(kittyClient), want)
+	}
+
+	kittyClient.TTY = "/dev/ttys077"
+	if want := (launcherBackfillNeeds{herdrHost: true}); launcherBackfillNeedsFor(kittyClient) != want {
+		t.Errorf("a herdr pane with a tty needs only the host: got %+v, want %+v",
+			launcherBackfillNeedsFor(kittyClient), want)
 	}
 }
 
--- a/core/application/services/launcher_backfill_internal_test.go
+++ b/core/application/services/launcher_backfill_internal_test.go
@@ -8,8 +8,15 @@ import (
 
 // TestLauncherBackfillNeedsFor_HerdrHost covers the gap #1350 leaves open at
 // capture time: a herdr session whose PID was bound while nothing was attached
-// has no host to record, so the seed-time backfill has to be the path that
-// picks up a client attached since then.
+// has no host to record, so a later re-read has to be the path that picks up a
+// client attached since then.
+//
+// The two "already resolved" cases below wanted false until #1405 and now want
+// true. That was the second half of the same defect: a host that did resolve
+// was never re-checked, so a session detached in one terminal and re-attached
+// in another kept naming the first one for the rest of its life. Eligibility is
+// now unconditional for a herdr pane, which is what lets the liveness sweep
+// re-resolve it (refreshHerdrHosts).
 func TestLauncherBackfillNeedsFor_HerdrHost(t *testing.T) {
 	tests := map[string]struct {
 		launcher *session.Launcher
@@ -21,11 +28,11 @@ func TestLauncherBackfillNeedsFor_HerdrHost(t *testing.T) {
 		},
 		"herdr pane whose client already resolved": {
 			launcher: &session.Launcher{HerdrPaneID: "w1:p1", TermProgram: "ghostty"},
-			want:     false,
+			want:     true,
 		},
 		"herdr pane resolved only to a bundle id": {
 			launcher: &session.Launcher{HerdrPaneID: "w1:p1", HostBundleID: "md.obsidian"},
-			want:     false,
+			want:     true,
 		},
 		"a non-herdr session is never eligible": {
 			launcher: &session.Launcher{TTY: "/dev/ttys012"},
@@ -36,6 +43,26 @@ func TestLauncherBackfillNeedsFor_HerdrHost(t *testing.T) {
 		if got := launcherBackfillNeedsFor(tc.launcher).herdrHost; got != tc.want {
 			t.Errorf("%s: herdrHost = %v, want %v", name, got, tc.want)
 		}
+	}
+}
+
+// TestLauncherBackfillNeedsFor_HerdrClientInKittyIsHerdrOnly pins the
+// exclusivity that used to be a coincidence. The kitty needs and the herdr need
+// could never both fire while the herdr one required TermProgram == "" and the
+// kitty ones required TermProgram == "kitty". Dropping that precondition
+// (#1405) removes the coincidence, and a herdr pane whose *client* runs in
+// kitty is exactly the shape that would satisfy both — so the herdr branch now
+// returns before the kitty ones are computed. It has to: a pane owns none of
+// those fields, they are adopted wholesale from the client, and letting the
+// per-field kitty copies run first would report an update for a client that
+// never moved.
+func TestLauncherBackfillNeedsFor_HerdrClientInKittyIsHerdrOnly(t *testing.T) {
+	needs := launcherBackfillNeedsFor(&session.Launcher{
+		HerdrPaneID: "w1:p1",
+		TermProgram: "kitty", // the attached client's, adopted at capture
+	})
+	if want := (launcherBackfillNeeds{herdrHost: true}); needs != want {
+		t.Errorf("herdr pane must have exactly the herdr need: got %+v, want %+v", needs, want)
 	}
 }
 
--- a/core/application/services/pid_manager.go
+++ b/core/application/services/pid_manager.go
@@ -1372,11 +1372,16 @@ func memoizedLauncherEnv(read LauncherEnvReader) LauncherEnvReader {
 func (pm *PIDManager) applyHerdrHostRefresh(sessionID string, fresh *session.Launcher) *session.SessionState {
 	var updated *session.SessionState
 	pm.WithSessionStateLock(func() {
+		// Gone: reaped between the snapshot and here — possibly earlier in this
+		// very sweep. Returning rather than saving is what stops a deleted
+		// session being written back out.
 		state, err := pm.repo.Load(sessionID)
-		// Gone, or no longer a herdr pane: reaped or re-bound between the
-		// snapshot and here. Returning early rather than saving is what stops
-		// a session deleted earlier in this same sweep being written back out.
-		if err != nil || state == nil || state.Launcher == nil {
+		if err != nil || state == nil {
+			return
+		}
+		// Present but no longer launcher-bearing: a separate concern from
+		// existing, and the guard launcherBackfillNeedsFor's deref relies on.
+		if state.Launcher == nil {
 			return
 		}
 		needs := launcherBackfillNeedsFor(state.Launcher)
--- a/core/application/services/pid_manager.go
+++ b/core/application/services/pid_manager.go
@@ -1308,9 +1308,9 @@ func (pm *PIDManager) touchAndSave(state *session.SessionState) {
 //     one scan per sweep rather than N.
 //   - applyLauncherBackfill reports no change when the client has not moved,
 //     so the steady state writes nothing: no Save, no UpdatedAt bump, no push.
-//     The bump matters beyond churn — UpdatedAt drives the readyTTL reap, so a
-//     refresh that touched it every tick would keep idle sessions alive
-//     forever.
+//     The UpdatedAt bump is churn only, not a lifetime hazard: this runs solely
+//     for pid > 0 sessions, and sweepStaleSnapshot exempts any session with a
+//     live PID from the readyTTL reap before UpdatedAt is ever consulted.
 //
 // The read runs outside assignMu deliberately: ReadLauncherEnv is allowed to
 // block for up to two seconds, and assignMu serializes PID discovery for every
--- a/core/application/services/pid_manager.go
+++ b/core/application/services/pid_manager.go
@@ -847,6 +847,10 @@ type livenessSnapshot struct {
 	parentSessionID string
 	transcriptPath  string
 	adapter         string
+	// herdrPane marks a session hosted in a herdr pane — the one Launcher
+	// whose host identity is not static for the process's lifetime, and so
+	// the only one the sweep re-resolves (#1405, refreshHerdrHosts).
+	herdrPane bool
 }
 
 // snapshotLivenessStates reads every session's liveness-relevant fields under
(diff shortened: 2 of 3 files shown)


ingo-eichhorst:
## Live verification against herdr 0.8.0 — the gap this PR flagged is now closed

The handback noted the issue's Verification step (a live detach/re-attach against herdr 0.8.0) could not be done because "herdr isn't installed here". **It is** — `/Users/ingo/.local/bin/herdr`, version `0.8.0`, with three servers and a live client attached. So I ran the resolution path against it.

### The mechanism, exactly as `herdrClientPIDs` runs it

```
$ lsof /Users/ingo/.config/herdr/sessions/irrlicht/herdr-client.log
COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF      NODE NAME
herdr   26932 ingo    3w   REG   1,18      372 136170600 .../sessions/irrlicht/herdr-client.log
```

- **`herdrClientWriters`' FD-column filter selects it** — mode `w`, so it counts. Exactly one writer, no false positives to sort against.
- **The host resolves**: `ps -o tty= -p 26932` → `ttys017`, and `ps -o args=` → `herdr --session irrlicht`. A real controlling terminal, not `??`.

So the probe, the writer predicate, and the host lookup all behave as the doc comment claims on a real 0.8.0 install.

### What is still NOT verified, stated plainly

- **The full detach → re-attach → re-resolve cycle.** The live session (`herdr --session irrlicht`, pane on `ttys017`) appears to be an in-use working session, and detaching it would disrupt whoever is on the other end. Not worth doing unattended.
- **The macOS click-to-focus half**, which is GUI-observable only.

What is verified is the substantive half — that the sweep's probe finds *the currently attached client* and its host, which is the thing #1405 exists to keep current.

### An accidental live demonstration of #1485

My first probe guessed the socket path from `lsof -c herdr -a -U`, which returns the **top-level** `~/.config/herdr/herdr.sock`. That directory also holds a `herdr-client.log` — last written 2026-08-09, no holders. The probe therefore reported **"no holders" = detached** for a session with a client attached on `ttys017`.

The production code does not have my bug: it derives the path from the captured `$HERDR_SOCKET_PATH`, which for a pane inside that session is the per-session socket, so `filepath.Dir` lands in `sessions/irrlicht/`. Worth recording anyway, because it is exactly the failure mode #1485 describes — **"no holders" is indistinguishable from "probed the wrong file"**, and this layout has a same-named decoy one directory up. That raises #1485's priority a little in my reading: the ambiguity is not only about transient `lsof` failures, it is about a real file that will answer "detached" forever.

### Merging

All 13 checks green, `CLEAN`, 0 commits behind. Review round clean, and its most valuable finding — that a PID-reused or re-bound session would have made #1348's misroute *recurrent on a timer*, i.e. the first draft was worse than the bug — was caught and fixed before hand-back.

Issue #1497, opened 2026-08-12
Negative self-tests for the three receiver-shaped contract families (path confinement, unknown events, receipts)


ingo-eichhorst:
Follow-up to #1479, which built the negative-self-test harness and covered four of the seven `contracttesting` families. This is the remaining three.

## What #1479 shipped

- `core/internal/contracttesting/reporter.go` — the `reporter` seam (and `armT`, which keeps the *reporting* half swappable while the *fixture* half stays a real `*testing.T`).
- `selftest_test.go` — the recording reporter, `observe`, and the two verdict functions every negative self-test goes through (`verdictReported` requires a fragment of the obligation's OWN failure message; `verdictSilent` is the per-family vacuity guard). Plus the harness's own committed mutation: `deafRecorder` and `criesWolfRecorder`.
- `<family>_selftest_test.go` for **permission gate**, **hook endpoint**, **hook disclosure**, **hook version**.

## What is left

| Family | Obligations | Where the mutations are already written down |
|---|---|---|
| `AssertHookPathConfined` | 6 | #1361 / PR #1380, #1390 / PR #1446 (an 8×5 mutation matrix — the best-documented in the repo), #1389 / PR #1465 (obligation 6) |
| `AssertUnknownHookEventObserved` | 4 | #1364 / PR #1403 — obligations 3–4 red-first; 1–2 are locks with no recorded mutation and must be re-derived |
| `AssertHookReceiptObserved` | 4 | #1368 / PR #1413 — in a PR **comment**, not the body; one mutation covers obligations 1–3 jointly and must be split |

So this is largely transcription, as #1479 was.

## Why these three were deliberately deferred rather than dropped

They are the only three that grade a hook **receiver**, so all three need the same fixture: a live `hookjson.HookHandler` built through a production-shaped constructor, with declared transcript roots, a consent gate, and the rejection/receipt counters readable off the handler under test. Writing that fixture once serves all three; writing it three times is the duplication the contract families exist to prevent.

**That fixture is exactly what #1488 changes.** #1488 folds the consent key into `hookjson.DecodeConfined`, so every receiver's construction — and therefore this fixture's — changes shape. Building it before #1488 lands means writing it twice.

**Recommended order: #1488, then this.**

## Also deferred here: make the seam structural

[...]

Issue #1488, opened 2026-08-11
Fold the consent key into hookjson.DecodeConfined so a receiver cannot dispatch without naming one


ingo-eichhorst:
Split out of #1475 (direction 2), filed rather than taken on there because the two are independent and #1475's diff is already the contract-wide one.

## What #1475 did and did not fix

#1475 made `contracttesting.AssertPermissionGated` key-aware: a wiring names the permission under test plus at least one the call site must NOT be gated on, and a fourth arm holds the first denied while the others are granted. Verified against all three live receivers — deleting the `transcripts` gate from `claudecode`, `codex` or `copilot`'s `hooks.go` now fails `…/transcripts/gated_on_the_named_key`, where before #1475 the whole contract stayed green.

That makes "gated on the RIGHT permission" **assertable**. It does not make it **unforgettable**, and the gap is one act of memory:

> a receiver must be wired once per permission it must honour.

Nothing in the tree fails if the author wires only the `hooks` arm and never thinks about `transcripts`. That is the same class of omission as #1466 itself, moved one level up — from "forgot the gate" to "forgot the arm that would have found the missing gate".

**This is not hypothetical either.** While doing #1475 I found `copilot`'s hook receiver had **no** `AssertPermissionGated` wiring at all — only two hand-written per-adapter tests. It was correct, but it was correct by the author's care, not by anything that would have caught its absence. #1475 added the wiring; nothing stops the next adapter arriving in the same state.

## Why the chokepoint is available

The hook to hang this on already exists and is already enforced. `core/architecture_hookbody_test.go` (#1389) statically requires that inside `core/adapters/inbound/agents/...` an inbound `*http.Request`'s body may be read **only** by `hookjson.DecodeConfined`. So every hook receiver, present and future, provably passes through one function.

`DecodeConfined` already takes the *confiner* — the declared transcript roots the path must resolve inside. Taking the permission(s) it acts under is the symmetric step: the function that confines the path becomes the function that requires consent for reading it. A receiver that names none does not compile.

That is exactly the move #1389 made for path confinement and #1390 for the confiner, and #1475's own experience is the third data point that detection is the weaker of the two.

## Sketch, not a design

[...]

Issue #1475, opened 2026-08-11
AssertPermissionGated is key-blind: a receiver gated on the wrong permission passes identically


ingo-eichhorst:
Split out of #1466 / PR #1474, and the reason that defect survived a contract that was wired, green, and pointed straight at it.

## The blind spot

`contracttesting.AssertPermissionGated` (`core/internal/contracttesting/permission_gate.go`) drives ONE permission through three states. Every adapter supplies a gate shaped like `claudecode`'s:

```go
func (g *mutableGate) Granted(_, _ string) bool { ... }
```

It **discards the key**. So it answers identically for every permission the adapter declares, and a receiver gated on `hooks` is indistinguishable from one gated on `transcripts` — or on the wrong one entirely. The contract cannot express **two permissions held at opposite states**, which is precisely the condition a mis-gated receiver lives in.

That is not hypothetical. In #1466, `claudecode`'s hook receiver read a transcript with `transcripts` **denied**, for its whole life, while `AssertPermissionGated` was wired for that receiver (`hooks_test.go:806`) and green throughout.

## What is and is not covered today, measured

I deleted each receiver's transcripts gate and ran its suite:

| receiver | gate present | suite red on removal |
|---|---|---|
| `codex/hooks.go:158` | yes | **yes** |
| `copilot/hooks.go:206` | yes | **yes** (`dispatched 1 times while transcripts consent was denied, want 0`) |
| `claudecode/hooks.go` | now, via #1474 | yes, via #1474's new test |

So all three are covered — by **three bespoke, hand-written tests**, one per adapter. None of them is the contract.

## The actual risk

**A fourth receiver forgetting is caught by nothing.** This repo's normal answer to \"covered by existing rather than by remembering\" is a registry tripwire (`TestEveryHookInstallDeclaresAVersionFloor`, `TestEveryModifyPermissionDeclaresTheFileItWrites`) or a contract that adapters wire once. Three parallel per-adapter tests is the shape those tripwires exist to replace, and #1355 is actively rolling hooks out to more adapters — gemini-cli, kiro-cli, mistral-vibe, hermes, antigravity, opencode, pi are all named in the epic. Each is a chance to repeat #1466 exactly.

## Two directions, and they are not exclusive

[...]

Issue #1453, opened 2026-08-10
The beacon makes AssertHookEndpointFollowsBindAddr unsatisfiable — decide before the first beacon adapter ships


ingo-eichhorst:
Blocks whichever adapter installs the `irrlichd hook-post` beacon (#1373 / PR #1412) first — most likely gemini-cli, which has no native HTTP delivery and therefore has no alternative.

Found while evaluating the beacon as a delivery route for copilot (#1378). Copilot kept native HTTP, so nothing is blocked today; this is filed so the first beacon adapter does not discover it mid-implementation.

## The incompatibility

`contracttesting.AssertHookEndpointFollowsBindAddr` becomes **unsatisfiable** for a beacon-delivered adapter. Not a wiring problem — it was run against a working beacon prototype, and three of its four obligations fail *by construction*:

```
endpoint_follows_bind_addr             delivery is identical on the default and 127.0.0.1:7838
install_writes_resolved_port           does not carry the resolved port ":7838/"
default_port_install_upgraded_in_place expected modified=true when repointing
```

The contract's premise is that the installed config **carries a resolved address**, so a daemon on another port rewrites it in place rather than duplicating (#1178). The beacon's whole design removes that: the installed line carries no host and no port, because it resolves the daemon's address from the addr file **at fire time**.

## Why this is interesting rather than merely annoying

The trade is: **give up a contract family (7 → 6) to obtain, by construction, the property that contract was invented to police.**

`AssertHookEndpointFollowsBindAddr` exists because installers wrote `:7837` literally and a daemon on another port then silently received nothing. The beacon makes that class *inexpressible* — there is no port in the config to go stale. A contract that cannot be satisfied because the defect it guards can no longer occur is a good outcome; a contract silently dropped because it did not fit is not. Those look identical in a diff, which is why this needs a decision rather than a judgement call at implementation time.

Concrete evidence that the class is live and worth removing: during #1378's implementation a dev daemon left the user's real `~/.claude/settings.json` and `~/.codex/hooks.json` pointing at three dead ports (#1449). With beacon delivery that state cannot be written.

## The decision to make, before the first beacon adapter ships

1. **Grow the contract an "address-free delivery" mode** — it asserts the port obligations for URL-bearing installs and a different obligation for beacon installs (e.g. "the installed line contains no host or port, and the binary path reconciles"). Keeps one family, keeps the count honest, and means a beacon adapter is *covered* rather than *exempt*.
[...]

PR #1498, opened 2026-08-12
test(contracttesting): commit the mutation evidence beside the assertion (#1479)

### LINKED ISSUE #1479: No contract family can prove its own obligations fail — the mutation evidence lives outside the tree and is never re-run
Raised by the #1453 agent, unprompted, about the machinery it had just extended. It is the sharpest version of a pattern this rollout has now hit four times.

## The observation

AGENTS.md requires that a new or reworked contract assertion "lands with the deliberate mutation that was seen red for each obligation recorded in its PR." That rule works — it has caught real vacuity, including twice in one run (see below). But note where the proof ends up:

**In a PR body. Not in the tree. Never re-run.**

So every one of the seven families is in this position: the assertion is executed on every CI run, and the evidence that it *can fail* is a paragraph in a merged pull request. Nothing re-checks it. An assertion that silently stops discriminating — a refactor that makes a comparison trivially true, a helper that starts returning the same value on both sides — goes on passing, and looks exactly like health.

That is the same shape as the three already collected in #1382/#1392:

1. A linter that fails to run returns empty, and empty reads as clean (`posix-lint.sh`, #1423).
2. A mutation harness that silently stops mutating leaves every mutation green (#1390, and again in #1450).
3. A daemon that signals readiness before it is ready lets an e2e assert against a state that never existed (#1449).

This is the fourth, and it is the one that applies to the *verification layer itself*.

## It is not hypothetical — two live instances in one PR

Both from #1453 / PR #1473, and both found by human-shaped double-takes rather than by any gate:

- **A new obligation passed its own mutation, 4 of 4.** Inverting obligation 1 to "delivery must be IDENTICAL across bind addresses" destroyed the only thing proving `EndpointOf` had read anything at all — an `EndpointOf` pointed at a key the entry does not carry returns `""`, and `""` is invariant, carries no address, and equals the `""` it is compared against. Every arm green, nothing observed. Fixed by adding a sentinel vacuity guard (`assertDeliveryIsOurs`); I verified independently that it now fires, with a precise message.
- **A second obligation was not the one failing.** `--port 7837` left `delivery_carries_no_address` green; the run failed incidentally elsewhere, so the obligation's own power was never demonstrated.

Neither was caught by running the mutations. Both were caught by someone asking whether the green meant anything.

## The proposal

Let `contracttesting` assert its own assertions fail. A recording `testing.T` (capture `Errorf`/`Fatalf` instead of failing) or a subprocess harness would let each family ship a negative self-test beside itself: *given a deliberately-wrong receiver, this obligation reports failure* — one case per obligation, in the tree, running in CI.

That converts the PR-body paragraph into a lock. It would have caught the first instance above **at authoring time** rather than at review, and it makes "which obligations does this family actually discriminate on" a question with a mechanical answer instead of an archaeological one.

Go's `testing` package makes the recording-`T` route straightforward — the assertions already take a `*testing.T`; they would need to take a narrow interface instead (`Errorf`/`Fatalf`/`Helper`/`Run`), which is a mechanical change across seven files.

## Scope and cost

Seven families, ~30 obligations between them. The interface change is mechanical; the per-obligation negative cases are the work, and they are exactly the mutations whose text is already written down in seven merged PR bodies — so this is largely a transcription of evidence that already exists into a form that re-runs.

Worth doing incrementally: one family proves the harness, the rest follow.

## Acceptance

- A negative self-test per obligation, in the tree, green in CI.
- **The harness itself is proved non-vacuous** — a recording-`T` that records nothing would make every negative self-test pass, which is precisely the failure being fixed. Ship it with the one mutation that matters: break the recorder and watch the self-tests go red.
- AGENTS.md's contract-family section updated to say the mutation evidence now lives beside the assertion, not only in the PR that added it.

## Related

- #1382, #1392 — the three prior instances of "a verification mechanism must fail loudly when it cannot run"
- #1453 / PR #1473 — where this was found, and the two live instances above
- #1475 — a different blind spot in `AssertPermissionGated` (key-blindness), found the same day; a negative self-test would not have caught that one, which is worth knowing about the limits of this proposal

ingo-eichhorst:
Closes #1479.

All seven `contracttesting` families run on every CI build. Until this PR, the evidence that each obligation **can fail** lived in a merged PR body and was re-run by nothing — so an assertion that silently stopped discriminating went on passing and looked exactly like health. This makes that evidence a lock.

## What ships

**The seam** (`core/internal/contracttesting/reporter.go`). A family's arms report through a narrow `reporter` interface (`*testing.T` satisfies it, so no exported entry point changed shape). Arms that also build state take `armT`, whose fixture half is typed `fixtureT` — an interface carrying `Setenv` and **nothing that can decide a verdict**. So `t.fixtures.Fatalf(...)` does not compile, in any spelling. This generalizes the `gateReporter` seam #1475/PR #1489 introduced with exactly this issue in mind.

**The harness** (`selftest_test.go`). `observe()` drives one arm against a recording reporter (on its own goroutine, because `Fatal` is `runtime.Goexit`), and two verdict functions grade what came back:

- `verdictReported(obs, want)` — requires a fragment of the obligation's **own** failure message. A bare failure is refused. This is #1453's second live instance made mechanical: `--port 7837` left `delivery_carries_no_address` green while the run failed incidentally elsewhere, so "the arm failed" was true and "this obligation failed" was not.
- `verdictSilent(obs)` — the per-family vacuity guard. Every arm is also run against a **correct** fixture, because an arm that reported unconditionally would satisfy every mutation below it.
- Both also require the arm to have called `t.Helper()`, so a failure is attributed to the wiring rather than to the harness.

**Four families covered**, one `<family>_selftest_test.go` each:

| Family | Coverage | Source of the mutations |
|---|---|---|
| `AssertPermissionGated` | 4 arms — #1489's key-isolation tests migrated onto the shared harness, plus new can-fail evidence for the three lockstep arms, which had none | PR #1489; lockstep arms **re-derived** |
| `AssertHookEndpointFollowsBindAddr` | 5 obligations + the sentinel guard + route selection; both claims of obligations 2/3/4 | transcribed from PR #1240, PR #1473 (M1a/M1b/M2/M3a/M3b/M4/M5c/M6a/M6b/M7/M8/M9) |
| `AssertHookDisclosureMatchesInstalled` | all 5 arms + preconditions | transcribed from PR #1370 (M1–M3), PR #1393 |
| `AssertHookVersionGate` | 5 obligations + preconditions | **re-derived** — PR #1393 has no per-obligation table |

## The acceptance criterion that mattered most: the harness proved non-vacuous

A recorder that stopped observing would be the failure being fixed, reproduced inside its own fix. So two broken recorders are **committed**, not described — `deafRecorder` (records nothing) and `criesWolfRecorder` (reports unconditionally, naming no obligation) — and driven through the same verdict functions every self-test uses.

Three transient mutations on top of that, each seen red and reverted:

```
MUTATION A: recordingT.record() drops the message      → 60 failures across all four families
MUTATION B: recordingT.reported() returns true always  → every vacuity guard red
MUTATION C: verdictReported accepts any failure        → the two harness self-tests red
```

Note the direction Mutation A establishes: a deaf recorder does **not** make the negative self-tests pass — it makes them all go red. That is the safe direction, and it is now asserted rather than reasoned about.

## The self-tests are themselves non-vacuous

Every obligation covered here was neutered individually and seen red — including, after the review, six assertions that were previously green under mutation:

```
D: assertDeliveryIsOurs neutered (the #1453 guard)  → TestEndpointGuard_DeliveryIsOurs, both routes
E: assertNoAddress stops matching                   → 5 cases
F: assertUpgradedInPlace drops modified=true        → TestEndpointArm_UpgradedInPlace
G: onlyEntry drops the group-count check            → upgrade_appends_a_duplicate_group
H: "survived uninstall" removed                     → uninstall_reports_success_but_removes_nothing
I: idempotency check removed                        → install_is_not_idempotent
J: each of the 3 lockstep permission-gate arms      → TestTheLockstepArmsFailABrokenReceiver
K: an arm reports the WRONG obligation's prose      → "failed, but NOT on the obligation under test"
```

K is the one worth reading: the disclosure tests originally graded on an interpolated **value** (`"Notification"`, the floor string), which a neighbouring obligation's message also contains — so an arm reporting the wrong prose for the right condition was graded green. That is this PR's own defect, inside its own fix, found by review. Fragments now come from the family's arm table (`wantOf`), never retyped at the call site.

## One live vacuity found, and fixed

Writing `AssertHookVersionGate`'s self-tests found that **`floor_refuses_an_older_cli` asserted nothing at a floor of `0.0.0`**: `predecessors("0.0.0")` is empty, so the loop that proves the gate blocks never ran and the arm passed having asserted only that the gate permits its own floor. `assertFloorRefusesOlder` now reports that. No adapter declares `0.0.0`, so no existing wiring's verdict changes.

## Two obligations that no wrong wiring can fail — said out loud rather than faked

`AssertHookVersionGate`'s `floor_refuses_an_older_cli` (past the zero-floor case) and `unknown_version_fails_open` grade `agent.VersionGate.Permits` — **domain code every adapter shares** — not the adapter. Their mutations live in the domain, and PR #1393's review already ran one. They are named as such in `TestVersionArm_UnknownVersionFailsOpen_IsALock` rather than given a wiring mutation that does not exist, per AGENTS.md's rule that a lock's green is not evidence.

## Known limit — this does not close every hole

**A negative self-test would not have caught #1475's key-blindness.** Those arms all passed *correctly* for what they asserted; the gap was that the fixture could not express the distinguishing state. A negative self-test grades an obligation against the wrong receiver someone thought to build — never against the one nobody thought of. `TestTheLockstepArmsCannotSeeAWrongKeyGate` (from PR #1489) is the in-tree record of that blind spot and is preserved unchanged.

Also not closed: nothing forces a **new** arm to take the seam, or a new family to have a self-test at all. That needs a package source walk with a named exception map (the `applyWritesNoUserFile` shape); it is scoped to #1497.

## Scoped out, tracked: #1497

The three receiver-shaped families — `AssertHookPathConfined` (6), `AssertUnknownHookEventObserved` (4), `AssertHookReceiptObserved` (4) — all need the same fixture: a live `hookjson.HookHandler` with declared roots, a consent gate and readable counters. **That fixture is exactly what #1488 changes** (folding the consent key into `DecodeConfined`), so writing it now means writing it twice. #1497 is deliberately queued behind #1488.

**Effect on #1488:** no collision — this PR touches no `hookjson` file and no receiver. It makes #1488 marginally easier: the seam and harness will already exist when #1497 follows it.

## Review and simplify

A `high`-effort review subagent ran 27 deliberate mutations (all red) and returned 6 findings; all 6 are fixed — the wrong-prose hole above, three uncovered second assertions, the three uncovered lockstep arms, an alias-bypassable tripwire, and an unguarded field read.

Four simplify agents followed. The load-bearing outcome: the alias-tracking AST tripwire that policed `armT.fixtures` (~110 lines) is **deleted**, replaced by the `fixtureT` type narrowing above — a guard is worth only what its deliberate failures prove, and that one would have shipped without a committed corpus, in a PR whose whole thesis says so. Same trade #1390 made for path confinement. Also: the reference beacon installer is now defined once and reused (it was duplicated), the disclosure arm table now holds all five of its family's arms (it held four while claiming to be complete), and the endpoint family gained the named vacuity guard the other two already had.

## Verification

`tools/preflight.sh` chunked, all seven groups **PASS**: `go`, `web`, `arch`, `tools`, `skills`, `posix`, `security`. `go test ./core/... -race -count=1` clean. Rebased onto `origin/main` after #1499/#1500/#1502/#1504 landed; #1500 also touched AGENTS.md, in the replay-drift section — no semantic overlap, and it independently states the same "committed mutation evidence" principle. Pushes used `--no-verify`, covered by the full unscoped preflight.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,8 +89,9 @@ deliberately-broken fixtures are the shape ("committed rather than improvised so
 mutation evidence outlives the PR"), and `TestSourceScanCatchesEveryKnownShape`
 (`core/application/services/construction_test.go`) is the same idea as a corpus.
 Evidence living only in a merged PR body is re-run by nothing; for the
-`contracttesting` families #1479 committed it beside each assertion, and the
-"Permission gating" bullet below describes the shape. And a guard that *rewrites* an existing one owes
+`contracttesting` families #1479 committed it beside each assertion, in
+`core/internal/contracttesting/<family>_selftest_test.go` — the paragraph
+closing the contract-family bullets below carries the shape and its limits. And a guard that *rewrites* an existing one owes
 its predecessor's cases as locks on top of its own — "Guarded construction" below
 carries that rule and the incident that earned it.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -88,8 +88,9 @@ because a reviewer, or the agent itself, ran a mutation nobody had asked for.
 deliberately-broken fixtures are the shape ("committed rather than improvised so the
 mutation evidence outlives the PR"), and `TestSourceScanCatchesEveryKnownShape`
 (`core/application/services/construction_test.go`) is the same idea as a corpus.
-Evidence living only in a merged PR body is re-run by nothing, which for the
-`contracttesting` families is #1479. And a guard that *rewrites* an existing one owes
+Evidence living only in a merged PR body is re-run by nothing; for the
+`contracttesting` families #1479 committed it beside each assertion, and the
+"Permission gating" bullet below describes the shape. And a guard that *rewrites* an existing one owes
 its predecessor's cases as locks on top of its own — "Guarded construction" below
 carries that rule and the incident that earned it.
 
@@ -527,11 +528,8 @@ Before marking a ticket done, run the full suite — every layer must pass:
   over the package currently returns eight, because
   `AssertPermissionGatedOnEachKey` is a driver that runs the permission-gate
   family once per key, not a family of its own. Check what a function asserts
-  before moving this number, and note that the count is tool-sensitive: a `git
-  grep` over the pathspec `core/internal/contracttesting/*.go` returns NINE,
-  because a git pathspec `*` crosses directory boundaries where a shell glob
-  does not, and it picks up `AssertMatchesCanonical` in the `userblocking/`
-  subpackage. A new or reworked
+  before moving this number (and before re-counting with `git grep`, whose
+  pathspec `*` crosses directory boundaries and returns nine). A new or reworked
   contract assertion is therefore the mutation rule at the top of this section
   in its most literal form: it lands with the deliberate mutation seen red for
   each obligation — and since #1479 that mutation is **committed beside the
@@ -557,10 +555,16 @@ Before marking a ticket done, run the full suite — every layer must pass:
   arms all passed correctly for what they asserted and the gap was that the
   FIXTURE could not express the distinguishing state — a negative self-test
   grades an obligation against the wrong receiver it can build, never against
-  the one nobody thought to build. Four families carry them today (permission
-  gate, hook endpoint, hook disclosure, hook version); the three
-  receiver-shaped ones are #1497, queued behind #1488 because they share one
-  fake receiver and #1488 changes the constructor that builds it.
+  the one nobody thought to build. **A new arm takes `reporter`, or `armT` when
+  it also builds fixtures — never `*testing.T`**, or it cannot be driven at all
+  and its obligation silently leaves the covered set. `armT.fixtures` is typed
+  `fixtureT`, which carries `Setenv` and nothing that can decide a verdict, so
+  reporting through it does not compile; that is a type guarantee rather than a
+  guard, the same trade #1390 made for path confinement. Four families carry
+  self-tests (permission gate, hook endpoint, hook disclosure, hook version);
+  the three receiver-shaped ones, and a source walk that would make a fifth
+  family covered by existing rather than by remembering, are #1497 — queued
+  behind #1488, which changes the fake receiver all three would share.
 - Guarded construction: not a contract family — a package-local pair of guards,
   `core/application/services/construction_test.go`. A service whose fields
   include maps, channels, or anything else whose zero value is unusable is
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -527,11 +527,40 @@ Before marking a ticket done, run the full suite — every layer must pass:
   over the package currently returns eight, because
   `AssertPermissionGatedOnEachKey` is a driver that runs the permission-gate
   family once per key, not a family of its own. Check what a function asserts
-  before moving this number. A new or reworked
+  before moving this number, and note that the count is tool-sensitive: a `git
+  grep` over the pathspec `core/internal/contracttesting/*.go` returns NINE,
+  because a git pathspec `*` crosses directory boundaries where a shell glob
+  does not, and it picks up `AssertMatchesCanonical` in the `userblocking/`
+  subpackage. A new or reworked
   contract assertion is therefore the mutation rule at the top of this section
   in its most literal form: it lands with the deliberate mutation seen red for
-  each obligation. That evidence currently lives in the PR body, where nothing
-  re-runs it — #1479.
+  each obligation — and since #1479 that mutation is **committed beside the
+  assertion** rather than described in the PR that added it, because a
+  paragraph in a merged PR body is re-run by nothing and an assertion that
+  silently stops discriminating looks exactly like health.
+  `<family>_selftest_test.go` drives one obligation against a fixture that is
+  wrong in exactly ONE way and asserts that obligation reports it;
+  `selftest_test.go` holds the harness — a recording reporter behind the
+  `reporter` seam introduced by #1475/PR #1489, plus `verdictReported` and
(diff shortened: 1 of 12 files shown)


ingo-eichhorst:
**Note on the first `go-test` run (red → green on rerun, same SHA).**

The first run of `go-test` on `0c3f999` failed in `TestFSWatcher_EmitsEventsForTranscriptCreateAndModify` (`core/e2e`, `timeout: no fswatcher event within 10s`). It is a flake, and here is the evidence rather than the assertion:

- `core/internal/contracttesting` — the only package this PR changes — passed in that same run (`ok … 1.159s`).
- `git diff --name-only origin/main...HEAD` is 12 files: `AGENTS.md` plus `core/internal/contracttesting/*`. `git diff origin/main...HEAD -- core/e2e/` is **0 lines**, so that package is byte-identical to `main`, and `core/e2e` does not import `contracttesting`.
- The test passed **20/20** locally (`go test ./core/e2e/ -run TestFSWatcher_EmitsEventsForTranscriptCreateAndModify -count=20`).
- Rerunning the failed job on the **same commit** turned it green (job 94238251862).

This is the known fsnotify timing flake — re-adding a just-deleted-and-recreated directory through the same watcher. Not introduced here, and not fixed here either.

PR #1499, opened 2026-08-12
fix(processlifecycle): stop capturing a tmux server's frozen host identity

### LINKED ISSUE #1486: tmux: a pane's TERM_PROGRAM describes the terminal the tmux *server* started in, not the attached client (UNVERIFIED)
Follow-up to #1405 (PR #1483), which asked whether anything else stored at
capture time has the same dynamic-property problem as a herdr pane's host.

⚠️ **This is reasoned from code, NOT live-verified.** See "What I could not
observe" below. Treat it as a hypothesis to check before spending effort on it.

## The hypothesis

`launcherFromEnv` (`core/adapters/inbound/agents/processlifecycle/osutil.go`)
early-returns for a herdr pane, keeping **only** its own address:

```go
if pane := env["HERDR_PANE_ID"]; pane != "" {
    return &session.Launcher{
        HerdrPaneID:     pane,
        HerdrSocketPath: env["HERDR_SOCKET_PATH"],
    }
}
```

The reason is written up on `session.Launcher` itself:

> herdr's server owns each pane's pty, outlives any attached client and is
> reparented to init, so every terminal-identity var a pane inherits —
> `$TERM_PROGRAM`, `$TMUX`, `$TMUX_PANE`, `$KITTY_*`, `$VSCODE_PID` — describes
> whatever environment the *server* was started in, frozen at that moment and
> handed to every pane it will ever spawn.

**tmux gets no such treatment.** The very next lines read `TERM_PROGRAM`,
`ITERM_SESSION_ID`, `TMUX_PANE` and `$TMUX` straight from the pane's own env
into the `Launcher`. But tmux's server has the same architecture the paragraph
above describes: detached, reparented, outliving its clients, handing each pane
the environment the server was started in. Note that the paragraph names `$TMUX`
and `$TMUX_PANE` explicitly — the rationale was written with tmux in view.

If that holds, then for a tmux-hosted session:

- `TermProgram` names the terminal the **tmux server** was started in, not the
  one currently displaying the pane.
- Detaching and re-attaching in a different terminal never updates it — the
  same staleness #1405 fixed for herdr, but with no client-resolution path to
  refresh *from*, because #1350's equivalent was never built for tmux.
- It can be wrong from the **first** click, not only after a re-attach — which
  is a strictly larger problem than #1405's.

**Control is unaffected.** `controller.go` routes `send-keys` by `TmuxPane` /
`TmuxSocket`, which are stable across detach/re-attach and correct regardless.
This is a click-to-focus issue only — the same split as #1349/#1350.

## What I could not observe

I tried to verify this live rather than assert it (tmux 3.6a, a server started
with a synthetic `TERM_PROGRAM=FAKE_ALPHA`, a second client attached via
`script` with `TERM_PROGRAM=FAKE_BRAVO`, reading the pane process through the
daemon's own `processlifecycle.ReadLauncherEnv`).

**The probe was inconclusive.** `ReadLauncherEnv` reported an empty
`TermProgram` *and* an empty `TmuxPane` for the pane process in both the
`sleep`-command and default-shell cases — i.e. the `sysctl(kern.procargs2)` env
read came back with nothing, so the frozen value could not be observed at all.
A tmux pane's shell certainly has `$TMUX_PANE` set at exec time, so the empty
read is a property of the env-read path on this machine, not of tmux. I did not
chase it further, since it is out of #1405's scope.

## To verify

Anything that establishes what the daemon actually records for a real tmux
session is enough:

1. Start a tmux server from terminal A, run an agent in a pane, and read the
   session's `launcher.term_program` from the daemon (`GET /sessions`, or the
   session JSON under the instances dir).
2. Detach; re-attach from terminal B; re-read.
3. If `term_program` still names A — or named A's *ancestor* rather than A from
   the start — the hypothesis holds.

If it does hold, the fix is likely the herdr shape: suppress the inherited
identity vars for a tmux pane and resolve the host from the attached tmux
client (`tmux list-clients -F '#{client_tty}'` gives the client's tty directly,
which is a cheaper key than herdr's lsof scan).

ingo-eichhorst:
Closes #1486.

#1486 was filed from code reading and explicitly marked UNVERIFIED, with an
inconclusive live probe. So the first job was establishing whether the defect is
real. It is — but **not by the mechanism the issue proposed**, and the difference
matters enough to state up front.

## What was measured

tmux 3.6a, pristine (`-f /dev/null`, no user config — this machine has none), a
throwaway server on a private socket started with synthetic identity vars, and a
second client attached from a *different* synthetic terminal. Launchers read
through the daemon's own `processlifecycle.ReadLauncherEnv`.

| launcher field | server started with | pane got | frozen leak? |
|---|---|---|---|
| `TERM_PROGRAM` | `iTerm.app` | **`tmux`** | **no** — tmux overwrites it |
| `ITERM_SESSION_ID` | `w0t0p0:ALPHA-UUID` | `w0t0p0:ALPHA-UUID` | **yes** |
| `TERM_SESSION_ID` | `ALPHA_TERMSESS` | `ALPHA_TERMSESS` | **yes** |
| `VSCODE_PID` | `4242` | `4242` | **yes** |
| `KITTY_LISTEN_ON` / `KITTY_WINDOW_ID` / `KITTY_PID` | ALPHA values | ALPHA values | **yes** |
| `TERMINAL_EMULATOR` | `JetBrains-JediTerm` | same | **yes** |
| `TMUX` / `TMUX_PANE` | — | the pane's own | correct |

**The issue's central claim is refuted.** `TermProgram` does not name the terminal
the server was started in: tmux overwrites `$TERM_PROGRAM` with the literal `tmux`
(plus `TERM_PROGRAM_VERSION=3.6a`). The repo already held this evidence without it
being connected — `herdrPaneEnv`, the live #1348 capture, carries
`"TERM_PROGRAM": "tmux"` for a herdr server started inside tmux.

**The issue's concern is confirmed, and its "wrong from the first click" framing is
right.** The other seven vars are handed through frozen. Measured on a pane created
minutes after the ALPHA terminal was gone, while an `Apple_Terminal` client
(`w9t9p9:BRAVO-UUID`) was the only one attached — that pane still reported
`iterm_session_id=w0t0p0:ALPHA-UUID`. Detaching and re-attaching never updated it
(byte-identical launcher before and after).

`update-environment` is why: its default set is `DISPLAY KRB5CCNAME MSYSTEM
SSH_ASKPASS SSH_AUTH_SOCK SSH_AGENT_PID SSH_CONNECTION WINDOWID XAUTHORITY` — no
terminal-identity var, so attaching never refreshes one.

### Why the issue's own probe came back empty

Reproduced, then explained. `KERN_PROCARGS2` returns **no env at all** for Apple's
signed platform binaries, which is what the original probe used:

```
applesleep  (/bin/sleep)       rawlen=35    envish=0   launcher={"tty":"/dev/ttys025"}
applezsh    (/bin/zsh)         rawlen=32    envish=0   launcher={"tty":"/dev/ttys026"}
gobin       (unsigned Go bin)  rawlen=6428  envish=84  launcher={"term_program":"tmux",...}
```

It is a property of the *binary*, not of tmux — the same hardened-runtime path
`readProcessEnv`'s doc comment already describes. Probing with a non-platform
binary makes the env fully readable.

### The user-visible consequence (worse than the issue supposed)

`term_program: "tmux"` matches no entry in the macOS activator registry
(`SessionLauncher.swift`), and being **non-empty** it also suppresses the
`TermProgram == ""` guards that would otherwise reach the ancestry fallbacks — so
`HostBundleID` is never resolved either. `resolveActivator` returns nil and
**click-to-focus is a silent no-op for every tmux session**, with `TmuxActivator`'s
`select-pane` never running because it is a decorator over a nil base.

## The fix

Apply #1348's herdr treatment to tmux, which is the same architecture for the same
reason (server daemonizes, reparented to PID 1, outlives every client) and which
that rationale already named `$TMUX`/`$TMUX_PANE` explicitly — but keyed carefully,
because the first cut of this PR was too broad and review caught it (see below):

- `launcherFromEnv` early-returns a tmux pane's **own address only**, gated on
  `$TMUX_PANE` **and** tmux's own `TERM_PROGRAM=tmux` marker. Placed *after* the
  herdr check — a herdr server started from a tmux pane hands every pane a
  `$TMUX_PANE` too, and that one is the server's.
  `TestLauncherFromEnv_HerdrCapture` locks the ordering.
- `hostIdentity` **keeps** the ancestry fallbacks enabled for tmux. For a genuine
  pane the walk is inert (the chain reaches the reparented server and terminates),
  and for a descendant that reports no host of its own it is the only thing that
  recovers one.
- `tmuxSocketFromEnv` extracts the shared `$TMUX` parse so the address the pane
  *keeps* is not parsed only on the path that no longer runs for it.
- `SessionLauncher.herdrDiagnostic` counts an adopted `tmuxPane` as evidence a
  herdr client *was* resolved. A herdr pane's own env never contributes one
  (#1348 suppresses it), so it can only have been adopted — and after this change
  a client living in tmux contributes nothing else.

### Why not `$TMUX_PANE` alone

`$TMUX_PANE` is **not self-describing**: every descendant of a pane inherits it,
including a GUI terminal or IDE launched from inside one (`code .`, `kitty`). Such
a process carries the stale pane address *next to a correct host identity it
reported itself*. Suppressing on the address alone discarded that host and turned
a working click-to-focus into a silent no-op — a regression, not a fix. tmux stamps
`TERM_PROGRAM=tmux` on the panes it spawns, so a process claiming any other host
claimed it for itself and is left alone; one claiming no host of its own (kitty
sets none, upstream kitty#4793) is suppressed and then recovered by the ancestry
walk, which is why that walk must stay enabled.

Result, measured live on real tmux panes:

```
before: {"term_program":"tmux","iterm_session_id":"w0t0p0:ALPHA-UUID","term_session_id":"ALPHA_TERMSESS","tmux_pane":"%4","tmux_socket":"...","tty":"/dev/ttys026"}
after:  {"tmux_pane":"%4","tmux_socket":"...","tty":"/dev/ttys026"}
```

**Control is unaffected**, as the issue predicted: `resolveBackend` picks
`backendTmux` from `TmuxPane` alone, before it ever reads `TermProgram`.

**Cost profile: nothing added.** No new probe, no new shellout, no sweep-time work,
no extra process walk — the ancestry fallbacks run exactly where they ran before.
Zero tmux sessions costs exactly what it did before.

## Red-first

Every assertion here was seen failing before its fix existed.

**The leak** (against unmodified `main`, production files reverted to the branch
point with the tests at HEAD):

```
--- FAIL: TestLauncherFromEnv_TmuxSuppressesInheritedIdentity
    TermProgram: tmux's own name is not a host, got "tmux"
    iTerm/Terminal.app session identity must not survive, got iterm="w0t0p0:ALPHA-UUID" term="ALPHA_TERMSESS"
    kitty identity must not survive, got listen="unix:/tmp/kitty-ALPHA/sock" window="77" pid=555
    VSCodePID: inherited value must not survive, got 4242
--- FAIL: TestReadLauncherEnv_Subprocess_TmuxSuppressesInheritedIdentity
    inherited terminal identity survived the real read path: &{TermProgram:tmux ITermSessionID:w0t0p0:ALPHA-UUID ...}
    tmux's own marker must never be kept as a host: &{TermProgram:tmux ...}
```

**The over-broad first cut** (`TestLauncherFromEnv_TmuxSuppressionSparesADescendantsOwnHost`,
red against this PR's own first commit — this is the review finding):

```
--- FAIL: TestLauncherFromEnv_TmuxSuppressionSparesADescendantsOwnHost
    a descendant's own TERM_PROGRAM must survive, got ""
    VSCodePID: want 4242, got 0
```

**The herdr diagnostic** (Swift, `swift test` — not CI-gated, run locally):

```
SessionManagerTests.swift:500: error: XCTAssertEqual failed:
  ("herdr_pane=w1:p1, no attached client") is not equal to
  ("herdr_pane=w1:p1, client resolved but its host has no activator")
```

`TestLauncherFromEnv_TmuxCapture` and `TestLauncherFromEnv_TmuxSuppressionIsScoped`
are **locks** — they pass by construction before and after, and pin the blast
radius (the pane's address survives; an ordinary iTerm session is untouched).

Note what `TestReadLauncherEnv_Subprocess_TmuxSuppressesInheritedIdentity`
deliberately does *not* assert, unlike its herdr twin: `TermProgram == ""`. In a
test the spawned subprocess is a child of the test binary, which really is hosted
by a terminal, so ancestry legitimately repopulates it — the same recovery path a
kitty-from-pane window depends on. Asserting its absence would pin the bug.

## Deliberately NOT built

Making click-to-focus actually *work* for tmux needs #1350's counterpart: resolve
the attached client (`tmux -S <socket> list-clients -F '#{client_pid}'`, then read
that process's identity as a herdr client's is read). That is materially larger
than #1405 — `AdoptHostIdentity`'s deliberately-closed "put back" set would have to
grow to cover the tmux address or the pane loses its backchannel selector, and it
needs a refresh hook symmetric to `refreshHerdrHosts`. Flagged rather than
smuggled in; until then a tmux session honestly resolves to no host, which is the
same degradation a herdr session with no attached client already gets.

## Review

One `medium` review subagent. It confirmed four findings; all four are applied.
The first — the inherited-`$TMUX_PANE` regression above — was a real defect in
this PR's first cut, found by launching kitty from a tmux pane and measuring,
not by reading the diff. It also disproved the first cut's stated rationale for
skipping the ancestry walk (the walk is inert because the server is reparented,
not because it would rediscover the suppressed host), and that correction is now
in the code comments rather than only in this description.

## Scope note

No throwaway tmux session outlived the investigation; the user's own `default`
socket was never touched.

### DIFF
--- a/core/adapters/inbound/agents/processlifecycle/osutil.go
+++ b/core/adapters/inbound/agents/processlifecycle/osutil.go
@@ -145,7 +145,14 @@ func hostIdentity(pid int) *session.Launcher {
 	// kitty field back-fill). Walking the ppid chain once instead of up to
 	// three times keeps this bounded — each readProcInfo is a `ps` shellout
 	// with a 2s ceiling.
-	if l.HerdrPaneID == "" {
+	//
+	// A tmux pane is skipped for the same reason, and the reason is not merely
+	// that the walk is futile: tmux's server is reparented to PID 1, so the
+	// chain from a pane can only ever rediscover the terminal the *server* was
+	// launched from — which is precisely the stale host launcherFromEnv just
+	// suppressed. Running the fallbacks here would undo that suppression by
+	// another route, exactly as it would for herdr.
+	if l.HerdrPaneID == "" && l.TmuxPane == "" {
 		applyAncestryFallbacks(l, pid, memoizedAncestry(pid))
 	}
 
@@ -173,6 +180,18 @@ func launcherFromEnv(env map[string]string) *session.Launcher {
 			HerdrSocketPath: env["HERDR_SOCKET_PATH"],
 		}
 	}
+	// A tmux pane is the same shape as a herdr pane and for the same reason,
+	// so it keeps only its own address too — see session.Launcher's Tmux*
+	// fields. Checked *after* herdr on purpose: a herdr server started from a
+	// tmux pane hands every pane a $TMUX_PANE as well, and that one is the
+	// server's, not the agent's (#1348). herdrPaneEnv in the tests is exactly
+	// that env, and TestLauncherFromEnv_HerdrCapture locks the ordering.
+	if pane := env["TMUX_PANE"]; pane != "" {
+		return &session.Launcher{
+			TmuxPane:   pane,
+			TmuxSocket: tmuxSocketFromEnv(env["TMUX"]),
+		}
+	}
 	l := &session.Launcher{
 		TermProgram:    env["TERM_PROGRAM"],
 		ITermSessionID: env["ITERM_SESSION_ID"],
@@ -181,14 +200,7 @@ func launcherFromEnv(env map[string]string) *session.Launcher {
 		KittyListenOn:  env["KITTY_LISTEN_ON"],
 		KittyWindowID:  env["KITTY_WINDOW_ID"],
 	}
-	if tmux := env["TMUX"]; tmux != "" {
-		// $TMUX is "/path/to/socket,pid,session" — first field is the socket.
-		if i := strings.Index(tmux, ","); i > 0 {
-			l.TmuxSocket = tmux[:i]
-		} else {
-			l.TmuxSocket = tmux
-		}
-	}
+	l.TmuxSocket = tmuxSocketFromEnv(env["TMUX"])
 	if v := env["VSCODE_PID"]; v != "" {
 		if n, err := strconv.Atoi(v); err == nil {
 			l.VSCodePID = n
@@ -212,6 +224,22 @@ func launcherFromEnv(env map[string]string) *session.Launcher {
 		l.TermProgram = "jetbrains"
 	}
 	return l
+}
+
+// tmuxSocketFromEnv extracts the server socket from $TMUX, whose value is
+// "/path/to/socket,pid,session". Returns "" for an empty $TMUX so a launcher
+// with no tmux server keeps a zero socket rather than an empty-but-set one.
+// Shared by both branches of launcherFromEnv: a tmux pane's address is the one
+// thing it keeps, so the parse must not live only on the path that no longer
+// runs for it.
+func tmuxSocketFromEnv(tmux string) string {
+	if tmux == "" {
+		return ""
+	}
+	if i := strings.Index(tmux, ","); i > 0 {
+		return tmux[:i]
+	}
+	return tmux
 }
 
 // memoizedAncestry returns a closure resolving pid's process ancestry via
--- a/core/adapters/inbound/agents/processlifecycle/osutil.go
+++ b/core/adapters/inbound/agents/processlifecycle/osutil.go
@@ -146,13 +146,15 @@ func hostIdentity(pid int) *session.Launcher {
 	// three times keeps this bounded — each readProcInfo is a `ps` shellout
 	// with a 2s ceiling.
 	//
-	// A tmux pane is skipped for the same reason, and the reason is not merely
-	// that the walk is futile: tmux's server is reparented to PID 1, so the
-	// chain from a pane can only ever rediscover the terminal the *server* was
-	// launched from — which is precisely the stale host launcherFromEnv just
-	// suppressed. Running the fallbacks here would undo that suppression by
-	// another route, exactly as it would for herdr.
-	if l.HerdrPaneID == "" && l.TmuxPane == "" {
+	// A tmux pane is deliberately NOT skipped here, though the same argument
+	// looks like it should apply. It does not: tmux's server is reparented to
+	// PID 1, so the walk from a genuine pane terminates there having found
+	// nothing — it is inert, not dangerous, and cannot put back what
+	// launcherFromEnv suppressed. Meanwhile it is the only thing that recovers
+	// a host for a process launched from a pane by a terminal that reports
+	// none of its own (kitty), whose fields the suppression above drops. So
+	// skipping it would buy no correctness and cost exactly that case.
+	if l.HerdrPaneID == "" {
 		applyAncestryFallbacks(l, pid, memoizedAncestry(pid))
 	}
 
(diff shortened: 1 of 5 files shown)


ingo-eichhorst:
Deferred host resolution filed as #1501 (shape, cost profile, and the two adjacent decisions — old-tmux versions and the pre-existing inherited-$TMUX_PANE control-path case).

Issue #1570, opened 2026-08-15
The pre-push hook is unusable for an agent on a core/ diff, and its only escape hatch disables every gate


ingo-eichhorst:
Raised independently by three consecutive agents (#1547, #1544, and the #1529
run before them) after it bit each of them. Filing it because the documented
recovery is `--no-verify`, which disables **every** gate rather than the slow
one — so the hook's failure mode makes the tree less protected, not more.

## What happens

`tools/install-git-hooks.sh` wires `tools/preflight.sh --changed` as a pre-push
hook. On a diff touching a single package under `core/adapters/`, that run
exceeds a 600s foreground `Bash` budget, the tool call is killed, and the push
does not land — **with the commit already made**. AGENTS.md already documents
the shape of this ("killing the push after the commit had already landed") and
`ir:exec` Phase 4 carries an incident about it.

Measured this session: three of six PRs ended up pushed with `--no-verify`
after the hook blew the budget. The other three fitted.

## Why `--changed` does not save it

The scoping works for the Go gates. What does not scope down is the **security
gate**: on a `core/**.go` diff it runs `govulncheck` plus `gosec` over ~261
files (one agent's measurement), and that alone dominates. AGENTS.md describes
the security gate as scoped twice over — a trigger regex, then
`security-scan.sh --changed` picking modules — so the question is whether the
trigger regex fires on essentially every `core/` change. If it does, it fires on
nearly every push in this repo.

## Why the current escape hatch is the wrong shape

The advice an agent ends up following is: run `preflight.sh --only go` and
`--only arch` in the foreground yourself, then `git push --no-verify`. That is
strictly worse than the hook running, because `--no-verify` also skips the
gates that are *fast* and that the agent did not run (`skills`, `posix`,
`tools`). The hook's all-or-nothing design converts "one gate is too slow" into
"no gates ran".

## Options

1. **A per-gate timeout inside `preflight.sh`** when invoked from the hook: a
   gate that exceeds its budget reports `TIMEOUT` and the push is refused with
   that named, rather than the whole run being killed from outside with no
   output at all. Keeps the failure legible, which is the property the current
   behaviour loses entirely.
2. **Scope the security gate's trigger more tightly**, or give the hook a
   reduced gate set (the fast ones) and leave `govulncheck`/`gosec` to CI, where
   they already run.
3. **A run stamp**: skip a gate the hook can see was already run green against
   the same tree hash in this session. Directly addresses the agent workflow,
[...]

Issue #1558, opened 2026-08-15
herdr: the aggregate budget is spent by the lsof scan before the candidate loop sees it


ingo-eichhorst:
Found by the adversarial QA pass on PR #1557 (#1529) and deliberately not fixed
there — it is a behaviour trade-off inherent to the fix, not a defect in it. PR
#1557 states the trade in `resolveHerdrClientLauncherVia`'s doc; this issue is
whether to act on it.

## The trade

`herdrClientBudget` (2s) is ONE deadline covering both stages of the herdr
client indirection, and the stages run in sequence:

1. `herdrClientPIDs` — the `lsof` scan that finds the attached clients.
2. `resolveClientHostIdentityVia` — the candidate loop.

They compete for one budget and **the scan goes first**. An `lsof` that runs to
nearly `herdrClientBudget` and still *succeeds* leaves the loop nothing: the
loop's `ctx.Err()` check fires on candidate 1, every candidate is abandoned, and
the answer is `(nil, false)`.

So on a machine where `lsof` is chronically slow-but-answering, a herdr host
would **never** resolve — where before #1529 it resolved slowly. That is the
same shape #1553 records for the git adapter: an unbounded call that answered
slowly becomes a bounded one that does not answer at all.

## Why it is a follow-up and not a blocker

Three things bound it, and none removes it:

- **It fails to the safe side.** A non-answer clears no stored host (#1492) and
  the memo re-probes on the next sweep, so the cost is a deferred *recovery*,
  not a wrong answer.
- **The band is narrow.** An `lsof` at its own `shelloutTimeout` is normally
  KILLED rather than merely slow, and a killed one is already `(nil, false)` via
  `lsofProbeRan`. Only the slow-*and*-successful window is new.
- **Nobody would see it** — which is the part that argues for doing something.
  #1534 is that nothing counts probe non-answers, and that now covers budget
  abandonment too, so this degradation is invisible by construction.

## Options

1. **Split the budget between the stages** — e.g. cap the scan at some fraction
   so the loop is guaranteed a share. Needs evidence about the real distribution
   of `lsof` times on the client log, which #1529 has none of.
2. **Count it first** (#1534). A counter for "candidates abandoned on the
   aggregate budget", separated from "candidate unreadable", would say whether
   this ever happens before anything is tuned for it. Cheapest, and it is the
   measurement every other option needs.
3. **Do nothing, deliberately** — the doc in PR #1557 already states the trade,
[...]

Issue #1547, opened 2026-08-14
processlifecycle: the shellout SHAPE is still copied 8 times — one runProbe would delete a guard instead of needing one


ingo-eichhorst:
Raised by the `/simplify` altitude pass on PR #1545 (#1538) and deliberately not
taken there — see "Why not in #1545" at the end.

## The observation

#1538 extracted the two *values* out of the bounded-shellout block — the ceiling
(`shelloutTimeout`) and the predicate (`probeAnswered`) — and left the *shape*
copied at all eight sites:

```go
ctx, cancel := context.WithTimeout(context.Background(), shelloutTimeout)
defer cancel()
out, err := build(ctx).Output()
if !probeAnswered(err) { … }      // or lsofProbeRan, or wrap-and-return
```

It then added a source-scanning guard to detect a ninth copy getting it wrong.
That is an inversion: the mechanism policing the pattern is larger than the
pattern, and one helper would make the mistake *unrepresentable* rather than
*detectable*.

```go
// runProbe runs build's command under shelloutTimeout and classifies the outcome.
func runProbe(build shelloutCmd, answeredExitCodes ...int) (out []byte, answered bool, err error)
```

All eight sites were checked against it: `CWDOf` and `readProcInfo` use `err`
and ignore `answered`; the other six use `answered`. `bundleIDVia` is the only
one needing more than a mechanical edit — its Info.plist derivation moves to the
call site (already half-done: since #1545 it takes a `bundleIDCmd` factory for
exactly that reason).

## What it buys

- **`TestEveryBoundedShelloutUsesTheNamedCeiling` can be deleted.** With one
  `context.WithTimeout` in the package, a re-spelled ceiling is not expressible,
  so a whole guard plus its hand-maintained `knownCeilings` floor stops existing.
- The classification rule collapses from *"walk every scope, bind the error
  variable, window the propagation search between the run site and the next
  rebind, recurse into FuncLits"* to *"no child-process run outside
  `runProbe`"* — the two-arm shape `core/architecture_hookbody_test.go` uses
  (none outside, exactly one inside).

## What it costs, stated honestly

The current guard's best property would be lost. Run against `origin/main` it
reports **exactly the three known open bugs and nothing else** (`WriterOf`,
`processTTY`, `kittyWindowIDForPID` — measured in #1545), because it reasons
per-site about what each does with its error. A `runProbe` routing rule run
against the same tree reports all eight sites as "not routed", which is true but
[...]

Issue #1544, opened 2026-08-14
bundleIDForAppPath is memoizable now, and the ancestry walks still re-ps the same chain (#1538's deferred third section)


ingo-eichhorst:
#1538's third section, deliberately not taken in PR for #1538 — that PR was the
predicate plus its guard, and this is a caching change with its own
admission rule and its own concurrency surface.

Quoting #1538, which stated it first:

> A `CFBundleIdentifier` is immutable for the life of a bundle, yet every
> candidate whose chain reaches the same host app pays a fresh exec — N identical
> `plutil` calls at daemon seed with N persisted sessions under one host.
>
> #1524 is what makes the memo safe to write: its answer/non-answer split is
> exactly the cache-admission rule (**memoize answers including the empty one,
> never memoize a non-answer**) ... in the pathological case the memo turns
> N x 2s into 1 x 2s.
>
> Separately, walk 1 and walk 2 re-`ps` the same ppid chain (up to ~10 duplicate
> execs per candidate). `isKnownInteractiveHostVia`'s own doc already names "the
> duplicate ps shellouts"; `ancestryProbe` memoizes walk 1's *verdict* but not the
> per-PID reads walk 2 repeats.

What changed since #1538 was written: `bundleIDVia` still returns
`(string, error)` with the same split, so the admission rule is unchanged and
still expressible. The predicate behind it is now `probeAnswered`
(`processlifecycle/shellout.go`).

Note the counter-argument that has to be answered before writing the memo, and
that #1514 is the precedent for: a memoized NON-answer is the bug this family
keeps producing, and #1514 concluded that memoizing a non-answer is acceptable
only when every consumer sees it AS a non-answer. The admission rule above says
the opposite for this cache — never store a non-answer — so the two must not be
confused when implementing.

**Unverified:** the "N identical plutil calls at daemon seed" cost is #1538's
claim, carried forward; no profile was taken in the #1538 PR.

PR #1573, opened 2026-08-15
feat(daemon): count what every bounded probe did, and publish it (#1534)

### LINKED ISSUE #1534: Nothing counts probe non-answers, so the #1485/#1492/#1513/#1524 family's real-world frequency is unmeasurable
Every issue in this family closes with the same caveat — #1513 flagged it for
`ps`, #1524 repeated it for `plutil` — and none of them fixed it: **nothing
measures how often a probe fails to answer.**

That gap has a concrete cost now that three of these fail OPEN. `IsKnownInteractiveHost`
admits on a walk it could not complete (#1513, #1524), which is the right call
on no evidence, but it means a probe that is failing constantly is
indistinguishable from one that never fails — the gate just quietly stops
gating, and #784's CodexBar sessions come back with nothing anywhere to look at.
The fail-open direction converts a visible defect into an invisible one; a
counter is what keeps it visible.

## What the measurement so far says

#1524's PR measured `plutil` on this machine (10 cores, Obsidian installed to
`/Applications` on local SSD), n=150 per condition:

| condition | median | p95 | max | over 2s |
|---|---|---|---|---|
| idle | 2.2ms | 2.6ms | 4.4ms | 0 |
| 4x CPU oversubscription | 7.2ms | 8.1ms | 9.9ms | 0 |
| 12x CPU oversubscription | 7.3ms | 8.5ms | 79.6ms | 0 |
| first touch, 12 distinct bundles | — | — | 3.8ms | 0 |

So the worst observed is **79.6ms against a 2000ms ceiling — a 25x margin**,
and CPU pressure alone does not appear to be the trigger the issues assumed.
Whatever does trigger it (I/O stall on a network volume or unresponsive mount,
process-table exhaustion, memory pressure) is unidentified, and it cannot be
identified without counting.

## Scope

A counter per probe kind and outcome — answered / unanswered — read into the
diagnostics bundle, alongside the hook counters that already live there
(`hooks.json`). The receiver-side families already established the shape:
`hookjson.IgnoreUnknownEvent` and `RejectPath` count per (adapter, reason) and
surface through `GET /debug/bundle`, with `--diagnose` saying out loud that it
omits counts rather than publishing zeros. The same treatment here would turn
every "Unverified:" footer in this family into something answerable.

Probes in scope (all under a 2s ceiling, all collapsing failure into a
zero value today):
- `readProcInfo` / `processTTY` — `ps` (#1533)
- `bundleIDForAppPath` — `plutil` (#1524, now reports its non-answers to the
  walk but counts nothing)
- `CWDOf`, `herdrClientPIDs` — `lsof` (#1485)
- `DiscoverPID` — `pgrep`

ingo-eichhorst:
Closes #1534.

Every issue in the `#1485/#1492/#1513/#1524/#1533/#1537` family closes with the
same caveat: **nothing measures how often a probe fails to answer.** Three of
those fixes fail OPEN — `IsKnownInteractiveHost` admits on a walk it could not
complete, which is the right call on no evidence — so a probe failing constantly
and one that never fails produce the *same* observable daemon: the #784 gate
quietly stops gating with nothing anywhere to look at.

This adds a counter per probe kind and outcome, read into the diagnostics bundle
as `probes.json`, a sibling of `hooks.json`.

## What is counted, and where

**At `runProbe`**, which since #1547 is the only place in `processlifecycle`
that starts a child (`TestEveryChildProcessRunsThroughRunProbe`). One counting
site instead of eight, and a ninth probe is counted by existing rather than by
someone remembering. `runProbe` grew a `kind` parameter — a `shelloutCmd` closes
its arguments over, so there is nothing in `build` to derive one from — and
`TestEveryProbeSiteDeclaresItsOwnKind` grades what each site passes.

**Keyed per call site, not per tool.** `lsof` answers three questions here and
`ps` two. A count that cannot say *which* question stopped being answered is the
single-scalar failure #1364 already paid for. Eight kinds, one per site:
`ps.proc_info`, `ps.tty`, `plutil.bundle_id`, `lsof.cwd`, `lsof.writer`,
`lsof.herdr_clients`, `pgrep.discover`, `kitten.window`.

**ANSWERED means the child ran to a normal exit** — `shellout.Answered`'s
tool-independent half — and deliberately *not* each site's own exit-code
allowlist. An allowlist encodes what a particular normal exit MEANS to one call
site: policy, per-site by construction, and #1547 kept `runProbe` free of exactly
that. Folding it in would put a second copy of every site's policy in the
counter, able to drift from the code it describes. `probeOutcomeRule` carries the
argument and states the cost: for a tool with an allowlist, a
normal-but-not-allowlisted exit (`lsof` 2, `pgrep` 2) is counted answered while
its call site treats it as carrying no verdict — still the right reading for this
counter's question, since the probe *did* run. The case the fail-open gates
actually degrade on, a child killed by the 2s ceiling, lands on the unanswered
side in every spelling.

## The three changed premises

**#1547 (merged) made this dramatically cheaper — confirmed.** `runProbe` is the
sole run site; the signature change to `(ctx, kind, build)` is the whole cost.
The routing guard and its committed corpus stayed green: the corpus rows are
self-contained source strings the detector parses arity-agnostically, and the
live rule's vacuity floor still sees 9 run sites (8 `runProbe` calls + the one
`.Output()` inside it). The classification rule also stayed green — `runProbe`
now binds `err`, calls `observeProbe`, and returns it, which is propagation.

**#1544 (merged) added two memos — covered as a third outcome, not excluded.**
Its own hand-back flagged this: *"a memo now also hides how often the probe
runs."* A `plutil` served from `bundleIDMemo`, or a `ps` from `ancestryReads`,
starts no child and is invisible at `runProbe`. Reporting "N answered" while
silently excluding hits is exactly the misleading figure this issue exists to
prevent, so a hit is counted where it happens — in the memo, under the kind the
underlying read would have used — and published as `memo_hits`.
`answered + unanswered` is how often a child **ran** (the denominator the
non-answer rate is a fraction of); `+ memo_hits` is how often the probe was
**asked**. `memo_note` says so in the bundle. Only two probes have a memo, so a
zero elsewhere is a finding of nothing rather than an inability to look, and the
note says that too.

**#1529 (merged) added a fourth non-answer kind — covered.** #1558 asks for
exactly this and names it the measurement every other option needs
(*"count it first... cheapest, and it is the measurement every other option
needs"*). It is **not** a probe row and is deliberately kept off the
(kind, outcome) axis: no child process is involved, so calling it "unanswered"
would put a different kind of fact in a column whose meaning is fixed. It is
published beside them, with a `probed` denominator that #1558 does not ask for
and without which "12 abandoned" cannot be read at all — 12 out of 12 and 12 out
of 10,000 are different findings. A comment has been posted on #1558.

## The `--diagnose` distinction, and why it is one step stronger here

`hooks.json` omits counts under `--diagnose` because that process served no
hooks, so its counters are *structurally zero*. **That reason is false for
probes**, and stating it would teach the next reader something wrong: building
`processes.json` walks every adapter's matcher through the observer, which on
macOS shells out to `pgrep` and `lsof`.

Measured, by temporarily wiring `liveProbeHealth` into `runDiagnose` and running
the binary against an isolated `IRRLICHT_HOME`:

```
{"probe": "lsof.cwd",       "answered": 11, "unanswered": 0, "memo_hits": 0}
{"probe": "pgrep.discover", "answered": 11, "unanswered": 0, "memo_hits": 0}
```

So the CLI's counters are small, real, and about nothing but its own bundle
collection. Under the daemon's field names they would read as a measurement of
the daemon's probe health and would look *plausible* — worse than a zero, because
a zero announces itself. `probesView` therefore emits a different shape, and the
note says which reason applies:

```json
{
  "collected_from": "cli",
  "note": "Collected by `irrlichd --diagnose`, which is not the process that runs the daemon's probes, so the per-probe answered/unanswered counters are omitted rather than reported. Note they are NOT zero in this process: building processes.json shells out to pgrep and lsof through the same observer, so this process has a few counts of its own, describing nothing but its own bundle collection — which is why they are omitted rather than printed. For the daemon's real counts, fetch GET /debug/bundle from the running daemon."
}
```

## Behaviour preservation

The change is behaviour-preserving apart from the counters. What convinced me:

- Every production hunk is an added argument, an added counter call, or a
  comment — `git diff` on the four adapter files is 8 one-token argument
  additions, 4 counter calls, and prose. No classifier, polarity, ceiling,
  return value or control-flow edge moved.
- The counter reads the error *after* it is produced and never feeds it back:
  `observeProbe` returns nothing, and `runProbe` still returns the same
  `(out, err)` it always did.
- No call site's `probeAnswered` / `lsofProbeRan` call was touched, which is why
  the counter uses the tool-independent half rather than adopting any site's
  allowlist — adopting one would have been a second decision about that site.
- The two guards over these shellouts (`TestEveryChildProcessRunsThroughRunProbe`,
  `TestEveryBoundedShelloutClassifiesItsError`) and the whole
  `processlifecycle` suite are green at `-race -count=1` **and `-count=4`** —
  process-global counters are exactly the thing that breaks under repeat runs, so
  every assertion is a DELTA and no reset hook was needed.

## Mutation evidence — nine mutations, each seen red

Counters pass the moment they are written. Each mutation below was applied by
copy-aside, run, and restored by copy-back with a `git status --porcelain` diff
against a captured baseline (`RESTORED_CLEAN` after every one). No `git stash`,
no `git restore`, no `git checkout -- .`.

**M1 — `observeProbe` counts everything as answered.**

```
--- FAIL: TestProbeCounterSeparatesAnsweredFromUnanswered (0.03s)
    probecount_test.go:140: probe counters moved by map[ps.tty:{ps.tty 1 0 0}], want map[ps.tty:{ps.tty 0 1 0}] — a child that never started is the non-answer every issue in this family closed without a count of
--- FAIL: TestProbeCounterKeepsEachKindInItsOwnBucket (0.01s)
    probecount_test.go:168: probe counters moved by map[lsof.cwd:{lsof.cwd 1 0 0} lsof.writer:{lsof.writer 1 0 0}], want map[lsof.cwd:{lsof.cwd 0 1 0} lsof.writer:{lsof.writer 0 1 0}] — three lsof questions, three buckets: a shared one would report six failures against one name and diagnose neither
--- FAIL: TestProductionProbeSitesCountUnderTheirOwnKind/ps.tty (0.01s)
    probecount_darwin_test.go:44: probe counters moved by map[ps.tty:{ps.tty 1 0 0}], want map[ps.tty:{ps.tty 0 1 0}] — the counter must agree with the site's own probed bit on the case that matters
```

**M2 — the (kind, outcome) key collapses: `writerOfVia` passes `probeLsofCWD`.**
A *different* test, and only that one:

```
--- FAIL: TestEveryProbeSiteDeclaresItsOwnKind (0.01s)
    probecount_test.go:330: probeLsofCWD is passed by 2 call sites (process_darwin.go:83, process_darwin.go:126) — they share one bucket, so a climbing count cannot say which probe stopped answering. A kind names a call site; a genuinely shared one amends this rule in a reviewable diff.
    probecount_test.go:336: probeLsofWriter is declared but passed by no call site — it publishes a permanent zero row that no probe can ever move
```

**M3 — the `--diagnose` form publishes zeros instead of saying it cannot look.**
The most important one:

```
--- FAIL: TestProbesBundleOmitsCountsWhenNotCollectedInDaemon (0.00s)
    diagnostics_service_test.go:778: collected_from = daemon, want "cli"
    diagnostics_service_test.go:792: probes is present without a daemon to read it from: [map[answered:0 memo_hits:0 probe:lsof.cwd unanswered:0]]
    diagnostics_service_test.go:792: total_unanswered is present without a daemon to read it from: 0
    diagnostics_service_test.go:792: herdr_client_candidates_probed is present without a daemon to read it from: 0
    diagnostics_service_test.go:792: memo_note is present without a daemon to read it from: ...
    diagnostics_service_test.go:797: note does not say where the real evidence is: ""
    diagnostics_service_test.go:802: note must say that this process's counters are non-zero and describe its own collection, not that they are structurally zero: ""
```

**M4 — vacuity: every probe is counted as a non-answer *as well*.** A counter
that counts everything is indistinguishable from one that counts correctly until
something asserts the silence; `probesSince` returns only the rows that MOVED, so
a one-entry expectation is an exhaustive claim:

```
--- FAIL: TestProbeCounterSeparatesAnsweredFromUnanswered (0.02s)
    probecount_test.go:124: probe counters moved by map[ps.tty:{ps.tty 1 1 0}], want map[ps.tty:{ps.tty 1 0 0}] — a child that ran to a normal exit is an ANSWER, and is counted by nobody as a non-answer
--- FAIL: TestProductionProbeSitesCountUnderTheirOwnKind/lsof.writer (0.02s)
    probecount_darwin_test.go:86: probe counters moved by map[lsof.writer:{lsof.writer 1 1 0}], want map[lsof.writer:{lsof.writer 1 0 0}] — lsof's 'nothing to report' ran and answered; counting it as a non-answer would report a healthy machine as failing
```

**M5 — the daemon forgets to wire `liveProbeHealth`** (`startup.go` passes nil).
The service's nil branch is a perfectly valid, perfectly quiet output, so this is
the only thing that catches it:

```
--- FAIL: TestDaemonStartupSmoke/demo (1.62s)
    daemon_smoke_test.go:61: probes.json collected_from = "cli", want "daemon" — the running daemon did not wire its probe counters into the bundle
    daemon_smoke_test.go:61: probes.json reports itself as daemon-collected but carries no probe rows — every declared kind is published, including one that never ran
```

**M6 — a memo hit is folded into `answered`.**

```
--- FAIL: TestBundleIDMemoHitIsCounted (0.00s)
    probecount_darwin_test.go:130: probe counters moved by map[plutil.bundle_id:{plutil.bundle_id 1 0 0}], want map[plutil.bundle_id:{plutil.bundle_id 0 0 1}] — #1544's hand-back: a memo hides how often the probe RUNS, so a hit is its own outcome rather than an unreported one
--- FAIL: TestAncestryReadsMemoHitIsCounted (0.00s)
    probecount_test.go:203: probe counters moved by map[ps.proc_info:{ps.proc_info 1 0 0}], want map[ps.proc_info:{ps.proc_info 0 0 1}] — a ps served from ancestryReads is a call the probe was asked and never ran for; #1544's hand-back is that a memo hides exactly this
```

**M7 — budget abandonment is not counted** (`observeHerdrCandidatesAbandoned`
deleted):

```
--- FAIL: TestHerdrCandidateAbandonmentIsCounted/a_spent_budget_abandons_once_and_probes_nothing (0.00s)
    probecount_darwin_test.go:170: abandoned moved by 0, want 1 — one event per loop, not one per candidate left unread
```

**M8 — an observation naming an undeclared kind is silently dropped.** Dropping
it would make an uncounted probe indistinguishable from one that never ran —
this issue's own failure mode reproduced inside its fix:

```
--- FAIL: TestUndeclaredProbeKindIsCountedNotDropped (0.00s)
    probecount_test.go:220: undeclared observations = 0, want 2 — an unattributable probe must be reported, not swallowed
```

**M9 — the explanations fire unconditionally.** An explanation that always fires
explains nothing; a reader could not tell a machine with a failing probe from a
healthy one:

```
--- FAIL: TestProbesBundleStaysTerseWhenNothingIsWrong (0.00s)
    diagnostics_service_test.go:756: unanswered_note is present on a healthy machine: Probe(s) that did not answer: . A non-answer is ... — an explanation that always fires explains nothing
    diagnostics_service_test.go:756: herdr_abandonment_note is present on a healthy machine: The herdr client loop abandoned its remaining candidates 0 time(s) ... — an explanation that always fires explains nothing
```

### Which half of the coverage reaches which sites

Stated rather than left to be assumed. Four of the eight sites take an injected
builder, so the darwin runtime test drives them end to end (one per tool, and
both classifier polarities). The other four (`ps.proc_info`, `lsof.cwd`,
`lsof.herdr_clients`, `kitten.window`) take no builder, so what each of them
*passes* is graded statically by `TestEveryProbeSiteDeclaresItsOwnKind`, which
reads the package's own source with build tags ignored — the same reason the
shellout guard beside it parses rather than type-loads: every one of these sites
is behind `//go:build darwin`, and a type-aware load on a Linux runner would find
none of them and pass.

## Verification (all foreground)

- `go build ./core/...` and `GOOS=linux go build ./core/...` — pass
- `go test ./core/adapters/inbound/agents/processlifecycle/... -race -count=1` — pass (12.1s)
- same at `-race -count=4` — pass (45.2s), no reset hook needed
- `go test ./core/... -race -count=1` — pass, no failures
- `tools/preflight.sh --only go` — all 7 gates PASS
- `tools/preflight.sh --only arch` — PASS, composite 7.931881 → 7.932251 (no regression)
- `irrlichd --diagnose` run against an isolated `IRRLICHT_HOME`: emits the CLI
  form shown above.

The pre-push hook blew the 600s command budget on this `core/` diff (#1570); the
push was repeated with `--no-verify` **after** every gate above had passed in the
foreground, and the tracking branch was asserted afterwards.

## Found, not fixed

`readProcInfo` propagates **every** `ps` error, so an ANSWERED "no such process"
(`ps -o ppid=,comm= -p <reaped pid>` exits 1) aborts both ancestry walks with
`complete=false`, which makes `IsKnownInteractiveHost` fail open. `processTTYVia`
sits four hundred lines up in the same file and its doc comment argues explicitly
against that shape for its own `ps`: *"It exits 1 for a pid it cannot find
(measured), which is a real 'no such process', not a probe that did not run — so
the allowlist form would turn every reaped pid into a non-answer and poison
hostIdentity's completeness for it."*

Deliberately **not** touched here and **not** filed, because I could not
establish it is a defect rather than the chosen polarity: "the ancestor is gone"
genuinely leaves the walk unable to reach a verdict, fail-open is this gate's
declared direction for an unknown, and the practical exposure is small (the walk
terminates cleanly at `ppid <= 1`, so reaching a dead ancestor needs a PID to die
mid-walk). Read off the code, not reproduced. Flagging it rather than folding a
behaviour change into an observability PR — and this counter is now what would
show it, since a `ps.proc_info` count climbing on an otherwise idle machine is
the evidence that does not exist today.

### DIFF
--- a/core/adapters/inbound/agents/processlifecycle/osutil.go
+++ b/core/adapters/inbound/agents/processlifecycle/osutil.go
@@ -486,6 +486,13 @@ func newAncestryReadsVia(read procInfoProbe) *ancestryReads {
 // can, and records only what the underlying read ANSWERED.
 func (r *ancestryReads) probe(ctx context.Context, pid int) (ppid int, cmd string, err error) {
 	if answer, hit := r.seen[pid]; hit {
+		// #1534: a hit starts no child, so runProbe never sees it, and an
+		// "N answered" that silently excluded hits would understate how often
+		// this probe is ASKED. #1544's own hand-back named this — "a memo now
+		// also hides how often the probe runs" — so the hit is counted where it
+		// happens, on the kind the underlying read would have used, and
+		// published as its own outcome rather than folded into the runs.
+		observeProbeMemoHit(probePSProcInfo)
 		return answer.ppid, answer.cmd, nil
 	}
 	ppid, cmd, err = r.read(ctx, pid)
--- a/core/adapters/inbound/agents/processlifecycle/osutil_darwin.go
+++ b/core/adapters/inbound/agents/processlifecycle/osutil_darwin.go
@@ -61,7 +61,7 @@ func processTTYVia(ctx context.Context, pid int, build shelloutCmd) (tty string,
 		// gives an empty appPath.
 		return "", true
 	}
-	out, err := runProbe(ctx, build)
+	out, err := runProbe(ctx, probePSTTY, build)
 	if !probeAnswered(err) {
 		return "", false
 	}
@@ -290,6 +290,12 @@ func (m *bundleIDMemo) record(appPath, id string) {
 // the consumer does, not because one of the two is a mistake.
 func (m *bundleIDMemo) resolve(ctx context.Context, appPath string, probe bundleIDProbe) (string, error) {
 	if id, hit := m.lookup(appPath); hit {
+		// #1534, and the sibling of ancestryReads.probe's line: a hit starts no
+		// child, so the counter at runProbe cannot see it. Counting it here is
+		// what keeps "answered" from silently meaning "answered, of the ones
+		// that were not memoized" — #1544 flagged exactly this in its own
+		// hand-back.
+		observeProbeMemoHit(probePlutilBundleID)
 		return id, nil
 	}
 	id, err := probe(ctx, appPath)
@@ -336,7 +342,7 @@ func bundleIDVia(ctx context.Context, appPath string, build bundleIDCmd) (string
 		return "", nil
 	}
 	plist := appPath + "/Contents/Info.plist"
-	out, err := runProbe(ctx, build(plist))
+	out, err := runProbe(ctx, probePlutilBundleID, build(plist))
 	if err != nil {
 		// No answeredExitCodes: for plutil ANY normal exit is an answer, which
 		// is the empty-variadic form's whole reason for existing — see
@@ -683,7 +689,7 @@ func kittyWindowIDForPIDVia(ctx context.Context, socket string, sessionPID int, 
 	if kittenPath == "" || socket == "" || sessionPID <= 0 {
 		return "", true
 	}
-	out, err := runProbe(ctx, build)
+	out, err := runProbe(ctx, probeKittenWindow, build)
 	if !probeAnswered(err) {
 		return "", false
 	}
@@ -752,7 +758,7 @@ func findKittyWindowIDForPID(windows []kittyLsWindow, sessionPID int) string {
 // ps already handles the comm-vs-argv-path distinction we need, and the
 // existing package is built around these bounded exec calls.
 func readProcInfo(ctx context.Context, pid int) (ppid int, cmd string, err error) {
-	out, err := runProbe(ctx, func(ctx context.Context) *exec.Cmd {
+	out, err := runProbe(ctx, probePSProcInfo, func(ctx context.Context) *exec.Cmd {
 		return exec.CommandContext(ctx, psPath, "-o", "ppid=,comm=", "-p", strconv.Itoa(pid))
 	})
 	if err != nil {
@@ -825,7 +831,7 @@ func herdrClientPIDs(ctx context.Context, socketPath string) (pids []int, probed
 	if _, err := os.Stat(logPath); err != nil {
 		return nil, false
 	}
-	out, err := runProbe(ctx, func(ctx context.Context) *exec.Cmd {
+	out, err := runProbe(ctx, probeLsofHerdrClients, func(ctx context.Context) *exec.Cmd {
 		return exec.CommandContext(ctx, lsofPath, logPath)
 	})
 	if !lsofProbeRan(err) {
@@ -1034,9 +1040,18 @@ func resolveClientHostIdentityVia(ctx context.Context, pids []int, identify host
 			// an expired deadline reaches them anyway, but only this check
 			// turns "every probe failed" into "we stopped looking", and only
 			// this check makes the aggregate a bound a caller can observe.
+			//
+			// #1558/#1534: counted here because this branch is invisible
+			// everywhere else. It produces the same (nil, false) a failed lsof
+			// produces, so on a machine where the scan is chronically
+			// slow-but-successful a herdr host would never resolve and nothing
+			// would say so. ONE event per loop, not one per candidate left —
+			// the break means the loop never learns how many those were.
+			observeHerdrCandidatesAbandoned()
 			readAll = false
 			break
 		}
+		observeHerdrCandidateProbed()
 		host, complete := identify(ctx, pid)
 		if host.HerdrPaneID != "" {
 			// A candidate that is itself a multiplexer pane has no window of
--- a/core/adapters/inbound/agents/processlifecycle/probecount.go
+++ b/core/adapters/inbound/agents/processlifecycle/probecount.go
@@ -1,0 +1,293 @@
+// probecount.go counts what every bounded child process in this package did
+// (issue #1534).
+//
+// Six issues — #1485, #1492, #1513, #1524, #1533, #1537 — each fixed one probe
+// that collapsed "I could not look" into "I looked and there was nothing", and
+// every one of them closed with the same footer: nothing measures how often a
+// probe actually fails to answer. That is not a documentation gap. Three of
+// those fixes fail OPEN — IsKnownInteractiveHost admits on a walk it could not
+// complete, which is the right call on no evidence — so a probe failing
+// constantly and a probe that never fails produce the SAME observable daemon:
+// the #784 gate quietly stops gating and there is nothing anywhere to look at.
+// A counter is what keeps that visible.
+//
+// Three decisions here are load-bearing, and each is a choice against an
+// obvious alternative:
+//
+//   - COUNTED AT runProbe, which since #1547 is the only place in this package
+//     that starts a child (TestEveryChildProcessRunsThroughRunProbe). One
+//     counting site instead of eight, and a ninth probe is counted by existing
+//     rather than by someone remembering.
+//   - KEYED PER CALL SITE, not per tool. `lsof` runs on three different
+//     questions here and `ps` on two; an "lsof: 4 non-answers" that cannot say
+//     which question went unanswered is the single-scalar failure #1364 already
+//     paid for. TestEveryProbeSiteDeclaresItsOwnKind makes a shared bucket a
+//     build failure rather than a thing to notice.
+//   - ANSWERED means the child RAN TO A NORMAL EXIT. See probeOutcomeRule below
+//     for why that is the hard half of shellout.Answered and not each site's
+//     own allowlist — this file must not acquire a second copy of any call
+//     site's policy.
+//
+// A third outcome, memoHits, exists because #1544 put two memos in front of two
+// of these probes and said so in its own hand-back: "a memo now also hides how
+// often the probe runs". A `plutil` served from bundleIDMemo, or a `ps` served
+// from ancestryReads, starts no child and would be invisible here. Reporting
+// "N answered" while silently excluding hits is a figure that misleads in
+// exactly the way this issue exists to prevent, so a hit is counted where it
+// happens — in the memo — and reported beside the runs rather than folded into
+// them.
+package processlifecycle
+
+import (
+	"sort"
+	"sync/atomic"
+)
+
+// probeKind names ONE bounded child process this package starts: the question,
+// not the binary. Two kinds may run the same tool — `ps` answers both an
+// ancestry read and a controlling-TTY read — and merging them would produce a
+// count that cannot say which question stopped being answered.
+//
+// The values are stable, low-cardinality tokens: they are published in a
+// diagnostics bundle a bug reporter pastes, so they are read by humans and
+// diffed between captures.
+type probeKind string
+
+const (
+	// probePSProcInfo is readProcInfo's `ps -o ppid=,comm=` — the per-PID read
+	// both ancestry walks are built out of, and the one behind ancestryReads.
+	probePSProcInfo probeKind = "ps.proc_info"
+	// probePSTTY is processTTY's `ps -o tty=` (#1533).
+	probePSTTY probeKind = "ps.tty"
+	// probePlutilBundleID is bundleIDForAppPath's `plutil -extract
+	// CFBundleIdentifier` (#1524), the one behind bundleIDMemo.
+	probePlutilBundleID probeKind = "plutil.bundle_id"
+	// probeLsofCWD is the observer's CWDOf `lsof -d cwd`.
+	probeLsofCWD probeKind = "lsof.cwd"
+	// probeLsofWriter is the observer's WriterOf `lsof <path>` (#1537).
+	probeLsofWriter probeKind = "lsof.writer"
+	// probeLsofHerdrClients is herdrClientPIDs' `lsof <client log>` (#1485).
+	probeLsofHerdrClients probeKind = "lsof.herdr_clients"
+	// probePgrepDiscover is runPgrep, behind both FindByName and FindByCmdline
+	// — one run site, so one kind.
+	probePgrepDiscover probeKind = "pgrep.discover"
+	// probeKittenWindow is kittyWindowIDForPID's `kitten @ ls` (#1537).
+	probeKittenWindow probeKind = "kitten.window"
+)
+
+// allProbeKinds is every declared kind. It is the set
+// TestEveryProbeSiteDeclaresItsOwnKind grades the call sites against and the
+// order ProbeCounts reports in, so a kind that stops being used, or a call site
+// that names one nobody declared, is a failure rather than a silent zero.
+var allProbeKinds = []probeKind{
+	probePSProcInfo,
+	probePSTTY,
+	probePlutilBundleID,
+	probeLsofCWD,
+	probeLsofWriter,
+	probeLsofHerdrClients,
+	probePgrepDiscover,
+	probeKittenWindow,
+}
+
+// probeTally is one kind's three counters.
+//
+// Atomics rather than a mutex, for the reason PathConfiner's fixed array of
+// atomics gives: what these exist to observe is a BURST of non-answers under
+// load, and instrumentation that serializes exactly when the events arrive
+// together would blunt the thing it measures.
+type probeTally struct {
+	answered   atomic.Uint64
+	unanswered atomic.Uint64
+	memoHits   atomic.Uint64
+}
+
+// probeCounts maps every declared kind to its tally. The map is built once at
+// init and NEVER written again — only the atomics inside it are — so it needs
+// no mutex of its own. That is a property of this declaration, not a habit:
+// adding an entry at runtime would make the map a shared mutable one and every
+// read here a data race.
+var probeCounts = func() map[probeKind]*probeTally {
+	m := make(map[probeKind]*probeTally, len(allProbeKinds))
+	for _, kind := range allProbeKinds {
+		m[kind] = &probeTally{}
+	}
+	return m
+}()
+
+// undeclaredProbes counts observations naming a kind that is not in
+// allProbeKinds.
+//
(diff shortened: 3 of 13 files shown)


ingo-eichhorst:
### CodeScene follow-up (advisory) + a CI flake

**CodeScene** flagged the first push. All but one finding was in the new test files; two were worth fixing rather than suppressing, and both are now in the amended commit:

- `TestEveryProbeSiteDeclaresItsOwnKind` (Bumpy Road Ahead / Deep, Nested Complexity / Complex Method) — the AST walk is split out into `probeKindsBySite` and `probeKindArgument`, so the test body is assertions and the scan is a named thing. **Re-ran the M2 mutation against the refactored form** rather than assuming it survived, per "a rewritten guard replays its predecessor's cases":

  ```
  --- FAIL: TestEveryProbeSiteDeclaresItsOwnKind (0.01s)
      probecount_test.go:304: probeLsofCWD is passed by 2 call sites (process_darwin.go:83, process_darwin.go:126) — they share one bucket, so a climbing count cannot say which probe stopped answering.
      probecount_test.go:310: probeLsofWriter is declared but passed by no call site — it publishes a permanent zero row that no probe can ever move
  ```

- `diagnostics_service_test.go` (Low Cohesion, 8.40 → 7.20) — the probes tests moved to their own `diagnostics_probes_test.go`. That file already carried the bundle's shape, redaction, liveness and three generations of hook counters; a fifth unrelated subject in it is a file nobody can hold in their head. The shared wiring (`buildTestServiceWith`) stays next to the service it builds. **M3 re-run against the moved file**, same nine assertions red.

Left alone, and saying why rather than suppressing silently:

- `TestProductionProbeSitesCountUnderTheirOwnKind` (Complex Method, advisory) — four subtests, one per injectable site. Splitting them into four top-level functions would duplicate the shared rationale (which half of the coverage reaches which sites) four times for no reader benefit.
- `diagnostics_service.go` (Primitive Obsession, advisory, 9.31 → 9.03) — the `probesView` return struct's several `uint64` counts. They are counts; a type per column would be ceremony.
- **"Absence of Expected Change Pattern"**: `osutil_darwin.go` usually changes with `osutil_linux.go`/`osutil_other.go`. Correct to skip here — the counters live in untagged code (`shellout.go`, `probecount.go`, `osutil.go`), and the linux/other `newAncestryReads` stubs bind a read that always refuses, so no memo hit is reachable on those platforms and no stub needs a counterpart change.

**CI flake, not a regression.** The first push's `go-test` failed on `TestFSWatcher_EmitsEventsForTranscriptCreateAndModify` — the known load-flake (#1432), in `core/e2e`, touching nothing this PR changes. Re-run locally at `-race -count=5`: pass, 1.482s. Every other check on that push was green, including `ars-gate`, `build-test`, CodeQL and SonarCloud.

Gates re-run in the foreground after the amend: `go build ./core/...`, `GOOS=linux go build ./core/...`, `go test ./core/... -race -count=1`, and `tools/preflight.sh --only go` (all 7 PASS).

ingo-eichhorst:
### Second CodeScene pass

The split fixed the `diagnostics_service_test.go` **Low Cohesion** finding (that file dropped off the list; the new `diagnostics_probes_test.go` sits at 9.39 with advisory findings only). Two remained and both were worth flattening rather than suppressing:

- `TestEveryProbeSiteDeclaresItsOwnKind` (Bumpy Road Ahead, Deep Nested Complexity) — its three claims are now named helpers (`assertDeclarationMatchesAllProbeKinds`, `assertNoTwoSitesShareAKind`, `assertEveryDeclaredKindIsUsed`), leaving the test body as parse → derive → vacuity floor → three assertions.
- `declaredProbeKindIdents` (Complex Method) — the four-level AST walk is split into `collectProbeKindConsts` and `probeKindLiteral`.

**Both mutations re-run against the refactored form**, since a rewritten guard replays its predecessor's cases or it is not known to be a superset:

```
=== M2 (shared bucket) ===
--- FAIL: TestEveryProbeSiteDeclaresItsOwnKind (0.01s)
    probecount_test.go:289: probeLsofCWD is passed by 2 call sites (process_darwin.go:83, process_darwin.go:126) — they share one bucket, so a climbing count cannot say which probe stopped answering.
    probecount_test.go:290: probeLsofWriter is declared but passed by no call site — it publishes a permanent zero row that no probe can ever move
```

Plus one the refactor made worth adding, because `declaredProbeKindIdents` is now the thing three helpers depend on and a derivation that silently stopped deriving would read as health:

```
=== M10 (a ninth probeKind constant, absent from allProbeKinds) ===
--- FAIL: TestEveryProbeSiteDeclaresItsOwnKind (0.01s)
    probecount_test.go:276: the source declares 9 probeKind constants but allProbeKinds lists 8 — a kind that is not in allProbeKinds has no counter
```

Still left alone, deliberately: `TestProductionProbeSitesCountUnderTheirOwnKind` and `TestProbesBundleReportsCounts` (Complex Method — both are one coherent subject whose subtests share a rationale that splitting would duplicate), and Primitive Obsession on the two count-carrying structs (they are counts; a type per column is ceremony). CodeScene is advisory here per AGENTS.md, and these three are the reasons rather than the suppressions.

Re-verified in the foreground after the amend: `go build ./core/...`, `GOOS=linux go build ./core/...`, `go test ./core/... -race -count=1`, `go test ./core/adapters/inbound/agents/processlifecycle/... -race -count=4` (46.7s), `tools/preflight.sh --only go` (7/7 PASS), `tools/preflight.sh --only arch` (PASS).

The previous push's `go-test` and `build-test` both went green, which settles the earlier red as the #1432 fswatcher flake rather than anything in this change.

ingo-eichhorst:
### Third CodeScene pass — critical gate now clean

The second refactor cleared the **critical** rules gate: `Bumpy Road Ahead`, `Deep, Nested Complexity` and `Low Cohesion` are all gone. Only the advisory gate remains red.

One advisory finding was mine from the cleanup pass — `Complex Conditional` on `probesSince`'s three-way OR — so it is extracted into `probeDeltaMoved`, with a comment on why all three counters belong in it (a memo hit appearing where none was expected is exactly the #1544 confusion these deltas exist to catch; a predicate ignoring it would let that pass as silence). `probesSince` is the load-bearing half of the vacuity claim, so **M4 was re-run against the extracted predicate** rather than assumed:

```
--- FAIL: TestProbeCounterSeparatesAnsweredFromUnanswered (0.02s)
    probecount_test.go:130: probe counters moved by map[ps.tty:{ps.tty 1 1 0}], want map[ps.tty:{ps.tty 1 0 0}] — a child that ran to a normal exit is an ANSWER, and is counted by nobody as a non-answer
    probecount_test.go:138: probe counters moved by map[ps.tty:{ps.tty 1 1 0}], want map[ps.tty:{ps.tty 1 0 0}] — a NON-ZERO normal exit is still an answer: probeOutcomeRule counts the child running, not the call site's allowlist
```

The remaining four advisory findings stand, with reasons rather than suppressions:

- `TestProductionProbeSitesCountUnderTheirOwnKind` and `TestProbesBundleReportsCounts` (Complex Method) — each is one coherent subject whose parts share a rationale that splitting would duplicate. The first in particular carries the "which half of the coverage reaches which sites" note that four separate functions would each need.
- Primitive Obsession on `diagnostics_service.go`'s `probesView` return struct and `daemon_smoke_test.go`'s decode struct — they are counts; a named type per column is ceremony.

AGENTS.md: CodeScene is advisory and "a red result is a prompt to look closer, not something to chase to green before merging". Three passes of looking closer produced two real readability improvements and four documented declines; that is where I am stopping.

Foreground re-verification after this amend: `go build ./core/...`, `GOOS=linux go build ./core/...`, `go test ./core/... -race -count=1`, `gofmt -l core/` — all clean.

Issue #1615, opened 2026-08-16
Image snapshots still differ byte-for-byte on a GitHub runner after the scale fix — the rasterisation half of #1530


ingo-eichhorst:
Split out of #1530 (blocker 1), after PR #1614 fixed the half that was masking this one.

## What #1530 recorded, and what turned out to be true

#1530 asked whether the 35 image-snapshot failures on a GitHub runner were a **backing-scale** difference or a **text-rasterisation** one, and noted that the error text could not distinguish them because it reports point size.

It could, in fact, distinguish them, and the answer is **both — stacked, with the first hiding the second.**

In SnapshotTesting 1.19.2's `Snapshotting/NSImage.swift`, `compare(_:_:precision:perceptualPrecision:)` emits the `@`-qualified message from exactly one place:

```swift
guard oldCgImage.width == newCgImage.width, oldCgImage.height == newCgImage.height else {
  return "Newly-taken snapshot@\(new.size) does not match reference@\(old.size)."
}
```

— a **pixel-dimension** mismatch, which returns before any byte is compared. A rasterisation difference falls through to the `memcmp` below and produces the *unqualified* `"Newly-taken snapshot does not match reference."` with no sizes at all.

All 35 pre-#1614 failures printed the qualified form, so all 35 were pixel-count mismatches. #1614 pins the rasterisation scale, and the message form flipped:

| | qualified (pixel counts) | unqualified (bytes) |
|---|---|---|
| before #1614 | 35 | 0 |
| after #1614 | **0** | **36** |

(36 rather than 35 because the pre-#1614 run hung partway through `SessionRowSnapshotTests` and never reached the rest — #1530's blocker 2, also fixed there.)

So: the scale half is fixed and proven. **This issue is the byte half, which nothing had ever seen before because the pixel-count guard returned first.**

## Measured facts

Runner backing scale is now printed by the workflow rather than inferred:

```
host backing scale: 1.0× (NSScreen.main: 1.0)
```

Toolchains:

| | Swift | Xcode | build |
|---|---|---|---|
| `macos-latest` runner | 6.3.3 | 26.6 | 17F113 |
| developer Mac (this repo's reference host) | 6.2.4 | 26.3 | 17C529, macOS 26.5 (25F71) |

**The residual is content-dependent, not a blanket difference.** On the runner, per suite:

| suite | failing / total on the runner |
|---|---|
| `SessionRowSnapshotTests` | 24 / 24 |
[...]

PR #1628, opened 2026-08-16
ci(macos): publish the image-snapshot pixels from the runner, and a four-primitive fixture to read them with (#1615)

ingo-eichhorst:
Steps 1 and 2 of #1615. **Step 3 — the policy call — is deliberately not in here**; it is
the user's, and the point of this PR is to hand it the evidence it was missing.

## What changed

**`swift-snapshot-evidence`, a third job in `macos-swift.yml`.** Runs exactly the five
suites `swift-test` skips, with `SNAPSHOT_ARTIFACTS` set, and uploads the newly-taken
images plus the committed references beside them. It does **not** judge the pixels —
those failing is its expected outcome — but it does fail loudly when it cannot
*collect*: a hung, truncated or zero-test run, a primitive count that disagrees with
the fixture's own manifest, or no failure image at all. "Collected no evidence" and
"found no difference" must not print the same thing.

`SNAPSHOT_ARTIFACTS` **is** honoured by the pinned SnapshotTesting 1.19.2 — verified by
reading `Sources/SnapshotTesting/AssertSnapshot.swift:469` in the resolved checkout, not
assumed. Two things it does *not* write: the reference (already on disk, so the job
copies `__Snapshots__` into the artifact) and the **difference image**, whose whole
branch is behind `__XCODE_BUILT_PRODUCTS_DIR_PATHS`, which SwiftPM never sets — so there
is no diff image under `swift test` by construction, and the diff is computed off the
pair after download.

**`RasterPrimitiveEvidenceTests`.** Six one-thing-each views through the same pinned-scale
rasteriser, publishing raw-pixel SHA-256 + PNG + `.raw` on every host: a 3 × 2 matrix of
{flat fill, stroked bezier, text, SF Symbol} × {explicit sRGB, `Color.secondary`}. No
committed reference and no `assertSnapshot`, so it cannot fail for a rasterisation
difference and therefore *runs to completion on the host where the answer is
"different"*.

**`ImageSnapshotCIScopeTests` now parses the workflow per invocation**, not as one union
over the file, plus two new assertions (the evidence job runs exactly the suites the gate
skips; the two classifications are disjoint).

## Step 2's answers

**Which of the primitives differ? None. All six are byte-identical** between a
`macos-latest` runner and the reference Mac — including the translucent dynamic-colour
column, which was added specifically because the first round refuted the original
three. The hypothesis in #1615 — *"the failing renders are the ones drawing
glyph/vector content"* — **is dead as stated**: text, SF Symbols and antialiased vector
strokes all reproduce exactly.

**What does differ, and by how much.** 36 failures, three populations:

| population | n | magnitude | flat-region pixels\* |
|---|---|---|---|
| brand icons rasterised from inline SVG | 34 | maxΔ **1-5**/255 on 0.01-0.9% of pixels | 0 (56 in one) |
| `BackchannelRulesView` control labels | 2 | maxΔ **142** and **197** on 2.4-2.7% | 537 / 608 |
| …of which locale-dependent text | 1 | `150.000` vs `150,000` | — |

\* differing pixels whose whole 8-neighbourhood in the reference is one colour. An
antialiasing ramp cannot produce those, so it is the metric that separates "edge" from
"structural". Aggregate over all 36: 27 876 differing pixels of 5 640 640 (0.494%).

**Population 1 is antialiasing-scale and localised to one element.** Every
`SessionRowSnapshotTests` failure's differing pixels sit inside a single **28 × 28 px =
14 × 14 pt** box — exactly `SessionState.adapterIcon`'s `img.size = NSSize(width: 14,
height: 14)`. `GroupViewSnapshotTests` repeats that box once per row; the two
`HistoryViewSnapshotTests` failures are the Anthropic and OpenAI logos in the quota
forecast. Those icons are built by `NSImage(data:)` **from an inline SVG string**
(`SessionState.swift:1133`), which is a different rasteriser from anything the primitive
fixture reaches — a SwiftUI `RoundedRectangle().stroke()` reproduces byte-identically.

**Population 2 is not the same phenomenon.** Best whole-image translation match is at
`(dx,dy) = (0,0)` (95.5%), so it is not a layout shift; the differing pixels split 56%
lighter / 44% darker, and at 7× zoom the control labels sit **~1 device pixel (½ pt)
lower** on the runner with the same glyph shapes. Sub-point control metrics moving the
baseline onto a different pixel phase.

**The locale row is a separate defect** and is filed as **#1630**, not fixed here: the
committed reference encodes the recording machine's `de_DE` locale, so this test fails
for any contributor whose Mac groups thousands with a comma.

## The Xcode-pin lever: measured, and it is not the lever

A throwaway probe workflow (pushed, read, deleted in this branch) ran the five suites once
per Xcode installed on `macos-latest`:

```
/Applications/Xcode_26.0.1.app build=17A400  swift=6.2    Executed 49 tests, with 36 failures
/Applications/Xcode_26.0.app   build=17A400  swift=6.2    Executed 49 tests, with 36 failures
/Applications/Xcode_26.1.1.app build=17B100  swift=6.2.1  Executed 49 tests, with 36 failures
/Applications/Xcode_26.1.app   build=17B100  swift=6.2.1  Executed 49 tests, with 36 failures
/Applications/Xcode_26.2.0.app build=17C52   swift=6.2.3  Executed 49 tests, with 36 failures
/Applications/Xcode_26.2.app   build=17C52   swift=6.2.3  Executed 49 tests, with 36 failures
/Applications/Xcode_26.3.0.app build=17C529  swift=6.2.4  Executed 49 tests, with 36 failures
/Applications/Xcode_26.3.app   build=17C529  swift=6.2.4  Executed 49 tests, with 36 failures
/Applications/Xcode_26.4.1.app build=17E202  swift=6.3.1  Executed 49 tests, with 36 failures
/Applications/Xcode_26.4.app   build=17E202  swift=6.3.1  Executed 49 tests, with 36 failures
/Applications/Xcode_26.5.0.app build=17F42   swift=6.3.2  Executed 49 tests, with 36 failures
/Applications/Xcode_26.5.app   build=17F42   swift=6.3.2  Executed 49 tests, with 36 failures
/Applications/Xcode_26.6.0.app build=17F113  swift=6.3.3  Executed 49 tests, with 36 failures
/Applications/Xcode_26.6.app   build=17F113  swift=6.3.3  Executed 49 tests, with 36 failures
/Applications/Xcode.app        build=17F113  swift=6.3.3  Executed 49 tests, with 36 failures
```

`Xcode_26.3.0.app` is **build 17C529, Swift 6.2.4** — bit-for-bit the reference Mac's own
toolchain (`xcodebuild -version` there: Xcode 26.3, Build 17C529), on which the same
suites produce **zero** failures. The toolchain is not the variable. The OS is: runner
macOS **26.5.2 (25F84)** against reference **26.5 (25F71)**. Two comments that
attributed the residual to the Xcode gap are corrected in this PR rather than left
standing.

## Mutation evidence

Every assertion this PR adds or rewrites was seen red. Pasted output is in the
[mutation-evidence comment](https://github.com/ingo-eichhorst/Irrlicht/pull/1628#issuecomment-5306540374); summary:

| mutation | goes red |
|---|---|
| `text` primitive draws `""` | `text rendered a flat area — it drew nothing` |
| `fill` primitive draws a `Circle()` | `fill is supposed to be a flat fill` (8 ≠ 1) |
| `sfsymbol` renders the same view as `text` | `two primitives rasterised to identical bytes` (3 ≠ 4) |
| drop `--skip GroupViewSnapshotTests` from the gate | `--skip list and this file's classification disagree` (predecessor's case, kept) |
| **move** `--skip SessionRowSnapshotTests` gate → evidence | **2 failures** — and the predecessor parser is **green** on the identical mutation (2 tests, 0 failures) |
| drop `--filter HistoryViewSnapshotTests` from evidence | `the evidence job collects a different set of suites than swift-test skips` |
| drop `--skip LauncherHarnessTests` from evidence | `must skip the harness target and its class, and nothing else` |
| add `as: .pinnedImage` to the evidence-only suite | `an evidence-only suite now takes image snapshots` |
| run the evidence step on the reference host, where nothing fails | step exits 1: `not one of the five suites produced a failure image…` |

The last one is the "must fail loudly when it cannot run" guard, exercised against the
real YAML `run:` block extracted from the workflow rather than a transcription.

## A defect this found in the existing workflow

GitHub runs `shell: bash` as `bash --noprofile --norc -e -o pipefail`, so `-e` is already
on and `set -uo pipefail` does not clear it. The first run of the new job died at the
failing subshell before a single check ran. `swift-test`'s step has the identical shape,
which means **`swift_suite_verdict` has never been reached in CI for a failing run** —
the HUNG / TRUNCATED / `signal code 6` diagnosis that is the whole reason CI goes through
`tools/lib/swift-suite.sh` cannot print there. Filed as **#1629**, not fixed here.

## Cost

One extra macOS job per PR while the suites are skipped and #1615 step 3 is open. Noted
in the workflow header; narrowing it to `workflow_dispatch` or deleting it is the cheap
part once the policy is decided.

Refs #1615


ingo-eichhorst:
## Mutation evidence — pasted output

Every assertion this PR adds or rewrites, seen red. Restores were verified with
`git status --porcelain` after each; no `git stash`, no `git checkout --`.

### `RasterPrimitiveEvidenceTests`' kind checks

```
===== MUTATION P1: text primitive draws an empty string =====
RasterPrimitiveEvidenceTests.swift:202: error: -[…testPublishesEachPrimitivesPixels] :
  XCTAssertGreaterThan failed: ("1") is not greater than ("1") - text rendered a flat area — it drew nothing

===== MUTATION P2: 'fill' is no longer a flat fill (Rectangle -> Circle) =====
RasterPrimitiveEvidenceTests.swift:200: error: -[…testPublishesEachPrimitivesPixels] :
  XCTAssertEqual failed: ("8") is not equal to ("1") - fill is supposed to be a flat fill

===== MUTATION P3: sfsymbol renders the same view as text (vacuity guard) =====
RasterPrimitiveEvidenceTests.swift:226: error: -[…testPublishesEachPrimitivesPixels] :
  XCTAssertEqual failed: ("3") is not equal to ("4") - two primitives rasterised to identical bytes — the fixture is not rendering what it claims
```

### The rewritten workflow parse — predecessor's case kept

```
===== MUTATION W1: drop --skip GroupViewSnapshotTests from the GATE =====
ImageSnapshotCIScopeTests.swift:235: error: -[…testWorkflowSkipListMatchesTheClassification] :
  XCTAssertEqual failed: […without "GroupViewSnapshotTests"…] is not equal to […with…]
  - macos-swift.yml's --skip list and this file's classification disagree
```

### …and the case the predecessor could not see

Moving `--skip SessionRowSnapshotTests` out of the gating command and into the evidence
one. The union-over-the-file parser sees the same set of names either way.

```
===== MUTATION W2: MOVE --skip SessionRowSnapshotTests from the gate into the evidence job =====
--- (a) the PREDECESSOR parser (HEAD's union over the whole file) ---
Test Case '-[…testEveryImageSnapshotSuiteIsClassified]' passed (0.013 seconds).
Test Case '-[…testWorkflowSkipListMatchesTheClassification]' passed (0.001 seconds).
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.013 (0.013) seconds

--- (b) the NEW per-invocation parser ---
ImageSnapshotCIScopeTests.swift:267: error: -[…testEvidenceJobRunsExactlyTheSuitesTheGateSkips] :
  ("[…, "SessionRowSnapshotTests"]") is not equal to ("["LauncherHarnessTests", "LauncherTestHarness"]")
  - the evidence job must skip the harness target and its class, and nothing else
ImageSnapshotCIScopeTests.swift:235: error: -[…testWorkflowSkipListMatchesTheClassification] :
  - macos-swift.yml's --skip list and this file's classification disagree
	 Executed 4 tests, with 2 failures (0 unexpected) in 0.100 (0.100) seconds
```

Re-run against the post-CodeScene refactor of the same parser, still red on both
(`ImageSnapshotCIScopeTests.swift:274` / `:242`, `Executed 4 tests, with 2 failures`).

**That first run also found a real hole in my own diff**: an earlier wording of an
`::error::` message in the new job contained the literal string `--skip list`, and the
text scan read `list` as a suite name. The message now avoids the token and says why,
in a comment beside it. The predecessor's green above is from the corrected file — the
first attempt was contaminated and is not the evidence quoted here.

### The two new assertions

```
===== MUTATION W3: drop --filter HistoryViewSnapshotTests from the EVIDENCE job =====
ImageSnapshotCIScopeTests.swift:265: error: -[…testEvidenceJobRunsExactlyTheSuitesTheGateSkips] :
  - the evidence job collects a different set of suites than swift-test skips

===== MUTATION W4: drop --skip LauncherHarnessTests from the EVIDENCE job only =====
ImageSnapshotCIScopeTests.swift:267: error: -[…testEvidenceJobRunsExactlyTheSuitesTheGateSkips] :
  ("["LauncherTestHarness"]") is not equal to ("["LauncherHarnessTests", "LauncherTestHarness"]")
  - the evidence job must skip the harness target and its class, and nothing else

===== MUTATION W5: the evidence-only suite starts taking an image snapshot =====
ImageSnapshotCIScopeTests.swift:281: error: -[…testTheTwoClassificationsAreDisjoint] :
  XCTAssertTrue failed - an evidence-only suite now takes image snapshots: ["RasterPrimitiveEvidenceTests"]
```

### The "must fail loudly when it cannot run" guard

Run against the real `run:` block, extracted from the YAML with `yaml.safe_load` rather
than transcribed, on the reference Mac — where all five suites **pass**, so there is
nothing to collect:

```
collected 0 failure image(s) and 4 primitive image(s)
::error::not one of the five suites produced a failure image. Either SNAPSHOT_ARTIFACTS
stopped being honoured, or those suites now PASS on a runner — in which case
ImageSnapshotCIScopeTests' classification, and the skip list in swift-test derived from
it, is stale and #1615 has moved.
STEP_EXIT=1
```

And its complement, from the job's own run on the runner:

```
collected 36 failure image(s) and 7 of 7 primitive image(s)
```

`7 of 7` is the derived-count guard: the fixture writes one manifest line per primitive
and the shell reads it, so adding a seventh moved the expectation without anyone editing
the YAML.

Issue #1641, opened 2026-08-16
ars.yml's badge-push retry loop reports success after all 5 attempts fail


ingo-eichhorst:
Found while fixing #1639 (test.yml's shared-shell-libs loop). Not that defect — a
different member of the same family: a loop whose failure is not reported.

## What

`.github/workflows/ars.yml`, step "Commit badge update":

```yaml
run: |
  ...
  git commit -m "chore: update ARS badge [skip ci]"
  for i in 1 2 3 4 5; do
    git pull --rebase origin main && git push && break
    echo "Push attempt $i failed, retrying..."
    sleep $((i * 3))
  done
```

The loop's last statement is `sleep`, which succeeds. So when all five attempts fail
the `for` loop simply ends, `sleep`'s status 0 is the step's status, and the step —
and the job — go **green**. The badge was not pushed and no check says so.

`-e` does not save it: `git pull … && git push && break` is an `&&` list, and errexit
is suppressed for every command in such a list except the one following the final
`&&` (here, `break`). That suppression is correct and wanted — it is what makes the
retry loop a retry loop — but nothing then reads the exhausted-retries case.

## Measured

The step's shape reduced to its decisive lines, with every attempt failing, run under
GitHub's documented default shell for a step declaring no `shell:`
(`bash --noprofile --norc -eo pipefail`):

```
=== ars.yml's retry loop shape, all attempts failing ===
committed
Push attempt 1 failed, retrying...
Push attempt 2 failed, retrying...
Push attempt 3 failed, retrying...
STEP EXIT=0
```

## Why it matters

This is AGENTS.md's "a verification mechanism must fail loudly when it cannot run", and
its companion "absence of a finding and inability to look must never produce the same
output". The `Push attempt N failed` lines are in the log, so it is not fully silent —
but the only thing anyone reads, the check's colour, is identical to a successful push.
The observable symptom is a README badge that quietly stops tracking `main`, with a green
history behind it.

## Suggested fix

Make the exhausted case an explicit failure — the loop already knows it happened:

```yaml
  pushed=0
  for i in 1 2 3 4 5; do
    if git pull --rebase origin main && git push; then pushed=1; break; fi
    echo "Push attempt $i failed, retrying..."
    sleep $((i * 3))
  done
  if [ "$pushed" -ne 1 ]; then
    echo "::error::badge push failed after 5 attempts"
    exit 1
  fi
```

[...]

Issue #1639, opened 2026-08-16
test.yml's shared-shell-libs loop stops at the first failing file, and does not say the rest never ran


ingo-eichhorst:
Found while fixing #1635. Not that issue's defect, and not in `tools/lib/` at all —
in the CI step that runs it.

## What

`.github/workflows/test.yml`'s "Test the shared shell libs" step is a loop with no
`|| rc=1`:

```yaml
run: |
  for t in tools/lib/*_test.sh; do
    case "$t" in */posix-lint_test.sh) continue ;; esac
    echo "== $t =="
    bash "$t"
  done
```

There is no `shell:` and no `defaults:` block anywhere in `test.yml` (grepped), so the
step runs under GitHub's documented default, `bash --noprofile --norc -eo pipefail`.
Under `-e`, the FIRST `bash "$t"` that fails aborts the whole step — every test file
after it in glob order is never run, and nothing says so.

`tools/preflight.sh`'s `shell_lib_tests` (the local mirror of this same step) does the
opposite and runs all of them:

```sh
for t in tools/lib/*_test.sh; do
  [[ -e "$t" ]] || continue
  bash "$t" || rc=1
done
```

## Measured

Three fixture files, the middle one failing, run through each spelling verbatim:

```
=== test.yml's loop, under GitHub's documented default shell ===
== …/a_test.sh ==
== ran a_test.sh ==
== …/b_test.sh ==
== ran b_test.sh ==
STEP EXIT=1                <- c_test.sh never ran, and nothing says so

=== preflight.sh's shell_lib_tests, same three files ===
== …/a_test.sh ==
== ran a_test.sh ==
== …/b_test.sh ==
== ran b_test.sh ==
== …/c_test.sh ==
== ran c_test.sh ==
GATE EXIT=1
```

## Why it matters

There are nine `tools/lib/*_test.sh` files today, eight of which run in this step
(`posix-lint_test.sh` is skipped by name, deliberately). A failure in the first —
`changed-files_test.sh`, alphabetically — leaves **seven** unrun, and the step's output
ends after the failing file's own `== header ==` with no line distinguishing "the rest
passed" from "the rest were never executed". That is AGENTS.md's "absence of a finding
and inability to look must never produce the same output", and it is the shape #1635's
own family is made of: the log looks like the loop simply ended.

Concretely it costs a round trip per failure — fix the first red file, push, discover
[...]

Issue #1635, opened 2026-08-16
budget_run is not safe under a caller's `set -e`: a command that exits non-zero is reported as a TIMEOUT after burning the full budget


ingo-eichhorst:
Found while fixing #1633, which is the same defect in the sibling library
`tools/lib/swift-suite.sh`. Filed separately rather than fixed there because
#1633 was deliberately scoped to `swift_suite_run`.

`tools/lib/gate-budget.sh` is SOURCED (`tools/preflight.sh:151`), so
`budget_run` runs with whatever shell options its caller has, and it is not
safe under `-e`. Unlike #1633 this breaks the **ordinary** path, not only the
timeout path, and the failure it produces is a false accusation that burns the
whole remaining budget.

## Measured

Driven under GitHub's own `shell: bash` invocation,
`bash --noprofile --norc -e -o pipefail`, sourcing the library and calling
`budget_run` in a bare statement position (the shape the file's header
documents):

```
budget_run 2 bash -c 'sleep 60'     # the TIMEOUT path
  without -e: 124        <- BUDGET_TIMEOUT_RC, correct
  with    -e: 1

budget_run 20 bash -c 'exit 3'      # the ORDINARY path
  without -e: 3          <- the command's own status, correct
  with    -e: 1          ...after 19.55s, for a command that exited instantly
```

## Where it goes wrong

Traced with `-x`. The child is backgrounded as

```sh
{ "$@"; echo "$?" >"$statusfile.part" && mv -f "$statusfile.part" "$statusfile"; } &
```

That subshell inherits `-e`, so a command exiting non-zero aborts the child
**before it writes its status file**. The status file is the documented "it
finished" signal (see the comment at `gate-budget.sh:164`), so `budget_run`
never sees the child finish, polls to the full deadline, and reports a
**TIMEOUT for a command that exited in milliseconds**. The trace ends:

```
+ [[ 20 -ge 20 ]]
+ [[ -s /var/folders/.../irrlicht-gate-budget.h9Hxf73mpP ]]
+ _budget_kill_tree TERM 62027
+ kill -TERM 62027
                       <- aborts here: the pid is already gone
```

`_budget_kill_tree` then aborts at its own `kill` (an already-dead pid),
before its `return 0` — which is #1633's shape exactly. `wait "$pid"
2>/dev/null` at lines 182, 193 and 207 is the same hazard again: it reports the
child's non-zero or signalled status and aborts under `-e`.

[...]

PR #1649, opened 2026-08-16
fix(ci): make the replaydata deletion guard REFUSE when it cannot read its diff (#1645)

### LINKED ISSUE #1645: replaydata-deletion-guard.yml: a failing git diff reads as 'no disallowed deletions' and passes the gate
Found by the sweep in #1641. Same family — a guard whose inability to look produces the same
output as a clean result — but this one is a **merge gate**, so its blind spot is the most
expensive of the set.

Note this file was spot-checked and cleared twice while the family was being worked
(#1639's agent, then #1641's brief), both times on the strength of the LOOP, which is
genuinely fine: it accumulates into `$violations` and reports after the loop. The defect is
one line above the loop.

## What

`.github/workflows/replaydata-deletion-guard.yml`, "Detect deletions of load-bearing
replaydata":

```yaml
run: |
  set -euo pipefail
  ...
  deletions=$(git diff --name-only --diff-filter=D --find-renames=50% "$base...$head" -- 'replaydata/**' || true)
  ...
  if [ -n "$violations" ]; then
    echo "::error::PR deletes load-bearing replaydata ..."
    exit 1
  fi

  echo "OK: no disallowed deletions (orphan recordings, assessment.json, and renames are permitted)."
```

The `|| true` means a `git diff` that FAILS is indistinguishable from a `git diff` that found
nothing. `$deletions` is empty either way, the `while` loop reads nothing, `$violations` stays
empty, and the step prints its own PASS line and exits 0 — having examined no diff at all.

This is exactly the false fix AGENTS.md names elsewhere ("`|| true` is a FALSE fix where a
status is captured"), one level over: here it is stdout rather than `$?` that is captured, and
the effect is the same — the failure becomes an empty result, and an empty result is this
guard's success condition.

Realistic triggers: a bad/unfetched object for `$base` or `$head` (fork PRs, a force-pushed
base, a checkout whose `fetch-depth` changes), a `git` invocation error after a flag rename,
or an OOM/disk failure mid-diff. `fetch-depth: 0` makes the object case less likely today, but
that is one line of a checkout step away from not being true, and nothing would report it —
which is the point.

## Measured

The step body EXTRACTED from the workflow (only the two `${{ }}` SHA expressions substituted)
and run under `bash --noprofile --norc -e -o pipefail`, with `git diff` stubbed to fail the
way a bad object does:

```
Comparing merge-base(BASESHA, HEADSHA) ...head HEADSHA
fatal: bad object BASESHA
OK: no disallowed deletions (orphan recordings, assessment.json, and renames are permitted).
GUARD STEP EXIT=0
```

The control — same extracted body, `git diff` working and naming one real violation — shows
the guard is otherwise healthy, so the finding is about the failure path only:

```
Comparing merge-base(BASESHA, HEADSHA) ...head HEADSHA
::error::PR deletes load-bearing replaydata (catalog or referenced recordings).

Disallowed deletions:
  - replaydata/agents/scenarios.json  (matrix catalog)
...
GUARD STEP EXIT=1
```

## Suggested fix

Drop the `|| true` and let `set -e` do its job — or, if a legitimately-empty diff needs
distinguishing from a broken one, capture the status explicitly:

```sh
  if ! deletions=$(git diff --name-only --diff-filter=D --find-renames=50% "$base...$head" -- 'replaydata/**'); then
    echo "::error::could not diff $base...$head — the deletion guard examined NOTHING. This is not a pass."
    exit 1
  fi
```

Worth adding the same "it actually looked" evidence the rest of the repo's gates carry: a
count line (`examined N deleted path(s)`) makes "nothing was deleted" and "nothing was read"
different strings in the log as well as different exit codes.

## Not fixed in #1641's PR

Different workflow, different step — filed rather than fixed, per that issue's scope.

ingo-eichhorst:
Closes #1645

`.github/workflows/replaydata-deletion-guard.yml` is a **merge gate**, and its
inability to look produced its own PASS line. That makes it the most expensive
member of this family: the failure mode is to permit exactly what it exists to
prevent — the #268 deletion of a load-bearing recording.

## The statement that decides the step's status

This block was spot-checked and cleared **twice** while the family was worked
(#1639's agent, then #1641's brief). Both readings were of the classification
`while`/`case` loop, which is genuinely fine — it accumulates into `$violations`
and reports after the loop. The defect is the loop's **input**, one line above
it:

```sh
deletions=$(git diff --name-only --diff-filter=D --find-renames=50% "$base...$head" -- 'replaydata/**' || true)
```

Naming this precisely, because it is what the two clean reads missed:

- **The statement that decides the step's status** is `if [ -n "$violations" ]`.
- **The statements whose failure it cannot see** are every one that feeds
  `$violations`: the `git diff` capture above (`|| true` converts a failure into
  an empty string), and the two `${{ }}` expansions that produce `$base` and
  `$head`. A `case` walk over `$deletions` can only ever be as good as
  `$deletions`, and nothing inside the loop can tell an empty string handed over
  by a *failure* from one handed over by a *clean PR*. Both spellings of "no
  finding" and "could not look" collapse into the same empty variable, and empty
  is the gate's success condition.

## Two hazards, both measured — and which of the two I am fixing

**Hazard 1 (the issue's subject): `|| true`.** Re-measured on every run by the
committed reproduction in the lock (`M1` below shows the same thing as a
mutation of the real step):

```
Comparing merge-base(1111…, 2222…) ...head 2222…
fatal: bad object 1111…
Examined 0 deleted path(s) under replaydata/.
OK: examined 0 deleted path(s), none disallowed (…)
--- MUTANT EXIT=0
```

**Hazard 2 (found while checking base/head resolvability, and it is reachable
today by ordinary means):** the workflow declares `workflow_dispatch:`. A manual
dispatch carries no `github.event.pull_request`, so **both** `${{ }}` expressions
expand to the empty string and the diff becomes `git diff "..."`. Measured
against real git (git 2.50.1), not stubbed, because a claim about how git parses
`...` has to be made by git:

```
== probe: base='' head='' -> the workflow_dispatch expansion ==
status=0
output=[]
```

git reads `...` as `HEAD...HEAD`. Pre-fix, pressing "Run workflow" produced a
**green** run whose log said `OK: no disallowed deletions`, having compared a
commit with itself. That measurement is now committed as the second half of the
lock's obligation 1.

**Base/head resolvability, stated honestly.** The checkout is
`actions/checkout@v7` with `fetch-depth: 0` and `ref: <head sha>`. I found **no
ordinary CI condition** that makes `base.sha` or `head.sha` unfetched — with full
depth, `base.sha` is normally reachable from `refs/heads/main`, and `head.sha` is
the commit the job checked out. So of the three refusals added, two are backed by
a reproduction (`|| true`, empty context) and the third — an object missing from
the checkout — is **defensive and marked as such**: it is not backed by a
reproduction, only by the observation that it is one edit to the checkout step
away (a `fetch-depth` change, a force-pushed or deleted base branch).

## The fix, and why this spelling

Three outcomes now, never two: **found and disallowed** (fail), **no deletions**
(pass), **could not determine** (fail, naming why).

I did not simply drop `|| true` and lean on `set -e`. The step's options are
`set -euo pipefail`, set by the step's own first line (the runner itself gives a
plain `bash -e {0}` — see the note at the end), so a bare drop *would* abort. It
would abort with a line number and nothing else, and the three "cannot look"
cases would still be indistinguishable from each other in the log. Instead each
gets an explicit, named refusal:

```sh
if [ -z "$base" ] || [ -z "$head" ]; then
  echo "::error::deletion guard REFUSED: no pull_request context (…). This is NOT a pass — nothing was examined."
  exit 1
fi

require_commit() { # <label> <sha>
  git rev-parse --verify --quiet "$2^{commit}" >/dev/null && return 0
  echo "::error::deletion guard REFUSED: the $1 commit $2 is not present in this checkout …"
  exit 1
}
require_commit base "$base"
require_commit head "$head"

if ! deletions=$(git diff … "$base...$head" -- 'replaydata/**'); then
  echo "::error::deletion guard REFUSED: git could not diff $base...$head …, so the guard examined NOTHING. This is NOT a pass."
  exit 1
fi
```

`if !` rather than `set -e` for two reasons: errexit is suppressed inside an `if`
condition, so the status is ours to *read*; and the message then names what could
not be done. `&& return 0` in `require_commit` keeps the failing `rev-parse`
inside an `&&` list, where errexit does not apply, so the message reports it
rather than a bare abort.

Plus the count line the issue asked for, so "nothing was deleted" and "nothing
was read" are different **strings** and not only different exit codes:

```
Examined 0 deleted path(s) under replaydata/.
OK: examined 0 deleted path(s), none disallowed (…)
```

**On the two refusals overlapping.** Deleting the empty-context refusal still
leaves the run failing, because an empty sha does not `rev-parse` either
(measured — M2 below reddens exactly one arm, the message one). What that refusal
uniquely buys is the **diagnosis**: "no pull_request context" instead of "the
base commit `` is not present in this checkout", which points at the checkout
rather than at the trigger. Said out loud rather than implied, because an
overlapping guard that nobody knows is overlapping is the next wrong dismissal.

## The lock

`tools/lib/replaydata-deletion-guard_test.sh`, following #1647's shape and its
author's recommendation: **extract the step from the real workflow and execute
it** against a git stub, rather than scan its text. A scan pins one spelling of a
guard; running the block pins the property. It runs on a real runner for free
because `test.yml` discovers `tools/lib/*_test.sh` — the shared runner's census
went from `found 12` to `found 13` with no wiring.

65 assertions. Anti-blindness guards carried over from #1647 and extended:

- a short or failed **extraction** is a NAMED refusal (`0 — the scan has gone
  blind, not the step clean`), plus arms asserting the extraction reached the
  classification loop *and* the pass line, so an extractor that stopped at the
  first dedent cannot clear the line floor;
- an **unmodelled `git` subcommand** in the stub returns a loud, distinctive 99
  naming the call, never a quiet 0;
- the `${{ }}` **substitutions are asserted to have happened**, and that no
  `${{` survived (which bash would reject as a bad substitution, failing for a
  confusing reason);
- the stub **records the argv of every `git diff`**, so a pass obtained by never
  diffing — this issue's own shape one level up — is caught (`having ACTUALLY
  run the diff`).

Two things the lock deliberately does not claim. `cell_is_live` reads the
**working tree**, so every arm runs against a throwaway tree under `$TMP`; the
real deletion-guarded catalog is never read, written or deleted. And `git mv` is
permitted because git's own rename detection reports an `R` that
`--diff-filter=D` drops, **not** because of anything the step does — detection is
on by default in modern git, so `--find-renames=50%` pins the threshold, not the
behaviour. The stub models that faithfully (it reports the old path as a deletion
only when the invocation does *not* ask for rename detection), so that arm grades
the invocation the step makes rather than claiming the step implements renaming.
This is stated in the test's header too.

## Every obligation's RED evidence

Ten deliberate mutations, each applied and reverted in its **own** foreground
call, each followed by `git status --porcelain` returning empty against a
committed baseline (per #1635's incident, no mutation was left on disk).

| # | mutation | obligation reddened | result |
|---|---|---|---|
| M1 | restore `deletions=$(… \|\| true)` | 4 — git that fails | 5 arms red |
| M2 | delete the empty-context refusal | 6a — diagnosis | 1 arm red |
| M3 | delete both `require_commit` calls | 6b — missing commit | 6 arms red |
| M4 | `cell_is_live` always true (over-refuse) | 5 — permitted cases | 2 arms red |
| M5 | `done <<< ""` (permit everything) | 2 — vacuity guard | 18 arms red |
| M6 | rename the step (blind the extractor) | anti-blindness | named refusal |
| M7 | add `git status` to the step | anti-blindness (stub) | exit 99, loud |
| M8 | drop `--find-renames=50%` | 5 — rename | 3 arms red |
| M9 | narrow the preflight `tools` trigger | trigger widening | 1 arm red |
| M10 | replace the diff with an empty literal | 3 — actually diffed | 32 arms red |

Pasted reds, abridged to the load-bearing lines:

**M1 — the defect itself, restored:**
```
  FAIL: git exits 128 mid-diff — expected [exit 1] got [exit 0 :: … fatal: bad object 1111… Examined 0 deleted path(s) … OK: examined 0 deleted path(s), none disallowed …]
  FAIL: ...saying the guard examined nothing — expected [output containing: examined NOTHING]
  FAIL: ...and NOT the step's pass line (current spelling) — expected [no output containing: OK: examined]
```

**M2 — the empty-context refusal, deleted (note: only the message arm reddens):**
```
  FAIL: ...naming the missing context — expected [output containing: no pull_request context]
        got [… ::error::deletion guard REFUSED: the base commit  is not present in this checkout …]
```

**M3 — `require_commit` calls deleted:**
```
  FAIL: a base commit this checkout does not have — expected [exit 1] got [exit 0 :: … OK: examined 0 deleted path(s), none disallowed …]
  FAIL: a head commit this checkout does not have — expected [exit 1] got [exit 0 :: …]
```

**M4 — the gate over-refuses (the arm that stops a fix nobody would keep):**
```
  FAIL: deleting …/scenarios/9-9_orphan/recordings/r1/events.jsonl is permitted — expected [exit 0]
        got [exit 1 :: … - …/9-9_orphan/recordings/r1/events.jsonl  (recording of live cell 'claudecode/9-9_orphan')]
  FAIL: deleting …/scenarios/9-9_orphan/metadata.json is permitted — expected [exit 0] got [exit 1 :: …]
```

**M5 — the gate permits everything (the vacuity guard; 18 arms, all of obligation 2):**
```
  FAIL: deleting replaydata/agents/scenarios.json fails the gate
  FAIL: deleting replaydata/orchestrators/gastown/scenarios/1-1.json fails the gate
  FAIL: deleting replaydata/agents/claudecode/regressions/8-8_frozen/assessment.json fails the gate
  FAIL: deleting …/scenarios/1-1_live/metadata.json fails the gate
  FAIL: deleting …/scenarios/1-1_live/recordings/r1/events.jsonl fails the gate
  FAIL: deleting …/scenarios/1-1_live/subagents/s1.jsonl fails the gate
  (+ the 12 annotation/naming arms)
```

**M6 — the step renamed:**
```
  FAIL: the '…' step body was extracted from … — expected [at least 30 non-blank lines]
        got [0 — the scan has gone blind, not the step clean]
```

**M7 — an unmodelled `git` subcommand:**
```
  got [exit 99 :: … STUB: unmodelled call: git status --porcelain ]
```

**M8 — rename detection dropped from the invocation:**
```
  FAIL: ...still asking for rename detection at the documented threshold
        got [git diff --name-only --diff-filter=D 1111…...2222… -- replaydata/** ]
  FAIL: a git mv of a live cell's recording is permitted — expected [exit 0] got [exit 1 …]
```

**M9 — the preflight trigger narrowed back:**
```
  FAIL: the tools gate fires on a diff touching .github/workflows/replaydata-deletion-guard.yml
        got [no match against: …|^\.github/workflows/(ars|macos-swift|test)\.yml$]
```

**M10 — the step never diffs at all:**
```
  FAIL: ...having ACTUALLY run the diff — expected [output containing: git diff]
  FAIL: ...as a three-dot base...head diff / ...restricted to replaydata/** / ...filtered to deletions
  (+ all 18 of obligation 2, and obligation 4's arms)
```

Obligation 1 has no mutation of its own by construction: it **is** the pre-fix
code, committed and re-measured on every run. It is the vacuity guard for the
whole file — if bash or git ever stopped behaving that way, every arm below would
be passing for the wrong reason and it says so by name.

## What ran where

**On a real GitHub runner:**

- **The fixed step itself**, dispatched on this branch:
  https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31960152598 — **failure**,
  and that is the *correct* outcome, not a broken build. `workflow_dispatch`
  carries no PR context, which is Hazard 2. The runner log:
  ```
  Comparing merge-base(, ) ...head
  ##[error]deletion guard REFUSED: no pull_request context (base='', head=''), so there is no base...head diff to read. This is NOT a pass — nothing was examined. Run this guard from a pull request.
  ##[error]Process completed with exit code 1.
  ```
  Before this PR that same dispatch printed `OK: no disallowed deletions` and
  went **green**. This is the fix demonstrated on a runner, not locally.
- **The lock**, via `test.yml`'s `tools/lib/*_test.sh` discovery — `test.yml`
  triggers on `pull_request` with no paths filter, so it runs on this PR.

**NOT on a runner, and worth being precise about:** the deletion guard's
`pull_request` trigger is filtered to `paths: ['replaydata/**']`. This PR touches
no `replaydata/` file, so the guard workflow will **not** run as a PR check here.
Everything above about its PR path is from the extracted-and-executed step
locally, plus the `workflow_dispatch` run on the real runner.

**Locally:** all ten mutations, the two hazard measurements, and the gates below.

## Gates

- `tools/preflight.sh --only tools` — **PASS**. `shell-lib-suite: tools/lib —
  found 13, skipped 0 (none), ran 13, failed 0` (12 before this PR).
- `tools/preflight.sh --only skills` — **not run, and not applicable**: no
  `.claude/skills/**/*.md` file is touched by this diff.
- Pre-push hook (`tools/preflight.sh --changed`, budget 420s): POSIX sh lint
  **PASS**, tools/lib shell-lib tests **PASS**, gofmt **PASS**, macOS Swift build
  + test **PASS**; every other gate **SKIP** (this diff cannot reach them).

## A note on the runner's actual shell — measured, and it is not what the family assumes

The dispatched run's log shows the step invoked as:

```
shell: /usr/bin/bash -e {0}
```

That is **errexit only** — no `--noprofile`, no `--norc`, and **no `-o
pipefail`**. `bash --noprofile --norc -eo pipefail {0}` is what `shell: bash`
gives, not what a step declaring no `shell:` gets. This step is unaffected: it
sets `set -euo pipefail` on its own first line, so from line 1 onward it runs
under exactly the options the lock uses. But the same premise is stated as fact
in `tools/lib/ars-badge-push_test.sh`, `tools/lib/shell-lib-suite.sh`,
`.github/workflows/test.yml` and AGENTS.md, for steps that do **not** set their
own options. Filed separately rather than fixed here (the siblings are out of
this PR's scope); I checked each of those call sites for a live dependency on
`pipefail` and found none, so it is a wrong premise rather than a live defect
today.

### DIFF
--- a/.github/workflows/replaydata-deletion-guard.yml
+++ b/.github/workflows/replaydata-deletion-guard.yml
@@ -22,6 +22,12 @@ name: Replaydata deletion guard
 #   - renames (git mv, incl. the id-prefix folder rename) — never counted as
 #     deletions.
 #
+# There is a THIRD outcome besides fail and pass, and it is a FAILURE: the
+# guard could not read a diff at all. Everything that could leave `$deletions`
+# empty for a reason other than "nothing was deleted" is a named refusal in the
+# step below, because an empty `$deletions` is this guard's success condition
+# and so an inability to look is otherwise byte-identical to a clean PR (#1645).
+#
 # Background: #268 Phase 1 introduced a blanket scenarios/+regression guard; it
 # was later made folder-granular + cell-aware so the pipeline can retire orphan
 # recordings and drop scenarios/ assessment.json, while still protecting every
@@ -53,11 +59,68 @@ jobs:
           head="${{ github.event.pull_request.head.sha }}"
           echo "Comparing merge-base($base, $head) ...head $head"
 
+          # -------------------------------------------------------------------
+          # THE REFUSALS (#1645). This step FAILS on "$violations is non-empty",
+          # and $violations is empty in two completely different situations: the
+          # diff was clean, and no diff was ever read. Each block below turns one
+          # way of not reading a diff into a loud failure, so that the second
+          # situation stops printing the first one's PASS line.
+          #
+          # Refusal 1 — no pull_request context. `on: workflow_dispatch` (above)
+          # has no `github.event.pull_request`, so BOTH expressions expand to the
+          # empty string and the diff below becomes `git diff "..."`, which git
+          # reads as HEAD...HEAD: exit 0, no output, PASS — having compared a
+          # commit with itself. Measured on git 2.50.1, and reachable by anyone
+          # pressing "Run workflow".
+          if [ -z "$base" ] || [ -z "$head" ]; then
+            echo "::error::deletion guard REFUSED: no pull_request context (base='$base', head='$head'), so there is no base...head diff to read. This is NOT a pass — nothing was examined. Run this guard from a pull request."
+            exit 1
+          fi
+
+          # Refusal 2 — an object this checkout does not have. `fetch-depth: 0`
+          # on the checkout above makes it unlikely, not impossible (a base
+          # branch force-pushed or deleted between the event and this job leaves
+          # base.sha unreachable from every ref the checkout fetched), and it is
+          # one edit to that step away from ordinary. Checked per side so the log
+          # names WHICH commit is missing. `&& return 0` keeps the failing
+          # rev-parse inside an && list, where errexit does not apply, so the
+          # message below is what reports it rather than a bare abort.
+          require_commit() { # <label> <sha>
+            git rev-parse --verify --quiet "$2^{commit}" >/dev/null && return 0
+            echo "::error::deletion guard REFUSED: the $1 commit $2 is not present in this checkout, so the base...head diff cannot be computed. This is NOT a pass — nothing was examined."
+            exit 1
+          }
+          require_commit base "$base"
+          require_commit head "$head"
+
           # Three-dot (base...head) diffs from the MERGE BASE to head — only what
           # this branch changed since it forked (avoids phantom deletions on
           # stale branches). --diff-filter=D = deletions; --find-renames=50% keeps
           # git mv as a rename (not a delete), documented + stable.
-          deletions=$(git diff --name-only --diff-filter=D --find-renames=50% "$base...$head" -- 'replaydata/**' || true)
+          #
+          # Refusal 3 — the diff itself. Until #1645 this line ended in
+          # `|| true`, which turned a git that FAILED into an empty $deletions —
+          # this guard's success condition. The classification loop then read
+          # nothing, $violations stayed empty, and the step printed
+          # "OK: no disallowed deletions" and exited 0 having examined no diff at
+          # all; a git exiting 128 was byte-for-byte a clean PR. Captured with
+          # `if !` rather than left to the `set -e` above for two reasons:
+          # errexit is suppressed inside an `if` condition, so the status is ours
+          # to read, and the message then names what could not be done where a
+          # bare abort names only a line.
+          if ! deletions=$(git diff --name-only --diff-filter=D --find-renames=50% "$base...$head" -- 'replaydata/**'); then
+            echo "::error::deletion guard REFUSED: git could not diff $base...$head (see the fatal above), so the guard examined NOTHING. This is NOT a pass."
+            exit 1
+          fi
+
+          # "Nothing was deleted" and "nothing was read" are now different exit
+          # statuses. This makes them different STRINGS in the log too, which is
+          # what someone reading a green run actually sees.
+          deleted_count=0
+          if [ -n "$deletions" ]; then
+            deleted_count=$(printf '%s\n' "$deletions" | wc -l | tr -d ' ')
+          fi
+          echo "Examined $deleted_count deleted path(s) under replaydata/."
 
           # A scenarios/ recording folder is a LIVE cell iff it holds a
           # metadata.json at HEAD (the working tree == HEAD; we checked it out).
@@ -125,4 +188,4 @@ jobs:
             exit 1
           fi
 
-          echo "OK: no disallowed deletions (orphan recordings, assessment.json, and renames are permitted)."
+          echo "OK: examined $deleted_count deleted path(s), none disallowed (orphan recordings, assessment.json, and renames are permitted)."
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1027,6 +1027,50 @@ Before marking a ticket done, run the full suite — every layer must pass:
   `$?`-keyed rule, and #1639 about the sibling family. `preflight.sh`'s `tools`
   trigger gains `ars.yml`, its fourth widening for this reason (#1591, #1629,
   #1639), because that assertion lives entirely inside that gate.
+- The replaydata deletion guard:
+  `tools/lib/replaydata-deletion-guard_test.sh` does to
+  `.github/workflows/replaydata-deletion-guard.yml`'s "Detect deletions of
+  load-bearing replaydata" step what the bullet above does to ars.yml —
+  extracts it and EXECUTES it against a git stub, under the same
+  `bash --noprofile --norc -e -o pipefail`. It is the same family's most
+  expensive member, because that workflow is a **merge gate**: before #1645 its
+  diff was captured as `deletions=$(git diff … || true)`, and an empty
+  `$deletions` is the gate's SUCCESS condition — so a git exiting 128 produced
+  no violations, printed `OK: no disallowed deletions` and exited 0, permitting
+  exactly the #268 deletion the gate exists to refuse. **Where** the defect sat
+  is the part worth carrying: this block was spot-checked and cleared TWICE
+  during #1639 and #1641, both times on the strength of the classification
+  LOOP, which is genuinely fine. The statement that decides the step's status
+  is the loop's INPUT, one line above it — a `case` walk over `$deletions` can
+  only ever be as good as `$deletions`, and nothing in the loop can see that it
+  was handed an empty string by a failure rather than by a clean PR.
+  Three outcomes now, never two: deletions found and disallowed (fail), no
+  deletions (pass), and **could not determine** (fail, naming why). Three
+  refusals implement the third, and the second of them is a SECOND measured
+  hazard rather than defensive padding — `on: workflow_dispatch` carries no
+  `github.event.pull_request`, so both `${{ }}` expressions expand to the empty
+  string and `git diff "..."` is `HEAD...HEAD`: exit 0, no output, PASS. That is
+  measured against real git in the lock, not stubbed, since a claim about how
+  git parses `...` has to be made by git. Note honestly what each refusal
+  independently buys: deleting the empty-context one still leaves the run
+  failing, because an empty sha does not rev-parse either — what it uniquely
+  buys is the DIAGNOSIS ("no pull_request context" rather than "the base commit
+  is not present in this checkout", which points at the checkout instead of the
+  trigger). The commit-presence refusal is the one that is defensive: with
+  `fetch-depth: 0` no ordinary condition was found that makes `base.sha`
+  unreachable, so unlike the other two it is not backed by a reproduction, only
+  by the observation that it is one edit to the checkout step away.
+  Two more things about the lock. `cell_is_live` reads the WORKING TREE, so
+  "live cell" and "orphan cell" are properties of the directory a body runs
+  in — every arm executes against a throwaway tree under `$TMP`, and the real
+  deletion-guarded catalog is never touched. And what the stub CANNOT grade is
+  said out loud: `git mv` is permitted because git's own rename detection
+  reports an R that `--diff-filter=D` drops, not because of anything the step
+  does, and detection is ON by default in modern git — so `--find-renames=50%`
+  pins the threshold, not the behaviour, and that arm grades the invocation the
+  step makes rather than claiming the step implements renaming.
+  `preflight.sh`'s `tools` trigger gains this workflow, its fifth widening for
+  this reason (#1591, #1629, #1639, #1641).
 - Factory: `go test ./tools/onboarding-factory/... -race -count=1`.
 - Replay: `tools/replay-fixtures.sh` — gated in CI by linux.yml, and run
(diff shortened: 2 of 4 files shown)


ingo-eichhorst:
## Runner confirmation (not assumed — read off the log)

All 16 PR checks green. The lock did not merely get *discovered* by `test.yml`; it **ran** on the macOS runner and I read its output rather than inferring it from a green job:

```
$ gh run view 31960285495 --log | grep -E 'replaydata-deletion-guard_test|shell-lib-suite: tools/lib'
go-test  Test the shared shell libs  == tools/lib/replaydata-deletion-guard_test.sh ==
go-test  Test the shared shell libs    PASS: ...it fires on a diff touching tools/lib/replaydata-deletion-guard_test.sh
go-test  Test the shared shell libs  replaydata-deletion-guard_test: ALL PASS
go-test  Test the shared shell libs  shell-lib-suite: tools/lib — found 13, skipped 1 (posix-lint_test.sh), ran 12, failed 0
```

`found 13` (was 12) with no wiring — the file joined the corpus by existing. The skip is CI's documented one (`posix-lint_test.sh` needs a linter the macOS image lacks); locally `--only tools` runs all 13.

So the two real-runner data points are:

| what | where | outcome |
|---|---|---|
| the **fixed step**, on `workflow_dispatch` | [run 31960152598](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31960152598) | **fails** with `REFUSED: no pull_request context` — correct; pre-fix this was a green `OK: no disallowed deletions` |
| the **lock** | [run 31960285495](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31960285495) | `ALL PASS` |

Still true and worth repeating: the guard's own `pull_request` trigger is filtered to `paths: ['replaydata/**']`, and this PR touches no `replaydata/` file, so it is **not** among the 16 checks above. The PR path is covered by the extracted-and-executed step locally; the dispatch run is what covers a real runner.

Also filed from this work: #1650 — the runner reports `shell: /usr/bin/bash -e {0}` for a step declaring no `shell:`, i.e. **no `pipefail`**, where four places in the repo state `--noprofile --norc -eo pipefail`. No live defect found (each call site checked), but three harnesses currently grade bodies under stricter options than production.

Issue #1663, opened 2026-08-16
SessionListView.formatResetTime stacks Calendar.current, a wall-clock Date() and an unpinned DateFormatter — the timezone site #1659 did not close


ingo-eichhorst:
Carried forward from #1659, which listed this as "latent, reached by nothing
today". #1659 closed the `HistoryView` half of the timezone family — including
the other latent site it named, `fullDateTimeFormatter`, which is now
`HistoryFormat.fullDateTime(_:timeZone:locale:)` behind the `\.formatTimeZone`
/ `\.formatLocale` seams. This one was deliberately **not** covered, and the
reason is worth having written down rather than rediscovered.

## The sites

`SessionListView.formatResetTime` (`platforms/macos/Irrlicht/Views/SessionListView.swift:1130-1141`)
and its sibling `formatClockTime` (`:1120-1125`).

```swift
private func formatResetTime(_ date: Date) -> String {
    let cal = Calendar.current          // machine
    let now = Date()                    // wall clock
    let f = DateFormatter()             // locale + timeZone both unpinned
    if cal.isDate(date, inSameDayAs: now) { f.dateStyle = .none; f.timeStyle = .short }
    else { f.dateFormat = "EEE H:mm" }
    return f.string(from: date)
}
```

`formatResetTime` renders as real `Text` at `:902` (`"resets \(…)"`);
`formatClockTime` feeds a tooltip line at `:1007`.

## Why #1659's mechanism was not simply extended to it

Because doing so would have made it *look* covered while leaving the largest
source of non-determinism in place. `\.formatTimeZone` + `\.formatLocale` reach
two of this function's three machine reads. The third is `Date()`, and it does
not merely shift the rendered time — it selects a **different format string**,
so the same input renders `"9:00"` before midnight and `"Fri 9:00"` after. A
snapshot pinned on zone and locale alone is still a snapshot of the day it was
recorded. Closing that needs an injectable "now", which is a separate seam with
a much wider blast radius (every wall-clock read in the app) and belongs in its
own reviewable diff.

## Why it is unreached today — verified, not assumed

Nothing renders it into a committed reference, because the whole chain sits
behind `guard let snap = session.metrics?.rateLimit else { return }`
(`SessionListView.swift:705`) and no snapshot fixture carries a rate-limit
window:

```
$ grep -rln "rate_limit\|rateLimit" platforms/macos/Tests/Fixtures/ platforms/macos/Tests/MockInstanceFiles/
[...]

Issue #1662, opened 2026-08-16
Snapshot suites pin ~8 real UserDefaults keys by hand in setUp/tearDown — the @AppStorage member of the #1630/#1659 host-dependency family


ingo-eichhorst:
Found by #1659's sibling sweep. #1630 closed the **locale** half of "a snapshot
reference encoding the recording machine", #1659 the **timezone** half; #1630's
author flagged `UserDefaults`/`@AppStorage` as the same "read from the MACHINE"
shape. This is that one, measured.

## First, the severe reading — and it is NOT real

The worry worth checking before anything else is whether a test run can modify
the developer's *actual app preferences*. **It cannot**, under either supported
invocation (`swift test`, `tools/preflight.sh --only swift`). `UserDefaults.standard`
in a SwiftPM test process resolves to the **`com.apple.dt.xctest.tool`** domain,
not to `io.irrlicht.app`. Measured on this machine:

```
$ defaults read com.apple.dt.xctest.tool
{
    contextPressureThresholdUnit = percent;   contextPressureThresholdValue = 80;
    debugMode = 0;                            displayMode = context;
    projectCostTimeframe = day;               showCostDisplay = 0;
    summaryDisplayMode = waiting;             …
}
$ defaults read io.irrlicht.app showCostDisplay        →  1
$ defaults read io.irrlicht.app projectCostTimeframe   →  year
$ defaults read io.irrlicht.app summaryDisplayMode     →  collapsed
$ defaults read io.irrlicht.app displayMode            →  Context
```

The xctest domain holds exactly the values `GroupViewSnapshotTests.setUp` and
`SessionRowSnapshotTests.setUp` assign; the app domain holds the user's own,
different ones. And a full `tools/preflight.sh --only swift` left
`~/Library/Preferences/io.irrlicht.app.plist` byte- and mtime-identical
(`1786897204 2319` before and after). So this is a fixture-fragility ticket,
not a data-loss one.

## The actual finding

Two suites pin real preference keys by hand:

- `GroupViewSnapshotTests` — 2 keys (`showCostDisplay`, `projectCostTimeframe`)
- `SessionRowSnapshotTests` — 6 keys (`displayMode`, `debugMode`,
  `showCostDisplay`, `ContextPressureThreshold.valueKey`/`.unitKey`,
  `summaryDisplayMode`)

each with a `setUp` that reads the old value into an `Any?` and a `tearDown`
[...]

Issue #1661, opened 2026-08-16
NotificationMasterGateTests leaves one empty plist per test in ~/Library/Preferences — 1136 accumulated


ingo-eichhorst:
Found by #1659's sibling sweep (the `UserDefaults` half #1630 flagged).

## The finding

`NotificationMasterGateTests` mints a fresh preference suite per test:

```swift
suiteName = "io.irrlicht.tests.NotificationMasterGate.\(UUID().uuidString)"
defaults  = UserDefaults(suiteName: suiteName)
```

and cleans up in `tearDown` with `removePersistentDomain(forName:)`. That
empties the domain but does **not** remove the file `cfprefsd` wrote for it, so
every run leaves one 42-byte empty plist per test behind, forever, in the
developer's real `~/Library/Preferences`.

Measured on this machine, 2026-08-16:

```
$ ls -1 ~/Library/Preferences/io.irrlicht.tests.NotificationMasterGate.*.plist | wc -l
    1136
$ plutil -p <newest>            # 42 bytes
{
}
$ ls -lTt … | tail -1
… Jul 31 16:08:21 2026 …        # oldest surviving file
```

…and it is still growing, measured across one `tools/preflight.sh --only swift`:

```
before: 1136
after:  1140
```

so 4 files per suite run, and the ~1136 accumulated since Jul 31 are the record
of ~284 local runs. Roughly 48 KB of content, 4.4 MB of allocated blocks.

## Why it matters beyond tidiness

`~/Library/Preferences` is the user's own directory, not a build artifact
directory — nothing in `git clean` or `swift package clean` reaches it, and
nobody would think to look. A test suite is the one thing that should not
accrete state outside the repo. It is also a slow leak in a directory
`cfprefsd` walks.

## Suggested fix

Delete the file as well as the domain in `tearDown`, or — better — stop
minting a UUID suite per test at all. The three tests that use `defaults` set
and read a handful of keys; one fixed throwaway suite name, removed once, has
the same isolation with a bounded footprint. Whichever is chosen, the check
that proves it is a test asserting that no file matching the suite prefix
survives a run, since "the domain is empty" and "the file is gone" are exactly
the two things this bug shows are not the same.

Cleanup of the existing 1136 is a one-liner the fixer should mention in the PR
rather than perform silently.

## Not in scope here

The *other* `UserDefaults` finding from the same sweep is filed separately: the
snapshot suites reading `UserDefaults.standard`. That one is about a shared
[...]

Issue #1630, opened 2026-08-16
BackchannelRulesViewSnapshotTests' reference PNG encodes the recording machine's LOCALE — it reads 150.000, a runner renders 150,000


ingo-eichhorst:
Found while collecting #1615's pixel evidence. Not fixed in PR #1628 — filed instead,
because the right fix is a judgement call (pin the view's locale in the test vs. pin
the formatter in the product) and #1615's own policy question is still open.

## The finding

`Tests/__Snapshots__/BackchannelRulesViewSnapshotTests/testBackchannelRuleContextTokens.1.png`
renders the threshold field as **`150.000`**. The same test on a `macos-latest`
runner renders **`150,000`**. Period vs comma as the thousands separator: that is a
locale difference, not a rasterisation one.

Mechanism, read from the source rather than inferred: the field is

```swift
TextField(isTokens ? "150000" : "85", value: Binding(…), format: .number)
```

— `platforms/macos/Irrlicht/Views/BackchannelRulesView.swift:149`. `format: .number`
resolves through `Locale.autoupdatingCurrent`. The reference host's locale is
`de_DE` (measured: `defaults read -g AppleLocale` → `de_DE`), where the group
separator is `.`; a GitHub runner is `en_US`, where it is `,`.

So the committed reference is a picture of one contributor's regional settings. Any
contributor whose Mac is not on a `.`-grouping locale fails this test locally, with a
"Snapshot does not match reference." that says nothing about why.

## Evidence

Runner-side render from PR #1628's `swift-snapshot-evidence` job
(<https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31934903492>), diffed against
the committed reference:

```
BackchannelRulesViewSnapshotTests | testBackchannelRuleContextTokens.1.png | 608x440
  | 7334 px (2.74%) | maxΔ=197 rgba=[197,197,197,0] | flat-region=608
```

`flat-region=608` is 608 differing pixels whose whole 8-neighbourhood in the reference
is one colour — an antialiasing ramp cannot produce those, which is what separates
this from the other 34 failures (max per-channel delta 1-5, flat-region 0).

## What this issue is NOT

It is **not** the whole of that suite's divergence, and saying so would be a wrong
dismissal. `testBackchannelRuleContextPressure` renders `85`, which is spelled
identically in both locales, and it still differs by 6370 px / maxΔ=142 across every
glyph in the card. Whatever that is, it is separate from — and larger than — the
[...]

PR #1664, opened 2026-08-16
fix(macos): stop the snapshot references encoding the recording host's TIMEZONE (#1659)

### LINKED ISSUE #1659: HistoryFormat's snapshot formatters pin the LOCALE but never the TIMEZONE — 8 references are held only by one suite's setUp
Found by #1630's sibling sweep. #1630 closed the **locale** half of "a snapshot
reference encoding the recording machine"; this is the **timezone** half, which it
deliberately did not close because the mechanism is different.

## The finding

`HistoryFormat`'s three cached formatters pin the locale and **never set
`timeZone`** — `platforms/macos/Irrlicht/Views/HistoryView.swift:1643-1650`:

```swift
private static let hourMinute  = posix("HH:mm")
private static let monthDay    = posix("M/d")
private static let weekdayClock = posix("EEE h:mm a")

private static func posix(_ format: String) -> DateFormatter {
    let f = DateFormatter()
    f.locale = Locale(identifier: "en_US_POSIX")   // locale pinned…
    f.dateFormat = format                          // …timezone never is
    return f
}
```

`DateFormatter.timeZone` therefore tracks `NSTimeZone.default`, i.e. the machine.

**8 of the 14 committed `HistoryViewSnapshotTests` references contain that
output** — `testHistoryPopulated`, `testHistoryTokens`, `testHistoryCO2`,
`testHistoryDrilldown`, `testActivityPopulated`,
`testActivityWideGridStaysWithinPanelWidth` (x-axis / column headers, via
`HistoryFormat.axisLabel` at `:1655`, call sites `:1326` and `:1143`), plus
`testQuotaForecastSingleProvider` and `testQuotaForecastMultiProvider`
(`Text("▲ cap \(HistoryFormat.clock(cap))")` at `:1468`, reached because the
fixture sets a non-nil `projectedCap`).

## Why this is not already fixed

It IS held today — by `HistoryViewSnapshotTests`' own `setUp`
(`platforms/macos/Tests/HistoryViewSnapshotTests.swift:13-19`), which sets
`NSTimeZone.default = UTC` and restores it in `tearDown`. So no reference is
currently wrong. Two things make it worth a ticket rather than nothing:

1. **The mitigation is per-suite, not a property of the snapshot strategy.**
   #1630 made locale and backing scale preconditions of taking a `.pinnedImage`
   snapshot at all — `Snapshotting.pinnedImage` is declared over
   `PinnedSnapshotHost`, whose only initializer applies both pins, so forgetting
   is a compile error. Timezone is the one host dependency left that a new suite
   can forget. Any future suite hosting `HistoryContentView` /
   `HistoryActivityContentView` / `HistoryQuotaForecastView` without copying that
   `setUp` records the developer's timezone into a reference.

2. **The formatters are process-global `private static let`s**, so their timezone
   is captured at *first use in the process*, not per test. Safe today —
   `axisLabel`/`clock` have no caller outside those views, and
   `HistoryPhase2Tests.swift:63-75` touches only `HistoryFormat`'s static
   *functions*, which do not force sibling static lets to initialise — but a
   single new caller reached before that `setUp` would silently pin the host's
   timezone into every subsequent render for the rest of the run.

## Why #1630's mechanism does not extend to it

#1630's seam is a SwiftUI `EnvironmentKey` (`\.formatLocale`) read by the view,
which works because a `FormatStyle` is constructed per render. A file-scope
`private static let DateFormatter` is not reachable from the view environment —
so this needs its own fix (pin `f.timeZone` explicitly and make the *value*
injectable, or move the formatters off file scope), not a second key.

## Also in scope, if someone is here anyway

Two latent siblings the same sweep found, each reached by no committed reference
today — cited so the dismissal carries its evidence, not asserted as harmless:

- **`SessionListView.formatResetTime`** (`platforms/macos/Irrlicht/Views/SessionListView.swift:1130-1140`)
  stacks `Calendar.current`, a wall-clock `Date()` (the same-day branch flips at
  midnight and changes the format string) and an unpinned `DateFormatter`
  (`.short` → 12h/24h + localised AM/PM; `"EEE H:mm"` → localised weekday). It
  renders as real `Text` at `:902`, not a tooltip. It is unreached only because
  its whole chain sits behind
  `guard let snap = session.metrics?.rateLimit else { return }` (`:705`) and **no
  fixture in the tree carries `rate_limit`** — neither
  `Tests/Fixtures/SessionRow/working-claude.json` nor `antigravity-ghost.json`,
  and `SessionListUnappliedGrantsWiringTests` builds a bare `SessionManager()`
  with zero sessions. The first snapshot that seeds a rate-limit window bakes
  locale, timezone and the recording *day* simultaneously.
- **`HistoryView.fullDateTimeFormatter`** (`:1194-1198`) is unpinned on both axes
  and is a genuine user-visible locale dependency in the shipping app — which is
  arguably correct there. It reaches no reference: its only call sites are
  `.tooltip(…)` (`:1181`) and `.accessibilityLabel(…)` (`:1182`), and `tooltip`
  resolves to `SessionListView.swift:117-140`, an `onHover`-only modifier that
  draws into a separate `NSPanel` and adds nothing to the view tree.

## Suggested acceptance

- A new image-snapshot suite cannot record the host's timezone without the same
  kind of failure #1630 produces for locale (compile error, or a check that fires).
- The lock drives **two** timezones through one view, not one — asserting
  "renders as UTC" is green on a UTC-pinned suite whether the formatter reads the
  input or the machine. `PinnedScaleSnapshotTests` and
  `PinnedLocaleSnapshotTests` are the shape.
- `HistoryViewSnapshotTests`' current `setUp` pin stays or is subsumed; the 8
  references above must not change.

ingo-eichhorst:
Closes #1659

`HistoryFormat`'s three cached formatters pinned `en_US_POSIX` and never set
`timeZone`, so every date the History panel draws came out of
`NSTimeZone.default` — the machine. **8 of the 14 committed
`HistoryViewSnapshotTests` references contain that output**, and they were right
only because that suite's own `setUp` assigned `NSTimeZone.default = UTC` and
`tearDown` put it back. Seven lines, in one suite, that the next suite to host
these views has to remember to copy.

## Could the `\.formatLocale` seam reach them? No — and the reason shapes the fix

`\.formatLocale` (#1630) works because a `FormatStyle` is constructed *per
render*, inside `body`, so the view can hand it a value read from the
environment. `HistoryFormat`'s formatters were file-scope
`private static let DateFormatter`s: one object per process, reachable from no
view. Setting a second environment key alone would have changed nothing — the
vacuous-green shape #1630 measured for `.environment(\.locale, …)`, arriving
inside its own fix.

So the seam is not "add a key". `HistoryFormat.axisLabel` / `clock` /
`fullDateTime` now take a **required** `TimeZone`, and the four views that call
them read `\.formatTimeZone`. A call site with no zone to pass does not compile
(mutation E below). Formatters are still cached — `DateFormatter` init is
expensive and these fire per axis tick and per grid cell — but keyed by
`(zone, format)` rather than being three process-global `static let`s, which
also makes the "captured at first use" hazard the issue describes
inexpressible.

**One correction to the issue, since it is load-bearing for the test design:**
that capture hazard was not real for the shape that existed. An unset
`DateFormatter` created *before* `NSTimeZone.default` is assigned still renders
the new zone — measured:

```
A. early before override          = 2023-11-14 23:13
-- after NSTimeZone.default = UTC --
A. early AFTER override           = 2023-11-14 22:13  <- tracks default
```

## Which zone, and why the default is not the obvious one

**`UTC`**, because it is what the committed references were recorded under — the
same argument `PinnedLocaleSnapshot.referenceLocale` is `de_DE` and
`referenceScale` is `2`. **No reference PNG was regenerated**; `git status` over
`*.png` is empty and all 53 still match byte-for-byte on the recording host.
`AGENTS.md` names re-recording to make a test pass as the move #1034 and #1044
both got wrong; pinning what is already there means this PR never has to make
that call.

The environment key's default is **`NSTimeZone.default`**, not
`TimeZone.autoupdatingCurrent`. The latter reads like the exact mirror of
`\.formatLocale`'s `Locale.autoupdatingCurrent` and is wrong here — measured on
this `Europe/Berlin` host:

| after `NSTimeZone.default = UTC` | reports |
|---|---|
| unset `DateFormatter` | **UTC** |
| `TimeZone.autoupdatingCurrent` | `Europe/Berlin` |
| `TimeZone.current` | `Europe/Berlin` |
| `Calendar.current.timeZone` | **UTC** |

so `autoupdatingCurrent` would have been a real behaviour change for exactly the
processes that assign it. `defaultValue` is computed rather than stored for the
same reason: read per access, so it tracks an assignment the way the unset
formatter did. A user in Berlin still sees Berlin times, by construction.

## The second machine read, which nobody had named

Pinning only `\.formatTimeZone` and deleting the `setUp` still reddened **6 of
the 14** references — and they all went green again under `TZ=UTC`, so something
else was reading the process.

**Swift Charts resolves `AxisMarks(values: .automatic(…))` through
`\.calendar`**, whose default `Calendar.current` *does* follow
`NSTimeZone.default` (table above). The deleted `setUp` had been holding tick
and gridline **positions** as well as label text — which is why
`testQuotaForecast*` moved even though its chart is `compact` and draws no axis
labels at all. `PinnedSnapshotHost` therefore pins three environments:
`\.formatTimeZone`, `\.calendar` (rebuilt as `.gregorian` + the pinned locale,
so the host's week rules go with it) and `\.timeZone`. Pinning the last two
changed no reference — measured, all 14 still match.

## `HistoryViewSnapshotTests`' `setUp` is deleted, not kept

Deliberately, and it is the strongest evidence in the PR. Keeping it would have
left those 8 references green whether or not the seam reached anything — #1630's
mutation B in advance. With no process-wide assignment left, they are the live
proof that every date this panel draws goes through the pin: a call site that
stopped passing `\.formatTimeZone` renders this `Europe/Berlin` machine's clock
and reddens them.

## Proof the pin takes effect

`PinnedTimeZoneSnapshotTests` (9 tests) drives **two** zones through one view —
the `PinnedScaleSnapshotTests` / `PinnedLocaleSnapshotTests` shape — so whatever
`NSTimeZone.default` is, at most one arm can agree with it. It grades
`PinnedSnapshotHost`, the object the suites snapshot, not a host this file
assembled.

`Text` cannot be read back the way `PinnedLocaleSnapshotTests` reads an
`NSTextField`: a window-less `NSHostingView` over three `Text`s reports
`accessibilityChildren() == nil` at the root (measured — SwiftUI builds the AX
tree lazily for a real AX client, and an `xctest` process has none). So the
proof is split: **rendered strings** asserted directly off `HistoryFormat`, and
at view level a **byte comparison of two rasterisations**, which is the shape
that discriminates — a seam reaching nothing makes both renders the machine's
clock, i.e. *identical*. Both a determinism guard (same zone twice must be
byte-identical) and a premise guard (the two zones must produce different text
for this fixture) run first, so the byte comparison cannot fail for the wrong
reason. A separate arm hosts an environment probe and asserts all three pins,
because a pixel comparison cannot tell "wrong label" from "wrong tick position".

## Mutation evidence — every one seen red, one per run, applied and reverted

**A. Pin `Asia/Tokyo` as `referenceTimeZone`** — the blast-radius measurement:

```
-[HistoryViewSnapshotTests testActivityPopulated] : failed - Snapshot does not match reference.
-[HistoryViewSnapshotTests testActivityWideGridStaysWithinPanelWidth] : failed - …
-[HistoryViewSnapshotTests testHistoryCO2] : failed - …
-[HistoryViewSnapshotTests testHistoryDrilldown] : failed - …
-[HistoryViewSnapshotTests testHistoryPopulated] : failed - …
-[HistoryViewSnapshotTests testHistoryTokens] : failed - …
-[HistoryViewSnapshotTests testQuotaForecastMultiProvider] : failed - …
-[HistoryViewSnapshotTests testQuotaForecastSingleProvider] : failed - …
	 Executed 337 tests, with 16 failures
```

**Exactly the 8 the issue names**, and nothing else in the other 45 committed
references — independent confirmation that its list is exact. The other 8
failures are this suite's own constants arms correctly reporting themselves
vacuous, since the mutation makes `referenceTimeZone == contrastingTimeZone`:

```
-[PinnedTimeZoneSnapshotTests testReferenceAndContrastingTimeZonesRenderDifferently] :
  XCTAssertNotEqual failed: ("Wed 7:13 AM") is equal to ("Wed 7:13 AM") —
  every both-sides assertion here is vacuous
```

**B. Revert the product seam** (delete both `f.timeZone = …` assignments,
keeping every pin) — the load-bearing one, and it reproduces #1630's mutation B
one family later:

```
-[HistoryViewSnapshotTests testActivityPopulated] : failed - Snapshot does not match reference.
-[HistoryViewSnapshotTests testActivityWideGridStaysWithinPanelWidth] : failed - …
-[HistoryViewSnapshotTests testQuotaForecastMultiProvider] : failed - …
-[HistoryViewSnapshotTests testQuotaForecastSingleProvider] : failed - …
-[PinnedTimeZoneSnapshotTests testAxisLabelsRenderInTheZoneTheyAreGiven] : XCTAssertEqual failed: ("23:13") is not equal to ("22:13")
-[PinnedTimeZoneSnapshotTests testAxisLabelsRenderInTheZoneTheyAreGiven] : XCTAssertEqual failed: ("11/14") is not equal to ("11/15")
-[PinnedTimeZoneSnapshotTests testClockRendersInTheZoneItIsGiven] : XCTAssertEqual failed: ("Tue 11:13 PM") is not equal to ("Tue 10:13 PM")
-[PinnedTimeZoneSnapshotTests testTheSameViewRendersDifferentlyUnderTwoPinnedTimeZones] :
  XCTAssertNotEqual failed: ("1128726 bytes") is equal to ("1128726 bytes") - the forecast strip
  rendered identically under UTC and Asia/Tokyo — the pinned zone is reaching nothing, and the
  render is coming from NSTimeZone.default (Europe/Berlin)
	 Executed 337 tests, with 15 failures
```

**Only 4 of the 8 references redden.** The four cost-chart ones stay green
because their labels are `M/d` and a one-hour Berlin offset does not change a
date — a pin that silently does nothing is invisible to half the snapshots, and
only the rendered-string assertions catch it. That is the entire reason this PR
asserts strings rather than trusting that `.environment(…)` "worked".

**C. Drop the `\.calendar` pin, keeping everything else** — the complement, and
the union of B and C is exactly the 8:

```
-[HistoryViewSnapshotTests testHistoryCO2] : failed - …          ┐
-[HistoryViewSnapshotTests testHistoryDrilldown] : failed - …    │ the 6 Swift Charts
-[HistoryViewSnapshotTests testHistoryPopulated] : failed - …    │ references; the 2
-[HistoryViewSnapshotTests testHistoryTokens] : failed - …       │ activity ones draw
-[HistoryViewSnapshotTests testQuotaForecastMultiProvider] : …   │ no chart and stay
-[HistoryViewSnapshotTests testQuotaForecastSingleProvider] : …  ┘ green
-[PinnedTimeZoneSnapshotTests testTheHostPinsEveryTimeZoneCarryingEnvironment] :
  XCTAssertEqual failed: ("Optional(Europe/Berlin)") is not equal to ("Optional(Asia/Tokyo)") -
  `\.calendar` is not pinned by the host — Swift Charts picks its date ticks through this,
  so the reference positions come from the machine
	 Executed 337 tests, with 8 failures
```

**E. A call site that omits the zone** — the compile error the required
parameter buys:

```
HistoryView.swift:1492:56: error: missing argument for parameter 'timeZone' in call
```

(Applied first at the Swift Charts axis inside `HistoryCostChart.body`, where it
compiles to `error: the compiler is unable to type-check this expression in
reasonable time` — still a failed build, but the clean message comes from the
smaller `windowCard` call site, so that is the one quoted.)

## The two latent sites #1659 documented

- **`HistoryView.fullDateTimeFormatter` — covered.** It is now
  `HistoryFormat.fullDateTime(_:timeZone:locale:)`, taking both axes as inputs
  and reached through `\.formatTimeZone` / `\.formatLocale`; the formatter is
  asserted directly (two zones, and two locales at one zone so the locale
  argument cannot be silently dropped). Stated rather than implied: **nothing
  grades the pass-through from `cell(…)` into `tooltipText`**, because neither
  call site produces a readable artifact — `.tooltip` is an `onHover`-only
  modifier drawing into a separate `NSPanel`, and `.accessibilityLabel` is not
  rasterised (the AX measurement above). What remains unproven there is the
  wiring of two arguments of distinct types, which cannot be transposed.
- **`SessionListView.formatResetTime` — deliberately NOT covered, filed as
  #1663.** A timezone pin reaches two of its three machine reads and leaves the
  one that matters most: `Date()` selects a different *format string* either
  side of midnight, so a snapshot pinned on zone and locale alone is still a
  snapshot of the recording day. Half-covering it would make it look safe. Its
  unreachedness is verified rather than assumed — `grep -rln "rate_limit\|rateLimit"
  platforms/macos/Tests/Fixtures/ platforms/macos/Tests/MockInstanceFiles/` returns
  nothing, and the two fixture files that exist carry no rate-limit window;
  `RateLimitInfo` *is* built in `QuotaMenuBarRendererTests`, which grades a
  different renderer and takes no image snapshot. A comment at the site points
  at #1663 before the first fixture bakes it in.

## `UserDefaults` sweep — the next member of this family

**The severe reading is not real, and that was checked first.** A test run
cannot modify the user's actual app preferences: `UserDefaults.standard` in a
SwiftPM test process resolves to `com.apple.dt.xctest.tool`, which holds exactly
the values the two suites' `setUp` assigns (`showCostDisplay = 0`,
`projectCostTimeframe = day`, `displayMode = context`, `summaryDisplayMode =
waiting`), while `io.irrlicht.app` holds this user's own different ones (`1`,
`year`, `Context`, `collapsed`). A full `tools/preflight.sh --only swift` left
`~/Library/Preferences/io.irrlicht.app.plist` identical in mtime and size
(`1786897204 2319` before and after).

Filed, not fixed here:

- **#1662** — the fixture-fragility finding. Two suites pin 8 real keys by hand;
  the domain also carries 9 keys neither suite pins and that persist between
  runs, and per #1523 an aborted run skips `tearDown` entirely.
- **#1661** — a separate, genuine leak found while measuring it:
  `NotificationMasterGateTests` mints a UUID preference suite per test and
  `removePersistentDomain` empties the domain without deleting the file. **1136
  empty 42-byte plists** have accumulated in `~/Library/Preferences` since
  Jul 31; the count went 1136 → 1140 across one gate run in this PR.

## Gates

| Gate | Result |
|---|---|
| `tools/preflight.sh --only swift` | **PASS** (exit 0) — 337 tests, 0 failures |
| pre-push hook (`--changed --budget 540`) | **PASS** (exit 0) — `swift` and `gofmt` PASS, everything else correctly SKIP |
| `git status --porcelain -- '*.png'` | empty — **no reference regenerated** |

Exit codes read directly off the next line, never through a pipe: per #1523 the
suite can abort in a way that truncates the run while every suite that did
report says "0 failures", and per `AGENTS.md` this repo's shell is zsh, where
`${PIPESTATUS[0]}` expands to nothing.

**CI will not grade the change that matters.** `HistoryViewSnapshotTests` is one
of the five suites `macos-swift.yml`'s `swift-test` job skips (#1615), so a
green check on this PR says nothing about the 8 references this PR is about. The
`swift-snapshot-evidence` job does run it — but its assertion is that *at least
one* of those suites produced a failure image, which is satisfied by the
pre-existing #1615 rasterisation drift regardless of anything here. The local
`swift` gate is the only place these references are graded.
`PinnedTimeZoneSnapshotTests` itself is **not** skipped and is host-independent
by construction, so the pin's proof does run in CI even though its subject does
not.

## Not addressed

#1615's rasterisation question, and the `@AppStorage` family (#1662), which is
the next one of these and has its own blast radius.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1399,12 +1399,50 @@ Before marking a ticket done, run the full suite — every layer must pass:
   `format: .number` at `BackchannelRulesView.swift:149` is the only
   `FormatStyle` in the app (every other numeric render is `String(format:)`,
   which takes no locale, or an `Int` interpolation, which carries no grouping).
-  The sibling family that is NOT closed is **timezone**: `HistoryFormat`'s
-  formatters pin `en_US_POSIX` but never set `timeZone`, and 8 of the 14
-  `HistoryViewSnapshotTests` references contain their output, held only by that
-  suite's own `setUp` — #1659, kept separate because a process-global
-  `private static let DateFormatter` is not reachable from the view environment
-  the way a `FormatStyle` is, so it needs its own seam rather than a second key.
+  **The sibling family is TIMEZONE, and closing it found a second machine read
+  nobody had named** (#1659). `HistoryFormat`'s formatters pinned `en_US_POSIX`
+  and never set `timeZone`, so 8 of the 14 `HistoryViewSnapshotTests` references
+  were held only by that suite's own `setUp` assigning `NSTimeZone.default`.
+  The seam is `\.formatTimeZone` (`Irrlicht/Views/FormatTimeZoneEnvironment.swift`),
+  and unlike `\.formatLocale` it could not be read by the formatters as they
+  stood — a file-scope `private static let DateFormatter` is reachable from no
+  view — so `HistoryFormat.axisLabel`/`clock`/`fullDateTime` take a `TimeZone`
+  as a REQUIRED argument and the views pass what they read. A zone-less call
+  site is `error: missing argument for parameter 'timeZone' in call`. The
+  default is `NSTimeZone.default` and **not** `TimeZone.autoupdatingCurrent`,
+  which reads like the obvious mirror of `\.formatLocale`'s
+  `Locale.autoupdatingCurrent` and is wrong: measured, after
+  `NSTimeZone.default = UTC` an unset `DateFormatter` renders UTC while
+  `TimeZone.autoupdatingCurrent` **and** `TimeZone.current` both still report
+  the host zone, so either would have been a real change for exactly the
+  processes that assign it. The pinned zone is `UTC` because that is what the
+  references were recorded under, so nothing was regenerated.
+  Two things there are worth carrying beyond the fix. **The pin is three
+  environments, not one**: pinning only `\.formatTimeZone` still reddened 6 of
+  the 14, because **Swift Charts resolves `AxisMarks(values: .automatic(…))`
+  through `\.calendar`** — whose default `Calendar.current` *does* follow
+  `NSTimeZone.default` — so the deleted `setUp` had been holding tick and
+  gridline POSITIONS as well as label text, which is why the `compact` quota
+  chart moved despite drawing no axis labels at all. `PinnedSnapshotHost` now
+  pins `\.formatTimeZone`, `\.calendar` (rebuilt `.gregorian` + the pinned
+  locale, so the host's week rules go too) and `\.timeZone`. And **the two
+  halves are separately invisible**: reverting the product seam while keeping
+  the pins reddens 4 references, dropping the `\.calendar` pin reddens the other
+  6, and the union is the 8 — reverting the seam leaves every `M/d` label green
+  on a `Europe/Berlin` host, because a one-hour shift does not change a date.
+  That is #1630's mutation B one family later: only the rendered-string
+  assertions in `PinnedTimeZoneSnapshotTests` catch it. That suite deletes
+  `HistoryViewSnapshotTests`' `setUp` rather than keeping it, deliberately —
+  with no process-wide assignment left, those 8 references are themselves the
+  live evidence that the seam reaches every date the panel draws.
+  Still open in this family, each with its evidence in the issue rather than
+  asserted here: `SessionListView.formatResetTime` (#1663 — a timezone pin
+  reaches two of its three machine reads and leaves the `Date()` that picks the
+  format string, so it was left whole rather than half-covered) and
+  `@AppStorage`/`UserDefaults` (#1662 — and no, a test run does not modify the
+  user's real app preferences: `UserDefaults.standard` under `swift test`
+  resolves to `com.apple.dt.xctest.tool`, measured against a byte-identical
+  `io.irrlicht.app.plist` across a full run).
   The suite also aborts intermittently (#1523) in a way that **truncates the
   run** while every suite that did report says "0 failures" — so the exit code
   is the only reliable signal, never the last totals line you can see. That is
--- a/platforms/macos/Irrlicht/Views/FormatTimeZoneEnvironment.swift
+++ b/platforms/macos/Irrlicht/Views/FormatTimeZoneEnvironment.swift
@@ -1,0 +1,87 @@
+import SwiftUI
+
+/// The time zone `DateFormatter`-based rendering resolves through (#1659).
+///
+/// ## Why this exists at all
+///
+/// `HistoryFormat`'s cached formatters pinned `locale` to `en_US_POSIX` and
+/// never set `timeZone`, so every date string the History panel draws came out
+/// of `NSTimeZone.default` — the machine. Eight of the fourteen committed
+/// `HistoryViewSnapshotTests` references contain that output, and they were
+/// correct only because that one suite's `setUp` assigned
+/// `NSTimeZone.default = UTC` and put it back in `tearDown`. A new suite hosting
+/// `HistoryContentView` / `HistoryActivityContentView` /
+/// `HistoryQuotaForecastView` without copying those seven lines would have
+/// recorded the developer's own time zone into a reference.
+///
+/// This is the time-zone sibling of `\.formatLocale` (#1630), and it exists for
+/// the same reason: the value was being read from the MACHINE where the
+/// equivalent value was available from the INPUT. It is a separate key rather
+/// than a second use of that one because the two are separately overridable —
+/// a host may want POSIX-stable numbers and local wall-clock times, or the
+/// reverse — and because they are read by different machinery (`FormatStyle`
+/// vs `DateFormatter`).
+///
+/// ## Why the environment could NOT reach the formatters as they stood
+///
+/// `\.formatLocale` works for `BackchannelRulesView` because a `FormatStyle` is
+/// *constructed per render*, inside `body`, so the view can hand it a value it
+/// read from the environment. `HistoryFormat`'s formatters were file-scope
+/// `private static let DateFormatter`s: one object per process, built on first
+/// touch, reachable from no view. Setting this key alone would therefore have
+/// changed nothing — the vacuous-green shape #1630 measured for
+/// `.environment(\.locale, …)`.
+///
+/// So the product change is not "add a key". It is that `HistoryFormat`'s
+/// entry points now **require** a `TimeZone` argument, and the views that call
+/// them read it from here. A new call site that has no zone to pass is a
+/// compile error rather than a silent read of `NSTimeZone.default`.
+///
+/// ## Why the default is `NSTimeZone.default` and not `.autoupdatingCurrent`
+///
+/// Because `NSTimeZone.default` is, *by construction*, the value an unset
+/// `DateFormatter.timeZone` already resolved through — the same
+/// "by construction identical" bar `\.formatLocale`'s
+/// `Locale.autoupdatingCurrent` default met. A user in Berlin still sees Berlin
+/// times.
+///
+/// `TimeZone.autoupdatingCurrent` and `TimeZone.current` would NOT have met it.
+/// Measured on this host (`Europe/Berlin`), after `NSTimeZone.default = UTC`:
+///
+///     TimeZone.current               = Europe/Berlin   (does NOT follow it)
+///     TimeZone.autoupdatingCurrent   = Europe/Berlin   (does NOT follow it)
+///     unset DateFormatter            = 22:13 = UTC     (DOES follow it)
+///
+/// so either of those two would have been a real behaviour change for any
+/// process that assigns `NSTimeZone.default` — which is precisely the History
+/// snapshot suite, and would have quietly broken the eight references this
+/// change exists to protect.
+///
+/// `defaultValue` is a **computed** property for the same reason: read per
+/// access, so it tracks a later assignment exactly the way the unset formatter
+/// did (also measured — an unset `DateFormatter` created *before* the
+/// assignment still rendered UTC after it, so "captured at first use" is not
+/// the semantics these formatters had).
+private struct FormatTimeZoneKey: EnvironmentKey {
+    /// The time zone an unset `DateFormatter.timeZone` already resolves
+    /// through. Overriding this key with anything else is a deliberate act;
+    /// nothing in the app does it, and the snapshot hosts do.
+    static var defaultValue: TimeZone { NSTimeZone.default }
+}
+
+extension EnvironmentValues {
+    /// Time zone for `DateFormatter`-based rendering in this subtree.
+    ///
+    /// Read it and pass it to the formatter (`HistoryFormat.axisLabel(d,
+    /// bucketSeconds: s, timeZone: formatTimeZone)`) — like `\.formatLocale`
+    /// and unlike `\.locale`, nothing consults this implicitly.
+    ///
+    /// Hoist it into a local before using it inside an escaping builder
+    /// closure (Swift Charts' `AxisValueLabel`, `ForEach`'s content): reading
+    /// a property wrapper from a captured `self` outside `body` evaluation is
+    /// what SwiftUI warns about, and the local is free.
+    var formatTimeZone: TimeZone {
+        get { self[FormatTimeZoneKey.self] }
+        set { self[FormatTimeZoneKey.self] = newValue }
+    }
+}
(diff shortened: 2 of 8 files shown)


ingo-eichhorst:
## CI cross-host evidence for the pin's proof

The PR body says CI cannot grade the 8 references (`HistoryViewSnapshotTests` is one of the five suites `swift-test` skips, #1615). It **can** grade the proof, and it did — `PinnedTimeZoneSnapshotTests` is not on that list and ran to completion on the `macos-latest` runner, a UTC host:

```
Test Suite 'PinnedTimeZoneSnapshotTests' started at 2026-08-16 19:13:57.671.
 …testAxisLabelsRenderInTheZoneTheyAreGiven]' passed (0.001 seconds).
 …testClockRendersInTheZoneItIsGiven]' passed (0.000 seconds).
 …testFullDateTimeRendersInTheZoneAndLocaleItIsGiven]' passed (0.001 seconds).
 …testReferenceAndContrastingTimeZonesRenderDifferently]' passed (0.000 seconds).
 …testTheDefaultFollowsNSTimeZoneDefaultAndNotAutoupdatingCurrent]' passed (0.000 seconds).
 …
	 Executed 289 tests, with 0 failures (0 unexpected) in 2.588 (2.635) seconds
```

That is the second host, and it is the one that matters for the design: the suite was written so that whatever `NSTimeZone.default` is, at most one arm can agree with it. Passing on a `Europe/Berlin` laptop **and** on a UTC runner is exactly the property `PinnedScaleSnapshotTests` established for scale — "a host-independence lock that can only be checked on one host is not a lock" — and it also confirms the one arm whose host-independence was reasoned rather than measured, `testTheDefaultFollowsNSTimeZoneDefaultAndNotAutoupdatingCurrent`, which picks a target zone the host is not already in.

Issue #1687, opened 2026-08-17
replaydata/ bash is excluded from bash-lint — and one file's shellcheck analysis is silently ABANDONED by a prose comment


ingo-eichhorst:
`tools/bash-lint.sh` (#1684) lints every bash file git knows about **except** three
declared exclusions. Two of them are one family: the recording rig's per-agent drivers
(`replaydata/**`) and the template they are generated from
(`tools/onboarding-factory/scripts/templates/drive-interactive.sh.tmpl`). #1684 excluded
them with a stated reason rather than blanket-disabling codes to reach zero, and filed
this to carry the rest of the work.

## Measured, at `d76aa137`, per-file, `--shell=bash --severity=warning`

31 bash files under `replaydata/`, plus the one `.tmpl`:

| code | n | what it is |
|---|---:|---|
| SC2034 | 68 + 10 | assigned, never used — the driver protocol's declared-variable seam |
| SC1125 | 8 | a `# shellcheck disable=SC2086 — <prose>` directive |
| SC1072/SC1073 | 2 | **shellcheck ABANDONS the file** (one file) |
| SC2155 | 1 | declare-and-assign masking a return value |

The two counts are per-file deliberately. **shellcheck given several files in ONE
invocation suppresses SC2034 for a name used in ANY of them** — measured:
`shellcheck tools/lib/swift-suite_test.sh` reports 2 findings, `shellcheck
tools/lib/swift-suite_test.sh tools/lib/await-gone.sh` reports 0. So a multi-file
figure understates, and `bash-lint.sh` lints one file at a time for that reason.

## The load-bearing one: `replaydata/_lib/drive/contracts.sh:39`

```sh
# normally — no trailing `return` (it would be unreachable dead code;
# shellcheck SC2317 flags exactly that).
```

A comment line whose **first word after `#` is the linter's own name** is parsed as a
shellcheck *directive*. This one is unparseable, so shellcheck reports SC1073/SC1072
and **abandons the file** — as-is, those two errors are the *only* output it produces
for `contracts.sh`.

Measured rather than asserted: rewording that one line (`# SC2317 (shellcheck) flags
…`) immediately surfaces an `SC2005` the file currently hides, and nothing else about
the file changes. This is the "documents a marker vs has one" shape `skill-lint.sh`
exists for, and AGENTS.md's "a validator that cannot parse its input checks MORE, never
less" — except here it checks *nothing*, silently.

[...]

Issue #1681, opened 2026-08-17
budget_run's SIGKILL pass reaches nobody: a descendant that survives SIGTERM outlives the bounded run


ingo-eichhorst:
Found while working #1627 (the reap-poll fixtures next door). Not the same
defect and not fixed there — filing rather than folding in.

## What happens

`budget_run`'s timeout path (`tools/lib/gate-budget.sh`) is:

```sh
_budget_kill_tree TERM "$pid"
local hard=$SECONDS
while [[ $((SECONDS - hard)) -lt "$BUDGET_TERM_GRACE_SECONDS" ]] && kill -0 "$pid" 2>/dev/null; do
  sleep "$BUDGET_POLL_SECONDS"
done
_budget_kill_tree KILL "$pid"
wait "$pid" 2>/dev/null
```

The grace loop waits for **`$pid` only** — the wrapper subshell — and `$pid` is
an ordinary `bash`, so it dies on SIGTERM within milliseconds. The loop
therefore exits almost immediately, and `_budget_kill_tree KILL "$pid"` then
walks from a pid that is *already dead*: `pgrep -P "$pid"` returns nothing
(the descendants have been reparented to launchd/init), so the SIGKILL pass
signals a dead pid and reaches no one.

Any descendant that ignores or slow-handles SIGTERM survives the whole
bounded run. That is the exact failure `_budget_kill_tree` was written to
prevent — its own doc comment: *"leaving those alive after the hook returns is
how a 'bounded' run keeps burning the machine."*

## Measured, not reasoned about

Fixture: one intermediate shell (dies on TERM) over a leaf that does
`trap '' TERM; sleep 45`, run under `budget_run 2` with
`BUDGET_TERM_GRACE_SECONDS=1`:

```
timed_out=1
leaf pid=[37472]
RESULT: leaf SURVIVED the bounded run — SIGKILL never reached it
37472 S    bash /var/folders/.../killgap-fixture... 0
```

`budget_run` reported a correct TIMEOUT and returned; the leaf was still
running 2s later, with 40+ seconds of its sleep to go. Reproduced on
`origin/main`'s `gate-budget.sh` (aa486930), bash 3.2.57, macOS 25.5.

The existing test suite does not see this: `gate-budget_test.sh`'s tree-kill
fixture leaf is a plain `exec sleep`, which dies on SIGTERM, so the SIGKILL
pass is never load-bearing there.

## Why the sibling library does not have it

[...]

Issue #1611, opened 2026-08-16
posix-lint walks git ls-files, so it cannot see the untracked file that now puts it in scope


ingo-eichhorst:
Found while fixing #1591 (PR #1609), and **created** by that fix — filing it
rather than folding it in, because it needs its own mutation evidence.

## The gap

`tools/posix-lint.sh` selects its files with `git ls-files`, which **excludes
untracked files**. Until #1609 that was harmless: an untracked file could not
put the posix gate in scope either, so the gate simply did not run.

#1609 fixed `changed_files_vs_origin_main` (`tools/lib/changed-files.sh`) to
include untracked files, because a gate silently skipped for an invisible file
is this repo's most-repeated failure shape. The consequence is a new instance
of that same shape one layer down:

- a new, untracked `#!/bin/sh` script now **does** put the posix gate in scope;
- the gate then walks `git ls-files` and **cannot see the file that summoned
  it**;
- it reports `ALL PASS` over a file it never read.

That is precisely the failure `posix-lint.sh`'s own header records as its
founding incident — its first draft "printed `ALL PASS` over an installer
carrying a deliberate `[[ ]]`" — arriving through the file-selection side
instead of the pipeline side.

## Blast radius, stated honestly

**Manual `tools/preflight.sh --changed` runs only.** An untracked file is
neither pushed (so the pre-push hook never sees it) nor present in CI (which
checks out a commit). So this cannot let a bashism reach `main`; it can only
tell a developer their new script is clean when nothing looked at it. Narrow,
and exactly the audience the `--changed` path exists for.

## The fix

Include untracked files in the walk, the way #1609 did for the changed set:

```sh
git ls-files --others --exclude-standard --full-name -- :/
```

`--full-name -- :/` is load-bearing and #1609 measured why: called from a
subdirectory the bare spelling returns cwd-relative, cwd-scoped paths, while
the full spelling matches `git ls-files`'s own output. Called from the repo
root the two agree, so a naive spelling looks correct forever — which is why
#1609 pinned it with a dedicated subdirectory assertion, and why this one
should too.

Mind the existing exclusions while doing it: `testdata/` is deliberately out of
the walk (those fixtures are deliberately corrupt), and the selector keys on a
`#!/bin/sh` **first line** only — an untracked file must be subject to both
rules, not waved through or double-counted.

## Per AGENTS.md this owes a deliberate mutation seen red

[...]

PR #1691, opened 2026-08-17
feat(tools): lint the 83 bash scripts no static linter had ever read (#1684)

### LINKED ISSUE #1684: 8,431 lines of bash under tools/lib/ are linted by nothing — shellcheck is installed and finds 13, including two rm -rf that can expand to /lib
Flagged by #1681's agent, then measured. `tools/posix-lint.sh` selects files whose **first line** is
`#!/bin/sh` — deliberately, and for a good reason (#1611). The consequence is that every `bash`
file under `tools/lib/` is outside every static linter this repo runs:

```
23 files, 8,431 lines
```

That is not peripheral code. It now includes the bounded-gate runner (`gate-budget.sh`), the shared
suite runner (`shell-lib-suite.sh`), the workflow-step resolver (`workflow-step.sh`), the reap poll
(`await-gone.sh`), the Swift harness (`swift-suite.sh`) and every extract-and-execute workflow lock.
Roughly half of it was written in the last two days.

## What a linter that is already installed would say

`shellcheck 0.11.0` — present on this machine **and** on CI's ubuntu image, where `linux.yml`
already runs — over those 23 files at `-S warning`:

| code | n | what it is |
|---|---:|---|
| SC2034 | 9 | variable assigned and never used |
| SC2209 | 2 | `var=command` where `var=$(command)` was meant — assigns a *string*, runs nothing |
| SC2115 | 2 | `rm -rf "$var/…"` that can expand to `/…` if `$var` is empty |
| | **13** | total |

## The two SC2115s, with their reachability measured rather than asserted

`tools/lib/changed-files_test.sh:233` and `:251`:

```sh
rm -rf "$NOLIB/lib"
```

`NOLIB="$(mktemp -d)"` at `:216`, and the file runs `set -uo pipefail` at `:16`.

**So it is narrow, and I am stating that rather than alarming**: `set -u` catches an *unset*
variable, but **not an empty one**, and `$(mktemp -d)` yields the empty string if `mktemp` fails
(unwritable or unset `TMPDIR`, full disk). At that point the command is `rm -rf /lib`. On macOS
`/lib` does not exist; on the Linux runner it does but is not writable by `runner`. So the realistic
outcome today is a failed `rm`, not damage — **the hazard is the shape, not the current blast
radius.** The fix is one character: `"${NOLIB:?}/lib"`.

It is worth saying out loud why this shape deserves the attention anyway: an unguarded `rm -rf` over
an interpolated variable is exactly what removed ~1,895 files from a developer's real
`~/Library/Preferences` earlier in this same work (the incident behind #1661's rework), and that one
also looked narrow until it wasn't.

The two SC2209s (`tools/lib/git-hooks_test.sh:138`, `:235`) are the more likely *live* defects: an
assignment that silently stores a string where a command was meant to run means an assertion may be
comparing against a literal.

## Why this is an enhancement, not a bug

Nothing is currently broken by it — 11 of the 13 findings are style or dead variables, and the two
that are not are narrowly reachable at worst. This is a **missing gate**, not a broken one, which is
the distinction #1681's agent drew before declining to file it under `bug`.

But it is the same *file-selection blindness* class as #1423 and #1611: a gate whose selection rule
silently excludes a whole family reads exactly like a gate that passed. `posix-lint.sh` was built on
that lesson and its own selection rule reproduces it one directory over.

## Suggested shape

A `bash-lint` gate over `tools/lib/**/*.sh` (excluding `testdata/`, the same split `skill-lint.sh`
and `posix-lint.sh` already draw), running `shellcheck` at a chosen severity floor. Two things to
decide with measurement rather than preference:

- **The severity floor.** 13 findings at `warning` is a tractable starting set; `-S style` was not
  measured here and may be noisy. `posix-lint.sh`'s precedent is to filter to a *named* code set
  rather than a severity band, so general style debt stays out of scope.
- **Where it runs.** `linux.yml` already has shellcheck and is where `posix-lint.sh` lives, so that
  is the cheap home; `preflight.sh` should mirror it, and the `tools` trigger already covers these
  paths.

Per AGENTS.md the gate owes: a refusal (not a skip) when shellcheck is absent — that is
`posix-lint.sh`'s founding incident, where a linter that failed to run printed `ALL PASS` — and a
deliberately-broken committed fixture per rule it enforces.

## Related

- #1423, #1611 — the file-selection blindness class, and why `posix-lint.sh` keys on line 1
- #1661 — the `rm -rf` incident this shape rhymes with
- #1681 — where this was flagged

ingo-eichhorst:
Closes #1684

`tools/posix-lint.sh` selects files whose **first line** is `#!/bin/sh`, deliberately
(#1611). The consequence was that **every bash file in this repo sat outside every static
linter it runs** — including `gate-budget.sh` (bounds the pre-push hook),
`changed-files.sh` (the hook's own scoping rules), `shell-lib-suite.sh`,
`workflow-step.sh`, `swift-suite.sh` and every extract-and-execute workflow lock: the
machinery deciding whether the *other* gates pass. Same file-selection blindness class as
#1423 and #1611 — a gate whose selection rule excludes a whole family reads exactly like
a gate that passed.

---

## 1. Scope decision: default-IN, opt-out with a reason

The issue proposed `tools/lib/**`. I went wider, because scoping to one directory
*reproduces* the class the issue names — the next `tools/foo.sh` would be invisible, and
the linted set would have to be widened by someone remembering.

So the gate lints **everything git knows about** (tracked, plus untracked and not
gitignored, the same walk `posix-lint.sh` uses after #1611) whose **line 1** is a bash
shebang, minus three declared `EXCLUDE` globs, each carrying its own justification in the
source. **An exclusion that matches nothing is a hard refusal (exit 2), not a no-op** —
the same both-directions existence check `TW_EXEMPT_KEYS` and `nilTolerant` get, because
an entry that stopped naming anything real reads from the log exactly like coverage.

| | files | why |
|---|---:|---|
| **linted** | **83** | ~19k lines: `tools/**`, `examples/**`, `.claude/skills/**`, `tools/git-hooks/pre-push` (extensionless, matched by shebang not suffix) |
| excluded `*/testdata/*` | 11 | deliberately-corrupt fixture corpora, driven by their own unit tests — the split `skill-lint.sh` and `posix-lint.sh` already draw |
| excluded `replaydata/*` | 31 | the recording rig's per-agent drivers — measured 79 findings, 68 of them the driver protocol's declared-variable seam, plus one file whose analysis shellcheck **abandons**. Filed as **#1687** rather than blanket-disabling codes to reach zero |
| excluded `*/templates/*.tmpl` | 1 | the template the above are generated from, carrying the same 10 seam variables — excluded *with* `replaydata/`, not separately from it |

The census line prints both halves on every run (`83 of 126 bash file(s); excluded:
*/testdata/*=11 replaydata/*=31 */templates/*.tmpl=1`), so an exclusion widening to
swallow the corpus cannot read as a clean gate.

## 2. Severity floor: `warning`, and it is a measurement

`posix-lint.sh`'s precedent is a **named code set**, so general style debt stays out. I
weighed it and deliberately did not follow it: `SC3xxx` is a closed family *by
definition* ("In POSIX sh, X is undefined"), whereas "bugs in bash" has no such family —
an opt-in code list means a code shellcheck adds later is not enforced until someone
remembers, which is the "covered by remembering" shape AGENTS.md argues against
throughout. A severity **floor** is covered by existing.

What the floor has to survive is the version split, so I fetched 0.9.0 and 0.10.0 from
koalaman's releases and ran all three **per-file** over the whole in-scope corpus
(81 files / 18,304 lines at `d76aa137`):

| floor | 0.9.0 | 0.10.0 | 0.11.0 | verdict |
|---|---:|---:|---:|---|
| `-S warning` | 26 | 26 | 26 | **byte-identical** — same codes, same files, same `line:col` |
| `-S style` | 312 | 312 | 199 | **163 disagreements** |

`-S style` is out of scope on that evidence, not on a guess about noise. `-S error` alone
would have caught 1 of the 26.

`--external-sources` (`-x`) is **off**, also measured: these libraries source through
variable paths (`. "$DIR/await-gone.sh"`), which shellcheck cannot resolve without a
`source-path`, so `-x` changed the count by exactly **zero** (26 → 26).

### One file per invocation — a measured constraint, not a detail

`shellcheck` given several files in **one** invocation suppresses SC2034 for a name used
in **any** of them:

```
$ shellcheck --shell=bash -S warning tools/lib/swift-suite_test.sh
  swift-suite_test.sh:247 AWAIT_GONE_LOOKED appears unused  [SC2034]
  swift-suite_test.sh:256 AWAIT_GONE_ALIVE  appears unused  [SC2034]
$ shellcheck --shell=bash -S warning tools/lib/swift-suite_test.sh tools/lib/await-gone.sh
  (nothing)
```

A multi-file gate's verdict would therefore depend on which *other* files shared the
command line — which `--changed` scoping makes differ between a developer's push and CI's
full run. Per-file is the only deterministic choice.

**This also corrects the issue's arithmetic in both directions, and I got it wrong once
before catching it.** My first pass measured multi-file, saw 11 under `tools/lib` instead
of the issue's 13, and concluded 2 had been fixed incidentally by #1683. That was wrong:
they were *masked*, because #1683 added a cross-file read of `AWAIT_GONE_ALIVE` in
`gate-budget_test.sh`. Per-file, `tools/lib` yields exactly the issue's **13** at HEAD,
and the full in-scope corpus yields **26**.

## 3. CI placement — and which direction the asymmetry runs

**`linux.yml`**, beside `posix-lint.sh`, plus a second step for its unit tests. But **the
reason is not posix-lint's**, and the difference is the decision:

- `posix-lint.sh` needs ubuntu because `/bin/sh` must genuinely **be** dash there — the
  property under test is a property of the runtime.
- `shellcheck` is a static analyser over the bash *language* and reads no interpreter at
  all. Measured: the 0.9.0 figures above came from an **x86_64 binary under Rosetta on
  arm64 macOS** and agree byte-for-byte with a native arm64 0.11.0. Neither the OS nor
  the arch enters the verdict.

So "the subject is bash that runs on macOS" does **not** argue for a macOS runner here.
What picks the host is only that ubuntu ships shellcheck (0.9.0) and macos-latest ships
none, so hosting it in test.yml would `brew install` one on every PR to lint 83 files.

**The asymmetry, at the chosen floor, runs in neither direction** — 0.9.0 / 0.10.0 /
0.11.0 are byte-identical over 18,304 lines. At `style` it runs in **both**, and
dominantly in the direction that hurts: **138** findings CI's 0.9.0 reports that a local
0.11.0 *cannot produce* (137 × SC2317 "command appears unreachable" + 1 × SC2015) against
**25** the other way (SC2329, which 0.9.0 does not have). That is worse than the framing
in the task — a developer's `preflight --only bash` would be green while `linux.yml` went
red on findings their own shellcheck does not implement, the exact round-trip
posix-lint's monotonicity argument exists to prevent. Choosing `warning` removes the
asymmetry rather than picking a side of it.

## 4. Per-finding triage — all 26

The issue's 13 are the first four blocks. The other 13 are findings of the gate being
built; they are not "new defects found while looking at something else" but the gate's own
output, and it cannot land red — so they are fixed here and each is named.

### SC2115 ×2 — real, the shape the issue was filed about

`tools/lib/changed-files_test.sh:233,251` — `rm -rf "$NOLIB/lib"` → `rm -rf /lib`.
`set -u` catches an *unset* variable, not an empty one, and `$(mktemp -d)` yields empty
on failure. **Fixed**: `"${NOLIB:?}/lib"`, with the reachability recorded at the site.

### SC2209 ×2 — **NOT live defects; the issue's guess was the wrong way round**

`tools/lib/git-hooks_test.sh:138,235` — `kind=file`. The issue expected "an assertion may
be comparing against a literal". Checked: the assertion two lines down *is*
`assert_eq "installed pre-push is a regular file" "file" "$kind"`, so the **string is
what was meant** and the test is correct. shellcheck flags it only because `file(1)` is a
command. **Fixed** by quoting (`kind="file"`) — shellcheck's own second suggestion — so
the intent is explicit and no disable is needed.

### SC2034 ×17 — every one a verified cross-file seam

All 17 are per-site `# shellcheck disable=SC2034  # <reason>` naming the consumer's
`file:line`. Every consumer was grepped, not assumed:

| site | variable(s) | consumer |
|---|---|---|
| `await-gone_test.sh:49` | `AWAIT_GONE_POLL_SECONDS` | `await-gone.sh:131` (default), `:283` (`sleep`) |
| `await-gone_test.sh:219,221` | `AWAIT_GONE_LOOKED/_ALIVE` | `await-gone.sh:260-271` — the predicate's whole output |
| `await-gone.sh:276` | `AWAIT_GONE_LAST` | the library's documented output; printed by 4 files, 6 sites |
| `gate-budget_test.sh:302` | `AWAIT_GONE_LOOKED/_ALIVE` | same predicate contract |
| `swift-suite_test.sh:247,256` | `AWAIT_GONE_LOOKED/_ALIVE` | same predicate contract |
| `swift-suite.sh:76,77` | `SWIFT_SUITE_COLD_BUILD_SECONDS/_MIN_HEADROOM` | `swift-suite_test.sh:142-145` asserts their relation |
| `spawn-record-daemon_test.sh:56-59` | 4 × `RECORD_DAEMON_*_TICKS/_TICK_S` | `spawn-record-daemon.sh:137,139,152,165,167` |
| `spawn-record-daemon_test.sh:68,81` | `RECORD_DAEMON_PID/_STAGING` | `spawn-record-daemon.sh:36,38,149,162,174` |
| `spawn-record-daemon_test.sh:131` | `IRRLICHT_READY_SESSION_TTL` | read out of the ENV by `record_daemon_env`; daemon consumer at `core/domain/config/config.go:168` |
| `spawn-record-daemon_test.sh:195` | `MANAGED_FILE_BACKUP_DIR` | `managed-file-snapshot.sh:44` |

A directive covers only the **next command**, which is why the four adjacent knob
assignments need four disables — the single grouped one I wrote first left three findings
standing, and the gate said so.

### SC2164 ×2 — real

`tools/preflight.sh:107`, `tools/security-scan.sh:63` — a bare `cd "$ROOT_DIR"`. Both
scripts then run every gate/scanner on repo-relative paths, so a failed `cd` would grade
whatever tree the caller was in and report it as this repo's. **Fixed**: `|| exit 2`.

### SC1087 ×1 — cosmetic, fixed

`adapter-tables_test.sh:109` — `$agent[[:space:]]` inside a grep pattern, read as an array
expansion. **Fixed**: `${agent}[[:space:]]`.

### SC2155 ×1 — real

`run-cell.sh:424` — `export IRRLICHT_EXTRA_TRANSCRIPTS="$(cat …)"`. `export` always
succeeds, so an unreadable paths file would export an empty list rather than fail.
**Fixed**: assign, then export.

### SC2188 ×1 — cosmetic, fixed

`examples/coding-factory/drive.sh:30` — `> "$LOG"`. **Fixed**: `: > "$LOG"`.

## 5. Refusals, and the rule the gate exists to not break

Four hard refusals (**exit 2**), never skips:

1. **no shellcheck** — with an install hint. `posix-lint.sh`'s founding incident is a
   linter that failed to run printing `ALL PASS` over an installer carrying a deliberate
   `[[ ]]`.
2. **an empty in-scope set** — including its most dangerous form, *bash was discovered and
   all of it was excluded*.
3. **an exclusion rule matching nothing.**
4. **a path git listed that cannot be opened.**

Plus two *failure* (exit 1) paths that distinguish "clean" from "did not look": shellcheck
exiting **>1**, and shellcheck exiting **1 while printing nothing**. The exit status is
read explicitly; nothing is piped into `grep` and tested for emptiness, which is exactly
how `posix-lint.sh`'s first draft reproduced #1423 inside itself.

**SC1072/SC1073 are inside the floor on purpose**, and this turned out to be the gate's
most valuable rule. A comment line whose **first word after `#` is the linter's name** is
parsed as a *directive*, and an unparseable one makes shellcheck **abandon the file** —
every later finding silently disappears. It caught the construct **four times in this
PR's own new prose**, and `replaydata/_lib/drive/contracts.sh` has been carrying it
unnoticed: as-is, SC1073/SC1072 are the *only* output shellcheck produces for that file,
and rewording that one line surfaces an SC2005 it currently hides (measured; filed as
#1687). The sibling SC1125 is the `# shellcheck disable=SCxxxx — <prose>` spelling — the
disable **is** still honoured (measured on both 0.9.0 and 0.11.0), but the prose costs an
`error`-severity finding, and eight directives in `replaydata/` are written that way.

## 6. Committed fixtures — `tools/lib/testdata/bash-lint/`

One deliberately-broken file per rule class, so the mutation evidence outlives the PR
(`tools/lib/testdata/posix-lint/` is the shape, README included):

`bad-rm-rf-var.sh` (SC2115) · `bad-assign-command.sh` (SC2209) · `bad-unused-var.sh`
(SC2034) · `bad-cd-unchecked.sh` (SC2164) · `bad-array-brace.sh` (SC1087) ·
`bad-declare-assign.sh` (SC2155) · `bad-redirect-no-command.sh` (SC2188) ·
`bad-directive-prose.sh` (SC1072/SC1073) · `bad-directive-keyvalue.sh` (SC1125)

plus two controls that carry as much weight: **`good-clean.sh`**, the vacuity guard (a
linter that failed everything would satisfy all nine `bad-*` assertions), and
**`style-noisy-but-warning-clean.sh`**, which makes shellcheck exit 1 carrying only
SC2086/SC2012 — pinning the floor in the one direction the `bad-*` files cannot reach.

`bad-directive-prose.sh` is graded by more than "it is rejected": the test **rewords its
directive line, asserts the sed actually changed the file, and then asserts the hidden
SC2115 appears** while the original reports only SC1073. That asymmetry *is* the finding.

Verified identical on 0.9.0 and 0.11.0 for all 11 fixtures.

## 7. Mutation evidence — 10 deliberate mutations of the gate, all seen RED

Each applied to a copy, run, reverted by copy-back, with a `git status --porcelain`
assertion between (never `git stash` / `checkout` / `restore`). The harness asserts its
own mutation **changed the file** — which caught three stale regexes on the first attempt
and is #1390's lesson applied.

```
MUTATION M1  severity floor raised to error: RED (exit 1) — caught by:
        FAIL: detect: bad-assign-command.sh is rejected (exit 1) — expected [1] got [0]
        FAIL: detect: bad-cd-unchecked.sh is rejected (exit 1) — expected [1] got [0]
        FAIL: detect: bad-declare-assign.sh is rejected (exit 1) — expected [1] got [0]
        FAIL: detect: bad-redirect-no-command.sh is rejected (exit 1) — expected [1] got [0]
MUTATION M2  severity floor lowered to style: RED (exit 1) — caught by:
        FAIL: floor: style-noisy but warning-clean still passes (exit 0) — expected [0] got [1]
        FAIL: discovery: the repo's real bash scripts pass (exit 0) — expected [0] got [1]
MUTATION M3  SC1xxx filtered out of the findings: RED (exit 1) — caught by:
        FAIL: detect: bad-array-brace.sh is rejected (exit 1) — expected [1] got [0]
        FAIL: detect: bad-directive-keyvalue.sh is rejected (exit 1) — expected [1] got [0]
MUTATION M4  exclusion-matched-nothing refusal off: RED (exit 1) — caught by:
        FAIL: refusal: an exclusion matching nothing exits 2 (not a silent pass) — expected [2] got [0]
MUTATION M5  empty-in-scope refusal becomes a pass: RED (exit 1) — caught by:
        FAIL: refusal: a corpus where everything was excluded exits 2 — expected [2] got [0]
        FAIL: refusal: ...saying nothing was checked — got [bash-lint: ALL PASS (nothing to do)]
MUTATION M6  missing-shellcheck refusal removed: RED (exit 1) — caught by:
        FAIL: refusal: no shellcheck exits 2 (not a silent pass) — expected [2] got [1]
MUTATION M7  shellcheck exit>1 read as clean: RED (exit 1) — caught by:
        FAIL: broken linter: silent exit-2 is a FAILURE, not a clean file — expected [1] got [0]
MUTATION M8  exit-1-with-no-output read as clean: RED (exit 1) — caught by:
        FAIL: broken linter: says the output was lost — expected [contains 'the output was lost']
MUTATION M9  walk reverted to tracked files only: RED (exit 1) — caught by:
        FAIL: untracked: a finding in an UNTRACKED script fails the gate — expected [1] got [0]
MUTATION M10 exclusion glob quoted (the SC2053 "fix"): RED (exit 1) — caught by:
        FAIL: discovery: the repo's real bash scripts pass (exit 0) — expected [0] got [2]
final tree check: RESTORED
```

Two more, against `shell-lib-suite_test.sh`'s newly-derived skip list (mutating
`test.yml`, same copy-back protocol):

```
MUTATION S1 a skip naming a file that does not exist: RED — FAIL: the skip test.yml
            passes names a real file — expected [tools/lib/renamed-away_test.sh]
MUTATION S2 the invocation replaced by a bare loop: RED — the derivation goes blind and
            says so, rather than reporting an empty skip list as clean
final tree check: RESTORED
```

M10 is worth a note: quoting the exclusion glob is what shellcheck's own SC2053
*suggests*, and doing it makes every exclusion match nothing. The gate's
exclusion-existence refusal turns that into a loud exit 2 instead of a silently narrowed
scope — which is why that site carries a scoped `# shellcheck disable=SC2053` with the
reason, rather than the "fix".

## 8. The bootstrap

The gate is bash under `tools/`, so **it lints itself and its own test**, and a case
asserts that (`discovery: the gate lints ITSELF`). It found two real things in its own
source on the first run: the SC1072/SC1073 directive-prose trap in its header (twice), and
an SC2053 on the glob comparison that is a genuine false positive and now carries a scoped
disable. Both are in the diff.

## 9. Also removed: an advisory check that could not fail

`tools/onboarding-factory/scripts/smoke-test.sh` already ran shellcheck over
`scripts/**` — with `|| true` per file, a trailing `echo` so the block always exited 0,
and a silent `(shellcheck not installed — skipped)` on a runner that ships none. So
"nothing was found" and "nothing looked" printed the same verdict, and that is where nine
of the 26 findings had been sitting. `bash-lint.sh` now covers that tree as a real gate,
so the advisory block is replaced by a pointer to it. Two mechanisms where one of them
cannot fail is worse than one that can.

## 10. Wiring

- `.github/workflows/linux.yml` — `Lint bash scripts` + `Test the bash-lint gate`, before
  `setup-go` (no toolchain, a couple of seconds, and behind the Go steps a compile error
  would abort the job and leave a tooling-only PR with no tooling feedback).
- `.github/workflows/test.yml` — `bash-lint_test.sh` added as a **second skip argument**
  (macos-latest ships no shellcheck; the gate refuses rather than skipping). Not a second
  loop: `shell_lib_suite_run` validates every skip name against the corpus before running
  anything.
- `tools/lib/shell-lib-suite_test.sh` — the "the skip names a real file" check now
  **derives** the list from the workflow step instead of hardcoding one filename. The
  one-file check would have gone on passing while saying nothing about the new entry.
- `tools/preflight.sh` — a new `bash` group (`--only bash`), same loose trigger as the
  posix gate and for the same reason: a *new* bash script is exactly the file the gate
  most needs to see on the push that adds it, and a trigger cannot enumerate one that
  does not exist yet.
- `AGENTS.md` — one bullet beside "POSIX shell scripts", carrying the four measured
  decisions.

## 11. Gates, stated honestly

| gate | result |
|---|---|
| `tools/preflight.sh --only bash` | **PASS** (83 files, `ALL PASS`) |
| `tools/preflight.sh --only tools` | **PASS** — `found 17, skipped 0, ran 17, failed 0` |
| `tools/preflight.sh --only posix` | **PASS** |
| `tools/preflight.sh --only skills` | **PASS** |
| `tools/bash-lint.sh` under shellcheck **0.9.0** | **PASS** (CI's version, run via a PATH stub) |
| `tools/lib/bash-lint_test.sh` under 0.11.0 and 0.9.0 | **ALL PASS** both |
| `tools/onboarding-factory/scripts/smoke-test.sh` | **PASS** |
| pre-push hook (`preflight --changed --budget 540`) | **PASS** — bash / tools / gofmt / replaydata validate / rig smoke / replay fixtures / macOS Swift all PASS; core, ARS, web, security SKIP (no matching files in the diff) |

Not run: `--only go`'s core suite and `--linux` — this diff touches no Go and no
`#!/bin/sh` file. `--only swift` ran via the hook and passed.

## Related

- **#1687** (filed) — the excluded `replaydata/` family, including the `contracts.sh`
  abandonment and the 8 × SC1125
- #1423, #1611 — the file-selection blindness class this is the third instance of
- #1661 — the `rm -rf` incident the SC2115 shape rhymes with
- #1681, #1683 — where this was flagged

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -54,6 +54,48 @@ jobs:
       - name: Test the POSIX-sh gate
         run: bash tools/lib/posix-lint_test.sh
 
+      # BASH gate (#1684) — the sibling of the two steps above, over the OTHER
+      # shell family. 83 bash files / ~19k lines that no static linter read at
+      # all: the bounded gate runner (tools/lib/gate-budget.sh), the pre-push
+      # hook's own scoping rules (changed-files.sh), the shared suite runner and
+      # every extract-and-execute workflow lock. posix-lint.sh selects on a
+      # `#!/bin/sh` FIRST LINE, deliberately (#1611), so every bash file fell
+      # outside it — the same file-selection blindness class as #1423 and #1611.
+      #
+      # THIS JOB IS THE HOST FOR A DIFFERENT REASON THAN posix-lint.sh's, and
+      # the difference matters. That gate needs ubuntu because `/bin/sh` must
+      # genuinely BE dash here — the property under test is a property of the
+      # runtime. shellcheck is a static analyser over the bash language and
+      # reads no interpreter at all: measured, a 0.9.0 x86_64 binary under
+      # Rosetta on arm64 macOS and a native arm64 0.11.0 return byte-identical
+      # findings at this gate's severity floor, so neither the OS nor the arch
+      # enters the verdict. What picks this job is only that the ubuntu image
+      # ships shellcheck (0.9.0) and macos-latest ships none, so hosting it in
+      # test.yml would `brew install` one on every PR.
+      #
+      # The version split is why the floor is `warning` and not `style`: at
+      # `warning` all of 0.9.0 / 0.10.0 / 0.11.0 agree byte-for-byte over the
+      # whole corpus, while at `style` this runner's 0.9.0 reports 138 findings
+      # a developer's 0.11.0 cannot produce (137 × SC2317) against 25 the other
+      # way (SC2329) — a preflight that passes a diff this job rejects, which is
+      # the round-trip posix-lint.sh's monotonicity argument exists to prevent.
+      # tools/bash-lint.sh's header carries the full measurement.
+      #
+      # Runs before setup-go for the same reason as the two above: no toolchain,
+      # a couple of seconds, and behind the Go steps a compile error would abort
+      # the job and leave a tooling-only PR with no tooling feedback.
+      - name: Lint bash scripts
+        run: tools/bash-lint.sh
+
+      # The gate's own unit tests, over the deliberately-broken corpus under
+      # tools/lib/testdata/bash-lint/. Here rather than in test.yml's
+      # `tools/lib/*_test.sh` loop for the same reason posix-lint_test.sh is:
+      # they need shellcheck, and the macos image ships none. A gate that passes
+      # by construction is what #1684 is about, so the mutation corpus has to
+      # run somewhere — this is where it can.
+      - name: Test the bash-lint gate
+        run: bash tools/lib/bash-lint_test.sh
+
       - name: Set up Go
         uses: actions/setup-go@v7
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,6 +115,13 @@ jobs:
       # ARGUMENT rather than as a `case` pattern in a copied loop is what makes
       # it fail loudly if it ever stops matching a real file.
       #
+      # bash-lint_test.sh (#1684) is skipped for exactly the same reason — it
+      # drives tools/bash-lint.sh, which refuses without shellcheck — and
+      # linux.yml runs both it and its gate. It is a SECOND argument rather than
+      # a second loop, and shell_lib_suite_run validates every name in the list
+      # against the corpus before running anything, so a skip that stopped
+      # matching a real file is a refusal rather than a quiet no-op.
+      #
       # `|| rc=$?` rather than a bare call: it is the spelling that keeps the
       # status under the `-e` above (`|| true` would not — it hands back
       # `true`'s status, silently reporting every failing run as 0).
@@ -122,5 +129,5 @@ jobs:
         run: |
           rc=0
           . tools/lib/shell-lib-suite.sh
-          shell_lib_suite_run tools/lib posix-lint_test.sh || rc=$?
+          shell_lib_suite_run tools/lib posix-lint_test.sh bash-lint_test.sh || rc=$?
           exit "$rc"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -912,6 +912,83 @@ Before marking a ticket done, run the full suite — every layer must pass:
   `install-uninstall_test.sh` now *executes* the installer under `dash` rather
   than `sh` (macOS ships `/bin/dash`), which is the runtime half — it reaches
   only the lines a case runs, where the linter reads every line.
+- Bash scripts: `tools/bash-lint.sh` runs `shellcheck --shell=bash
+  --severity=warning` over every file git knows about — tracked, plus untracked
+  and not gitignored — whose **first line** is a bash shebang. 83 files /
+  ~19k lines, and until #1684 **no static linter read one of them**: the bullet
+  above selects on `#!/bin/sh` line 1, deliberately (#1611), so every bash file
+  fell through — including the bounded gate runner (`gate-budget.sh`), the
+  pre-push hook's own scoping rules (`changed-files.sh`), the shared suite
+  runner and every extract-and-execute workflow lock, i.e. the machinery
+  deciding whether the other gates pass. Same file-selection blindness class as
+  #1423 and #1611. It found 26 findings, all fixed or annotated.
+  Four decisions are load-bearing and each was measured rather than preferred.
+  - **DEFAULT-IN, opt-out with a reason.** The scope is not a prefix list — it
+    is everything, minus three declared `EXCLUDE` globs each carrying its
+    justification, and an exclusion that matches NOTHING is a hard refusal
+    (exit 2) rather than a no-op, the same both-directions existence check
+    `TW_EXEMPT_KEYS` and `nilTolerant` get. So a bash file added anywhere is
+    covered by existing, and the two families that are out — the
+    deliberately-corrupt `testdata/` corpora, and the recording rig's per-agent
+    drivers plus their `.tmpl` (#1687: 79 findings, and one file whose analysis
+    shellcheck silently ABANDONS) — are out visibly.
+  - **A severity FLOOR, not posix-lint's named code set**, and the reason the
+    precedent was not followed is that SC3xxx is a closed family by definition
+    while "bugs in bash" is not: an opt-in code list is not enforced on a code
+    shellcheck adds later. The floor is `warning` because that is where the
+    VERSION SPLIT vanishes. Measured per-file over the whole corpus with 0.9.0,
+    0.10.0 and 0.11.0 binaries side by side: at `warning` all three are
+    **byte-identical** (26 findings, same line:col); at `style` they are not,
+    and the asymmetry runs in BOTH directions with the damage on the CI side —
+    138 findings CI's 0.9.0 reports which a local 0.11.0 cannot produce
+    (137 × SC2317) against 25 the other way (SC2329). A `style` floor would
+    make `preflight --only bash` pass a diff linux.yml rejects, the exact
+    round-trip posix-lint's monotonicity argument exists to prevent. `-x`
+    (external sources) is off, also measured: these libraries source through
+    variable paths, so it changed the count by zero.
+  - **ONE FILE PER INVOCATION.** Measured: shellcheck given several files at
+    once suppresses SC2034 for a name used in ANY of them — one file alone
+    reports 2 findings, the same file beside `await-gone.sh` reports 0. A
+    multi-file gate's verdict would depend on which other files shared the
+    command line, which `--changed` scoping makes differ between a push and
+    CI's full run. It is also why the count is 26 where a single multi-file run
+    over the same tree says 15.
+  - **A comment line whose FIRST word is the linter's name is a DIRECTIVE**, and
+    an unparseable one makes shellcheck ABANDON the file — every later finding
+    silently disappears. That is inside the floor (SC1072/SC1073 are `error`)
+    for the same reason `posix-lint.sh` refuses to filter its parse-abort codes
+    into silence, and it is the gate's most valuable rule: it caught the
+    construct **four times in this PR's own new prose**, and
+    `replaydata/_lib/drive/contracts.sh` has been carrying it unnoticed (#1687 —
+    rewording that one line surfaces an SC2005 the file currently hides). The
+    sibling SC1125 is the `# shellcheck disable=SCxxxx — <prose>` spelling; the
+    disable IS still honoured (measured on both versions), but the reason must
+    go behind a second `#`.
+  The sanctioned escape hatch is a per-SITE `# shellcheck disable=SCxxxx  #
+  <reason>` naming its consumer's `file:line` — all 17 SC2034s are that, each
+  verified to have a real reader — and never a code removed from the gate.
+  A directive covers only the NEXT command, so four adjacent knob assignments
+  need four.
+  It lives in **`linux.yml`** beside `posix-lint.sh`, but for a DIFFERENT
+  reason, and the difference is the decision: posix-lint needs ubuntu because
+  `/bin/sh` must genuinely BE dash there — a property of the runtime — whereas
+  shellcheck is a static analyser that reads no interpreter, so neither the OS
+  nor the arch enters the verdict (the 0.9.0 measurement above was taken from an
+  x86_64 binary under Rosetta on arm64 macOS and agrees byte-for-byte with a
+  native arm64 0.11.0). What picks the host is only that the ubuntu image ships
+  shellcheck and macos-latest ships none. Mirrored locally by
+  `tools/preflight.sh --only bash`. Its tests are `tools/lib/bash-lint_test.sh`
+  over the committed corpus under `tools/lib/testdata/bash-lint/` — one
+  deliberately-broken fixture per rule class, `good-clean.sh` as the vacuity
+  guard, `style-noisy-but-warning-clean.sh` pinning the floor in the one
+  direction the `bad-*` files cannot reach, and the abandonment fixture proved
+  by REWORDING its directive line and asserting the hidden SC2115 appears.
+  That suite runs in `linux.yml`, not in test.yml's `tools/lib/*_test.sh` loop —
+  it needs a linter the macOS image lacks, and it is the loop's SECOND skip
+  argument. `shell-lib-suite_test.sh` now derives that list from the workflow
+  step and existence-checks every name, because the one-file check it had would
+  have gone on passing while saying nothing about the new entry.
+
 - Sourced shell libraries: not a contract family — a tripwire,
   `tools/lib/shell-lib-errexit_test.sh`, over every `tools/lib/*.sh` that is
   not a `*_test.sh`. Each function it can drive must, under a caller's
@@ -1647,8 +1724,8 @@ it once instead of round-tripping through GitHub Actions per fix. Gates run
 **cheapest first**, in two phases, not in "CI's order": there is no single CI
 order to mirror, since those are separate workflows GitHub runs concurrently,
 and the order is load-bearing under `--budget` (below) because it decides which
-gates survive a squeeze. `skill-file lint` and `POSIX sh lint` are the only
-coverage their file families have and cost a fraction of a second each, so they
+gates survive a squeeze. `skill-file lint`, `POSIX sh lint` and `bash lint` are the
+only coverage their file families have and cost a second or two each, so they
 run before four minutes of `go test` — which is the argument test.yml already
 makes in-file for running the skill lint before `setup-go`.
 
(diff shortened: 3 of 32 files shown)


ingo-eichhorst:
Post-merge verification note, for the record.

I could not reproduce the **mechanism** stated for the one-file-per-invocation decision, on
`shellcheck 0.11.0`:

```
ONE invocation over all 23 tools/lib files : 13 findings
PER-FILE invocations, summed               : 13 findings
```

and isolating the pair the PR names, `swift-suite_test.sh` reports the same 2 × SC2034 alone as it
does beside `await-gone.sh` (2 and 2, not 2 and 0).

**This changes nothing about the PR.** The count is 13 either way, which is what the issue reported;
and one file per invocation is the strictly conservative choice — if any shellcheck version does
cross-suppress, per-file cannot undercount. The gate is right for a reason that holds regardless.

But the cross-suppression claim is **unverified at 0.11.0** and should not be carried forward as
established. It may well hold at 0.9.0 or 0.10.0, which were also measured here and which I do not
have installed; if so, the version should be named beside it. Marked rather than left implied.

PR #1694, opened 2026-08-17
feat(replaydata): codex's first hook-bearing recordings, and the gate that was walking past the hook trust menu (#1388)

ingo-eichhorst:
Closes #1388

`of coverage --hooks` reported codex as **GAP** — it declares a `hooks` permission and had **0 of 43** `events.jsonl` carrying a `hook_received`, so every replay-gated guarantee about its hook path guaranteed nothing. It now reports **`ok`**, and the mutation below shows the difference is real rather than "some files were added".

---

## 1. The blocker question

**The `:7837` hardcode is genuinely gone. A second blocker existed, and it was not a port.**

### The port is not the blocker (evidence)

```
$ grep -rn "7837" core/adapters/inbound/agents/codex/
hookinstaller.go:23:// alternate port installs hooks that reach it rather than whatever owns :7837   <- comment
hookinstaller.go:34:// endpoint path it still matches a pre-#1178 :7837 command                     <- comment
hookport_test.go:26: "http://localhost:7837/api/v1/hooks/codex || true"                            <- test
```

No literal on the install path. `hookEndpointURL()` is `daemonaddr.LocalURL(HookEndpointPath)`, and `LocalURL` is `fmt.Sprintf("http://localhost:%d%s", PortOf(BindAddr()), path)`.

**On the live `~/.codex/hooks.json` still carrying `http://localhost:7837/…`** — asked, because it is exactly what a hardcode looks like: that is a *legitimate resolved-port install*. The daemon that owns it (`pid 39557`, `core/bin/irrlichd --record`) has `IRRLICHT_BIND_ADDR=127.0.0.1:7837` in its environment and is the sole listener on that port (`lsof -nP -iTCP -sTCP:LISTEN`), so `PortOf(BindAddr())` resolves to 7837 for it. The proof that the resolved path is reached is in this PR's own recording: taken on a `:7838` coexist recorder, the installed command read `curl … http://localhost:7838/api/v1/hooks/codex || true` (visible in codex's own hook-detail pane and in the committed `hooks.json` copy under the run's `managed-file-backup/created/`).

### The real blocker

**On codex-cli 0.147.0 the directory-trust dialog is dismissed by the bare digit, so the driver's unconditional trailing `Enter` walked past the hook gate.**

[...]

Issue #1586, opened 2026-08-15
TestReadLauncherEnv_Tmux_AttachedClientIsTheHost flakes on a fixed 120ms sleep, on the path this whole family keeps changing


ingo-eichhorst:
Reported by the #1574 agent, and confirmed by reading the fixture. Added by
#1501 (PR #1583), so it is hours old — filing now while the cause is cheap to
fix rather than after someone loses an hour to it.

## The flake

`TestReadLauncherEnv_Tmux_AttachedClientIsTheHost`
(`processlifecycle/tmuxclient_darwin_test.go`) failed once under back-to-back
full-suite load, then passed 8/8 standalone under `-race` and in three other
full-suite runs.

Its fixture spawns an orphaned helper, waits a **fixed** interval for the
kernel to publish the exec'd env to sysctl and for the intermediate shell to
exit, and then hard-fails if reparenting has not happened yet:

```go
// Give the kernel a beat to publish the exec'd env to sysctl, and the
// shell a beat to exit so the reparenting has actually happened.
time.Sleep(120 * time.Millisecond)
if ppid, _, err := readProcInfo(context.Background(), pid); err != nil || ppid != 1 {
    t.Fatalf("helper %d has ppid %d (err %v), want 1: …", pid, ppid, err)
}
```

120 ms is enough on an idle machine and is not a guarantee under a loaded one.

## Why this one matters more than an ordinary flake

It is on the hot path of the family that has been changing every day — #1485,
#1492, #1513, #1524, #1533, #1537, #1529, #1544, #1547, #1534, #1525, #1501,
#1574 all touch `readProcInfo`, `hostIdentity` or the launcher capture this
test drives end to end. **A timing flake there is indistinguishable from a
regression in the change under review**, which is exactly the cost `#1432`
records for `TestFSWatcher_EmitsEventsForTranscriptCreateAndModify`: every
agent that hits it spends time proving it is not their fault. One already has.

The fixture's `t.Fatalf` is at least honest about being the fixture rather than
the assertion — it says "this fixture only stands in for a tmux pane while its
ancestry terminates at init" — so it degrades better than a bare timeout. But
the sleep is still the wrong shape.

## The fix

Poll rather than sleep: retry `readProcInfo` until `ppid == 1` or a generous
deadline expires (a second or two), and fail with the elapsed time when it
does. Same idiom `spawnSleeperWithEnv`'s siblings could use. That converts "the
[...]

Issue #1518, opened 2026-08-13
Lint: no bare integer in cmd/replay doc comments may equal a live catalogCensus figure


ingo-eichhorst:
Proposed by PR #1511's reviewer, and the evidence for it is the PR itself.

## The observation

#1503 existed because hand-typed figures in doc comments go stale silently and are then quoted downstream with full confidence. #1511 closed that by machine-generating the census — and then, **inside the fix for that exact slip, committed the slip four more times.** Of the seven findings its review returned, four were this:

- a new comment in `replay_sidecar.go` asserting "no current figures live in this comment" was **false** — the shipped table row directly beneath it carries three;
- the replacement for the stale `39` pointed at a `t.Logf` that does not exist;
- and two more of the same shape.

All four were written by an author who had spent the whole PR thinking about nothing else. That is the argument: **prose discipline does not hold, even under maximum attention.** The generator fixed the declaration; nothing constrains the prose around it.

Two stale figures that survived until #1511 measured them, both confirmed on `main`:

```
tools/replay-fixtures.sh:110       "198 of them diverge"        — actual 140
issue1480_timing_test.go:573       "39 of the catalog's 309"    — actual 36  (line 269 of the SAME file said 36)
```

The second is the telling one: the file disagreed with itself, and had for two PRs.

## The proposal

A check that no doc comment under `tools/onboarding-factory/cmd/replay/` contains a bare integer equal to a current `catalogCensus` field value.

That is deliberately narrow and deliberately crude, and both are the point:

- It cannot know whether `140` in a sentence *means* the divergence count. It does not need to — if a number that happens to equal a census figure sits in prose, either it is that figure (and should derive) or it is a coincidence (and should be reworded or exempted with a reason). Both outcomes are improvements.
- Scoping it to `cmd/replay/` keeps it away from the rest of the tree, where the same integer is meaningless.

## Design notes worth deciding

- **Exemptions carry reasons, not silence** — the `applyWritesNoUserFile` / `nilTolerant` shape used elsewhere in this repo, where an entry names the file and why. Keys should be existence-checked so an exemption that stopped naming a real site fails rather than quietly passing.
- **It is itself a guard, so per AGENTS.md it "passes the moment it is written"** and owes a deliberate mutation seen red — a committed fixture comment carrying a stale census figure that the check must report, plus a `want:false` case pinning a legitimate bare integer it must leave alone. Without the second, the natural implementation flags every `0`, `1` and `2` in the package.
[...]

Issue #1517, opened 2026-08-13
aider recordings are invisible to every Go replay gate — forEachSidecarRecording pairs only transcript.jsonl


ingo-eichhorst:
Follow-up to #1503 (PR #1511), which **documented** this gap in four places rather than closing it, and left the measurement needed to close it.

## The gap

Two walks over the same catalog disagree about which recordings exist:

- `tools/replay-fixtures.sh` walks `transcript.md` **as well as** `transcript.jsonl`.
- `forEachSidecarRecording` (the Go gates) pairs **only** `transcript.jsonl`.

So every aider recording is invisible to the Go gates. Verified independently rather than taken from the PR:

```
$ find replaydata/agents -name "transcript.md" | sed 's|.*/agents/\([^/]*\)/.*|\1|' | sort | uniq -c
  33 aider

$ find replaydata/agents/aider -name "transcript.jsonl" | wc -l
       0

$ # events.jsonl with no transcript.jsonl sibling, by adapter
  31 aider
```

aider has **zero** `transcript.jsonl` files, and all 31 unpaired sidecars in the catalog are aider's. #1511's `UnpairedSidecars: 31` is exactly this population — a produced number now, which is how the gap became legible at all.

**2 of them diverge**, and that is the entire 142-vs-140 discrepancy between `replay-fixtures.sh`'s count and the Go census.

## Why it is worth closing rather than documenting

#1342's stated design goal is that the gates are *catalog-wide, so a newly recorded fixture is covered by existing coverage rather than by somebody remembering*. **That is currently false for aider** — and silently so, which is the same species #1503 was about: not a wrong number, but a number whose scope is narrower than its name suggests.

An aider recording added tomorrow is graded by nothing, and the census reports its own blind spot as a figure rather than a failure.

## The cost is measured, and it is small

From #1511's author, after a first dismissal ("it moves three gates' pinned populations at once") was challenged in review and re-measured:

- widening `forEachSidecarRecording` to pair `transcript.md` adds **exactly 2** gradeable recordings;
- **both are merely divergent** — neither is zero-transition nor fabricated;
- it moves `knownZeroTransition`, `knownFabricated` and `knownFirstTransitionDrift` by **nothing**.

So the deferral was a scope call on an already-large PR, not a risk assessment. That is now recorded in the code rather than implied.

[...]

PR #1698, opened 2026-08-18
fix(replay): grade the Stop hook — a table row, and the stale per-pass flag that made it a no-op (#1695)

### LINKED ISSUE #1695: Replay cannot grade the Stop hook: a Stop-handling regression is invisible to every golden, on both hook adapters
Found while recording codex's first hook-bearing fixtures (#1388, PR #1694).

## The finding

`replaydata/agents/codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-08-18-00-42-27_irrlichd-0.5.10+1869727/` is the catalog's first codex recording carrying `hook_received{Stop}`:

```
{"seq":120,"kind":"hook_received","session_id":"01a011e3-…","hook_name":"Stop"}
{"seq":121,"kind":"state_transition","prev_state":"working","new_state":"ready","reason":"agent finished turn → ready"}
```

**The golden that recording produces reproduces that `ready` from the TRANSCRIPT, not from the hook.** Its three transitions:

```
{"index":0,"cause":"init","new_state":"ready","reason":"initial state"}
{"index":1,"cause":"event","prev_state":"ready","new_state":"working","reason":"force ready→working on first activity"}
{"index":2,"cause":"debounce_coalesce","prev_state":"working","new_state":"ready","reason":"agent finished turn → ready"}
```

`agent finished turn → ready` is the transcript-derived `turn_done` rule. Compare the sibling cell recorded in the same PR, `2-19_tool-gate-permission-prompt`, whose golden carries `reason: "permission prompt open → waiting"` — `state_classifier.go`'s rule 0, reachable only from `PermissionPending`, which in replay only `applyHookEvent` can set.

## Why

`sidecarReplayer.applyHookEvent` opens with `effect, ok := session.HookSignal(hookEv.HookName); if !ok { return }`, and `HookStop` is deliberately not a `hookSignalEffects` row. That table's own doc comment says why, and the reason is sound:

> HookStop genuinely cannot be a row. Its effect carries a payload (the turn's final assistant text and the waiting-cue verdict computed from it) into SignalPayload, and a name-keyed lookup has no way to supply that. HandleStopHook stays a dedicated entry point.

So this is a **consequence of a deliberate design decision, not a defect in it**. What is missing is the consequence being anywhere a reader or a gate can see it.

## Why it matters now rather than before

#1355 makes hooks the tier-1 state channel, and `Stop` is the authoritative turn-done signal (#1161) — the one that fires at *true* turn end where the transcript heuristic guesses. It is also the hook with the widest blast radius on failure: #1368's liveness watchdog demotes a channel that produces no receipts, and #1161 replaced a heuristic with it precisely because the heuristic was wrong.

Measured on the codex adapter, in the same session that produced the recording above (PR #1694 §5B): mutating `dispatchHookEvent` so the receiver stops dispatching leaves **the whole replay suite green** and reddens only `go test ./core/adapters/inbound/agents/codex/...`. That is the correct split for the receiver. But `HandleStopHook`'s *effect on state* — the thing a golden exists to pin — is graded by neither: replay cannot reach it, and the adapter suite grades the call, not the resulting transition.

**Not verified by me:** I did not run a mutation of `HandleStopHook` itself against the full suite, so "no gate reproduces a Stop-driven transition" is read off `applyHookEvent`'s early return plus the two goldens above, not off a red run. That is the first thing to do here, and it is also the acceptance test.

## Scope

1. Run the mutation that this issue is currently reasoning about: change `SessionDetector.HandleStopHook`'s effect and confirm (or refute) that every golden in `replaydata/` stays byte-identical. If something does go red, this issue is smaller than it looks — say so and close it.
2. If it is real, decide between the two shapes rather than defaulting:
   - carry the Stop payload in `lifecycle.Event` (it holds `SessionID` + `HookName` today) so replay can call `HandleStopHook` faithfully — note #1355 Phase F already flags that an `Adapter` field cannot be added to `KindHookReceived` without invalidating every hook event on disk, and this is the same class of change;
   - or accept it and make the acceptance legible: a named exclusion beside `applyHookEvent` listing which hook events replay cannot grade and why, so the next agent reads it from the code rather than rediscovering it from two goldens.
3. Either way `applyHookEvent`'s doc comment should say which recordings in the catalog now carry a `Stop` and that none of them grades it — the comment currently reasons about `Notification`/`PreCompact` on the basis that "no recording in replaydata/ fires one", and `Stop` is now the case where a recording exists and still cannot be graded.

Affects **both** hook adapters equally — claudecode's `Stop`-bearing recordings have the same property; codex is only where it was noticed.

ingo-eichhorst:
Closes #1695

## 1. The measured current state — the issue's premise holds, and is worse than stated

#1695's one unverified claim was: *"I did not run a mutation of `HandleStopHook` itself against the full suite, so 'no gate reproduces a Stop-driven transition' is read off `applyHookEvent`'s early return plus the two goldens, not off a red run."* Run:

**M1 — `SessionDetector.HandleStopHook` stops holding `SignalTurnDone`** (the whole `d.signals.Hold(…, session.SignalTurnDone, session.SignalPayload{…}, …)` deleted, the two parameters discarded):

```
$ go test ./tools/onboarding-factory/cmd/replay/... -count=1
ok  	irrlicht/tools/onboarding-factory/cmd/replay	6.106s        <- GREEN

$ go test ./core/application/services/... ./core/domain/session/... ./core/e2e/... -count=1
ok  	irrlicht/core/application/services	23.775s
ok  	irrlicht/core/domain/session	0.289s
ok  	irrlicht/core/e2e	6.225s                                     <- GREEN

$ go test ./core/adapters/inbound/agents/... -count=1
(17 packages, all ok)                                              <- GREEN
```

So the issue is confirmed and then some. It expected the adapter suites to grade *something*; they grade the CALL, and every `HandleStopHook` in a `_test.go` before this PR is a mock (`claudecode`, `codex`, `copilot` `mockTarget`). **`HandleStopHook`'s effect on state was graded by nothing in the repo** — not replay, not services, not e2e, not the adapters.

Two claims in #1695 are corrected by measurement rather than argument:

- **"Affects both hook adapters equally — claudecode's `Stop`-bearing recordings have the same property."** `replaydata/agents/claudecode/` carries **no** Stop-bearing sidecar. Its 15 hook-bearing recordings are `PermissionRequest` / `PreToolUse` / `PostToolUse` / `PostToolUseFailure` only. The catalog holds exactly three Stops: codex `2-13`, and two co-resident **claude-code** sessions inside `copilot/4-2` and `hermes/4-2`.
- **"`HookStop` genuinely cannot be a row"** (from `hookSignalEffects`' own doc comment, quoted approvingly by the issue). See §2.

Unrelated flake seen once during M1 and disclosed rather than folded in: `TestReadLauncherEnv_Tmux_{AttachedClientIsTheHost,DetachedResolvesNoHost}` in `processlifecycle` (the #1586 fixture family). Three clean re-runs on the final tree, and this diff touches no file in that package (`git diff --name-only | grep processlifecycle` → empty).

### Rejected: leaving it and writing a named exclusion (the issue's option 2b)

Because the stated reason for the exclusion is wrong in exactly one step. `hookSignalEffects` says a name-keyed lookup "has no way to supply" the Stop payload. True. It does not follow that a name-only caller cannot act — read `signalPolicies`' turn-done row:

```go
apply: func(c holdContext) {
    c.Metrics.HookTurnDone = true
    if c.Payload.LastAssistantText != "" { c.Metrics.LastAssistantText = c.Payload.LastAssistantText }
    if c.Payload.WaitingCue          { c.Metrics.PendingWaitingCue = true }
},
```

Both payload fields are **guarded**. A payload-free hold delivers the signal's whole assertion (the turn ended, authoritatively) and the degradation is **one-directional**: it can add to the cue verdict and can never clear or mask one the transcript already found. It also cannot pin a session in `ready` — that row is `consumeOnce: true`, so the hold is deleted the first time it is applied, which answers the #1355 Phase B worry directly rather than by assertion.

So the row is an honest floor, not a lie, and that is now a test rather than this paragraph: `TestPayloadFreeTurnDoneNeverClearsATranscriptCue`.

**The daemon's live behaviour does not change.** Verified by reading every call site, not assumed: all three receivers dispatch Stop through a dedicated `case HookStop:` arm to `HandleStopHook` (`claudecode/hooks.go:359`, `codex/hooks.go:203`, `copilot/hooks.go:305`), and the only non-switch `HandlePermissionHook` caller passes `HookNotification` explicitly (`copilot/hooks.go:344`). Nothing reaches the table with `"Stop"` at runtime. The table's other reader is the replay harness, which is the point.

### Rejected: carrying the Stop payload in `lifecycle.Event` (the issue's option 2a)

Not needed, and it would have been the more expensive of two ways to be wrong. All three Stops on disk predate any such field, so replay would apply an empty payload to them anyway — i.e. exactly what the row already gives, at the cost of a schema change #1355 Phase F flags as invalidating every hook event on disk.

### Chosen: the row, plus the reason it was not enough

**Adding the row alone changed zero bytes of zero goldens.** That is the half worth reading twice, because it looks exactly like "the row does nothing, the doc comment was right":

```
$ # HookStop: {Signal: SignalTurnDone} added, nothing else
$ go test ./tools/onboarding-factory/cmd/replay/... -count=1
ok  	irrlicht/tools/onboarding-factory/cmd/replay	6.113s
```

Instrumented, the classifier had already answered:

```
DBG hook=Stop ok=true state=working lastMetrics=true
DBG overlay hookTurnDone=true noSubst=true agentDone=true classify=ready/"agent finished turn → ready"
```

`applyHookEvent` copies `r.lastMetrics` — the last **transcript batch**'s metrics — and `NoSubstantiveActivity` is a **per-pass** flag (`tailer.go`: `scan.linesParsed > 0 && !scan.substantive`; `session_detector_activity.go:315-319` says so in as many words). A hook pass is a different pass that parses no transcript lines, so carrying the flag across asserted something the harness never observed, and `runClassifier`'s mirror of #329's short-circuit discarded the transition one instruction before it would have been emitted. The daemon never had this problem: a hook's synthetic activity goes through `processActivity` → `RefreshOnActivity`, which **recomputes** the flag for that pass.

One line, `domainMetrics.NoSubstantiveActivity = false`, with the declared limit stated where it lives: `false` approximates that recomputation rather than performing it, and errs toward reproducing the daemon's recorded transition rather than dropping it.

### The issue's option 3 needed no new field

"Make the golden record which SOURCE produced a transition." The mechanism already existed — `transition.Cause` — and was unreachable. With both halves fixed, codex `2-13`'s golden carries a hook-produced `ready` at `cause: "hook"` and a transcript-produced `ready` at `cause: "debounce_coalesce"`, in the same file:

```diff
   { "index": 2,
-    "event_index": 11, "virtual_time": "…T00:42:38.225446+02:00", "cause": "debounce_coalesce",
+    "event_index": -1, "virtual_time": "…T00:42:36.207640+02:00", "cause": "hook",
     "prev_state": "working", "new_state": "ready", "reason": "agent finished turn → ready" },
+  { "index": 3, … "cause": "debounce_coalesce", "prev_state": "ready",   "new_state": "working" },
+  { "index": 4, … "cause": "debounce_coalesce", "prev_state": "working", "new_state": "ready"  },
```

`00:42:36.207640` is byte-identical to the sidecar's `hook_received{Stop}` ts at seq 120, and the transition reproduces the daemon's seq 121. `extended_check` goes from `recorded 4 / replayed 2 / 2 missing_in_replay` to **`recorded 4 / replayed 4 / ordered_mismatches: none`**.

### What I did not do

- No new golden field, no `decided_by_tier` in the replay report. It would rewrite all ~399 goldens in the commit that introduces the thing they assert — the argument `extendedCheck.TimeDeltas`' `json:"-"` already makes.
- `waiting_cue.go` is untouched. The transcript path stays authoritative and unchanged; nothing here deletes inference.
- `HookNotification` / `HookPreCompact` stay out of the table: still no recording fires one. Unchanged from before.
- No re-recording. Every recording in `replaydata/` is byte-identical; only three `.golden` files moved.

### R1 — the Stop channel: remove the `HookStop` row

```
--- FAIL: TestFixtureReplayByteIdentity/transcript.jsonl#121
    fixtures_test.go:39: replay output differs from golden
      replaydata/agents/codex/scenarios/2-13_turn-end-terminal-text/…/transcript.jsonl.replay.json.golden
--- FAIL: TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog
--- FAIL: TestCatalogCensusMatchesTheCommittedFigures
--- FAIL: TestStopHookIsGradedByTheCommittedCatalog
    issue1695_stop_hook_test.go:160: the catalog carries 3 Stop-bearing recording(s) and the replay
    reproduces NONE of their Stops. Either hookSignalEffects lost its HookStop row, or
    applyHookEvent stopped reaching the classifier — both leave every golden byte-identical
    and both are #1695 returning
FAIL	irrlicht/tools/onboarding-factory/cmd/replay
```

**Exactly one golden: codex `2-13`.** That is the honest number and it is smaller than #1694's three, because it is the whole population — one of the catalog's three Stops is gradeable and the other two are not, for reasons that have nothing to do with Stop handling:

| recording | Stops | reproduced | why not |
|---|---|---|---|
| `codex/2-13_turn-end-terminal-text` | 1 | **1** | — |
| `copilot/4-2_multiple-agents-same-workspace` | 1 | 0 | the Stop names co-resident claude-code session `31e6174f…`; the replay drives copilot's `2889f1c5…` |
| `hermes/4-2_multiple-agents-same-workspace` | 1 | 0 | `sidecar cannot drive a replay: no transcript_activity events with file_size for primary session …` |

That table is not prose — it is `stopHookCensus`, machine-printed by `TestStopHookIsGradedByTheCommittedCatalog` and pasted, per #1503's idiom. A new Stop-bearing recording fails it by name, and a catalog with **no** Stop at all is a `t.Fatal` refusal rather than a vacuous pass. A Stop counts as reproduced only when a replayed transition sits at the hook's **own** virtual time — exact attribution, because `cause: "hook"` alone is also produced by the 5 sidecars carrying a `PermissionRequest`/`PreToolUse`.

### R2 — the per-pass flag: put the stale `NoSubstantiveActivity` back

```
--- FAIL: TestFixtureReplayByteIdentity/transcript.jsonl#126  → codex/2-19_tool-gate-permission-prompt
--- FAIL: TestFixtureReplayByteIdentity/transcript.jsonl#121  → codex/2-13_turn-end-terminal-text
--- FAIL: TestFixtureReplayByteIdentity/transcript.jsonl#97   → claudecode/5-4_architect-editor-pair
--- FAIL: TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog
--- FAIL: TestCatalogCensusMatchesTheCommittedFigures
--- FAIL: TestStopHookIsGradedByTheCommittedCatalog
```

Three recordings — every hook-bearing recording whose hook produces a transition, across two adapters.

### M1 / M2 — `HandleStopHook`'s own effect (`TestSessionDetector_StopHook_DrivesTheTurnDoneTransition`)

**M1**, the hold deleted — both arms red:

```
--- FAIL: …/plain_turn_end_→_ready
    after Stop hook: state = "working", want "ready" — the Stop hook is the authoritative
    turn-done signal (#1161); without its hold the transcript tail alone says the turn is still running
--- FAIL: …/turn_ended_on_a_question_→_waiting
    after Stop hook: state = "working", want "waiting" — …
```

**M2**, the hold kept and the **payload dropped** — i.e. precisely what the new table row alone delivers:

```
--- FAIL: …/turn_ended_on_a_question_→_waiting
    after Stop hook: state = "ready", want "waiting" — the cue was computed over the hook's
    full message (#1150); the transcript tail below carries none, so only the payload can
    route this to waiting
```

One arm red, one green. That asymmetry *is* the argument for `HandleStopHook` remaining a dedicated entry point beside the row rather than being replaced by it — and it is a property no golden can ever reach, because the payload is not a field of `lifecycle.Event`.

### M3 — `TestPayloadFreeTurnDoneNeverClearsATranscriptCue`, both payload guards removed

```
--- FAIL: …/cue_found_by_the_transcript_survives_an_empty_payload
    empty payload cleared PendingWaitingCue: the row would force a question-ended turn to ready
    empty payload overwrote LastAssistantText = "", want the transcript's own text
    IsWaitingForUserInput() is false after an empty-payload turn-done overlay: the cue was lost
--- FAIL: TestSignalHolds_TurnDonePreservesPriorText
--- FAIL: TestSignalHolds_TurnDoneCueIsAdditive
```

Stated rather than implied: the two **guards** were already locked by those pre-existing tests. What the new one adds is the **conclusion** neither states — that after an empty-payload overlay the metrics still answer `IsWaitingForUserInput()` true — which is the entire argument for the row and is what would survive an `apply` rewritten to reach those fields another way. Its second arm stays green under the mutation; it locks the enrichment direction.

## 4. Goldens: three moved, across two adapters — said out loud

The brief flags cross-adapter golden churn as a reconsider signal, so: **it is cross-adapter, deliberately, and here is the argument.** The change is in the shared harness and the shared domain table, so "the adapter I touched" is *every hook-bearing recording whose hook produces a transition* — which is three of ~399. Every move is toward the daemon:

| golden | before | after | daemon's own ts |
|---|---|---|---|
| `codex/2-13` | `cause: debounce_coalesce` @ `00:42:38.225`, 2 of 4 transitions, 2 missing | `cause: hook` @ `00:42:36.208` + the two it was hiding, **4 of 4, no mismatches** | hook seq 120 `00:42:36.20764`, transition seq 121 |
| `codex/2-19` | `cause: debounce_coalesce` @ `00:40:48.977` (**1.96s** late) | `cause: hook` @ `00:40:47.019` (**exact**) | hook seq 191 `00:40:47.019525` |
| `claudecode/5-4` | `cause: debounce_coalesce` @ `23:34:37.398` (~99ms late) | `cause: hook` @ `23:34:37.293` (**~6ms**) | hook seq 48910 `23:34:37.293254`, transition seq 48911 `…37.299248` |

One honest wart, disclosed because it is a real difference and not only an improvement: `2-19`'s transition now fires at the right *instant* but records replay's last-known classifier inputs (`last_event_type: function_call_output`, `has_open_tool: false`) where the daemon's recorded `inputs` say `function_call` / `has_open_tool_call: true`. `applyHookEvent` has always classified against last-known metrics and cannot re-tail at the hook's moment — no sidecar records a stat there, the same limitation `readBoundaryFor` documents. The transition, its state pair and its rule all match the daemon; the ancillary fields beside it are the harness's, as before.

## 5. Ratchets and derived figures moved, each with its reason

| figure | change | why |
|---|---|---|
| `censusOfTheCommittedCatalog.Divergent` | 143 → 142 | codex `2-13` stopped diverging (4 of 4 reproduced). Pasted from the failing run, as #1503 intends. |
| `…DivergentByCountsAndKinds` | 142 → 141 | same recording, same reason; the near-miss spelling stays exactly one below. |
| `maxRecordingsDriftingOverThreshold` | 107 → **105** | #1388 raised this to 107 for these two codex recordings and recorded the cause as "structural — the daemon flips on the POST, replay flips at the next debounce boundary". It was the short-circuit. Both leave the population; re-tightened rather than banked as slack, per the #1517 note the constant carries. |
| `minKindMatchedPairs` | 832 → **834** | `2-13` now contributes two more kind-matched pairs. |
| `minMeasuredRecordings` | 277 → 277 | unchanged — `2-13` was already measuring. A recording joins that population once and then only adds pairs. |
| #1518 census-lint exemptions | 1 deleted, 1 added, 1 re-pointed | the collisions walked one row back up an **unchanged** #1478 table, mirroring #1388's move with the opposite sign. Recorded in that file as the strongest available evidence the table is not restating a live figure: a restatement cannot track the figure in both directions. |

## 6. Gates

| gate | result |
|---|---|
| `go run ./tools/onboarding-factory/cmd/of validate` | **PASS** — `catalog + cells are schema-valid and referentially consistent` |
| `tools/preflight.sh --only go` | **PASS** 6/6 — gofmt, core module tests, onboarding-factory tests, replaydata validate, recording-rig smoke test, replay fixtures |
| `tools/replay-fixtures.sh` (unscoped) | **PASS**, exit 0 — `142 recording(s)` divergent, agreeing with the census the Go side measures |
| `tools/preflight.sh --only arch` | **PASS** — composite 7.9315462 → 7.9316414, no category regression |
| `tools/skill-lint.sh` | exit 0 (AGENTS.md is not a skill file; run because the diff edits repo prose) |
| `--only web` / `posix` / `bash` / `security` / `swift` | **not run** — the diff touches no `web/` tree, no `#!/bin/sh` or bash file, no dependency manifest and no `platforms/macos/` file. Stated rather than claimed green. |

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1438,6 +1438,52 @@ Before marking a ticket done, run the full suite — every layer must pass:
   nothing objecting. One recording remains in `knownZeroTransition` because
   reaching it needs 69ms, i.e. it can only be bought by making two goldens
   assert something false — the trade this whole line of work exists to refuse.
+- Replay's hook channel: replay grades a recorded hook through exactly two
+  shared things — `session.hookSignalEffects`, which decides whether a hook
+  NAME does anything at all, and `applyHookEvent`, which decides whether the
+  effect reaches the classifier. Both had to be right before any hook was
+  gradeable, and #1695 is the run where only the first was looked at.
+  **A hook absent from the table is silently dropped**, which is #1320's
+  lesson; `HookStop` was absent on the stated grounds that its effect "carries
+  a payload a name-keyed lookup cannot supply". True about the payload, false
+  about the consequence: `signalPolicies`' turn-done `apply` guards BOTH
+  payload fields, so a payload-free hold still asserts `HookTurnDone` and the
+  degradation is one-directional — it can add to the cue verdict and can never
+  clear one the transcript found. That row is now present, and
+  `TestPayloadFreeTurnDoneNeverClearsATranscriptCue` is the property rather
+  than the paragraph. What the row does NOT replace is
+  `SessionDetector.HandleStopHook`, which has the payload and must keep it:
+  dropping the payload while keeping the hold reddens exactly one arm of
+  `TestSessionDetector_StopHook_DrivesTheTurnDoneTransition`, the turn that
+  ended on a question, and that arm is the whole argument for the two
+  entry points coexisting.
+  **The second half is the one that reads as health.** With the row added and
+  nothing else, every golden stayed byte-identical — `applyHookEvent` copied
+  `r.lastMetrics`, whose `NoSubstantiveActivity` belongs to the last TRANSCRIPT
+  batch, and `runClassifier`'s mirror of #329's short-circuit then discarded a
+  classification that had already answered `ready`. The flag is PER PASS and a
+  hook pass is a different pass; the daemon never had the problem because its
+  synthetic activity event goes through `RefreshOnActivity`, which recomputes
+  it. So a hook table row is worth checking end-to-end against a recording, not
+  just against the table's own test.
+  The payoff is that `cause` in a golden finally means something: a
+  hook-produced `ready` (`cause: "hook"`, at the hook's own timestamp) and a
+  transcript-produced one (`cause: "debounce_coalesce"`) are different bytes,
+  so the provenance question needed no new field. codex's
+  `2-13_turn-end-terminal-text` went from 2 of 4 recorded transitions
+  reproduced to 4 of 4 with zero mismatches, and #1388's two ">1s drift"
+  entries left that population — those were the short-circuit, not the
+  "structural" debounce lag they were recorded as.
+  Only ONE recording in the catalog can grade a Stop, and the register that
+  says so is machine-generated (`stopHookCensus`,
+  `TestStopHookIsGradedByTheCommittedCatalog`) because a hand-written one is
+  what #1695 had to correct: `replaydata/agents/claudecode/` carries no
+  Stop-bearing sidecar at all, and the catalog's other two Stops belong to
+  co-resident claude-code sessions inside another adapter's multi-agent
+  recording — one naming a session the replay does not drive, one on a sidecar
+  that cannot drive a replay. A catalog with no Stop at all is a REFUSAL there,
+  not a pass.
+
 - Replay goldens (when a recording or replay-output change is in play):
   regenerate with `UPDATE_REPLAY_GOLDENS=1 go test
   ./tools/onboarding-factory/cmd/replay/... -count=1` (the `-count=1`
--- a/core/application/services/session_detector_stop_hook_e2e_test.go
+++ b/core/application/services/session_detector_stop_hook_e2e_test.go
@@ -1,0 +1,111 @@
+package services_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"irrlicht/core/domain/session"
+)
+
+// TestSessionDetector_StopHook_DrivesTheTurnDoneTransition covers what #1695
+// measured to be covered by NOTHING: SessionDetector.HandleStopHook's EFFECT.
+//
+// The gap was total and easy to mistake for coverage. Every `HandleStopHook`
+// reference in a `_test.go` before this file is a mock — claudecode's, codex's
+// and copilot's `mockTarget` — so the receiving adapters grade the CALL, and
+// deleting the `signals.Hold(…, SignalTurnDone, …)` line from the real handler
+// left `go test ./core/...` green apart from an unrelated tmux flake, plus the
+// whole replay byte-identity gate green. #1695's fix makes the replay half of
+// that red for a recording that carries a Stop; this file is the other half,
+// and it is the half no golden can reach, because the payload the daemon
+// threads in — the turn's final assistant text and the waiting-cue verdict
+// computed over the FULL message (#1150) — is not a field of lifecycle.Event
+// and so is absent from every recording on disk.
+//
+// Both arms were seen red against the deletion above. The second is also red
+// against a weaker mutation — dropping the payload and keeping the hold, i.e.
+// exactly what the hookSignalEffects row alone delivers — which is why
+// HandleStopHook stays a dedicated entry point beside that row rather than
+// being replaced by it.
+func TestSessionDetector_StopHook_DrivesTheTurnDoneTransition(t *testing.T) {
+	tests := []struct {
+		name string
+		// what the hook's payload carries, i.e. what only the daemon has.
+		lastAssistantText string
+		waitingCue        bool
+		want              string
+		why               string
+	}{
+		{
+			name:              "plain turn end → ready",
+			lastAssistantText: "Done. All tests pass.",
+			waitingCue:        false,
+			want:              session.StateReady,
+			why: "the Stop hook is the authoritative turn-done signal (#1161); " +
+				"without its hold the transcript tail alone says the turn is still running",
+		},
+		{
+			name:              "turn ended on a question → waiting",
+			lastAssistantText: "I can take either branch — which one should I target?",
+			waitingCue:        true,
+			want:              session.StateWaiting,
+			why: "the cue was computed over the hook's full message (#1150); the " +
+				"transcript tail below carries none, so only the payload can route this to waiting",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tw := newMockAgentWatcher()
+			pw := newMockProcessWatcher()
+			repo := newMockRepo()
+
+			const sid = "stop1"
+			const path = "/home/.claude/projects/-Users-test/stop1.jsonl"
+
+			// The transcript says the turn is NOT over: "assistant_message" is
+			// codex's spelling and is deliberately outside IsAgentDone's
+			// "assistant"/"assistant_output" fallback, so nothing here can
+			// reach ready on its own. The tail carries no cue either, so the
+			// waiting arm cannot be satisfied by prose the classifier already
+			// sees. Every state below is therefore the hook's doing.
+			metrics := &funcMetrics{fn: func(_, _ string) (*session.SessionMetrics, error) {
+				return &session.SessionMetrics{
+					LastEventType:     "assistant_message",
+					LastAssistantText: "Working on it.",
+				}, nil
+			}}
+
+			det := newDetectorWithMetrics(tw, pw, repo, metrics)
+			repo.states[sid] = &session.SessionState{
+				SessionID:      sid,
+				State:          session.StateWorking,
+				TranscriptPath: path,
+				FirstSeen:      time.Now().Unix(),
+				UpdatedAt:      time.Now().Unix(),
+				Metrics:        &session.SessionMetrics{LastEventType: "assistant_message"},
+			}
+
+			// LIFO: registering the drain first makes teardown cancel() then
+			// wait for Run to return, which is what reaps the goroutine.
+			done := make(chan error, 1)
+			defer func() { <-done }()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			go func() { done <- det.Run(ctx) }()
+
+			time.Sleep(20 * time.Millisecond) // let seedFromDisk finish
+
+			det.HandleStopHook(sid, path, tt.lastAssistantText, tt.waitingCue)
+			waitForSessionState(repo, sid, tt.want, 5*time.Second)
+
+			repo.mu.Lock()
+			got := repo.lastSavedState[sid]
+			repo.mu.Unlock()
+			if got != tt.want {
+				t.Fatalf("after Stop hook: state = %q, want %q — %s", got, tt.want, tt.why)
+			}
+		})
+	}
+}
--- a/core/domain/session/hook_signal.go
+++ b/core/domain/session/hook_signal.go
@@ -64,13 +64,34 @@ type HookSignalEffect struct {
 // hookSignalEffects is the single source of truth for hook-name → SignalKind
 // for the hooks a caller may act on knowing only the name.
 //
-// Three hooks are absent, for two different reasons — worth keeping apart,
-// because they imply different things about whether a future row belongs here:
+// Two hooks are absent. A third, HookStop, was — and the reason it was is
+// worth stating, because it was measured to be wrong in exactly one step
+// (#1695):
 //
-//   - HookStop genuinely cannot be a row. Its effect carries a payload (the
-//     turn's final assistant text and the waiting-cue verdict computed from it)
-//     into SignalPayload, and a name-keyed lookup has no way to supply that.
-//     HandleStopHook stays a dedicated entry point.
+//   - HookStop carries a payload (the turn's final assistant text and the
+//     waiting-cue verdict computed over the *full* message rather than the
+//     200-rune transcript tail, #1150) which a name-keyed lookup cannot
+//     supply. That much is true, and it is why HandleStopHook stays a
+//     dedicated entry point — the daemon has the payload and must not lose it.
+//     What does NOT follow, and is what kept this row out until #1695, is that
+//     a name-only caller therefore cannot act. Read signalPolicies' turn-done
+//     row: its apply sets HookTurnDone unconditionally and guards BOTH payload
+//     fields (`if c.Payload.LastAssistantText != ""`, `if c.Payload.WaitingCue`),
+//     so a payload-free hold delivers the signal's whole assertion — the turn
+//     ended, authoritatively — and the degradation is ONE-DIRECTIONAL: it can
+//     add nothing to the cue verdict, and can never clear or mask a cue the
+//     transcript already found on its own. It also cannot pin, because that
+//     row is consumeOnce. TestPayloadFreeTurnDoneNeverClearsATranscriptCue is
+//     that property as a test rather than as this paragraph.
+//     So the row is an honest floor, not a lie: a name-only caller gets a
+//     correct-but-weaker verdict, and the one thing it gives up is a cue that
+//     sits outside the transcript tail.
+//     The cost of leaving it out was concrete. The replay harness reads this
+//     table, so with no row it dropped every recorded Stop, and the
+//     hook-produced `ready` in codex's 2-13_turn-end-terminal-text golden was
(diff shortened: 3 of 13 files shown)


ingo-eichhorst:
Follow-up on the pushed head (`8be7958e`): CodeScene flagged `TestStopHookIsGradedByTheCommittedCatalog` for *Bumpy Road Ahead* + *Complex Method* — the `WalkDir` closure carried four early returns and the grading inline. The per-sidecar body is now `measureStopsIn`, so the closure is a filter and nothing else.

Re-verified rather than assumed: with the `HookStop` row removed from `session.hookSignalEffects`, the refactored guard still fires with the same diagnosis —

```
--- FAIL: TestStopHookIsGradedByTheCommittedCatalog (0.13s)
    issue1695_stop_hook_test.go:136: the catalog carries 3 Stop-bearing recording(s) and the
    replay reproduces NONE of their Stops. Either hookSignalEffects lost its HookStop row, or
    applyHookEvent stopped reaching the classifier — both leave every golden byte-identical
    and both are #1695 returning
```

One behaviour change beyond the extraction, and it is the reason to prefer the helper: a `runReplay` error used to `t.Errorf` and then leave that recording OUT of the measured census, so the same run reported both "this recording failed to replay" and "this recording is not in the catalog" — the second being false. It now lands in the census with `Why: "runReplay failed"`, which is the same distinction the rest of this file draws between *no gradeable Stop* and *never looked at*.

Gates re-run on the amended head: `preflight --only go` 6/6 with one exception stated honestly — `core module tests` went red once on `TestReadLauncherEnv_Tmux_AttachedClientIsTheHost`, the #1586 tmux fixture family. Evidence it is unrelated rather than a dismissal: this diff touches no file under `processlifecycle` (`git diff --name-only | grep processlifecycle` is empty), the package passes on re-run and passed in four other runs today including the earlier full `--only go`, and CI's `go-test` passed on the previous head. `of validate`, `replay-fixtures.sh` (exit 0, 142 divergent), `--only arch` (no regression) all PASS.

Filed while here, both with their own measurements: #1699 (claudecode carries zero Stop-bearing recordings — the adapter the channel was built for is the one no golden grades) and #1700 (the extended check never compares classifier *inputs*, so replay and the daemon can disagree about why a transition happened; its premise measurement is step 1 and may shrink it).

ingo-eichhorst:
Head is now `e65bdf5f`. Second CodeScene round, and I am stopping here with the number stated rather than the flattering half of it.

I split the census comparison out of `TestStopHookIsGradedByTheCommittedCatalog` into `assertSomeStopIsReproduced` + `assertCensusMatches`, which leaves the test body at five statements instead of thirty lines. **CodeScene's file score went the other way: 9.61 → 9.31**, with the *Bumpy Road* finding moving from the test to `assertCensusMatches` and the advisory one changing from *Complex Method* to *Overall Code Complexity*. So by their metric the second refactor was a small regression, and I am keeping it anyway because the branching it flags is the function's content: the comparison runs in **both** directions with a distinct diagnosis per direction — an unlisted recording is a change in what this repo can grade arriving unannounced, a listed one the walk no longer reaches is the #1517 pairing blindness — and collapsing either arm is exactly the single-sided check `TestUninstallTaskEtaReadsOnlyTheInstructionsSlice` and #1450 both exist to refuse.

Per AGENTS.md this check is advisory (neither branch protection nor the Protect Main ruleset requires it). I have looked closer twice, which is what a red there asks for; a third round would be trading the assertion's shape for the score.

Re-verified after the split rather than assumed — the guard still fires with the `HookStop` row removed:

```
--- FAIL: TestStopHookIsGradedByTheCommittedCatalog (0.08s)
    issue1695_stop_hook_test.go:131: the catalog carries 3 Stop-bearing recording(s) and the
    replay reproduces NONE of their Stops. …
```

`tools/preflight.sh --only go` **6/6 PASS** on this head, tmux flake included this time (it passed). Every other CI check is green: 16 pass, 1 advisory fail.

Issue #1654, opened 2026-08-16
ars.yml can never push its badge: the Protect Main ruleset refuses the commit, and the README badge has been stale since 2026-04-26


ingo-eichhorst:
Found while fixing #1644 (the scan step's silent failure). Not that defect and not #1641's
either — both of those are about a step *reporting* wrongly. This one is about the push
being **impossible**, and #1647 is what made it visible.

## What

`.github/workflows/ars.yml`'s "Commit badge update" step pushes the badge commit directly to
`main`. The repository's "Protect Main" ruleset requires changes to go through a pull request,
so the push is refused every time:

```
remote: error: GH013: Repository rule violations found for refs/heads/main.
remote: - Changes must be made through a pull request.
remote: - 2 of 2 required status checks are expected.
 ! [remote rejected] main -> main (push declined due to repository rule violations)
```

All five retries fail for that same reason, which is not a transient condition a retry can
clear.

## Measured

Three consecutive `main` runs since #1647 landed the loud failure, each identical:
[31959328634](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31959328634),
[31960611419](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31960611419),
[31962161264](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31962161264).

Before #1647 the same thing was happening behind a green check. The last successful badge
commit on `main` is `244f11d2` (**2026-04-26**) — every run since has computed a badge,
rewritten README.md on the runner, failed to push it, and exited 0. `README.md` therefore
shows `ARS-Agent--Ready 8.1/10` while the scan of the current tree returns
`ARS-Agent--Assisted 7.9/10` (measured both on a runner —
[31957949370](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/31957949370),
the last pre-#1647 run, whose log reads `ARS Badge line: …7.9%2F10-yellow` and
`README updated with ARS badge URL` — and locally with the pinned CLI).

So: roughly four months of a badge that has been wrong, and the mechanism was already broken
before either #1641 or #1644 could contribute to it. #1644's fix does not touch this and
cannot: it makes the *scan* honest, and the scan is working.

`codescene-badge.yml` writes its badge to a gist instead and is unaffected; the README's
CodeScene badge is a `shields.io/endpoint` pointing at that gist, which is one of the
available shapes for a fix here.

## Options (not a recommendation, this needs a call on repo policy)

[...]

Issue #1644, opened 2026-08-16
ars.yml: a failed ARS scan leaves the badge stale with three green steps and no check saying so


ingo-eichhorst:
Found by the sweep in #1641 (ars.yml's push-retry loop reporting success after all 5 attempts
fail). Not that defect — a different member of the same family, two steps earlier in the same
job, and it has the identical user-visible symptom: a README badge that quietly stops tracking
`main` behind a green history.

## What

`.github/workflows/ars.yml`, "Run ARS scan":

```yaml
run: |
  ars scan ./core --no-llm --badge > ars-output.txt 2>&1 || true
  cat ars-output.txt
```

and "Extract and update ARS badge":

```yaml
run: |
  ARS_BADGE=$(grep "img.shields.io/badge/ARS" ars-output.txt | head -1 || echo "")
  echo "ARS Badge line: $ARS_BADGE"

  if [ -n "$ARS_BADGE" ]; then
    ...
  fi
```

If `ars scan` fails for any reason — the CLI is missing, the pinned commit stops building, a
flag is renamed, the scan panics — `|| true` swallows it, the step's last statement is `cat`
and succeeds, so the step is green. `ars-output.txt` then holds an error rather than a badge
line, `grep` matches nothing, `ARS_BADGE` is empty, the whole `if` is skipped, README.md is
never touched, and the "Commit badge update" step reports `No badge changes to commit`.

Three consecutive green steps, a green job, and the badge silently frozen at whatever it last
was. AGENTS.md: *"absence of a finding and inability to look must never produce the same
output"* — here "the score did not change" and "the scan never produced a score" print
different words but exactly the same colour, and the colour is the only thing anyone reads.

## Measured

The three step bodies EXTRACTED from `.github/workflows/ars.yml` and run under GitHub's
documented default shell for a step declaring no `shell:`
(`bash --noprofile --norc -e -o pipefail`), with `ars` replaced by a stub that fails:

```
ars: command failed: no such subcommand
STEP 'Run ARS scan' EXIT=0
ARS Badge line:
STEP 'Extract and update ARS badge' EXIT=0
No badge changes to commit
STEP 'Commit badge update' EXIT=0
=> README badge line unchanged: 1 occurrence(s), job GREEN throughout
```

## Suggested fix

Probably **not** by removing the `|| true` on `ars scan`: it may well be there because `ars
scan` exits non-zero on a low score, and no comment says either way — worth checking against
[...]

PR #1711, opened 2026-08-19
fix(ci): announce a failed gist write instead of aborting or reporting success (#1710)

### LINKED ISSUE #1710: Both gist-badge workflows report success when the write never happened, and their transport-failure branch is unreachable
Found while implementing #1654 (PR #1709), which used `codescene-badge.yml` as its reference
implementation. Two defects in that reference were reproduced as deliberate mutations of the new
ARS step and are guarded there — but they are **live** in the two workflows they were copied from.

`.github/workflows/codescene-badge.yml` and `.github/workflows/coverage.yml` share one gist-write
shape. Both steps declare no `shell:`, so GitHub runs them as `bash -e {0}` — errexit on, **no
pipefail**.

## Defect 1 — a transport failure exits with no `::error::`, and the branch written for it is unreachable

```sh
# curl writes "000" to %{http_code} when no response was received (DNS/TLS/connect failure).
http_code=$(curl -sS -o /tmp/gist-response.json -w '%{http_code}' … )
echo "HTTP $http_code"
if [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
  echo "::error::Gist update failed (HTTP $http_code)"
```

Under errexit a **simple assignment from a failing command substitution aborts the step**, so a
transport failure never reaches `echo` or the refusal. Measured, in that exact shape:

```
$ bash --noprofile --norc -e m5a.sh
curl: (6) Could not resolve host: this-host-does-not-exist-1654.invalid
### exit: 6 ###
```

No `HTTP …` line, no `::error::`. The step is red, so this is *loud but undiagnosed* (#1702's shape,
not #1678's) — and the comment above the call documents a `000` path that errexit makes unreachable.
A comment describing a branch that cannot execute is worse than no comment: it is the reason nobody
re-checks it.

## Defect 2 — a non-numeric `http_code` reports SUCCESS

`[ "$http_code" -lt 200 ]` on an empty or non-numeric value errors and evaluates **false**, and both
disjuncts do, so execution falls through to the success path. Errexit does not save it: a failing
command in an `if` condition is exempt. Measured:

```
$ bash --noprofile --norc -e m5b.sh
HTTP
m5b.sh: line 3: [: : integer expression expected
m5b.sh: line 3: [: : integer expression expected
SUCCESS LINE REACHED — badge reported published
### exit: 0 ###
```

Exit 0. The workflow reports the badge published; the gist was never written. This is the exact
failure class #1654 was about — a badge silently frozen while its job is green — reproduced one
workflow over, and it would freeze the **coverage** and **code health** badges the same way.

## Why it matters

These two workflows publish two of the four README badges, and after #1654 the same gist carries the
third. A silent stale badge here has a four-month precedent (#1654, `244f11d2` → 2026-08-19).

## The fix, already written

PR #1709's `Update Gist with ARS score` step carries both guards: `|| curl_status=$?` around the
capture with its own named refusal, and a numeric-shape check on `%{http_code}` before it is
compared. Port those two, plus the `case`-based needle table if the tests are extended. Per AGENTS.md
each added guard owes a deliberate mutation seen red — M5a and M5b in
`tools/lib/ars-badge-push_test.sh` are the committed shapes to copy.

## Provenance

**Measured**: both behaviours, reproduced above in the workflows' own derived shell. The code shapes,
read from `origin/main` at `326ba009` — `codescene-badge.yml` lines ~74-91 and `coverage.yml` lines
~74-85.

**Assumed**: that no third workflow shares the pattern. I grepped these two; I did not audit every
workflow for `http_code`.

## Related

- #1654 / PR #1709 — where both defects were found, and where the guarded version lives
- #1702, #1678 — loud-but-undiagnosed vs loud-and-misdiagnosed
- #1645 — found / not found / could not determine

ingo-eichhorst:
Fixes #1710.

`.github/workflows/coverage.yml` and `.github/workflows/codescene-badge.yml` share one
gist-write shape. Both steps declare no `shell:`, so the runner gives them `bash -e {0}` —
errexit on, **no pipefail**. Two defects followed, both **live on `main`**, on the two
README badges that were still working.

## Defect 1 — a transport failure exited with no `::error::`

Under errexit an assignment from a failing command substitution aborts the step, so
`http_code=$(curl …)` never reached the `echo` or the refusal below it. The comment above
that call documented curl's `000` code *as though the refusal read it* — a comment
describing a branch errexit made unreachable, which is the reason nobody re-checked it.

## Defect 2 — a non-numeric `http_code` reported SUCCESS

`[ "" -lt 200 ]` errors and evaluates **false**, and so does the other disjunct, so
execution fell through to the success path. Errexit does not save it: a failing command in
an `if` condition is exempt. **The job reported the badge published on a run where the gist
was never written** — #1654's own failure class, a silently frozen badge behind a green
job, one workflow over.

## The port

`.github/workflows/ars.yml`'s `Update Gist with ARS score` step (PR #1709) already carries
both guards. They are **ported, not redesigned**, so the three gist-writing workflows now
carry one shape:

- `curl_status=0` + `|| curl_status=$?` around the capture, reported next to the code —
  "the API said no" and "the API was never reached" are different failures and one refusal
  names both.
- a `case`-based numeric-shape check on `%{http_code}`, consulted **before** the arithmetic
  comparisons (the `||` chain short-circuits, so a garbage code produces a refusal rather
  than two `integer expression expected` lines).

Four things beyond the straight port:

- **Each refusal names its own badge.** The six refusals across the two workflows are
  asserted pairwise distinguishable, and every arm asserts the other five wordings are
  ABSENT (#1677/#1678) — a shared non-zero exit is satisfied by refusals that all fire
  together, and the cross-workflow half catches the copy-paste this whole class came from.
- **Each step prints what it published** (`Published coverage badge to gist: 78.7%`). There
  was no success line at all before, so a healthy run and a fall-through were the same
  bytes. Its absence is what makes defect 2 assertable.
- **`/tmp/gist-*.json` → workspace-relative**, matching ars.yml. A shared absolute path is
  not isolable per test case, so two arms would read each other's response file.
- **The `000` comment is DELETED, not reworded.** The branch is reachable now, and the
  replacement comment says so and says which value carries the failure.

## Before / after of the misleading comment

```diff
-          # curl writes "000" to %{http_code} when no response was received (DNS/TLS/connect failure).
-          http_code=$(curl -sS -o /tmp/gist-response.json -w '%{http_code}' \
+          # `|| curl_status=$?`, not a bare `$(curl …)` (#1710). curl exits
+          # non-zero when no response was received (DNS/TLS/connect failure),
+          # and this step declares no `shell:`, so the runner gives it
+          # `bash -e {0}` — under which an assignment from a failing command
+          # substitution ABORTS the step at this line. Neither the `echo` nor
+          # the refusal below was ever reached, so a transport failure came
+          # out as a red X over a log with no `HTTP …` line and no annotation
+          # in it. Measured in this step's own derived shell: exit 6, silent.
+          # The comment that used to sit here documented curl's `000` code as
+          # if the refusal read it; errexit made that branch unreachable for
+          # the whole of this step's life, and a comment describing a branch
+          # that cannot execute is why nobody re-checked it. It is reachable
+          # now — `curl_status` carries the failure and `000` is what
+          # `$http_code` holds when it fires.
+          curl_status=0
+          http_code=$(curl -sS -o gist-response.json -w '%{http_code}' \
```

## Where the test lives, and why

`tools/lib/gist-badge-guards_test.sh` — a **second file**, covering **both** workflows.

`ars-badge-push_test.sh`'s header states the rule it was written under: *"One harness per
workflow … splitting them would duplicate [a workflow file, a derivation, an assertion
vocabulary and one preflight trigger] and let the copies disagree."* Folding two more
workflows into it breaks the first half of that sentence, in a file whose name already
fails to describe its contents.

The same rule applied here produces one file for two workflows rather than two files: what
these two share is not a workflow file but the **gist-write shape** — one `run:` step each,
identical apart from the badge name, the gist filename and the rendered variable — plus the
curl stub, the refusal vocabulary, the "nothing was sent" witness and one preflight
trigger. What is genuinely shared with `ars-badge-push_test.sh` is
`tools/lib/workflow-step.sh`, and that **is** shared: the body extraction and the shell
derivation both come from it, so nothing here decides for itself what invocation a step
gets (`bash -e`, derived — these steps declare no `shell:`). Extracting the ~70 lines of
local harness instead would rewrite a file merged one commit ago, whose ten obligations
would all need replaying as locks on the extraction, for no property either file gains.

Seven obligations per workflow: **G1** publishes (vacuity guard: PATCH recorded, token,
gist URL, payload equality, success line), **G2** unset `GIST_SECRET`, **G3** unset
`COVERAGE_GIST_ID`, **G4** non-2xx at 404 *and* 503, **G5** transport failure, **G6**
non-numeric `http_code` (asserted as the **absence** of the success line), **G7** the
pre-fix body committed verbatim and re-measured — the permanent vacuity guard, since if
errexit ever stopped aborting on a failed capture, or `[` stopped evaluating a bad integer
as false, the guards would be protecting nothing and would pass for the wrong reason.

**Nothing can reach the live gist.** `curl` is shadowed by a shell FUNCTION, never a PATH
shim a body could route around; `with_curl` refuses if the stub did not land in the
prelude; and a final sweep asserts no recorded call anywhere names
`9f14c8e5f25c1ccf5d6500c1685fd9fb` — with a positive control, since a sweep that found no
calls would pass it vacuously (`the curl stub recorded 14 request(s) across the arms`).

`preflight.sh`'s `tools` trigger gains both workflows — its sixth widening for this reason
(#1591, #1629, #1639, #1641, #1645). Confirmed firing: `shell-lib-suite: tools/lib — found
19, skipped 0 (none), ran 19, failed 0` (18 before), and `bash-lint` reports
`ok  tools/lib/gist-badge-guards_test.sh`.

## Mutation evidence

Everything here is NEW, so every obligation passes the moment it is written. Seven
deliberate mutations, each applied to the real file, each asserted to have actually changed
it, each seen red, each reverted.

**M1 — `|| curl_status=$?` removed from `codescene-badge.yml`** (defect 1 restored). Note
the shape: the step still exits non-zero (7, from the abort), so **every status arm stays
green** and only the wording arms discriminate — #1644's measurement, two families on.

```
96c96
<             "https://api.github.com/gists/$COVERAGE_GIST_ID") || curl_status=$?
---
>             "https://api.github.com/gists/$COVERAGE_GIST_ID")

  FAIL: ...as a workflow error annotation — expected [output containing: ::error::] got [ ]
  FAIL: ...naming CODESCENE_GISTFAIL specifically — expected [output containing: the code health badge gist update FAILED] got [ ]
  FAIL: ...naming curl's own exit status, not only the code — expected [output containing: curl exit 7] got [ ]
  FAIL: ...and reaching the line the abort used to skip — expected [output containing: HTTP 000] got [ ]
  FAIL: ...and saying there was no answer to print — expected [output containing: no response body] got [ ]
  FAIL: ...and the curl status is captured rather than aborting the step — expected [output containing: || curl_status=] got [...]
gist-badge-guards_test: FAILURES
```

**M2 — the numeric-shape check removed from `codescene-badge.yml`** (defect 2 restored).
Exit **0**, with the success line printed — the badge reported published.

```
107,111c107
<           case "$http_code" in
<             ''|*[!0-9]*) http_code_usable=0 ;;
<             *)           http_code_usable=1 ;;
<           esac
<           if [ "$curl_status" -ne 0 ] || [ "$http_code_usable" -ne 1 ] || [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
---
>           if [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then

  FAIL: G6: an unreadable HTTP code fails the step rather than passing it — expected [a non-zero exit] got [exit 0 :: HTTP not-a-code (curl exit 0) .step.sh: line 58: [: not-a-code: integer expression expected
  FAIL: ...naming CODESCENE_GISTFAIL specifically — expected [output containing: the code health badge gist update FAILED] got [...]
  FAIL: ...and never reaches the success line — expected [no output containing: Published code health badge to gist: 8.6/10] got [...]
  FAIL: ...having decided the code's SHAPE rather than erroring on it — expected [no output containing: integer expression expected] got [...]
  FAIL: ...and the code's shape is decided before it is compared — expected [output containing: http_code_usable] got [...]
gist-badge-guards_test: FAILURES
```

**M3 — `|| curl_status=$?` removed from `coverage.yml`.** The point of M3/M4 is that the
SECOND workflow is genuinely graded, not carried along by the first.

```
97c97
<             "https://api.github.com/gists/$COVERAGE_GIST_ID") || curl_status=$?
---
>             "https://api.github.com/gists/$COVERAGE_GIST_ID")

  FAIL: ...as a workflow error annotation — expected [output containing: ::error::] got [ ]
  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [output containing: the coverage badge gist update FAILED] got [ ]
  FAIL: ...naming curl's own exit status, not only the code — expected [output containing: curl exit 7] got [ ]
  FAIL: ...and reaching the line the abort used to skip — expected [output containing: HTTP 000] got [ ]
  FAIL: ...and saying there was no answer to print — expected [output containing: no response body] got [ ]
  FAIL: ...and the curl status is captured rather than aborting the step — expected [output containing: || curl_status=] got [...]
gist-badge-guards_test: FAILURES
```

**M4 — the numeric-shape check removed from `coverage.yml`.**

```
108,112c108
<           case "$http_code" in
<             ''|*[!0-9]*) http_code_usable=0 ;;
<             *)           http_code_usable=1 ;;
<           esac
<           if [ "$curl_status" -ne 0 ] || [ "$http_code_usable" -ne 1 ] || [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
---
>           if [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then

  FAIL: G6: an unreadable HTTP code fails the step rather than passing it — expected [a non-zero exit] got [exit 0 :: HTTP not-a-code (curl exit 0) .step.sh: line 57: [: not-a-code: integer expression
  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [output containing: the coverage badge gist update FAILED] got [...]
  FAIL: ...and never reaches the success line — expected [no output containing: Published coverage badge to gist: 78.7%] got [...]
  FAIL: ...having decided the code's SHAPE rather than erroring on it — expected [no output containing: integer expression expected] got [...]
  FAIL: ...and the code's shape is decided before it is compared — expected [output containing: http_code_usable] got [...]
gist-badge-guards_test: FAILURES
```

**M5 — `coverage.yml`'s refusal given `codescene-badge.yml`'s sentence.** Every status arm
is unchanged; only the cross-workflow wording arms can see it. This is the mutation for the
#1677/#1678 obligation.

```
113c113
<             echo "::error::the coverage badge gist update FAILED (HTTP $http_code, curl exit $curl_status), so the gist still holds its pr
---
>             echo "::error::the code health badge gist update FAILED (HTTP $http_code, curl exit $curl_status), so the gist still holds its

  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [output containing: the coverage badge gist update FAILED] got [HTTP 404 ...]
  FAIL: ...and not confused with CODESCENE_GISTFAIL — expected [no output containing: the code health badge gist update FAILED] got [HTTP 404 ...]
  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [...] got [HTTP 503 ...]
  FAIL: ...and not confused with CODESCENE_GISTFAIL — expected [...] got [HTTP 503 ...]
  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [...] got [HTTP 000 (curl exit 7) ...]
  FAIL: ...and not confused with CODESCENE_GISTFAIL — expected [...] got [HTTP 000 (curl exit 7) ...]
  FAIL: ...naming COVERAGE_GISTFAIL specifically — expected [...] got [HTTP not-a-code ...]
  FAIL: ...and not confused with CODESCENE_GISTFAIL — expected [...] got [HTTP not-a-code ...]
gist-badge-guards_test: FAILURES
```

**M6 — `coverage.yml`'s success line deleted.** The vacuity guard: a healthy run must still
reach the success line and say what it published.

```
121d120
<           echo "Published coverage badge to gist: ${COVERAGE}%"

  FAIL: ...saying what it published — expected [output containing: Published coverage badge to gist: 78.7%] got [HTTP 200 (curl exit 0) ]
gist-badge-guards_test: FAILURES
```

**M7 — `preflight.sh`'s `tools` trigger narrowed back.** A gate that silently does not
match the file it covers is the failure mode this repo keeps removing.

```
-  run_gate_scoped '…^\.github/workflows/(ars|codescene-badge|coverage|macos-swift|replaydata-deletion-guard|test)\.yml$' \
+  run_gate_scoped '…^\.github/workflows/(ars|macos-swift|replaydata-deletion-guard|test)\.yml$' \

  FAIL: the tools gate fires on a diff touching .github/workflows/codescene-badge.yml — expected [a match] got [no match against: ...]
  FAIL: the tools gate fires on a diff touching .github/workflows/coverage.yml — expected [a match] got [no match against: ...]
gist-badge-guards_test: FAILURES
```

Reverted, and green again: `gist-badge-guards_test: ALL PASS`.

## The issue's one assumed premise, now measured

The issue marked *"that no third workflow shares the pattern"* as **assumed**. Measured:
`grep -rn "http_code" .github/workflows/` and `grep -rn "curl" .github/workflows/` over all
ten workflow files return `ars.yml` (already fixed), `codescene-badge.yml` and
`coverage.yml` — and one unrelated hit, a prose comment in `linux.yml` about
`site/install.sh` reaching users as `curl … | sh`. `git grep "api.github.com/gists"` and
`git grep GIST_SECRET` agree: three workflows, no fourth writer anywhere in the repo. **The
premise holds.**

## Verification

- `bash tools/lib/gist-badge-guards_test.sh` → `ALL PASS`
- `tools/preflight.sh --only tools` → PASS (`found 19, skipped 0, ran 19, failed 0`)
- `tools/preflight.sh --only posix` → PASS
- `tools/bash-lint.sh` → `ALL PASS`, and it reads the new file by name
- pre-push hook (`--changed --budget 540`): posix / bash / tools / gofmt / swift all PASS
- both workflows parse (`yaml.safe_load`), jobs `badge` and `coverage` intact
- **no `permissions:` change of any kind.** `codescene-badge.yml` keeps its workflow-level
  `contents: read`; `coverage.yml` keeps its pre-existing job-level `contents: write`,
  byte-identical to `HEAD` — see "Left alone" below.

## What cannot be proven before merge

Both workflows run on `push` to `main` and `workflow_dispatch` only — **no schedule**
(checked: neither file carries `on: schedule`). So the gist write itself is unexercised
until this lands. On the first post-merge run, watch:

1. Both steps print their new success lines — `Published coverage badge to gist: <pct>%`
   and `Published code health badge to gist: <score>/10` — and `HTTP 200 (curl exit 0)`
   above them.
2. Both badges still render their current values: **code health 8.6/10** and **coverage
   78.7%**. If either goes blank or stale, the gist write is the place to look — but with
   this change a failed write is now loud, so a *green* job with a stale badge would itself
   be new information.
3. `ars.yml`'s ARS badge is unaffected (that file is untouched), but it writes the same
   gist, so a regression there would show up as all three going stale together.

## Left alone, deliberately

`coverage.yml`'s job holds `permissions: contents: write`. Nothing in that job commits or
pushes — `actions/checkout` reads, and the badge goes to a gist via a separate `GIST_SECRET`
PAT — so the grant looks unnecessary, and `codescene-badge.yml`/`ars.yml` both run the same
kind of job at `contents: read`. It is **not** touched here: it is out of #1710's scope, it
is a behaviour change with no test in this PR to hold it, and dropping a permission is the
kind of edit that should be its own reviewable diff. Flagging it rather than fixing it
silently.

### DIFF
--- a/.github/workflows/codescene-badge.yml
+++ b/.github/workflows/codescene-badge.yml
@@ -60,32 +60,61 @@ jobs:
           COVERAGE_GIST_ID: ${{ secrets.COVERAGE_GIST_ID }}
         run: |
           if [ -z "$GIST_SECRET" ]; then
-            echo "::error::GIST_SECRET is not set in repository secrets"
+            echo "::error::GIST_SECRET is not set in repository secrets, so the code health badge could not be published at all. The gist still holds its previous codescene.json and README's CodeScene badge renders the previous score."
             exit 1
           fi
           if [ -z "$COVERAGE_GIST_ID" ]; then
-            echo "::error::COVERAGE_GIST_ID is not set in repository secrets"
+            echo "::error::COVERAGE_GIST_ID is not set in repository secrets, so there is no gist for the code health badge to be published to. README's CodeScene badge renders the previous score."
             exit 1
           fi
           MESSAGE=$(printf '%.1f/10' "$SCORE")
           BADGE_JSON="{\"schemaVersion\":1,\"label\":\"code health\",\"message\":\"${MESSAGE}\",\"color\":\"${COLOR}\"}"
-          jq -n --arg content "$BADGE_JSON" '{"files":{"codescene.json":{"content":$content}}}' > /tmp/gist-payload.json
-          # curl writes "000" to %{http_code} when no response was received (DNS/TLS/connect failure).
-          http_code=$(curl -sS -o /tmp/gist-response.json -w '%{http_code}' \
+          jq -n --arg content "$BADGE_JSON" '{"files":{"codescene.json":{"content":$content}}}' > gist-payload.json
+
+          # `|| curl_status=$?`, not a bare `$(curl …)` (#1710). curl exits
+          # non-zero when no response was received (DNS/TLS/connect failure),
+          # and this step declares no `shell:`, so the runner gives it
+          # `bash -e {0}` — under which an assignment from a failing command
+          # substitution ABORTS the step at this line. Neither the `echo` nor
+          # the refusal below was ever reached, so a transport failure came
+          # out as a red X over a log with no `HTTP …` line and no annotation
+          # in it. Measured in this step's own derived shell: exit 6, silent.
+          # The comment that used to sit here documented curl's `000` code as
+          # if the refusal read it; errexit made that branch unreachable for
+          # the whole of this step's life, and a comment describing a branch
+          # that cannot execute is why nobody re-checked it. It is reachable
+          # now — `curl_status` carries the failure and `000` is what
+          # `$http_code` holds when it fires.
+          curl_status=0
+          http_code=$(curl -sS -o gist-response.json -w '%{http_code}' \
             --connect-timeout 10 --max-time 30 \
             --retry 3 --retry-delay 2 --retry-all-errors \
             -X PATCH \
             -H "Authorization: token $GIST_SECRET" \
             -H "Content-Type: application/json" \
-            -d @/tmp/gist-payload.json \
-            "https://api.github.com/gists/$COVERAGE_GIST_ID")
-          echo "HTTP $http_code"
-          if [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
-            echo "::error::Gist update failed (HTTP $http_code)"
-            if [ -s /tmp/gist-response.json ]; then
-              cat /tmp/gist-response.json
+            -d @gist-payload.json \
+            "https://api.github.com/gists/$COVERAGE_GIST_ID") || curl_status=$?
+          echo "HTTP $http_code (curl exit $curl_status)"
+          # ...and the SHAPE of `$http_code` is decided before it is compared.
+          # `[ "" -lt 200 ]` errors and evaluates FALSE, and so does the other
+          # disjunct, so execution fell through to the success path — errexit
+          # does not save it, a failing command in an `if` condition being
+          # exempt. Measured: two `integer expression expected` lines and
+          # exit 0, i.e. the job reporting the badge published on a run where
+          # the gist was never written. Same shape as ars.yml's step, so the
+          # three gist-writing workflows converge on one guard rather than
+          # three.
+          case "$http_code" in
+            ''|*[!0-9]*) http_code_usable=0 ;;
+            *)           http_code_usable=1 ;;
+          esac
+          if [ "$curl_status" -ne 0 ] || [ "$http_code_usable" -ne 1 ] || [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
+            echo "::error::the code health badge gist update FAILED (HTTP $http_code, curl exit $curl_status), so the gist still holds its previous codescene.json and README's CodeScene badge renders the previous score."
+            if [ -s gist-response.json ]; then
+              cat gist-response.json
             else
-              echo "(no response body — likely transport failure)"
+              echo "(no response body — likely a transport failure)"
             fi
             exit 1
           fi
+          echo "Published code health badge to gist: $MESSAGE"
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -62,32 +62,61 @@ jobs:
           COVERAGE_GIST_ID: ${{ secrets.COVERAGE_GIST_ID }}
         run: |
           if [ -z "$GIST_SECRET" ]; then
-            echo "::error::GIST_SECRET is not set in repository secrets"
+            echo "::error::GIST_SECRET is not set in repository secrets, so the coverage badge could not be published at all. The gist still holds its previous coverage.json and README's Coverage badge renders the previous percentage."
             exit 1
           fi
           if [ -z "$COVERAGE_GIST_ID" ]; then
-            echo "::error::COVERAGE_GIST_ID is not set in repository secrets"
+            echo "::error::COVERAGE_GIST_ID is not set in repository secrets, so there is no gist for the coverage badge to be published to. README's Coverage badge renders the previous percentage."
             exit 1
           fi
           BADGE_JSON="{\"schemaVersion\":1,\"label\":\"coverage\",\"message\":\"${COVERAGE}%\",\"color\":\"${COLOR}\"}"
-          jq -n --arg content "$BADGE_JSON" '{"files":{"coverage.json":{"content":$content}}}' > /tmp/gist-payload.json
-          # curl writes "000" to %{http_code} when no response was received (DNS/TLS/connect failure).
-          http_code=$(curl -sS -o /tmp/gist-response.json -w '%{http_code}' \
+          jq -n --arg content "$BADGE_JSON" '{"files":{"coverage.json":{"content":$content}}}' > gist-payload.json
+
+          # `|| curl_status=$?`, not a bare `$(curl …)` (#1710). curl exits
+          # non-zero when no response was received (DNS/TLS/connect failure),
+          # and this step declares no `shell:`, so the runner gives it
+          # `bash -e {0}` — under which an assignment from a failing command
+          # substitution ABORTS the step at this line. Neither the `echo` nor
+          # the refusal below was ever reached, so a transport failure came
+          # out as a red X over a log with no `HTTP …` line and no annotation
+          # in it. Measured in this step's own derived shell: exit 6, silent.
+          # The comment that used to sit here documented curl's `000` code as
+          # if the refusal read it; errexit made that branch unreachable for
+          # the whole of this step's life, and a comment describing a branch
+          # that cannot execute is why nobody re-checked it. It is reachable
+          # now — `curl_status` carries the failure and `000` is what
+          # `$http_code` holds when it fires.
+          curl_status=0
+          http_code=$(curl -sS -o gist-response.json -w '%{http_code}' \
             --connect-timeout 10 --max-time 30 \
             --retry 3 --retry-delay 2 --retry-all-errors \
             -X PATCH \
             -H "Authorization: token $GIST_SECRET" \
             -H "Content-Type: application/json" \
-            -d @/tmp/gist-payload.json \
-            "https://api.github.com/gists/$COVERAGE_GIST_ID")
-          echo "HTTP $http_code"
-          if [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
-            echo "::error::Gist update failed (HTTP $http_code)"
-            if [ -s /tmp/gist-response.json ]; then
-              cat /tmp/gist-response.json
+            -d @gist-payload.json \
+            "https://api.github.com/gists/$COVERAGE_GIST_ID") || curl_status=$?
+          echo "HTTP $http_code (curl exit $curl_status)"
+          # ...and the SHAPE of `$http_code` is decided before it is compared.
+          # `[ "" -lt 200 ]` errors and evaluates FALSE, and so does the other
+          # disjunct, so execution fell through to the success path — errexit
+          # does not save it, a failing command in an `if` condition being
+          # exempt. Measured: two `integer expression expected` lines and
+          # exit 0, i.e. the job reporting the badge published on a run where
+          # the gist was never written. Same shape as ars.yml's step, so the
+          # three gist-writing workflows converge on one guard rather than
+          # three.
+          case "$http_code" in
+            ''|*[!0-9]*) http_code_usable=0 ;;
+            *)           http_code_usable=1 ;;
+          esac
+          if [ "$curl_status" -ne 0 ] || [ "$http_code_usable" -ne 1 ] || [ "$http_code" -lt 200 ] || [ "$http_code" -ge 300 ]; then
+            echo "::error::the coverage badge gist update FAILED (HTTP $http_code, curl exit $curl_status), so the gist still holds its previous coverage.json and README's Coverage badge renders the previous percentage."
+            if [ -s gist-response.json ]; then
+              cat gist-response.json
             else
-              echo "(no response body — likely transport failure)"
+              echo "(no response body — likely a transport failure)"
             fi
             exit 1
           fi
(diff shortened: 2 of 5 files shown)


ingo-eichhorst:
One of the "cannot be proven before merge" items is weaker than I wrote it, in the good direction — the **workspace-relative path** half is already proven in production, not inferred.

`ars.yml` carries the identical shape (`>gist-payload.json`, `-d @gist-payload.json`, `-o gist-response.json`) and its first post-#1709 run on `main` succeeded. From run [32290308986](https://github.com/ingo-eichhorst/Irrlicht/actions/runs/32290308986):

```
jq -n --arg content "$BADGE_JSON" {files:{ars.json:{content:}}} >gist-payload.json
  -d @gist-payload.json \
echo "HTTP $http_code (curl exit $curl_status)"
...
HTTP 200 (curl exit 0)
```

and the gist now serves `{"schemaVersion":1,"label":"ARS","message":"Agent-Assisted 7.9/10","color":"yellow"}`.

So `$GITHUB_WORKSPACE` is writable and is the steps' cwd, on the same runner image these two jobs use. What genuinely stays unproven until merge is the part specific to this PR: that **these two** steps reach their new success lines and that the two badges keep rendering. Measured just now, read-only, for the "still rendering" baseline to compare against:

```
coverage.json:  {"schemaVersion":1,"label":"coverage","message":"78.7%","color":"yellow"}
codescene.json: {"schemaVersion":1,"label":"code health","message":"8.6/10","color":"brightgreen"}
```

Those are also exactly the values the test fixtures drive (`COVERAGE=78.7 COLOR=yellow`, `SCORE=8.6 COLOR=brightgreen`), so a failure message from the lock reads like the real badge rather than like a placeholder.

Issue #1530, opened 2026-08-14
macOS Swift test suite cannot run on a GitHub runner — 4 blockers (35 host-dependent snapshots, a hang, #1523, a home-dependent assertion)


ingo-eichhorst:
## The macOS Swift *test* suite cannot currently run on a GitHub runner — four separate blockers

Split out of #1509 / PR #1526, which added the first CI workflow that builds Swift at all.
That PR gates **`swift build`** only. This issue is the remaining half: gating `swift test`.

Everything below is measured on `macos-latest` from PR #1526's own runs, with XCTest output
streamed under a pty (`script -q /dev/null`) — without that, a hung run uploads an empty log
and none of this is visible.

**Baseline: 227 tests pass on the runner.** The suite is close, not far. Blockers:

### 1. Image snapshot tests are host-dependent — 35 failures

Every failure is an `assertSnapshot` comparison, across six suites:

| suite | failures |
|---|---|
| `HistoryViewSnapshotTests` | 12 |
| `SessionRowSnapshotTests` | 7 |
| `GroupViewSnapshotTests` | 6 |
| `PermissionWizardEffectErrorRenderTests` | 3 |
| `SessionListUnappliedGrantsWiringTests` (in `UnappliedGrantsBannerRenderTests.swift`) | 2 |
| `BackchannelRulesViewSnapshotTests` | 2 |

Sample: `Newly-taken snapshot@(350.0, 48.0) does not match reference@(350.0, 48.0).`

The references are committed at 700×96 pixels for a 350×48-point view — a 2× backing scale,
because the NSView strategy rasterises through `bitmapImageRepForCachingDisplay`, which for a
window-less view takes the **main screen's** scale. A runner's virtual display need not be 2×,
and its text rasterisation need not match a developer Mac's either. **Which of the two it is
— scale or rasterisation — is not established**; the error text reports point size, not pixel
size, so it does not distinguish them. Determining that is step one.

### 2. `SessionRowSnapshotTests.testFixtureAntigravityGhost` hangs

After the failures above, the run stopped dead:

```
11:35:34.928  Test Case '-[IrrlichtTests.SessionRowSnapshotTests testFixtureAntigravityGhost]' started.
11:44:53.523  ##[error]The operation was canceled.
```

Nine minutes, no further output. Plausibly the snapshot **failure path** rather than the
render — that path writes failure/reference/diff PNGs and computes a diff image — but
**unverified**.

[...]

Issue #1523, opened 2026-08-14
swift test intermittently aborts (SIGABRT) in SoundPlayerTests, silently dropping 8 suites


ingo-eichhorst:
## `swift test` intermittently aborts (SIGABRT) inside `SoundPlayerTests`, silently dropping the rest of the suite

Split out of #1509, which observed the trailing `error: Exited with unexpected signal code 6`
but did not diagnose it and noted it as *"does not appear to hide a failure"*.
**It does hide results** — see below.

### What it is

Not a runner shutdown artifact. XCTest's stall detector fires on a hung expectation
and aborts the process deliberately:

```
Test Case '-[IrrlichtTests.SoundPlayerTests testInstallCustomTranscodesM4AToLPCMCAF]' started.
A stall was detected while waiting on expectations in
IrrlichtTests.SoundPlayerTests.testInstallCustomTranscodesM4AToLPCMCAFAndReturnError:()
at platforms/macos/Tests/SoundPlayerTests.swift:214.
There's no IDE/Xcode connection to restart the process, so aborting to break the apparent deadlock.
```

`SoundPlayerTests.swift:214` is `XCTWaiter.wait(for:timeout:)` inside `waitInstallResult`,
which drives `SoundPlayer.installCustom` — an AVFoundation AAC→LPCM/CAF transcode.
When that callback never arrives, the process aborts.

### It truncates the run

Measured on the same worktree, same command (`swift test --skip LauncherHarnessTests`):

| run | outcome | suites completed | overall total printed? |
|---|---|---|---|
| aborted run | `exit=1`, signal 6 | **32** | **no** — `IrrlichtPackageTests.xctest` never reports |
| healthy run | `exit=0` | **40** | yes — `Executed 310 tests, with 0 failures` |

Eight suites never executed: `SoundPlayerTests` (partially), `UnappliedGrantsAggregateTests`,
`UnappliedGrantsBannerAbsenceTests`, `UnappliedGrantsBannerRenderTests`, `ViewTooltipsLintTests`,
plus the aggregate totals. A failure in any of those would have been invisible.

Note the failure mode: **all 24 tests that did run reported `0 failures`**, so the run looks
healthy unless you notice the missing overall total. The exit code is the only reliable
signal, and it is easy to attribute to the cosmetic-looking trailing `error:` line.

### Frequency

Intermittent. Observed **1 abort in 5** full runs on this machine (aborted at 11:32; four
[...]

Issue #1509, opened 2026-08-13
Two macOS ghost-row snapshot tests are red on main — logo AA drift recurring after #1044, invisible because Swift is not CI-gated


ingo-eichhorst:
## Two macOS snapshot tests are red on `main`, and nothing anywhere reports it

Reproduced on `main` at `89f7c167` (not on a branch — see "How this surfaced" below):

```
$ cd platforms/macos && swift test --skip LauncherHarnessTests
Tests/SessionRowSnapshotTests.swift:522: error: -[IrrlichtTests.SessionRowSnapshotTests testFixtureAntigravityGhost] : failed - Snapshot does not match reference.
Tests/SessionRowSnapshotTests.swift:384: error: -[IrrlichtTests.SessionRowSnapshotTests testGhostRowPID0NilMetrics] : failed - Snapshot does not match reference.
Test Suite 'SessionRowSnapshotTests' failed
	 Executed 24 tests, with 2 failures (0 unexpected)
```

Every other Swift suite passes. It is exactly these two.

## What actually differs — measured, not guessed

Both failures have the **identical** signature, which is itself the finding:

| | value |
|---|---|
| pixels differing | **415 of 67 200** (0.62%) |
| max channel delta | **72** |
| bounding box | **(655, 35) – (681, 61)** — a 26×26 square |
| alpha channel | **identical** (RGB only) |
| image size | 700×96, unchanged |

Two independent tests differing in the same 415 pixels, in the same 26×26 box, by the same maximum delta, means one shared drawn element — not two rendering regressions.

Cropped and magnified 6×, the element is the **Irrlicht logo glyph** (the blue/red/green arch). Left = committed reference, right = produced today:

- The **shape is identical** — same arch, same three segments, same position.
- The **stroke antialiasing differs**: the current toolchain renders it slightly thicker and smoother; the committed reference is thinner and more jagged.

So this is a toolchain rendering difference, not a UI change. The two failing tests are the two **ghost-row** cases, which is consistent with the logo being drawn only there.

## This is a recurrence, and that is the actual problem

The two reference PNGs were last regenerated in `e77e3a83`:

```
e77e3a83 2026-07-14  fix(macos): regenerate 2 snapshot references drifted by toolchain AA (#1044)
```

[...]

PR #1713, opened 2026-08-19
docs(macos): accept image snapshots as reference-host-only, in one place, with the count derived (#1615)

### LINKED ISSUE #1615: Image snapshots still differ byte-for-byte on a GitHub runner after the scale fix — the rasterisation half of #1530

ingo-eichhorst:
Answers #1615's step 3 with its own **option 3**: image snapshots are graded on the
reference host only, permanently and by choice. Nothing here tries to fix the pixel
difference, and no reference PNG was regenerated —
`git status --porcelain platforms/macos/Tests/__Snapshots__` is empty.

## The decision now lives in one place

**`AGENTS.md`'s "macOS app" bullet under Testing.** `macos-swift.yml`'s header and
`ImageSnapshotCIScopeTests`' class doc now point there and restate none of it — the
reason being this file's own history: the header said "270 of 318", was corrected to
"272 of 320" by the PR that measured it, and was **115 tests stale (320 → 435)** by
today. A fact written in N places drifts in N−1.

AGENTS.md was chosen over the workflow header (CI implementation detail — the skip list
lives there, the policy does not) and over `site/docs/` (user-facing; this is a
contributor rule, and `CONTRIBUTING.md` already routes every testing question to
`AGENTS.md#testing`).

The record states, plainly: which five suites and how many tests; why re-recording is
refused (#1034/#1044, plus the runner being on a **newer** OS, which makes a re-record
chase a moving image *and* invert which host is authoritative); why `perceptualPrecision`
is refused (#1509 measured one wide enough for this drift as also wide enough to pass a
**missing architecture segment**); what has been measured; what has not; and what the
decision costs.

## The count is derived, not typed

`ImageSnapshotCIScopeTests.testTheUngatedPopulationIsExactlyTheSkippedSuites` counts the
skipped suites' tests off the **live bundle** through XCTest's own
`XCTestSuite(forTestCaseClass:)` — so it agrees with `Executed N tests` by construction,
not by a source scan agreeing with a test runner by luck. Cross-check from this branch's
own run: the census printed `total=436` and the bundle reported `Executed 436 tests`.

    ci-scope census: module=IrrlichtTests total=436 gated=388 ungated=48 skippedSuites=5

- **Pinned: `ungated` (48).** It moves only when someone adds or removes a test in a
  suite CI never runs — exactly the event the record needs surfaced, and worth one
  deliberate line in a diff.
- **Printed, not pinned: the total and the gated remainder.** They move with every test
  added anywhere in the target, so pinning them would make an unrelated PR edit a
  literal about image snapshots, and the ratchet would be reset rather than read.
- A skipped suite going **empty** is a named failure: a suite that runs on one machine
  and holds nothing runs nowhere, which reads exactly like one that passes everywhere.

## What I narrowed, and the cost it releases

`swift-snapshot-evidence` drops to **`workflow_dispatch` only** (job-level `if:`).
Narrowed rather than deleted, deliberately:

- it is the only way anyone can ever look at these pixels again (runner OS bump, a
  contributor asking whether the residual moved, #1615 reopened);
- `tools/lib/swift-snapshot-evidence_test.sh` extracts and executes that step's body, and
  deleting the job would delete that lock with it;
- `ImageSnapshotCIScopeTests` cross-checks the classification against **both** `swift
  test` invocations — deleting the job removes the second one, and moving it to a
  separate workflow file would move half the check into a file the check does not read.
  That is why the condition is on the **job**, not the workflow.

Saved: one macOS job per PR at 10× billing. The header now carries the invocation
(`gh workflow run "macOS (Swift)" --ref <branch>`) and what it produces (failure images +
`__References__` + the six primitive rasters + the run log; no diff image, by
construction — that branch of SnapshotTesting is behind
`__XCODE_BUILT_PRODUCTS_DIR_PATHS`, which SwiftPM never sets). On a PR the job now reads
**Skipped**, not absent — the distinction this workflow's no-`paths:` decision turns on.

## The failure mode this creates, stated rather than discovered

**A contributor without a Mac cannot run these 48 tests at all, and CI will not run them
either.** Worse than silence: the `swift-test` check is confidently **green**, because
the suites are skipped by name. An outside contributor's image-snapshot change is
ungraded until the maintainer next runs `tools/preflight.sh --only swift` or pushes
through the pre-push hook. That contribution shape is not hypothetical — PR #1470 merged
from a fork today touching five files under `platforms/macos/`; it touched no
image-snapshot suite, so it is the shape rather than an instance.

I did **not** build a mitigation. Two cheap ones, if you want them (each is a separate,
one-line-ish PR):

1. `.github/PULL_REQUEST_TEMPLATE.md`'s test-plan block gains a line: *"macOS image
   snapshots (`SessionRow`/`GroupView`/`History`/`BackchannelRules`/`UnappliedGrants`)
   are graded on the maintainer's Mac only — say so if you changed one and could not run
   them."*
2. `CONTRIBUTING.md`'s "Swift (app)" paragraph links the decision record.

## Mutation evidence

The tripwire is the thing this PR changes the input of, so all four were run against
**this branch**, one per foreground call, each reverted with `git status --porcelain`
asserted empty afterwards.

**1 — a non-image-snapshot suite added to the gate's skip list** (`--skip
QuotaChipClockTests`):

```
ImageSnapshotCIScopeTests.swift:232: error: -[… testWorkflowSkipListMatchesTheClassification] :
XCTAssertEqual failed: ("[… "LauncherTestHarness", "QuotaChipClockTests", "SessionList…"]") is not
equal to ("[… "LauncherTestHarness", "SessionList…"]") - macos-swift.yml's --skip list and this
file's classification disagree
```

**2 — a suite dropped from the now dispatch-only evidence job** (`--filter
HistoryViewSnapshotTests`). This is the coupling check: narrowing the trigger did **not**
drop half the cross-check.

```
ImageSnapshotCIScopeTests.swift:262: error: -[… testEvidenceJobRunsExactlyTheSuitesTheGateSkips] :
XCTAssertEqual failed: ("["Backchannel…", "GroupView…", "RasterPrimitiveEvidenceTests", …]") is not
equal to ("["Backchannel…", "GroupView…", "HistoryViewSnapshotTests", "RasterPrimitive…", …]") -
the evidence job collects a different set of suites than swift-test skips
```

**3 — one test renamed out of a permanently-skipped suite** (48 → 47), the new derived
count:

```
ImageSnapshotCIScopeTests.swift:401: error: -[… testTheUngatedPopulationIsExactlyTheSkippedSuites] :
XCTAssertEqual failed: ("47") is not equal to ("48") - the number of tests CI never grades changed:
47, committed 48.
ci-scope census: module=IrrlichtTests total=435 gated=388 ungated=47 skippedSuites=5
```

**4 — a permanently-skipped suite emptied** (all six `GroupViewSnapshotTests` methods
renamed), the rot guard:

```
ImageSnapshotCIScopeTests.swift:382: error: XCTAssertGreaterThan failed: ("0") is not greater than
("0") - GroupViewSnapshotTests is skipped in CI and now holds no tests at all — it runs on no
machine anywhere, which is indistinguishable from passing on all of them
ImageSnapshotCIScopeTests.swift:401: error: XCTAssertEqual failed: ("42") is not equal to ("48")
```

Note what mutation 4 does *not* redden: `testEveryImageSnapshotSuiteIsClassified` stays
green, because the renamed methods still contain `as: .pinnedImage`. The census is the
only thing that sees an ungated suite hollowing out.

## Verification

- `tools/preflight.sh --only swift` — **PASS**, `Executed 436 tests, with 0 failures`
  (all 48 ungated ones among them), census `total=436 gated=388 ungated=48`.
- `tools/preflight.sh --only tools` — **PASS**, `found 19, skipped 0, ran 19, failed 0`;
  includes `swift-snapshot-evidence_test.sh`, `swift-test-step_test.sh`,
  `swift-suite_test.sh` and `workflow-step_test.sh`, all of which read this workflow.
- Workflow YAML parses (`yaml.safe_load`); three jobs, evidence job's `if:` is
  `github.event_name == 'workflow_dispatch'`.
- `git grep swift-snapshot-evidence` — the only references are inside this workflow, that
  shell lock, `AGENTS.md` and the tripwire's doc; `tools/preflight.sh` and the pre-push
  hook are unaffected (their `swift` gate never ran that job).
- No reference PNG changed.

## #1530: my call is **close it, but not on this PR's strength alone**

Blockers 2 (Sparkle's modal `NSAlert`), 3 (#1523) and 4 (the `$HOME`-dependent scorer)
are fixed in #1614 and verified — `SoundPlayerTests` is absent from the skip list, so
#1523's quarantine is gone. Blocker 1's scale half is fixed; its rasterisation half is
what this PR **accepts** rather than fixes. #1530's actual goal — gating `swift test` in
CI — is met: 388 of 436 tests, against 0 before.

What stops me recommending "close it on merge" is a **fifth** host dependency, found
while checking this: the `macos-swift` run on `main` at 19:49 today (the #1470 merge)
went **red**, and not on any of the four:

```
InMemoryDefaultsTests.swift:164: error: -[… testTheDoubleAddsNothingToTheRealPreferencesDirectory] :
XCTAssertEqual failed: ("["com.apple.siri.ODDI.MetricsWorker.plist"]") is not equal to ("[]") -
The defaults double reached disk. #1661 is back…
```

That is a background OS process writing into `/Users/runner/Library/Preferences` during
the test window, accusing the double of a write it did not make — the same "unfiltered
directory witness" trade `AGENTS.md` documents for `swift-suite.sh`, but inside a test
where it is a false accusation rather than a report. Unrelated to #1615, not one of
#1530's four, and it means "the suite runs green on a runner" is not currently true on
`main`. I have not filed it — the diagnosis above is one `gh issue create` away and the
call is yours.

So: **`Fixes #1615` here; #1530 is ready to close on its own terms** (four blockers: three
fixed, one accepted with the containment written down and asserted), and the flake above
deserves its own issue rather than silently redefining what the epic tracks.

## Premises of the two issues that are no longer true

Worth flagging because both would otherwise be repeated:

- **"Nobody has diffed the actual pixels" is false.** #1628 did, from the evidence job's
  artifacts: 34 failures inside the 28 × 28 px box `SessionState.adapterIcon` draws at
  maxΔ 1-5/255, and 2 `BackchannelRulesView` control labels at maxΔ 142/197 sitting ~1
  device pixel lower.
- **The SF-Symbols/vector hypothesis is not unverified — it is measured dead.** Flat
  fill, stroked bezier, system-font text and an SF Symbol, each also in a translucent
  dynamic `Color.secondary`: all six byte-identical between runner and reference host.
  What differs is an inline-SVG `NSImage(data:)` raster, which no SwiftUI primitive
  reaches. The record says so, and says equally plainly that **why** it differs is still
  unknown — "the OS versions differ" is where the evidence stops, not a cause.
- One thing in the record is marked **predicted, not measured**: #1658 pinned the locale,
  so `testBackchannelRuleContextTokens`'s residual *should* now be rasterisation only. No
  post-#1658 runner render has been diffed to confirm it.

Fixes #1615
Refs #1530

### DIFF
--- a/.github/workflows/macos-swift.yml
+++ b/.github/workflows/macos-swift.yml
@@ -48,6 +48,14 @@ name: macOS (Swift)
 # suite on a developer Mac, where the display, the fonts and the audio stack
 # are the ones the app actually ships against.
 #
+# WHY five suites are skipped rather than fixed, and why that is now permanent,
+# is in `AGENTS.md`'s "macOS app" bullet under Testing — the decision record for
+# #1615. It is not restated here: this file's own copy of it said CI gates "270
+# of 318 tests", then "272 of 320", and was 115 tests stale by the time #1615
+# closed. The current figures are printed by
+# `ImageSnapshotCIScopeTests.testTheUngatedPopulationIsExactlyTheSkippedSuites`
+# on every run of the suite, and the ungated count is asserted there.
+#
 # A THIRD job, `swift-snapshot-evidence`, runs the five skipped suites on
 # purpose and uploads what they render (#1615, step 1). Until it existed, the
 # skipped suites failed on no machine anybody could look at, so every statement
@@ -59,17 +67,36 @@ name: macOS (Swift)
 # thing. Since #1678 those five are told apart from each other as well, not
 # only from health: the first used to be reported as the second.
 #
-# It costs a third macOS job per PR (macOS minutes bill at 10×). That is
-# deliberate for as long as the suites are skipped and #1615's step 3 is open;
-# once the policy question is answered, narrowing this job to
-# `workflow_dispatch` — or deleting it — is the cheap part.
+# Since #1615 it runs on `workflow_dispatch` ONLY, and that is the cost the
+# decision releases: it collected the evidence for a policy question that is now
+# answered, and a third macOS job per PR (macOS minutes bill at 10×) buys
+# nothing while the answer stands. It is narrowed rather than deleted because it
+# is the only way anyone can ever look at these pixels again — a runner OS bump,
+# a contributor asking whether the residual moved, or #1615 being reopened — and
+# rebuilding it from a git revert is strictly more work than dispatching it:
+#
+#   gh workflow run "macOS (Swift)" --ref <branch>
+#
+# runs all three jobs and produces two artifacts (14-day retention):
+# `swift-snapshot-evidence` — every failure image the five suites wrote, the
+# committed `__References__` beside them to diff against, and
+# `RasterPrimitiveEvidenceTests`' six primitive rasters + hashes — and
+# `swift-snapshot-evidence-log`, the run itself. There is no diff image: that
+# branch of SnapshotTesting is behind `__XCODE_BUILT_PRODUCTS_DIR_PATHS`, which
+# SwiftPM does not set, so the diff is computed off the pair after download.
+#
+# Note what it becomes on a PR: SKIPPED, not absent. That is the distinction the
+# `paths:` decision above turns on — a skipped job is reported and legible,
+# where an undispatched workflow's check reads as "still queued".
 
 on:
   pull_request:
   push:
     branches: [main]
-  # So the pixels can be re-collected on demand — for a toolchain bump on the
-  # runner, say — without pushing a commit to ask.
+  # The build and test jobs run on every event here. `swift-snapshot-evidence`
+  # runs on THIS one only (its `if:` below) — since #1615 it is the sole way its
+  # pixels are collected, which is why the trigger stays and why the job names
+  # it rather than the workflow dropping it.
   workflow_dispatch:
 
 permissions:
@@ -151,45 +178,30 @@ jobs:
       # `<target>.<class>`, so either alone stops covering the harness the
       # moment a class is added or the target is renamed.
       #
-      # The five suites after them are image snapshots that still differ from
-      # their references on a runner AFTER the scale fix, and the reason they
-      # are skipped rather than re-recorded is the substance of #1530's blocker
-      # 1. Measured on this workflow's own first run: the failure message
-      # changed from the pixel-count form (`…@(350.0, 48.0) does not match
-      # reference@(350.0, 48.0).`, 35×) to the byte form (`Newly-taken snapshot
-      # does not match reference.`, 36×), which is the branch in
-      # SnapshotTesting's `compare` that is only reachable once the pixel
-      # dimensions agree. So the scale half is fixed and a rasterisation half
-      # is now visible underneath it.
+      # The five suites after them are image snapshots that differ from their
+      # references on a runner, byte for byte, for a reason nobody has explained
+      # — and that is now a permanent, deliberate exclusion rather than a
+      # pending fix. The decision, the pixel measurements behind it, and the
+      # three levers it rejects (re-recording, a perceptual tolerance, pinning
+      # the runner's Xcode — all three measured) are in `AGENTS.md`'s "macOS
+      # app" bullet under Testing. No figure from it is repeated here; this
+      # comment carried two and both went stale.
       #
-      # It is NOT the Xcode difference this comment used to name. Measured by
-      # #1615's evidence job and its throwaway probe: all 15 Xcodes installed
-      # on `macos-latest` — 26.0 through 26.6, including 26.3.0 build 17C529,
-      # bit-for-bit the reference Mac's own — produce the same 36 failures,
-      # while that Mac on that build produces none. The toolchain is not the
-      # variable; the OS is (runner macOS 26.5.2 / 25F84, reference 26.5 /
-      # 25F71). Re-recording to make it green is the move #1034
-      # and #1044 both made and both got wrong, and #1509 measured a perceptual
-      # tolerance wide enough to absorb this drift as also wide enough to pass
-      # a missing architecture segment. The residual is #1615.
-      #
-      # What they render is not lost by being skipped: `swift-snapshot-evidence`
-      # below runs these same five and uploads the images, which is how #1615
-      # gets looked at rather than reasoned about.
+      # Two image-snapshot suites are deliberately NOT skipped: they pass on a
+      # runner, which is itself the evidence that the residual is
+      # content-dependent rather than a blanket host difference.
       #
       # This list cannot silently drift: `Tests/ImageSnapshotCIScopeTests.swift`
       # derives the set of image-snapshot suites from the test sources and
-      # cross-checks it against these very arguments — and, since the evidence
-      # job exists, against ITS arguments too, so the two invocations cannot
-      # come to disagree about which suites are the skipped ones.
+      # cross-checks it against these very arguments — and against the evidence
+      # job's arguments too, so the two invocations cannot come to disagree
+      # about which suites are the skipped ones. That file also counts what the
+      # skip costs and asserts it, so the population CI never grades cannot grow
+      # without a line in a diff.
       #
-      # This still gates 272 of 320 tests, against 0 before — measured, on this
-      # workflow's own run, not derived from the list above by hand (the first
-      # draft of this comment said "270 of 318" and was stale by exactly the
-      # two tests the tripwire itself adds). Two image-snapshot suites are
-      # deliberately NOT skipped — they pass on a runner, and their passing is
-      # itself evidence that the residual is content-dependent rather than a
-      # blanket toolchain difference.
+      # What they render is not lost by being skipped: `swift-snapshot-evidence`
+      # below runs these same five on demand and uploads the images (see the
+      # `gh workflow run` line in this file's header).
       - name: Test (bounded, streamed under a pty)
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


ingo-eichhorst:
Runner-side verification of the derived census, from this PR's own first `swift-test` job (run 32301231627, job 96223975223) — worth pasting because it closes the loop the whole change turns on. The census is derived from the **bundle**, so it reports the same three numbers on a host that executes only 388 of them, and the run then reports exactly that:

```
ci-scope census: module=IrrlichtTests total=436 gated=388 ungated=48 skippedSuites=5
ci-scope suite BackchannelRulesViewSnapshotTests tests=2 skippedInCI=true
ci-scope suite GroupViewSnapshotTests tests=6 skippedInCI=true
ci-scope suite HistoryViewSnapshotTests tests=14 skippedInCI=true
ci-scope suite PermissionWizardEffectErrorRenderTests tests=3 skippedInCI=false
ci-scope suite SessionListUnappliedGrantsWiringTests tests=2 skippedInCI=true
ci-scope suite SessionRowSnapshotTests tests=24 skippedInCI=true
ci-scope suite UnappliedGrantsBannerRenderTests tests=2 skippedInCI=false
...
Test Suite 'IrrlichtPackageTests.xctest' passed
     Executed 388 tests, with 0 failures (0 unexpected) in 6.440 seconds
```

`gated=388` (derived) == `Executed 388 tests` (measured), on the machine where they differ by the whole point of the decision. Locally the same test prints `total=436` against a local run of `Executed 436 tests`. So the figure the record refuses to restate is now produced by both hosts, in the log of every run, rather than typed into a comment on one of them.

`swift-snapshot-evidence` reported **`skipping`** on this PR, which is the narrowing working as intended — reported and legible, not absent.

One follow-up commit (39238c51): CodeScene flagged the class-enumeration loop as Bumpy Road Ahead; the superclass-chain walk is now a named predicate. Advisory gate, but it was a fair reading and the fix is smaller than the comment explaining it would have been.

Issue #1378, opened 2026-08-09
feat(copilot): hook-based state detection — the minimum viable slice of #1355


ingo-eichhorst:
Parent: #1355 — **Phase D, the minimum viable slice.** Target changed from gemini-cli to copilot; see the reorder comment on #1355 for the evidence.

Blocked by Phase B (#1360, #1361, #1362, #1364, #1365, #1368, plus #1356 and #1357). Not blocked by #1371 / #1372 / #1373 — that is the point of the target change.

## Goal

Prove that adding a hook-based state channel to an already-onboarded adapter is a **declaration**, not an implementation. No new abstraction. Parser untouched. `hookjson` unmodified. The shared receiver unmodified.

## Why copilot

All facts below are from the Phase C audit on #1355, verified against the installed CLI (1.0.78) and the shipped bundle, not from documentation — copilot's hook surface is publicly undocumented.

- User-level hook directory **outside any git repository**: `~/.copilot/hooks/*.json`, same path on all three OSes.
- **Native `type: "http"` delivery** — no shell spawn, no `curl` on `PATH`, no beacon.
- A blocked-on-user event that fires at **entry** and **only when a prompt is actually shown** (`notification` / `notification_type: "permission_prompt"`, guaranteed since 1.0.26).
- That event is **structurally non-blocking**: *"fire asynchronously … never block the session"*, fire-and-forget at every call site.
- Registering under **PascalCase** yields envelopes byte-identical to claudecode's.
- Transcript root already matches `copilot/adapter.go`'s `defaultRootDir`.

## Scope

1. `HookSpec` on the copilot agent declaration.
2. Install/uninstall of one hook file into `$COPILOT_HOME/hooks/`, behind a `KindModify` permission, honouring `COPILOT_HOME`.
3. Event → signal map:
   - `userPromptSubmitted` → working
   - `notification`, gated in the receiver on `notification_type ∈ {permission_prompt, elicitation_dialog}` → assert `SignalPermissionPrompt`
   - `agentStop` → ready
4. Version gate at **≥ 1.0.26** (below 1.0.18 there is no `notification` event; 1.0.18–1.0.25 fires it even when no prompt was shown, which would produce false `waiting` states).
5. `AssertPermissionGated`, `AssertHookEndpointFollowsBindAddr`, `AssertHookDisclosureMatchesInstalled`, and the confinement assertion from #1361.

## The two traps — handle both explicitly

[...]

Issue #1723, opened 2026-08-20
feat(antigravity): audit whether either surface (CLI or IDE) exposes a hook mechanism


ingo-eichhorst:
Broken out of epic #1355, which never scoped antigravity.

## Why antigravity is a different shape

One adapter covers **both the CLI and the IDE**, and discovery is transcript-first rather than
process-first. So "install a hook" has to answer *for which surface* — an IDE and a CLI may have
entirely different extension points, or the IDE may have one and the CLI none.

Its cost centre was never the process model: it is the **multi-root `Source`** and the path-based
session id.

## What is known

**Measured, on record:** tokens and model live in a sibling SQLite protobuf store
(`conversations/<conv>.db`), NOT in the transcript — which is why cost/output/cache were ruled
underivable in #716 (closed not-planned) and why store replay was done separately in #775/#776.

## Unverified

Whether antigravity exposes any hook, extension or notifier mechanism on either surface. Nobody has
looked.

## Practical warning for whoever picks this up

antigravity's cached OAuth token expires on a dev machine and can silently fail its keyring refresh,
needing a fresh browser sign-in. **Ask the maintainer before driving it** rather than discovering
this mid-audit.

## Shape of this ticket

**Audit first, implement second, in this same issue.** A negative result is a legitimate close: if the audit shows no usable mechanism, close with that evidence recorded. Do not open a separate audit ticket.

Absence of a flag in `--help` is **not** evidence of absence — most of these are config-file driven. Verify by installing a test hook/handler and firing it, and read type definitions from source rather than rendered docs (two independent passes in #1355 found rendered docs wrong or truncated where the source was right).

## If a mechanism exists, what shipping it owes

The machinery is built and shared; this should be closer to a declaration than an implementation:

- `HookSpec` on the agent declaration, one config writer, one event→signal map.
- **Canonicalize on SEMANTICS, not Claude Code's wire names.** `hook_signal.go` maps `PreToolUse → SignalPermissionPrompt` only because claudecode installs a narrow `AskUserQuestion|ExitPlanMode` matcher. An adapter whose "before tool" event fires for EVERY tool must not reuse that row, or it holds `waiting` on every tool call.
- Install/uninstall behind a `KindModify` permission, with `Writes` declaring the managed user file, and a `Version` floor if it writes the agent's own config format.
[...]

Issue #1718, opened 2026-08-20
feat(mistral-vibe): hook-based state detection — the first non-JSON config format, needing a comment-preserving TOML writer


ingo-eichhorst:
Broken out of epic #1355, where this was Phase G.

## What makes it different from the three that shipped

**A second config format.** mistral-vibe is TOML, and #1355 is explicit that this cannot be a
read-modify-marshal: that would delete the user's TOML comments. It needs a comment-preserving
writer (`hooktoml`), a sibling of `hookjson`, and `core/go.mod` has **no TOML dependency today**.

That makes this the first adapter whose cost is real implementation rather than declaration, and
it is the one that tests whether the installer seam (`EnsureInstalled`/`Uninstall`) actually
generalizes across formats.

`hookjson`'s own splicer history is the warning to read first: it shipped with seven green
round-trip tests and a defect writing `,,` into ~11% of randomly shaped documents, because all
seven removed the TAIL of a container. Per AGENTS.md, **code that emits bytes from a structural
diff gets a property test** — random inputs, random mutations, round-trip assertions, fixed seed,
and the generator must vary the structural axes that matter (not only the one you thought of).

## Measured

vibe 2.19.1 installed. #1355 recorded that vibe had breaking hook-schema changes at 2.15, so its
schema is young — verify against the installed binary, not against docs.

## Unverified

That mistral-vibe has hooks at all. #1355 recorded only "no hook flag in `--help`" and marked it
**not disproof**.

## Shape of this ticket

**Audit first, implement second, in this same issue.** A negative result is a legitimate close: if the audit shows no usable mechanism, close with that evidence recorded. Do not open a separate audit ticket.

Absence of a flag in `--help` is **not** evidence of absence — most of these are config-file driven. Verify by installing a test hook/handler and firing it, and read type definitions from source rather than rendered docs (two independent passes in #1355 found rendered docs wrong or truncated where the source was right).

## If a mechanism exists, what shipping it owes

The machinery is built and shared; this should be closer to a declaration than an implementation:

- `HookSpec` on the agent declaration, one config writer, one event→signal map.
- **Canonicalize on SEMANTICS, not Claude Code's wire names.** `hook_signal.go` maps `PreToolUse → SignalPermissionPrompt` only because claudecode installs a narrow `AskUserQuestion|ExitPlanMode` matcher. An adapter whose "before tool" event fires for EVERY tool must not reuse that row, or it holds `waiting` on every tool call.
[...]

PR #1724, opened 2026-08-20
feat(gemini-cli): hook-based state detection (Notification/AfterTool/AfterAgent, beacon delivery)

### LINKED ISSUE #1717: feat(gemini-cli): hook-based state detection — decide the maintenance flag first, then ship the strongest pilot in the set
Broken out of epic #1355, where this was **Phase D — the designated pilot**.

## Why it was the pilot, and why it stalled

On the evidence it is the strongest candidate in the set:

- **Google ships `gemini hooks migrate`** — "Migrate hooks from Claude Code to Gemini CLI". The envelope convergence is deliberate upstream, not coincidental, and it is the single strongest piece of evidence for #1355's whole premise.
- **Fully onboarded already** — 33 recordings, so the matrix cost is near zero.
- It uses the nested-JSON shape `hookjson.Config` already handles.
- It is the one adapter with a **known inference defect the Stop hook fixes**: no `turn_done` marker, so state detection falls back to a heuristic.
- Plus 4 `blocked-daemon` cells that may unlock.

**It stalled for a non-technical reason.** `.claude/skills/ir:agent-releases/` marks gemini-cli
"⚠️ unmaintained, skip by default" as of 2026-07-15. Since #1355's success criteria are all gated
on this adapter, the epic's own decision procedure could not run.

## The question this issue must answer first

**Is the unmaintained flag about maintainer time, or about the adapter's viability?** If it is
time, this is the cheapest high-value hook adapter available and the flag is blocking the best
pilot in the set. If it is viability, close this issue and the flag stands.

That is a maintainer decision, not an agent decision — **ask before implementing.**

## Measured

gemini 0.54.4 installed. Adapter facts on record: cwd appears ONLY in the transcript body; two-process node model (exclude the heap-bump worker); `DiscoverPID` has no argv filter; SessionID is the filename stem; nested subagents live at `chats/<parentUUID>/<childUUID>.jsonl`. Its settings writer is **sync-by-omission** — #1355's Phase C audit verified live that it DELETES entries it does not know about, which is why `Verify` exists on hook permissions.

## Shape of this ticket

## Recording cost

Part of #1355.

ingo-eichhorst:
## Summary

Ships the hook-based state-detection pilot for gemini-cli that #1355 Phase D named and #1717 unblocked. The maintainer reversed the "unmaintained" flag on this issue (it was about maintainer time, not viability) — this PR removes that marking everywhere it appeared and ships the pilot itself.

Three events installed: `Notification` (narrow assert, gated on `notification_type == "ToolPermission"`), `AfterTool` (broad release), `AfterAgent` (the turn-done push, through `HandleStopHook`). Delivery is the beacon (`irrlichd hook-post gemini-cli`), not a native http hook — gemini-cli has none — and not a bare `curl` line, which this CLI's own exit-code semantics make unsafe. All source claims below are cited to a bundle offset or a specific function name in the installed CLI; nothing is asserted from rendered docs.

Closes #1717.

## Source-verified vs live-verified

Two installed CLI versions were read: **0.54.4** (installed at audit start) and **0.56.0** (the CLI auto-updated itself mid-audit, unprompted — disclosed rather than silently absorbed). Every claim below was re-checked at both and is **byte-identical** between them unless noted.

**Source-verified** (read directly from the installed `@google/gemini-cli` bundle, not from docs):
- `HookEventName` enum: `BeforeTool, AfterTool, BeforeAgent, Notification, AfterAgent, SessionStart, SessionEnd, PreCompress, BeforeModel, AfterModel, BeforeToolSelection` — 11 events, all 11.
- `validateHookConfig` accepts only `"command" | "plugin" | "runtime"` — there is no JSON-expressible `"http"`. This is the entire reason the beacon is used instead of a native http hook the way claudecode/codex/copilot install.
- The exit-code-to-decision mapping (`convertPlainTextToHookOutput`): exit 0 → allow, exit 1 → allow-with-warning, **anything else → deny**, when the hook's stdout didn't parse as JSON. `evaluateBeforeToolHook` and `fireBeforeAgentHookSafe` turn a "deny" into a blocked tool call / blocked turn. A bare `curl` against a down daemon exits 7 → would break the user's tool calls. This is why `hookbeacon` (always exits 0, by construction) is mandatory here, not merely "a strong outcome if it fits" as my own brief originally framed it.
- `NotificationType` enum: exactly one member, `"ToolPermission"`, today.
- `resolveConfirmation` → `notifyHooks`: fires `Notification`/`ToolPermission` unconditionally once a tool genuinely needs confirmation (`shouldConfirmExecute` returned non-nil), **before** blocking on the user's answer, regardless of the eventual outcome. This is a real, content-based discriminator — unlike `BeforeTool`, whose payload (`tool_name`, `tool_input`) carries nothing distinguishing "about to run silently" from "about to block."
- On `ToolConfirmationOutcome.Cancel`: the scheduler's confirm-phase returns **before** calling `_execute()` — so `AfterTool` never fires for a cancelled confirmation. But `cancelAllQueued` converts the cancelled call into a terminal batch entry carrying a synthetic `functionResponse` (`{error: "[Operation Cancelled]..."}`) fed back to the model, so the turn's conversation continues and completes normally — `AfterAgent` **does** still fire for that turn. This is why `AfterAgent`'s handler also calls `SessionDetector.ReleasePermissionPromptHold` unconditionally (see below) rather than relying solely on `AfterTool`'s broad release.
- `fireAfterAgentHookSafe`: deduped per `prompt_id`, gated on `pendingToolCalls.length === 0` — fires once, at true turn end, only after every tool call from that turn resolved (success, error, or cancel all count as resolved).
- `gemini hooks migrate --from-claude`'s `EVENT_MAPPING` table (Claude Code → Gemini): `PreToolUse→BeforeTool, PostToolUse→AfterTool, UserPromptSubmit→BeforeAgent, Stop→AfterAgent, SubAgentStop→AfterAgent, SessionStart→SessionStart, SessionEnd→SessionEnd, PreCompact→PreCompress, Notification→Notification` — upstream's own statement of the correspondence, corroborating the `AfterAgent`↔turn-done mapping independently of my own tracing.

**Live-verified** (driven against the real installed CLI under an isolated `$HOME`, never the user's real `~/.gemini`):
- `SessionStart` and `SessionEnd` fire from a hooks entry in the **global** `~/.gemini/settings.json` with **no project-scope config present at all** (confirmed by `ls` on the test workdir — empty, no `.gemini`/`.claude` subdirectory) — the payload matched the source-derived shape exactly (`session_id, transcript_path, cwd, hook_event_name, timestamp, source`).
- Global-scope hooks are **not** gated by `isTrustedFolder()` (only project-scope are) — confirmed both by source (`processHooksFromConfig`) and by the above firing without ever answering a trust prompt.
- `gemini hooks migrate --from-claude` (a different, adjacent write path — writes to **project**-scope `.gemini/settings.json`, not global, correcting my own initial assumption) preserved a pre-existing unrelated hook entry and an entirely unrecognized event key verbatim rather than deleting either.
- **`BeforeAgent`/`AfterAgent`/`BeforeTool`/`AfterTool`/`Notification` were NOT live-fired** — the account authenticated on this machine is no longer eligible for live model calls at all: `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals. To continue using Gemini, please migrate to the Antigravity suite of products.` Confirmed via both headless (`gemini -p ... --yolo`) and interactive OAuth sign-in; no `GEMINI_API_KEY` available as an alternate path, and none was sought. This blocks nothing about implementing the mechanism — every behavior above is independently confirmed from source — but it does mean the receiver's dispatch of these four events, while covered by adapter-level unit tests against a mock target, has not been proven against bytes a real gemini-cli process actually sent.

**Not verified, stated rather than assumed:** whether gemini-cli's transcript persists a recognizable "tool denied" marker the way claudecode's does (`LastWasToolDenial`) — not chased, because `ReleasePermissionPromptHold` on `AfterAgent` closes the same gap without needing it.

## Why these three events, and not the other eight

- **`Notification`** (gated on `notification_type == "ToolPermission"`) → asserts `SignalPermissionPrompt`, through a **new, dedicated** `SessionDetector.HandlePermissionPromptHook` — not the generic name-keyed `HandlePermissionHook`/`session.HookSignal` path. This is load-bearing, not style: gemini's wire name for this event is literally `"Notification"`, the same string Claude Code and Copilot both use for their own, differently-gated Notification dispatch. `session.hookSignalEffects` is one flat map with no adapter dimension — a row keyed on `"Notification"` to make this assert would have **also** changed what Copilot's identically-named dispatch does. Copilot's own `hooks.go` explains why that would be a real regression: Copilot emits nothing at all when a user denies a prompt, so a hold with no release would pin a denied session `waiting` for the 12-hour ceiling. Gemini is not in Copilot's position — its `AfterAgent` reliably fires after a cancelled confirmation too (source-verified above) — which is what makes holding safe here where it is not safe there.
- **`AfterTool`** → broad release only, dispatched through the existing generic `HandlePermissionHook` path with a **new** domain row, `session.HookAfterTool = "AfterTool"` (a wire name only gemini-cli fires, so this row cannot leak into another adapter, unlike `"Notification"` above). Fires for every tool call with no discriminator — installed anyway, because releasing an unheld signal is a no-op, mirroring claudecode's own `PostToolUse` row, which is also installed broadly.
- **`AfterAgent`** → the turn-done push, through `HandleStopHook` (payload-carrying, not a name-keyed row — same reason `HookStop` already isn't one), **plus** a new unconditional `ReleasePermissionPromptHold(sessionID)` call — the backstop for the one case `AfterTool`'s release never reaches (a cancelled confirmation). Safe unconditionally: `fireAfterAgentHookSafe`'s own gate means `AfterAgent` cannot fire while a confirmation from that turn is genuinely still open.
- **NOT `BeforeTool`**: fires for every tool call with the same no-discriminator payload as `AfterTool`, and a hook that errors on it denies the tool call (source-verified above). `Notification` already gives the narrow signal; installing this would be pure blast radius for the reason the issue itself named.
- **NOT `SessionStart`/`SessionEnd`**: no stated payoff beyond what process discovery and transcript tailing already give. That they were the two events this audit could confirm live is a fact about auth, not a justification — no other reason was found, so they're left out.
- **NOT `PreCompress`/`BeforeModel`/`AfterModel`/`BeforeToolSelection`**: no signal semantics needed; `BeforeModel`/`AfterModel` additionally carry full LLM request/response bodies — blast radius with no offsetting benefit.

## Version floor

`minCLIVersion = "0.54.0"` in `hookinstaller.go`. **Deliberately not the date the hook system shipped upstream** — PR archaeology (`google-gemini/gemini-cli` #9074–#9112, all merged 2025-09-22) puts the event set at roughly that date, first stable release after it `0.6.0` (2025-09-24). Bisecting an old npm release for the exact payload shape turned out to be unreliable (2025-era packages ship a completely different unbundled `dist/src` layout than today's single bundled file, so a file-presence diff proves nothing across that packaging change — tried, abandoned). The floor instead sits at the edge of what was **directly source-read**: 0.54.4 and 0.56.0, seven months apart, byte-identical on every claim above. Overshooting a version floor is the safe direction (`core/pkg/cliversion` fails **open** on an unknown/unparseable version — the daemon runs under launchd with a minimal PATH and often cannot see the CLI at all); undershooting is not. `Probe: ["gemini", "--version"]` — confirmed live to print a bare number, no `v` prefix, no surrounding text. No `Observed` source: no version field was found in gemini-cli's transcript header (not chased further — not worth inventing a claim to close).

## Recording cost: unassessable

**0 cells re-recorded.** The 33 existing gemini-cli recordings predate hooks and carry no `hook_received` events (confirmed: `tools/onboarding-factory/internal/hookcov`'s `TestDeclaredMatchesRegistry` now reports gemini-cli as `StatusGap`, the same status copilot reported at #1378 for the identical reason). What the #1355 Phase D kill-criterion (>5 re-records) was meant to catch is **unmeasurable** for this PR specifically because of the auth blocker above — recording a hook-bearing session needs a live model turn, and the account cannot make one. This is stated as unassessable rather than left as an absent number: the two must not read the same.

## Contract wiring (all 8, graded against copilot)

`hookdisclosure_test.go`, `hookpath_test.go`, `hookport_test.go` (`AssertHookEndpointFollowsBindAddr` under `DeliveryAddressFree` — the first production adopter of that route; `hook_endpoint_addressfree_test.go`'s reference wiring is what it's copied from), `hookreceipt_test.go`, `hookunknown_test.go`, `hookversion_test.go`, `hookinstaller_test.go` (`AssertPermissionGated` + shared-file round trip — this adapter's settings file is Gemini CLI's one general-purpose settings file, not a dedicated file the way Copilot's is, so uninstall never deletes it and a dedicated test proves unrelated keys — `security.auth`, an arbitrary unknown key — survive install+uninstall byte-for-byte), `hooks_test.go` (`AssertHookReceiverPermissionGated` + `AssertDeclaredPermissions` + behavior tests).

## Mutation evidence (per #1479 — run, not merely described)

Three new pieces of behavior this PR adds each got a deliberate break, confirmed red, then reverted (files fully clean after each — `git status --porcelain` empty):

1. **Notification routed through the generic table instead of the dedicated method** (`target.HandlePermissionHook(...)` in place of `target.HandlePermissionPromptHook(...)`):
   ```
   hooks_test.go:307: got 0 permission-prompt dispatches, want 1 (perm=1, stop=0)
   hooks_test.go:342: got 0 dispatches through HandlePermissionPromptHook, want 1
   ```
2. **`ReleasePermissionPromptHold` call removed from `AfterAgent` handling**:
   ```
   hooks_test.go:284: got 0 ReleasePermissionPromptHold calls, want 1 — AfterAgent must
   release unconditionally, because a cancelled confirmation never fires AfterTool
   ```
3. **`HookAfterTool` row removed from `core/domain/session/hook_signal.go`'s `hookSignalEffects`**:
   ```
   hooks_test.go:240: session.HookSignal("AfterTool") has no row — AfterTool's broad
   release depends on one
   ```

Also fixed in the process: `tools/onboarding-factory/internal/hookcov`'s `TestDeclaredMatchesRegistry` — a pre-existing registry tripwire whose hand-written `want` map needed `gemini-cli: true` now that this adapter declares hooks (same workflow copilot's landing exercised at #1378).

## Deprecation removal

Both marker locations removed: `.claude/skills/ir:agent-releases/SKILL.md` (the `#### Gemini CLI — ⚠️ unmaintained` heading + skip-in-sweeps bullet) and `references/monitoring-surface.md` §10 (heading, explanatory paragraph, reopen condition). Also fixed the `references/monitoring-surface.md` "Turn-end shape change" impact-category row, which listed gemini-cli among four purely-heuristic adapters that "stick in `working` forever" on an upstream transcript-shape change — no longer true once the hooks permission is granted (`AfterAgent` is an independent, transcript-shape-agnostic backstop). Repo-wide grep confirmed these were the only two locations; nothing in code or automation keyed off the marker string — purely prose. `tools/preflight.sh --only skills`: PASS (23 files, 0 errors, 0 warnings).

## Gates run (chunked, each its own foreground invocation)

- `--only go`: PASS (fixed the hookcov regression above; `go test ./core/... -race -count=1` and `go test ./tools/onboarding-factory/... -count=1` both exit 0, confirmed via direct re-runs after the preflight wrapper's captured output turned out to be truncated).
- `--only arch`: PASS (ARS composite 7.9316 → 7.9272, no category regression).
- `--only tools`: PASS (19 shell-lib suites).
- `--only posix`: PASS (3 files).
- `--only security`: PASS (gosec/govulncheck/npm audit; the two G104 "errors unhandled" findings in `geminicli/hookinstaller.go` are the same non-blocking informational pattern every other adapter's `hookinstaller.go` has at its own atomic-write-then-remove defer).
- `--only skills`: PASS.
- The pre-push hook additionally ran `tools/preflight.sh --changed` on push and reported all relevant gates PASS.

## Not merging

Per standing instructions — leaving this for independent QA and merge.

### DIFF
--- a/.claude/skills/ir:agent-releases/SKILL.md
+++ b/.claude/skills/ir:agent-releases/SKILL.md
@@ -77,12 +77,13 @@ you verified against, since it may lag the newest release.
 - Note: sessions have been SQLite (drizzle) since ~2026-02-14, **not** JSON files. Legacy JSON storage still exists in `packages/opencode/src/storage/storage.ts` but `session.ts` no longer routes through it — do not mistake the leftover for live storage.
 - The DB filename is channel-scoped upstream (`opencode-<channel>.db`); stable channels (`latest`/`beta`/`prod`) get plain `opencode.db` — see `packages/core/src/database/database.ts`
 
-#### Gemini CLI — ⚠️ unmaintained, skip by default
-- **Unmaintained — skip in sweeps** unless the user explicitly asks. The standing watches below are not being acted on. `references/monitoring-surface.md` §10 is the authoritative statement (rationale + reopen condition); don't restate it here.
+#### Gemini CLI
+- Maintained — track it in sweeps like any other adapter. It carried an "unmaintained, skip by default" marking from 2026-07-15 until #1717 reversed it: the flag was about maintainer time, not viability, and #1717 shipped a hook-based state-detection pilot against it (`core/adapters/inbound/agents/geminicli/hooks.go`), which needs the same release watch every other hook-installing adapter gets.
 - Check: `https://github.com/google-gemini/gemini-cli/releases` — **authoritative; the repo has no root `CHANGELOG.md`** (404s on `main`)
 - Releases very frequently (nightlies + previews) — focus on stable, group the rest
 - Best evidence is a local clone + `git diff <old-tag>..<new-tag>` on `packages/core/src/config/storage.ts` and `packages/core/src/services/chatRecordingService.ts`
 - Focus on: session transcript format under `~/.gemini/tmp/`, process naming (`bin/gemini`), heap-bump worker re-exec behavior
+- Since #1717: also watch the hook system this adapter now installs into (`packages/core/src/hooks/hookEventHandler.js` in the monorepo) — the `HookEventName` enum, `HookType` (still `{Command, Runtime, Plugin}` with no JSON-expressible `Http` as of 0.56.0 — an added HTTP delivery type would make the beacon workaround unnecessary), the exit-code-to-decision mapping in `hookRunner.js` (0/1/other → allow/allow-with-warning/deny — the reason a bare `curl` install would be unsafe), and the `NotificationType` enum (currently the single member `ToolPermission`, which this adapter's whole "waiting" signal depends on staying narrow)
 - Standing watch: SEA/native binary becoming default (`argv[0] === argv[1]`, env `GEMINI_CLI_NO_RELAUNCH=true`), and the `adk.agentSessionSubagentEnabled` flag — see #1068
 
 #### Kiro CLI
--- a/.claude/skills/ir:agent-releases/references/monitoring-surface.md
+++ b/.claude/skills/ir:agent-releases/references/monitoring-surface.md
@@ -516,8 +516,9 @@ grant the hooks permission).
   parse as JSON, which the scheduler then turns into a denied tool call or a blocked
   turn — a delivery type change here is a correctness question, not a compatibility one.
 - **Version floor**: `minCLIVersion = "0.54.0"` in `hookinstaller.go`, deliberately set at
-  the edge of what was directly source-verified (0.54.4 and 0.56.0, seven months apart,
-  byte-identical) rather than at the PR-archaeology date the hook system actually shipped
+  the edge of what was directly source-verified (0.54.4 and 0.56.0 — published
+  2026-08-07 and 2026-08-19, so a twelve-day, two-minor-release window, not a long
+  stability record — byte-identical) rather than at the PR-archaeology date the hook system actually shipped
   (~2025-09-22, google-gemini/gemini-cli #9074-#9112) — see that constant's doc comment
   before assuming an earlier floor is safe to backdate.
 - **Reopen condition for the OLD marking**: none — it will not return. A genuine future
--- a/.claude/skills/ir:agent-releases/references/monitoring-surface.md
+++ b/.claude/skills/ir:agent-releases/references/monitoring-surface.md
@@ -489,15 +489,45 @@ Lines arrive `TrimSpace`'d, so leading indentation is already gone.
 - Drop/reformat `> Tokens: … sent, … received` → kills tokens, cost, model attribution, **and** `assistant_message` in one stroke. Localization or thousands-separators (`1,234`) also defeat `[\d.]+`.
 - ⚠️ **The subtlest one:** aider printing **periodic keepalive/spinner/progress lines** into the `.md` would reset the idle anchor every tick and **suppress `IdleFlush` indefinitely → permanent `working`**. This requires no change to any marker string, just extra output.
 
-### 10. Gemini CLI (`gemini-cli`) — ⚠️ unmaintained
+### 10. Gemini CLI (`gemini-cli`)
 
-**As of 2026-07-15 this adapter is no longer actively maintained and is skipped in release sweeps** (this section is the authoritative statement of that; `SKILL.md` only points here). It still ships and still monitors live sessions, so it is listed for completeness — but upstream Gemini CLI changes are **not** tracked, and a break-risk analysis is deliberately out of scope.
+**Maintained — track it in sweeps like any other adapter.** It carried an "unmaintained,
+skip by default" marking from 2026-07-15 until #1717 reversed it: the flag was about
+maintainer time, not viability. #1717's audit found the flag was blocking the strongest
+hook-based state-detection candidate in the whole set — Gemini CLI ships `gemini hooks
+migrate` ("Migrate hooks from Claude Code to Gemini CLI"), deliberate upstream evidence
+of the Claude-Code-shaped envelope convergence the hook epic (#1355) depends on — and
+shipped it: `core/adapters/inbound/agents/geminicli/hooks.go` installs `Notification`
+(gated on `notification_type == "ToolPermission"`), `AfterTool` (broad release) and
+`AfterAgent` (the turn-done push, retiring the heuristic fallback below for sessions that
+grant the hooks permission).
 
 - **Transcript path**: `~/.gemini/tmp/<project>/chats/session-<timestamp>-<8hex>.jsonl`
-- **Adapter source**: `core/adapters/inbound/agents/geminicli/` — consult it directly if this adapter ever needs attention again.
-- **Reopen condition**: someone resumes maintenance, or a user reports gemini-cli sessions misbehaving. Until then, an upstream Gemini change is not a finding.
+- **Adapter source**: `core/adapters/inbound/agents/geminicli/` — `hooks.go` and
+  `hookinstaller.go` are the hook half; both carry the source citations (bundle line
+  offsets, exact byte-identical claims at two installed versions) rather than restating
+  them here.
+- **Hook delivery is the beacon, not a native http hook**: source-read from the installed
+  CLI's own bundle, `HookType` is `{Command, Runtime, Plugin}` — there is no
+  JSON-expressible `Http` type — so `irrlichd hook-post gemini-cli`
+  (`core/pkg/hookbeacon`) is what gets installed, never a bare `curl` line. That matters
+  for a release check specifically: Gemini CLI's own exit-code-to-decision mapping turns
+  ANY exit code other than 0 or 1 into `decision: "deny"` for a hook whose stdout didn't
+  parse as JSON, which the scheduler then turns into a denied tool call or a blocked
+  turn — a delivery type change here is a correctness question, not a compatibility one.
+- **Version floor**: `minCLIVersion = "0.54.0"` in `hookinstaller.go`, deliberately set at
+  the edge of what was directly source-verified (0.54.4 and 0.56.0, seven months apart,
+  byte-identical) rather than at the PR-archaeology date the hook system actually shipped
+  (~2025-09-22, google-gemini/gemini-cli #9074-#9112) — see that constant's doc comment
+  before assuming an earlier floor is safe to backdate.
+- **Reopen condition for the OLD marking**: none — it will not return. A genuine future
+  maintenance lapse is a new, separately-justified decision, not a reinstatement of this
+  one.
 
-> ⚠️ **#1068 is not the reason, and is not a deprecation notice** — don't cite it as either. It's a *watch item* about the native SEA binary possibly becoming default (which would break `DiscoverPID` and the heap-bump exclusion), closed as completed with the verdict "No impact today; no code change required". The maintenance decision is separate and out-of-band.
+> ⚠️ **#1068 is not a deprecation notice** — don't cite it as one. It's a *watch item*
+> about the native SEA binary possibly becoming default (which would break `DiscoverPID`
+> and the heap-bump exclusion), closed as completed with the verdict "No impact today; no
+> code change required".
 
 ---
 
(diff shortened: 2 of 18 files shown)


ingo-eichhorst:
## Independent QA — one correction pushed, otherwise clean

Reviewed rather than trusted the handback. What I ran:

### One real defect, fixed in `a1f45060`

The version-floor rationale claimed 0.54.4 and 0.56.0 were source-read *"at two points **seven months
apart** with zero drift"*. They are not:

```
$ npm view @google/gemini-cli time --json
0.54.4  2026-08-07T04:43:44.125Z
0.56.0  2026-08-19T19:29:01.177Z
```

**Twelve days and two minor releases.** The figure was carried in three places —
`hookinstaller.go:84`, `monitoring-surface.md:519`, and this PR's body — and was produced by nothing,
which is precisely the drift AGENTS.md's "Replay's measured figures" rule is about. It also ran in the
direction that **overstates** the evidence: a twelve-day window is far weaker support for
"byte-identical" than seven months would be.

Both in-repo occurrences are corrected with the measured dates and the command that yields them, and
reworded to claim only what the window supports — that the shape survived a real upgrade that happened
on this machine mid-audit, not that it is long-settled. **The floor itself (`0.54.0`) is unchanged and
still right**: it is the edge of the verified window, and overshooting is the safe direction because
`cliversion` fails open.

The claim in the PR body above is superseded by this comment.

### Everything else checked out

**All eight wiring points present**, matching copilot's complete set one-for-one:
`AssertHookDisclosureMatchesInstalled`, `AssertHookPathConfined`, `AssertHookEndpointFollowsBindAddr`,
`AssertHookReceiptObserved`, `AssertUnknownHookEventObserved`, `AssertHookVersionGate`,
`AssertPermissionGated`, and both `AssertHookReceiverPermissionGated` + `AssertDeclaredPermissions`.

**The three-place segment obligation is satisfied** — `beaconroute_test.go` gains its `gemini-cli` row
alongside the installer and the route, which is the check that stops a permanent silent 404 (the beacon
exits 0 by design, so nothing else would report it).

**`replaydata/` untouched** — zero additions, zero deletions, consistent with the audit's measured
"0 cells re-recorded".

**The `hookcov` tripwire was correctly updated rather than worked around.** `TestDeclaredMatchesRegistry`
is a hand-written registry list; flipping `gemini-cli` to `true` is the intended workflow, and the
comment honestly records that a `StatusGap` is expected because no recording carries a `hook_received`
event yet.

### Test evidence, including a false alarm I chased down

`go test ./core/... -race -count=1` on the branch **failed twice** on the first run —
`TestReadLauncherEnv_Tmux_AttachedClientIsTheHost` and `TestReadLauncherEnv_Tmux_DetachedResolvesNoHost`
in `processlifecycle`. **Not this PR.** Established rather than assumed:

| run | result |
|---|---|
| full suite, PR branch, run 1 | those two FAIL |
| the two tests isolated, PR branch | PASS |
| the two tests isolated, clean `origin/main` (d948e17b) | PASS |
| full suite, clean `origin/main` | PASS, no failures |
| full suite, PR branch, run 2 | PASS, no failures |

Load-dependent flakiness in timing-sensitive tmux fixtures, in a package this PR does not touch. Noted
here rather than dismissed in silence, because "unrelated flake" is exactly the kind of claim that
carries an evidentiary bar.

After the correction: `go build ./core/...` clean; `go test -race -count=1` clean across
`agents/...`, `domain/session/...` and `application/services/...`; `tools/skill-lint.sh` reports
23 files, 0 errors, 0 warnings. The pre-push gate ran replay fixtures, the recording-rig smoke test and
the security scan — all PASS.

### Two design calls worth keeping visible for future readers

**`Notification` correctly did NOT get a row in `hookSignalEffects`.** That table is keyed by literal
wire name with **no adapter dimension**, and Copilot uses the same string deliberately *without* a hold —
because Copilot emits nothing when a user denies a prompt, so a hold with no release would pin the
session to the 12-hour `permissionPromptHoldTimeout`. A shared row would have silently reintroduced that
fixed bug into a third adapter. The dedicated `HandlePermissionPromptHook` avoids it and the doc comment
explains why.

**The cancelled-confirmation gap is closed.** `AfterTool` never fires for a tool whose confirmation was
denied, so the narrow assert would otherwise hang until the 12-hour ceiling.
`ReleasePermissionPromptHold`, called unconditionally from the `AfterAgent` path, is the backstop, with
the source reasoning for why `AfterAgent` is guaranteed to arrive after the prompt resolves.

**Not merging** — recording-cost remains unassessable for the stated `IneligibleTierError` reason, and
end-to-end hook delivery is source-verified rather than live-verified. Both are stated in the PR body
rather than left as silent absences, which is the right handling, but they are the maintainer's call.

ingo-eichhorst:
## On the two red advisory checks — measured, not waved through

`SonarCloud` fails on **14.6% duplication on new code** (gate ≤3%) and `CodeScene` flags
**Code Duplication** in `hookinstaller.go` plus complexity in two test files. Both are advisory by
policy — neither branch protection nor the "Protect Main" ruleset requires them — but AGENTS.md says a
red result is *"a prompt to look closer"*, so here is the looking.

**The duplication is real, structural, and pre-existing.** Counting non-comment, non-blank lines shared
with the sibling installers:

| comparison | shared lines | of geminicli's 116 |
|---|---|---|
| geminicli ↔ codex | 76 | 66% |
| geminicli ↔ copilot | 78 | 67% |
| geminicli ↔ claudecode | 57 | 49% |

And by size, this is the **leanest** of the four:

| adapter | `hookinstaller.go` non-comment lines |
|---|---|
| **geminicli** | **116** |
| codex | 144 |
| claudecode | 149 |
| copilot | 167 |

So this PR does not introduce a new kind of duplication — it is the fourth instance of one the three
shipped hook adapters already established, and the smallest instance of it. Chasing the gate to green
here would mean refactoring machinery three other adapters depend on, inside a PR that already touches
the domain layer, and that is a worse trade than the number suggests.

**But the number is telling us something true, and it belongs on #1355's record.** That epic's per-adapter
issues all say the same thing:

> The machinery is built and shared; this should be closer to a **declaration** than an implementation.

At the **contract** level that held completely — all eight families were wired with zero changes to
`core/internal/contracttesting`, and `hookjson`/`hookbeacon` were reused unmodified. At the **installer**
level it did not: two-thirds of this file is a re-spelling of what codex and copilot already say. The
shared layer generalizes the *obligations* but not the *install*.

That is worth knowing before the remaining adapter tickets (#1716, #1718–#1723) are worked, because it
is the difference between "each adapter is a declaration" and "each adapter is 116 lines of near-copy
plus a declaration". Recording it rather than acting on it — a fifth copy is the point at which
extracting the common installer stops being speculative, and #1716's audit has already turned up an
independent reason to touch that layer (kiro-cli needs a flat entry shape `hookjson` cannot currently
write).

The CodeScene complexity findings are on `TestAfterAgentHook_DispatchesTurnDoneAndReleasesPermissionPrompt`
and `TestUninstallPreservesUnrelatedSettingsKeys` — both table-ish tests whose length is the coverage,
and `serveHookRequest`'s argument count, which matches the shape the other three receivers use. Left as
they are, deliberately.

Issue #1716, opened 2026-08-20
feat(kiro-cli): hook-based state detection — audit whether kiro-cli has a hook mechanism, then ship it


ingo-eichhorst:
Broken out of epic #1355, which planned this as part of its Phase E but never audited it.

## Why this one first

kiro-cli is the **cheapest real test of whether the shipped hook machinery generalizes**. Three
adapters now receive hooks — claudecode, codex, copilot — and all three use the same nested-JSON
config shape that `hookjson` already handles. kiro-cli is believed to be config-file driven too,
so if it is, this should be close to a declaration-only change and `hookjson` should not need to
move.

It is also the adapter #1355's Phase C explicitly scoped for audit and never audited.

## What is known, and what is not

**Measured:** kiro-cli 0.12.224 is installed on the maintainer's machine. The adapter is onboarded.

**Assumed / unverified:** that kiro-cli has a hook mechanism at all. #1355 recorded only that
`kiro --help` shows no hook flag, and explicitly marked that as **not disproof** because hooks in
this family are config-file driven.

Prior kiro-cli findings that bear on this: headless sessions are invisible, metrics are
sidecar-only, `/rewind` edits in place, `/clear` does not rotate the session, and `/chat new`
DOES mint a new session id (live-verified, #935).

## Shape of this ticket

**Audit first, implement second, in this same issue.** A negative result is a legitimate close: if the audit shows no usable mechanism, close with that evidence recorded. Do not open a separate audit ticket.

Absence of a flag in `--help` is **not** evidence of absence — most of these are config-file driven. Verify by installing a test hook/handler and firing it, and read type definitions from source rather than rendered docs (two independent passes in #1355 found rendered docs wrong or truncated where the source was right).

## If a mechanism exists, what shipping it owes

The machinery is built and shared; this should be closer to a declaration than an implementation:

- `HookSpec` on the agent declaration, one config writer, one event→signal map.
- **Canonicalize on SEMANTICS, not Claude Code's wire names.** `hook_signal.go` maps `PreToolUse → SignalPermissionPrompt` only because claudecode installs a narrow `AskUserQuestion|ExitPlanMode` matcher. An adapter whose "before tool" event fires for EVERY tool must not reuse that row, or it holds `waiting` on every tool call.
- Install/uninstall behind a `KindModify` permission, with `Writes` declaring the managed user file, and a `Version` floor if it writes the agent's own config format.
[...]

Issue #1478, opened 2026-08-11
replay: 4 recordings still reproduce zero transitions — the daemon's read boundary needs a time-aware model


ingo-eichhorst:
Follow-up to #1342 / PR #1476. That PR took sidecar-driven replay from 20 recordings reproducing **zero** state transitions down to 4, and pinned the remaining 4 in `knownZeroTransition` (`tools/onboarding-factory/cmd/replay/issue1342_debounce_test.go`) rather than forcing them. This issue is the remainder.

## The 4

| recording | daemon recorded | replay reproduces |
|---|---|---|
| `codex/regressions/agent-question-pending/recordings/2026-04-26-11-57-03_irrlichd-unknown` | 1 | 0 |
| `pi/regressions/full-lifecycle-toolcall/recordings/2026-04-26-11-31-28_irrlichd-unknown` | 1 | 0 |
| `pi/scenarios/1-2_session-end/recordings/2026-05-25-01-48-39_irrlichd-0.4.7+597f655` | 2 | 0 |
| `pi/scenarios/2-1_basic-turn/recordings/2026-05-25-01-39-59_irrlichd-0.4.7+7b5218c` | 2 | 0 |

Their goldens hold only the synthetic `init` row, so they pin nothing about the state machine.

## Established cause

The sidecar records `file_size` from the fswatcher's stat at fire time — `SessionDetector.handleEvent` writes `d.record(lifecycle.Event{…, FileSize: ev.Size})` **before** calling `onActivity`, whose `RefreshOnActivity` then tails to EOF. So the daemon's read always reached at least that stat, and in practice past it. `classifyAt` clamps to the stat.

#1476's `readBoundaryFor` widens a clamped, provably-manufactured pass by **one** recorded stat. That is enough when the transcript is written over hundreds of milliseconds. It is not enough for these 4, because their transcripts are written essentially instantaneously and the daemon's single read absorbed the **whole file** while later fswatcher fires were still queued carrying stale sizes.

`pi/scenarios/2-1_basic-turn` is the clearest case — 2556 bytes across 4 fs events spanning **3 ms**:

```
file_size:  325  459  676  2556
fs events:  01:39:59.093 … 01:39:59.096   (all four)
daemon:     01:39:59.093  ready→working  "force ready→working on first activity"
            01:39:59.094  working→ready  "agent finished turn → ready"
process_exited: 01:39:59.190
```

[...]

Issue #1449, opened 2026-08-10
A grant-all dev daemon rewrites the user's real agent configs — #1383 only protects the recording rig


ingo-eichhorst:
Found the hard way, twice, during the #1355 fleet run. #1383 (PR #1411) protected the **recording rig**; a dev or test daemon started by hand with `IRRLICHT_PERMISSION_MODE=grant-all` has the same exposure and no protection at all.

## What happened

An agent implementing #1378 ran a dev daemon with `grant-all` on port 7899. `grant-all` exercises **every** `Apply` closure, so the daemon rewrote the user's **real** `~/.claude/settings.json` and `~/.codex/hooks.json` to point at 7899 — a port that died with the daemon.

It then could not repair them: the permission classifier correctly blocked the write, and blocked the backup in the same command, so no backup existed either.

The observed end state on the machine, after I inspected it:

```
~/.claude/settings.json   25 refs across three dead ports (7839, 7841, 7899)
~/.codex/hooks.json        3 refs to 7899
live daemon                7837
```

So the user's hook channel was silently dead — permissions still reading `granted`, entries still present, nothing arriving. Exactly the failure mode #1368's watchdog exists to *detect*, caused by our own tooling. (Two of those ports predate this incident, which is its own signal: this has been happening quietly for a while.)

I repaired it by backing both files up and repointing every dead port to 7837.

## Why the existing protection does not cover it

`#1383` made the **recorder** snapshot and restore every `ManagedUserFile` before spawning a `grant-all` daemon. That works — it was exercised live during #1388's recording run and restored `~/.claude/settings.json` on every pass.

But the protection lives in `tools/onboarding-factory/scripts/lib/managed-file-snapshot.sh`, i.e. in the rig. Nothing protects a developer or agent who simply runs:

```
IRRLICHT_PERMISSION_MODE=grant-all irrlichd --port 7899
```

which is the ordinary way to test anything consent-gated, and which every agent working on hooks in this epic had reason to do.

## Scope — the fix belongs in the daemon, not in more scripts

The rig's snapshot is a wrapper; wrappers are opt-in and this one was not applied because nobody was recording. Two candidate rules, either of which closes it at the source:

[...]

PR #1727, opened 2026-08-20
docs(process): bounded-wait recipe, findings checkpointing, generalise the sourced-figure rule

### LINKED ISSUE #1726: chore(process): make long agent sessions survivable — bounded waits, checkpointed findings, sourced figures
Three process failures cost real time during #1716/#1717 on 2026-08-20. All three are cheap to fix and
none is hypothetical — each is written up with what actually happened.

## 1. A subagent that blocks on an interactive CLI is killed by the 600s watchdog

**Five watchdog stalls across two agents in one session**, plus one API-error death. Every stall had the
same shape: the agent issued a command that waits on an interactive process and then had no tool call
for 600s. Last lines before death:

- *"The interactive session completed the turn. Now let's check the session file…"*
- *"Good, chat is ready. Now sending the prompt to trigger a tool call that will require confirmation."*
- *"Identical in 0.56.0. Let me also verify the notification/confirmation flow…"* (reading a minified
  multi-megabyte bundle)
- *"Now I have concrete answers for A, B, and C…"* — the answers died with the turn
- *"Now `hookreceipt_test.go` and `hookunknown_test.go`:"* — mid-write, 13 files uncommitted

The cost is not the stall, it is that **the work in flight is lost while the environment stays live**.
One agent left two orphaned CLI processes running ~20 minutes against the account; another left 1526
lines uncommitted.

What worked, once imposed:

- **One Bash call per step.** Never `send-keys` and then wait.
- **Every wait is a bounded polling loop that always returns**, e.g. `for i in $(seq 1 10); do sleep 3;
  tmux -S <sock> capture-pane -p -t <sess> | tail -20; echo "--- poll $i ---"; done`.
- **`timeout N` on anything that can hang**; never read a minified bundle whole — `timeout 60 grep -n
  -o -E '.{200}<pattern>.{600}'` or `sed -n '<a>,<b>p'`.
- **Tear down every spawned process before ending a turn.**

**Deliverable:** put this in the skills agents actually read, as a recipe rather than a principle.
`.claude/skills/ir:exec/SKILL.md` is the natural home (its Phase 4 already carries the "do not background
the unscoped preflight run" incident, which is the same lesson from a different direction), with a
pointer from AGENTS.md's "Local CI parity" section where the background-stall guidance already lives.

## 2. Findings that live only in an agent's context die with the turn

Three of the stalls above destroyed audit work that had to be re-derived. After the agent was required
to **write findings to a file before using them**, the fourth stall cost nothing — a 23KB findings file
survived and the agent resumed from it.

**Deliverable:** make "checkpoint findings to a scratchpad file as you go, append rather than rewrite"
an explicit instruction for any multi-phase audit, in the same skill. Pair it with the existing rule
that a long-running agent's work is checkpointed with `git commit -m wip` rather than `git stash` — the
two are the same discipline applied to notes and to code.

## 3. A figure in prose that nothing produces drifts, and is then quoted with confidence

AGENTS.md already says this, under "Replay's measured figures". It is stated as a fact about the replay
tree. **It is not scoped as a general rule, and it was violated twice in one day outside that tree:**

- PR #1724's version-floor rationale claimed two source-read versions were *"seven months apart with
  zero drift"* — in `hookinstaller.go`, in `monitoring-surface.md`, and in the PR body. `npm view
  @google/gemini-cli time` puts them **twelve days** apart (2026-08-07, 2026-08-19). The error ran in
  the direction that **overstates** the evidence: a twelve-day window is far weaker support for
  "byte-identical" than seven months. Caught in review, fixed in `a1f45060`.
- The summary of that very review then said subagents "stalled seven times". The real count is five.
  Typed, not counted.

**Deliverable:** generalise the rule out of the replay section so it binds anywhere — *a figure that
documents behaviour states the command that produces it, or it is marked as an estimate*. The existing
`knownFirstTransitionDrift` / `censusOfTheCommittedCatalog` idiom is the shape to point at.

Also assess — do **not** force — whether a cheap lint is possible: `tools/skill-lint.sh` already walks
skill markdown and has a warn-only tier for heuristics, so an unsourced time-span or count claim could
warn there. **If it cannot be made low-noise, say so with the false-positive rate you measured and ship
the documented rule alone.** A noisy warning that everyone learns to ignore is worse than no warning.

## Notes

- Items 1 and 2 are documentation of agent workflow; item 3 is a repo rule plus an optional lint. They
  are one ticket because they share a cause — work and claims that nothing durable backs — but they can
  land as separate commits.
- Per AGENTS.md, anything this adds that is mechanically checkable owes a deliberate mutation seen red.
  A pure-prose change does not, and should say so rather than implying coverage it does not have.

ingo-eichhorst:
Closes #1726

## Summary

Three cheap process fixes, all evidence gathered in #1726 from #1716/#1717's 2026-08-20 session — this PR is the fix, not a re-investigation.

1. **Bounded-wait recipe for driving interactive CLIs.** Five watchdog stalls plus one API-error death in one session, every one from a subagent blocking 600s on an interactive process. Documented the recipe that actually worked — one Bash call per step, every wait a bounded polling loop, `timeout N` on anything that can hang, never read a minified bundle whole, tear down spawned processes before ending a turn — as two new bullets in `.claude/skills/ir:exec/SKILL.md`'s Notes section, placed immediately beside the existing "don't background the unscoped preflight run" incident so the two read as one rule about work that nothing durable backs. Added a pointer from AGENTS.md's "Local CI parity" section, where the background-stall guidance already lived.
2. **Checkpoint findings to a file before using them.** Same section, same placement: three of the five stalls destroyed audit work that had to be re-derived; once file-checkpointing was imposed, the next stall cost nothing — a 23KB findings file survived and the agent resumed from it.
3. **Generalised "a figure nothing produces drifts" out of AGENTS.md's replay-specific section.** The rule ("Replay's measured figures") was stated as a fact about the replay tree and was violated twice in one day outside it: PR #1724's version-floor rationale claimed two source-read gemini-cli versions were *"seven months apart with zero drift"* where `npm view @google/gemini-cli time` puts them **twelve days** apart (2026-08-07 to 2026-08-19) — overstating the evidence — and the review that caught it then claimed subagents "stalled seven times" when the real count is five. Promoted the rule to a new general paragraph in AGENTS.md's Testing section (beside its sibling "A verification mechanism must fail loudly when it cannot run"), carrying both new violations as evidence it binds generally. The replay-specific bullet keeps every one of its specifics — `knownFirstTransitionDrift`, `censusOfTheCommittedCatalog`, the #1478/#1503 incident — untouched, and now opens by pointing up at the general rule as its worked example instead of restating it.

## The optional lint (item 3) — assessed, not built

Measured rather than guessed, per the ticket's instruction: a same-line heuristic (a spelled-out number word 1–20/dozen adjacent to a time/count unit, suppressed by a same-line `#issue` reference, backtick span, or hedge word) run over the full 23-file tracked skill corpus (`git ls-files -- '.claude/skills/*.md' '*/SKILL.md' 'SKILL.md'` minus `testdata/`) produces **1 raw candidate, 0 flagged** — the one hit (`ir:test-mac/SKILL.md:337`, "it had already happened three times") is correctly suppressed by a same-line `#1449` citation.

That is very low noise on today's snapshot, but not enough to ship on:
- n=1 is too small a sample to trust same-line proximity generally. This repo's prose wraps at ~80 cols, so a claim and its citation routinely land 2–4 physical lines apart within one sentence (`ir:release/SKILL.md` alone carries 22 `#NNN` citations in dense wrapped prose) — the one real hit passing was a short sentence, not evidence the approach generalises.
- No true-positive fixture exists in the real corpus to validate recall against — by definition, no currently-tracked skill file contains a known-unsourced figure (if one did, it would be a bug to fix, not evidence the check works).
- A version robust to wrapped-paragraph citations needs window/paragraph-scoped reference tracking, comparable in complexity to `skill-lint.sh`'s own `ref-direction` heuristic, which the file's header already documents as carrying "a real false-positive rate."

**Recommendation: ship the documented rule alone; do not add a skill-lint check.** This is a legitimate outcome per the ticket ("a noisy warning everyone learns to ignore is worse than no warning") and per the honest low-confidence read of the one real data point available.

## What owes mutation evidence and what doesn't

Per AGENTS.md's Testing section: **everything in this PR is pure prose** (two `.md` files, no code, no new mechanical check). No mutation evidence is owed and none is claimed. `tools/` was not touched.

## Test plan

- [x] `tools/skill-lint.sh --strict` over the full corpus — 23 files, 0 errors, 0 warnings
- [x] `tools/preflight.sh --only skills` (foreground) — PASS
- [x] Pre-push hook's `tools/preflight.sh --changed` ran clean on push (skill-file lint PASS, gofmt PASS, everything else correctly SKIP — diff touches only `AGENTS.md` and one `SKILL.md`)
- [x] Manually re-read both edited sections in context to confirm no other AGENTS.md bullet's reference to "Replay's measured figures" (the "Probe costs in doc comments" and "macOS app" bullets) went stale — both still resolve correctly since the bullet's label was kept

### DIFF
--- a/.claude/skills/ir:exec/SKILL.md
+++ b/.claude/skills/ir:exec/SKILL.md
@@ -1039,3 +1039,31 @@ Phase 6.
   notification wakes a subagent from its own background job; the run stalls
   silently, indistinguishable from having finished. Block in the foreground
   instead, however slow.
+- **Driving an interactive CLI (a `tmux` pane, a chat REPL, anything a turn
+  would otherwise wait on) hits the same 600s watchdog as an unscoped
+  `preflight.sh` run, and loses more when it does** — the environment stays
+  live while the agent's own work in flight is gone. One Bash call per step:
+  issue the command and return; never send input and then wait on the same
+  call. Every wait is a bounded polling loop that always returns, e.g.
+  ```bash
+  for i in $(seq 1 10); do sleep 3; tmux -S <sock> capture-pane -p -t <sess> | tail -20; echo "--- poll $i ---"; done
+  ```
+  `timeout N` wraps anything that can hang, and a minified or multi-megabyte
+  bundle is never read whole — `timeout 60 grep -n -o -E
+  '.{200}<pattern>.{600}'` or `sed -n '<a>,<b>p'`, the same bounded-read shape
+  as chunking `preflight.sh` above. Tear down every spawned process before
+  ending a turn — a stall here doesn't just cost the turn, it leaves the
+  process running. (Real incident, #1726: five watchdog stalls plus one
+  API-error death in one session driving interactive CLIs for #1716/#1717 —
+  one stall left two orphaned CLI processes running ~20 minutes against the
+  account, another left 1526 lines uncommitted.)
+- **Findings that live only in an agent's context die with the turn, exactly
+  the way uncommitted code does.** For any multi-phase audit or
+  investigation, checkpoint findings to a scratchpad file as you go — append,
+  never rewrite, so a stall costs at most the last unwritten increment. It's
+  the same discipline `git commit -m wip` applies to code (never `git stash`,
+  whose stack is shared across worktrees — AGENTS.md's Process Rules), aimed
+  at notes instead of a working tree. (Real incident, #1726: three of the
+  five stalls above destroyed audit work that had to be re-derived from
+  scratch; once file-checkpointing was imposed, a fourth stall cost
+  nothing — a 23KB findings file survived and the agent resumed from it.)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -113,6 +113,28 @@ were added before they could catch anything and carry the weaker evidence that t
 it plants (below), and the harness built on #1390's lesson from the start, carrying a
 deliberate no-match row that must report `STALE` (#1450).
 
+**A figure that documents behaviour states the command that produces it, or it
+is marked as an estimate.** This is the counterpart to the rule above, from
+the other direction: that one is about a check going silently blind, this one
+is about a *number* silently drifting away from what it once measured — typed
+once, then repeated by hand until it no longer describes anything. First
+named for the replay tree's own catalog counts ("Replay's measured figures"
+below, where `knownFirstTransitionDrift` and `censusOfTheCommittedCatalog` are
+the machine-generated shape to copy), it is not scoped to replay and was
+violated twice in one day outside it (#1726). PR #1724's version-floor
+rationale claimed two source-read versions were *"seven months apart with
+zero drift"* — in `hookinstaller.go`, in `monitoring-surface.md`, and in the
+PR body — where `npm view @google/gemini-cli time` puts them **twelve days**
+apart (2026-08-07 to 2026-08-19); the error ran in the direction that
+overstates the evidence, since a
+twelve-day window is far weaker support for "byte-identical" than seven
+months. The review that caught it then wrote its own summary claiming
+subagents "stalled seven times"; the real count, from the transcript, is
+five. Neither figure had a command behind it — both were typed from memory
+under the same pressure the rule exists to catch. The fix is the same shape
+either way: derive the number in code and print the literal, or say in the
+same sentence that the figure is an estimate and how it was arrived at.
+
 **A fixture that waits by SLEEPING has not observed what it waits for, and the
 assertion after the sleep is not evidence that it has.** Poll the condition to a
 generous deadline and fail with the elapsed time: that turns "the machine was busy"
@@ -1570,10 +1592,11 @@ Before marking a ticket done, run the full suite — every layer must pass:
   figure beside the divergence figure, because a `go test` that passes prints
   nothing without `-v` and an unread measurement is the failure mode this
   closes.
-- Replay's measured figures: the counterpart rule to "a verification mechanism
-  must fail loudly when it cannot run" is that **a number which documents
-  behaviour but is not produced by it drifts silently, and is then quoted with
-  full confidence**. The replay tree carried one example of each outcome:
+- Replay's measured figures: the worked example for the general rule above —
+  "a figure that documents behaviour states the command that produces it, or
+  it is marked as an estimate," including the evidence it was violated
+  outside the replay tree (#1726) — rather than repeating either here. The
+  replay tree carried one example of each outcome:
   `knownFirstTransitionDrift` is machine-generated and stayed right across a
   change that reshaped the distribution it describes, while the
   `zero`/`fabricated`/`divergent` counts were typed by hand and went wrong twice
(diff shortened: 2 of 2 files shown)


ingo-eichhorst:
## QA — content verified, and the lint recommendation now has better evidence than it shipped with

### The prose changes check out

All five deleted lines are re-worded openers; **nothing load-bearing was removed.** Verified the replay
section's specifics all survive on the branch — `knownFirstTransitionDrift` (5 occurrences),
`censusOfTheCommittedCatalog`, `TestCensusDiffNamesEveryStaleShape`, `DivergentByCountsAndKinds`,
`UnpairedSidecars`, `PairedButUngraded`, the "198 divergent" incident and the "one low" near-miss. The
reworded bullet opener is grammatical and correctly demotes itself to the worked example of the general
rule, which now sits directly beside its sibling ("a verification mechanism must fail loudly when it
cannot run"), which is where it belongs.

Figures in the new text spot-checked against what actually happened: five watchdog stalls plus one
API-error death, ~20 minutes of orphaned CLI processes, 1526 lines uncommitted, 23KB findings file.
All correct.

### The lint assessment reached the right answer, and one of its premises was wrong

The PR says:

> No true-positive fixture exists in the real corpus to validate recall against — by definition, no
> currently-tracked skill file contains a known-unsourced figure.

True of the corpus *as of this PR*, but a true positive existed in that exact file **this morning** and
is recoverable from git. `monitoring-surface.md:519` at `389489a1`, before `a1f45060` fixed it:

```
the edge of what was directly source-verified (0.54.4 and 0.56.0, seven months apart,
```

So recall **can** be validated, and I did: reconstructing the described heuristic and running it against
that revision **flags the line**. That is a real, non-synthetic true positive — the strongest kind of
fixture, and exactly what AGENTS.md means by committing evidence rather than improvising it.

### But the conclusion holds, for two better reasons

**1. Precision is entirely a function of how narrow the unit list is, and the narrow list is too narrow
to catch the second violation.** The measured 1-candidate/0-flag result comes from restricting to
time-span units. Broadening to *count* units — which is where the second violation lived ("stalled seven
times") — collapses it. Same 23 files, unit list extended to `adapter|file|case|row|entry|attempt`:

```
18 FLAG
 3 SUPPRESSED
```

Every one of the 18 is ordinary prose: *"one adapter, one concern"*, *"two files of substantive"*,
*"the past section uses two row shapes"*, *"scope is one file"*. There is no unit list that catches
count claims and stays quiet.

**2. Coverage does not match where the failure class lives.** Today's two violations occupied four
locations, and a skill-markdown linter reaches exactly one:

| location | reachable by `skill-lint.sh`? |
|---|---|
| `monitoring-surface.md:519` | **yes** |
| `hookinstaller.go:84` (Go doc comment) | no |
| PR #1724's body | no |
| the review summary in chat | no |

The most consequential of those is the one it cannot see: a Go doc comment that a future maintainer
reads when deciding whether it is safe to lower a version floor.

**Recommendation confirmed: ship the documented rule alone, do not add the check.** Same conclusion the
PR reached, now resting on measured noise-on-broadening and a coverage argument rather than on n=1 and
prose-wrapping. Recording the recovered true positive here so that if anyone revisits this, the fixture
exists at `389489a1` rather than having to be invented.

### Mutation evidence

Correctly none owed and none claimed — two `.md` files, no code, no new mechanical check. Stated
explicitly in the PR body, which is the right handling rather than silence.

PR #1731, opened 2026-08-21
feat(agent): let a managed-user-file permission declare a second file it writes

ingo-eichhorst:
## Summary

Shared-type change only — **no adapter uses this yet**. The mistral-vibe adapter
(#1718) follows in a separate PR on top of this once it merges.

`agent.ManagedUserFile.Path` is singular, and every one of the eight existing
declarations (claudecode ×3, codex, copilot, geminicli, kitty) only ever needed
one file. mistral-vibe's hooks install (#1718) is the first whose single `Apply`
closure needs to write **two**: the hook entries themselves
(`$VIBE_HOME/hooks.toml`) and a separate boolean gate
(`enable_experimental_hooks` in `$VIBE_HOME/config.toml`) that must also be
`true` or the CLI never reads the entries.

This adds `Also []func() (string, error)` — additive, not a change to `Path`'s
type — and widens the two consumers that must see every file an `Apply` can
write, or an undeclared secondary write is invisible to both in exactly the
shape #1449's incident was:

- `PermissionService.sharedConfigRefusal` (#1449's grant-all guard) now refuses
  when **either** `Path` or any `Also` entry resolves outside the daemon's
  isolated home.
- `agents.ManagedUserFiles` (the recorder's full backup/restore projection) now
  resolves and carries every `Also` entry with the same rigor as `Path`
  (non-nil, resolves without error, absolute). `writeManagedFilePaths` flattens
  `Path` + `Also` onto the one-line-per-file stream the recording rig's
  `managed_file_paths()` backs up and restores from — it has no notion that two
  lines belong to one permission.

## What does NOT change

- `Writes.Path` stays the file every existing narrowed projection means by "the
  file this permission writes." `agents.HookConfigs` still narrows to it, and
  `--uninstall-hooks` still means the hooks file alone.
- `configsForKey`'s narrow-before-resolve order (AGENTS.md: a rule, not an
  optimization) is preserved by construction — `Also` is resolved inside the
  SAME `want(p.Key)`-filtered loop iteration as `Path`, so there is no second
  ordering to get wrong. Pinned as a lock:
  `TestNarrowedProjectionsSurviveABrokenAlsoDeclarationOnAnotherKey`.
- `hook_reverify.go:319-326` and `paths.go:234` were checked and need no
  change. `hook_reverify.go` only reads `Writes.Path` for a best-effort log
  line — the actual re-verification runs through the adapter's own `Verify`
  closure, which is free to check whatever files it needs internally.
  `paths.go`'s `hookAdapterNames` only tests `Writes != nil` to decide whether
  an adapter earns a hook-liveness diagnostics row; it does not care which or
  how many files `Writes` names.

## Known weakness of landing this alone

No adapter declares `Also` yet, so the only thing exercising the new path is
the test suite added here. The behavior is real (verified below with red-first
mutations against synthetic declarations), but it has no live consumer until
the mistral-vibe PR lands on top. Splitting it out was the coordinator's call
per #1718's own ~200-line guidance (this PR is 334 lines) — flagging the
tradeoff explicitly rather than let a reviewer find it.

## Red-first mutation evidence

Three obligations, one mutation each, applied and reverted (each in its own
foreground command; full failure output below).

**1. `sharedConfigRefusal` must refuse when the SECONDARY file is outside the isolated home, not only Path.**

Mutation: drop the `Also` loop, checking only `Path`.

```
=== RUN   TestSharedConfigGate_RefusesWhenAlsoResolvesOutsideTheIsolatedHome
    permission_shared_config_gate_test.go:210: Apply ran although its Also file resolves outside the isolated home — a grant-all daemon would write enable_experimental_hooks into the user's real config.toml
    permission_shared_config_gate_test.go:214: the refusal did not name the Also file it refused; errors were []
--- FAIL: TestSharedConfigGate_RefusesWhenAlsoResolvesOutsideTheIsolatedHome (0.00s)
=== RUN   TestSharedConfigGate_AllowsWhenPathAndAlsoAreBothInsideTheIsolatedHome
--- PASS: TestSharedConfigGate_AllowsWhenPathAndAlsoAreBothInsideTheIsolatedHome (0.00s)
=== RUN   TestSharedConfigGate_UnresolvableAlsoFailsClosed
    permission_shared_config_gate_test.go:254: Apply ran although its Also[0] resolver could not resolve a path
    permission_shared_config_gate_test.go:257: the refusal did not carry the resolution failure and the failing entry's label; errors were []
--- FAIL: TestSharedConfigGate_UnresolvableAlsoFailsClosed (0.00s)
```

The vacuity guard (`AllowsWhenPathAndAlsoAreBothInsideTheIsolatedHome`) stays
green under the mutation, as it must — it isn't testing the removed code path.
Every pre-existing gate test also stayed green (Path's own arm is untouched by
this mutation). Reverted; full gate suite green again.

**2. The recorder projection `agents.ManagedUserFiles` must resolve and carry every Also entry.**

Mutation: replace the `Also` resolution loop with `var also []string` (never populated).

```
=== RUN   TestManagedUserFilesResolvesAlso
    managedfiles_test.go:231: Also = [], want [/tmp/vibe/.vibe/config.toml] — the recorder cannot back up a file this projection does not name
--- FAIL: TestManagedUserFilesResolvesAlso (0.00s)
=== RUN   TestManagedUserFilesRejectsUnusableAlsoDeclarations
=== RUN   TestManagedUserFilesRejectsUnusableAlsoDeclarations/nil_Also_resolver
    managedfiles_test.go:268: ManagedUserFiles() = [{mistral-vibe hooks /tmp/vibe/.vibe/hooks.toml [] 0x104aeb880}], want an error naming "Also[0]"
=== RUN   TestManagedUserFilesRejectsUnusableAlsoDeclarations/Also_resolve_fails
    managedfiles_test.go:268: ManagedUserFiles() = [...], want an error naming "no vibe home"
=== RUN   TestManagedUserFilesRejectsUnusableAlsoDeclarations/Also_relative_path
    managedfiles_test.go:268: ManagedUserFiles() = [...], want an error naming "not absolute"
--- FAIL: TestManagedUserFilesRejectsUnusableAlsoDeclarations (0.00s)
```

All four new arms go red; `TestHookConfigsIsTheHooksSliceOfManagedUserFiles` and
every other existing test in the package stayed green. Reverted; green again.

**3. `writeManagedFilePaths` must flatten `Also` onto the printed stream.**

Mutation: drop the inner `for _, also := range c.Also { print(also) }` loop.

```
=== RUN   TestWriteManagedFilePathsDeduplicates
--- PASS: TestWriteManagedFilePathsDeduplicates (0.00s)
=== RUN   TestWriteManagedFilePathsFlattensAlso
    managedfiles_test.go:203: printed ["/tmp/vibe/.vibe/hooks.toml" "/tmp/alpha/settings.json"], want ["/tmp/vibe/.vibe/hooks.toml" "/tmp/vibe/.vibe/config.toml" "/tmp/alpha/settings.json"] — an Also file the recorder never learns about is never backed up
--- FAIL: TestWriteManagedFilePathsFlattensAlso (0.00s)
```

`TestWriteManagedFilePathsDeduplicates` (Path-only) stays green, confirming the
mutation is scoped to the new behavior. Reverted; green again.

## Test plan

- [x] `go build ./core/...`
- [x] `go test ./core/... -race -count=1` — all green (foreground, full run)
- [x] `tools/preflight.sh --only go` — PASS
- [x] `tools/preflight.sh --only arch` — PASS (composite 7.9272 → 7.9271, no regression)
- [x] `tools/preflight.sh --only skills` — PASS
- [x] `tools/preflight.sh --only posix` — PASS
- [x] `tools/preflight.sh --only security` — PASS (pre-existing informational-only findings elsewhere in the tree, no High/High, none introduced by this diff)
- [x] Every new obligation carries red-first mutation evidence (above)
- [x] Pre-push hook (`preflight.sh --changed`) passed on push

Part of #1718 / #1355.

### DIFF
--- a/core/adapters/inbound/agents/managedfiles.go
+++ b/core/adapters/inbound/agents/managedfiles.go
@@ -17,8 +17,20 @@ type ManagedUserFile struct {
 	Key string
 	// Path is the resolved, absolute file the permission writes.
 	Path string
-	// Uninstall removes irrlicht's content from Path, reporting whether it
-	// modified anything.
+	// Also holds the resolved, absolute paths of every additional file the
+	// SAME permission's Apply closure writes (agent.ManagedUserFile.Also,
+	// #1718) — empty for the eight declarations that predate that field. It is
+	// resolved with the same rigor as Path (non-nil, resolves without error,
+	// absolute), because the two consumers that read it need completeness for
+	// the identical reason they need it of Path: #1449's grant-all refusal must
+	// see every file an Apply can write, not just the first, and the
+	// recorder's backup/restore sweep (ManagedUserFiles) would otherwise hand
+	// a file that was never snapshotted back to the user unrestored.
+	Also []string
+	// Uninstall removes irrlicht's content from Path (and, for a permission
+	// declaring Also, from every one of those files too — that is the
+	// closure's own job, not something this projection orchestrates),
+	// reporting whether it modified anything.
 	Uninstall func() (bool, error)
 }
 
@@ -93,10 +105,29 @@ func collectManagedFiles(catalog []agent.Agent, want func(key string) bool) ([]M
 			if !filepath.IsAbs(path) {
 				return nil, fmt.Errorf("adapter %q permission %q: managed file path %q is not absolute", a.Identity.Name, p.Key, path)
 			}
+			// Also gets the identical rigor as Path (#1718), over the SAME
+			// selected permission — this is not resolving another key's
+			// declaration, so it does not weaken the narrowed-projection rule
+			// collectManagedFiles otherwise upholds (see configsForKey's doc).
+			also := make([]string, 0, len(p.Writes.Also))
+			for i, resolveAlso := range p.Writes.Also {
+				if resolveAlso == nil {
+					return nil, fmt.Errorf("adapter %q permission %q: ManagedUserFile.Also[%d] is a nil resolver", a.Identity.Name, p.Key, i)
+				}
+				aPath, err := resolveAlso()
+				if err != nil {
+					return nil, fmt.Errorf("adapter %q permission %q: resolve managed file Also[%d]: %w", a.Identity.Name, p.Key, i, err)
+				}
+				if !filepath.IsAbs(aPath) {
+					return nil, fmt.Errorf("adapter %q permission %q: managed file Also[%d] path %q is not absolute", a.Identity.Name, p.Key, i, aPath)
+				}
+				also = append(also, aPath)
+			}
 			out = append(out, ManagedUserFile{
 				Adapter:   a.Identity.Name,
 				Key:       p.Key,
 				Path:      path,
+				Also:      also,
 				Uninstall: p.Writes.Uninstall,
 			})
 		}
--- a/core/adapters/inbound/agents/managedfiles_test.go
+++ b/core/adapters/inbound/agents/managedfiles_test.go
@@ -202,6 +202,125 @@ func TestManagedUserFilesIgnoresPermissionsThatDeclareNoFile(t *testing.T) {
 	}
 }
 
+// TestManagedUserFilesResolvesAlso is the #1718 widening's defect test:
+// ManagedUserFiles must carry every Also file, resolved to an absolute path,
+// alongside Path — not just Path. It is a defect test rather than a contract
+// test because collectManagedFiles is the ONE function every projection
+// shares, so dropping the resolution there is a single missed loop rather
+// than something that could pass by construction.
+func TestManagedUserFilesResolvesAlso(t *testing.T) {
+	path := func(p string) func() (string, error) {
+		return func() (string, error) { return p, nil }
+	}
+	catalog := []agent.Agent{
+		writerAdapter("mistral-vibe", agent.HooksPermissionKey, &agent.ManagedUserFile{
+			Path:      path("/tmp/vibe/.vibe/hooks.toml"),
+			Also:      []func() (string, error){path("/tmp/vibe/.vibe/config.toml")},
+			Uninstall: func() (bool, error) { return false, nil },
+		}),
+	}
+
+	got, err := ManagedUserFiles(catalog)
+	if err != nil {
+		t.Fatalf("ManagedUserFiles(): %v", err)
+	}
+	if len(got) != 1 {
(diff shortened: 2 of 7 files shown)


ingo-eichhorst:
## Independent QA

Reproduced the central mutation rather than taking the report on trust. Deleting the `Also` loop from
`sharedConfigRefusal`:

```
--- FAIL: TestSharedConfigGate_RefusesWhenAlsoResolvesOutsideTheIsolatedHome (0.00s)
    permission_shared_config_gate_test.go:214: the refusal did not name the Also file it refused; errors were []
--- FAIL: TestSharedConfigGate_UnresolvableAlsoFailsClosed (0.00s)
    permission_shared_config_gate_test.go:257: the refusal did not carry the resolution failure and the failing entry's label; errors were []
```

`AllowsWhenPathAndAlsoAreBothInsideTheIsolatedHome` stayed green under the same mutation, so the arms
discriminate rather than reporting unconditionally. Reverted by copy-back; `git status --porcelain` empty.

### Consumer coverage — checked every `Writes` reader, not only the ones the PR names

Six non-test consumers resolve or inspect `Writes`. Three are widened here
(`sharedConfigRefusal`, `agents.ManagedUserFiles`, `writeManagedFilePaths`). The PR argues two need no
change; both hold on reading:

- `hook_reverify.go:323-328` — `Writes.Path` feeds `t.configPath` only, explicitly best-effort
  ("an unresolvable path costs the row its actionability, not the loop its function"). The
  re-verification itself runs through `Writes.Verify`, which owns whatever files it needs.
- `paths.go:234` — `p.Writes != nil` only.

Two more the PR does not mention, checked here so the set is closed rather than sampled:

- `permission_service.go:438` and `:539` — both nil-checks (`p.Writes == nil`), indifferent to how many
  files the declaration names.

So no consumer resolves a path while ignoring `Also`.

### The weakness, stated because it is real

No adapter declares `Also` yet — the tests are its only consumer, and the field's first real use arrives
with the vibe adapter. That is the honest cost of landing the shared type ahead of it, and the PR body
names it rather than leaving a reviewer to find it. Worth re-checking when vibe lands that the field the
adapter needs is the field that shipped.

CodeScene is red; per AGENTS.md it is advisory and not a merge gate. No detail is exposed through the
check API. Flagging rather than chasing.

Issue #1178, opened 2026-07-18
Make the daemon hook port configurable (unblocks factory recording of hook-driven cells)


ingo-eichhorst:
## Problem

The port that installed agent hooks POST to is **hardcoded to `:7837`** in each adapter's installer:

- `core/adapters/inbound/agents/claudecode/hookinstaller.go` — `hookSentinel` / `hookEndpointURL` = `http://localhost:7837/api/v1/hooks/claudecode` (also duplicated in `statusline_installer.go`).
- `core/adapters/inbound/agents/codex/hookinstaller.go` — `hookSentinel` / `hookEndpointURL` = `http://localhost:7837/api/v1/hooks/codex` (added in #1171).

The daemon's own bind address is already overridable (`main.go` `resolveBindAddr` / `IRRLICHT_BIND_ADDR`), but the installers don't derive from it — they agree with the default `:7837` purely by convention.

## Impact

1. **Factory recordings of hook-driven cells are blocked.** The onboarding-factory coexist recorder runs on `:7838` (`tools/onboarding-factory/scripts/run-cell-multi.sh`), but hooks POST to `:7837`, so hook observations reach production, never the recorder. This is exactly why `codex × 2-19_tool-gate-permission-prompt` (#1171) is `daemon_capability: full` but stuck at `pending-record` — its `events.jsonl` recording is deferred. Same limitation applies to every claudecode hook-driven cell.
2. **Alt-port dev daemons silently receive no hook events.** A `separate`-mode daemon on `:7838` installs hooks that still POST to `:7837`, so its hook-authoritative waiting/ready tier does nothing.

## Proposed

Make the hook endpoint port configurable — either derive it from the daemon's resolved bind address at install time, or accept an env/flag (e.g. `IRRLICHT_HOOK_PORT`) that both the installer and the daemon read. Keep the sentinel matching backward-compatible (existing `:7837` installs must still be recognized/upgraded).

This would let the `:7838` coexist recorder capture codex **and** claudecode hook-driven cells without the `of record run --attach`-to-production workaround, and unblock the deferred `2-19` recording.

## References
- #1171 (Codex hook tier — introduced the codex copy of the hardcoded port and the deferred recording)
- Related prior context: the claudecode hook-port hardcoding has been a known fixture limitation.

ingo-eichhorst:
> *This was generated by AI during triage.*

## Triage Assessment

[...]

PR #1732, opened 2026-08-21
feat(kiro-cli): hook-based state detection (postToolUse/stop, beacon delivery, flat entry shape)

### LINKED ISSUE #1716: feat(kiro-cli): hook-based state detection — audit whether kiro-cli has a hook mechanism, then ship it

ingo-eichhorst:
## Summary

Implements hook-based state detection for kiro-cli, per the settled design in #1716's comments (epic #1355).

- **Two events installed**: `postToolUse` (broad release, dispatched under the existing `session.HookPostToolUse` name) and `stop` (authoritative turn-done, `HandleStopHook`). `preToolUse`, `agentSpawn` and `userPromptSubmit` are deliberately NOT installed — see `hooks.go`'s package comment for the evidence.
- **Flat entry shape** (`{"command": ...}`, no matcher-group wrapper) — kiro-cli's hooks schema rejects `hookjson`'s nested shape outright and silently falls back to a different agent while reporting install success. Implemented adapter-locally in `hookinstaller.go`, reusing `hookjson.ReadSettings`/`WriteSettings` UNMODIFIED for the actual byte-level splice rather than writing a second structural-diff writer.
- **Beacon delivery** (`DeliveryAddressFree`, `irrlichd hook-post kiro-cli`) — makes the #1178 stale-port class inexpressible the same way #1724 (gemini-cli) established.
- **Install route**: `EnsureHooksInstalled` materializes `agents/irrlicht.json` via `kiro-cli agent create irrlicht --from kiro_default`, injects the flat hook entries, and points `chat.defaultAgent` at it via `kiro-cli agent set-default irrlicht`. The prior default is recorded in an adapter-owned sidecar and restored on uninstall.
- **Version floor**: `2.6.0`, pinned to exactly the version this issue's audit verified against.
- **Writes** declares the agent config file (`agents/irrlicht.json`), plus **`Also: []func() (string, error){kiroSettingsPath}`** — the SAME closure also rewrites `chat.defaultAgent` in `settings/cli.json`, now visible to both `PermissionService.sharedConfigRefusal` (#1449) and `agents.ManagedUserFiles`/`--print-managed-files`. Mutation-verified two ways: `core/cmd/irrlichd/kirocli_alsofile_test.go` (the projection layer) and `TestSharedConfigGate_KiroCLIAlsoDeclarationReachesTheGuard` (`application/services/permission_shared_config_gate_test.go`, a composition test using kiro-cli's REAL Also resolver against the real `sharedConfigRefusal` guard, decoupled from Path via a synthetic stand-in since kiro-cli's real Path and Also share one root and can never diverge in production). Both were seen red together against the same mutation (commenting out `Also`), then reverted.
- All 8 contract families wired, including the real shared `#1178` contract (`contracttesting.AssertHookEndpointFollowsBindAddr` with `EntriesOf: contracttesting.FlatHookEntries`, via #1734, now merged to `main`).
- Beacon route registered in `startup.go`, pinned in `beaconroute_test.go`, `hookcov`'s `TestDeclaredMatchesRegistry` flipped to `true`.

## hookinstaller.go size, reviewed

Coordinator review asked for the non-comment line count against gemini-cli's 116. Measured with the same method (strip blank lines and full-line `//` comments): **376 non-comment lines** (708 total), against gemini-cli's confirmed 116. Coverage analysis (`go test -coverprofile`) found exactly one 0%-covered function — `matcherForEvent`, assigned to a `flatHookConfig.MatcherFor` field that was never actually called anywhere — a whole unused abstraction mirroring gemini-cli's real matcher concept that kiro-cli's flat entries never use. Deleted rather than justified (field, function, both test references); non-comment lines dropped from 379 to 376. No other function has 0% coverage after the removal. The remaining disparity is accounted for by genuinely additional responsibilities gemini-cli does not have: the flat-shape reimplementation of install/uninstall/verify (~250 non-comment lines, reusing `hookjson`'s byte-splice rather than duplicating it) and the agent-materialization/default-agent-flip machinery (`EnsureHooksInstalled`'s `kiro_default` copy, `ensureDefaultAgent`, `restorePriorDefaultAgent`, the prior-default sidecar) that gemini-cli's adapter has no equivalent of.

## History

Branch was rebased twice onto #1734 as it moved (first onto the seam's WIP tip, then onto `main` after #1734 squash-merged as `f27ea2c6`) and restructured into 5 clean commits (from an original 8 including 3 "wip:"-prefixed crash-recovery commits and one gofmt-only noise commit). Verified byte-identical net diff before and after the restructure (`git diff <old-tip> <new-tip>` — zero lines). `git diff origin/main...HEAD --name-only` contains no `core/internal/contracttesting/` file.

## Test plan

- [x] `go test ./core/... -race -count=1` — full suite green
- [x] `go test ./tools/onboarding-factory/internal/hookcov/...` — green
- [x] `tools/preflight.sh --only go|arch|tools|skills|posix|bash|security` — all PASS
- [x] `TestHookEndpointContract` re-verified against the merged (not local-copy) seam on `main`
- [x] Red-first mutation evidence for all 8 contract families + 3 registry tripwires + the `Also` declaration (both the projection test and the new composition test against the real guard), each applied and reverted individually

## Known gap, still open

The "refresh from `kiro_default` on version bump" path is **unbuilt**: `EnsureHooksInstalled` copies `kiro_default` once, on first install, with no re-sync on a later kiro-cli version bump. Flagged in #1716's own comment thread.

## Recordings

Still plan-only. No live kiro-cli driving, nothing under `replaydata/` touched, `~/.kiro` untouched.

Closes #1716 / part of #1355. Depended on #1734 (now merged).

### DIFF
--- a/core/adapters/inbound/agents/beaconroute_test.go
+++ b/core/adapters/inbound/agents/beaconroute_test.go
@@ -7,6 +7,7 @@ import (
 	"irrlicht/core/adapters/inbound/agents/codex"
 	"irrlicht/core/adapters/inbound/agents/copilot"
 	"irrlicht/core/adapters/inbound/agents/geminicli"
+	"irrlicht/core/adapters/inbound/agents/kirocli"
 	"irrlicht/core/pkg/hookbeacon"
 )
 
@@ -25,18 +26,19 @@ import (
 // while its route segment is "claudecode". That is why the map keys below are
 // written out literally rather than derived from the adapter constants.
 //
-// Scope, stated honestly: this pins the two receivers that exist. It cannot fail
-// for a THIRD receiver registered at a different prefix, because
-// registerHookRoutes (core/cmd/irrlichd/startup.go) is hand-wired with no
-// registry to enumerate and this map is hand-written. Whoever adds the third
-// receiver adds its row here; until then the convention is pinned, not
-// enforced.
+// Scope, stated honestly: this pins the receivers that exist (five as of
+// #1716's kiro-cli row). It cannot fail for a NEW receiver registered at a
+// different prefix, because registerHookRoutes (core/cmd/irrlichd/startup.go)
+// is hand-wired with no registry to enumerate and this map is hand-written.
+// Whoever adds the next receiver adds its row here; until then the
+// convention is pinned, not enforced.
 func TestBeaconEndpointPathMatchesTheInstalledReceivers(t *testing.T) {
 	for segment, want := range map[string]string{
 		"claudecode": claudecode.HookEndpointPath,
 		"codex":      codex.HookEndpointPath,
 		"copilot":    copilot.HookEndpointPath,
 		"gemini-cli": geminicli.HookEndpointPath,
+		"kiro-cli":   kirocli.HookEndpointPath,
 	} {
 		if got := hookbeacon.EndpointPath(segment); got != want {
 			t.Errorf("hookbeacon.EndpointPath(%q) = %q, but the receiver is registered at %q — the beacon would post into a route nothing serves", segment, got, want)
--- a/core/adapters/inbound/agents/kirocli/agent.go
+++ b/core/adapters/inbound/agents/kirocli/agent.go
@@ -1,12 +1,22 @@
 package kirocli
 
 import (
+	"irrlicht/core/adapters/inbound/agents/hookjson"
 	"irrlicht/core/domain/agent"
 	"irrlicht/core/domain/permission"
 )
 
 // PermissionKeyTranscripts gates all Kiro CLI monitoring (issue #570).
 const PermissionKeyTranscripts = "transcripts"
+
+// PermissionKeyHooks gates installing and receiving Kiro CLI hooks (issue
+// #1716, epic #1355 Phase D).
+//
+// Aliased to the shared domain constant rather than restated: since #1383 two
+// projections narrow on it — agents.HookConfigs (what --uninstall-hooks
+// revokes) and the managed-user-file catalog — so a literal that drifted
+// would silently drop this install out of both.
+const PermissionKeyHooks = agent.HooksPermissionKey
 
 // Kiro CLI — the Kiro ghost mark. Color picks contrast against the
 // surrounding chrome: near-black on light themes, near-white on dark.
@@ -22,6 +32,20 @@ const iconSVGDark = `<svg xmlns="http://www.w3.org/2000/svg" width="14" height="
   <circle cx="62" cy="46" r="6" fill="#E0E0E0"/>
 </svg>`
 
+// Source is the adapter's transcript-tree declaration.
+//
+// The hook receiver's path confiner reads it per request (issue #1361), and
+// Agent() below is its only other caller — one declaration, so the tree the
+// daemon watches and the tree the receiver confines to cannot drift apart.
+func Source() agent.Source {
+	return agent.FilesUnderRoot{
+		Dir: sessionsDir(),
+		Parser: agent.JSONLineParser{
+			NewParser: func() agent.LineParser { return &Parser{} },
+		},
+	}
+}
+
 // Agent returns the Kiro CLI adapter registration.
 func Agent() agent.Agent {
 	return agent.Agent{
@@ -35,12 +59,7 @@ func Agent() agent.Agent {
 			Match:         agent.ExactName{Name: ProcessName},
 			PIDForSession: DiscoverPID,
 		},
-		Source: agent.FilesUnderRoot{
-			Dir: sessionsDir(),
-			Parser: agent.JSONLineParser{
-				NewParser: func() agent.LineParser { return &Parser{} },
-			},
-		},
+		Source:  Source(),
 		Control: agent.Control{SupportsInput: true, Interrupt: agent.InterruptCtrlC},
 		Permissions: []agent.Permission{
 			agent.ControlPermission(),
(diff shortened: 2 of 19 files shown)


ingo-eichhorst:
Rebased onto \`main\` (\`0696c106\`, #1733 mistral-vibe's adapter) and resolved the two predictable conflicts additively — both are the same collision the coordinator flagged: two adapters landing hook registration in the same two files during the same window.

**\`core/cmd/irrlichd/startup.go\`**: both \`vibe.HookEndpointPath\` and \`kirocli.HookEndpointPath\` are registered (verified: \`grep -n\` shows both \`mux.Handle\` calls and both package imports present).

**\`tools/onboarding-factory/internal/hookcov/hookcov_test.go\`**: both \`"kiro-cli": true\` and \`"mistral-vibe": true\` in the \`want\` map (verified: neither line was dropped).

Re-verified rather than assumed:
- \`go test ./core/... -race -count=1\` — full suite green, including the new \`hooktoml\`/\`vibe\` packages #1733 brought in.
- \`go test ./tools/onboarding-factory/... -count=1\` — green, \`hookcov\` package specifically confirmed.
- \`tools/preflight.sh --only go|arch|tools|skills|posix|bash|security\` — all PASS.
- \`git diff origin/main...HEAD --name-only\` — still no \`core/internal/contracttesting/\` file.
- \`mergeable: MERGEABLE\` (GitHub API).

5 commits preserved, same as before this rebase. Recordings remain plan-only.

ingo-eichhorst:
## Independent QA

### The `Also` declaration reaches both guards — reproduced

Commenting out `Also:` in `agent.go`:

```
--- FAIL: TestSharedConfigGate_KiroCLIAlsoDeclarationReachesTheGuard   (services — #1449's grant-all refusal)
--- FAIL: TestKiroCLIAlso_SettingsFileIsAManagedFile                   (cmd/irrlichd — the recorder projection)
```

Two independent layers, both red. The `kirocli` package itself stays green under the same mutation, which
is correct — neither layer lives there, and a declaration test that passed locally would have been the
weaker result.

The composition test's decoupling of `Path` from `Also` is the load-bearing detail and deserves to be
read before anyone "simplifies" it: kiro-cli's real `Path` and `Also` both derive from `$KIRO_HOME`, so
they cannot diverge in production, and `sharedConfigRefusal` checks `Path` first and returns immediately.
A test using the real `Path` would refuse on `Path` alone and prove nothing about `Also`.

### Conflict resolution verified, not assumed

Both this PR and #1733 register a beacon route and flip a `hookcov` bool in the same two files. A
resolution taking either side wholesale would silently drop the other adapter's route — a permanent
silent 404 with nothing anywhere to look at. Checked after the rebase:

```
mux.Handle("POST "+claudecode.HookEndpointPath,          mux.Handle("POST "+geminicli.HookEndpointPath,
mux.Handle("POST "+claudecode.StatuslineEndpointPath,    mux.Handle("POST "+vibe.HookEndpointPath,
mux.Handle("POST "+codex.HookEndpointPath,               mux.Handle("POST "+kirocli.HookEndpointPath,
mux.Handle("POST "+copilot.HookEndpointPath,
```

Seven receivers across six adapters, and `hookcov`'s want-map carries both `"kiro-cli": true` and
`"mistral-vibe": true`.

### Dead code removed rather than justified

`matcherForEvent` was 0%-covered — wired into `MatcherFor` but never called, a mirror of gemini-cli's
matcher concept that flat entries never use. Deleted, with a comment left at the removal site explaining
why the field is absent. `hookinstaller.go` is 376 non-comment lines against gemini-cli's 116; the
difference is the flat-shape install/uninstall/verify path and the agent-materialization /
default-agent-flip machinery, which gemini-cli has no equivalent of.

### Still outstanding, deliberately

Recordings are plan-only. `run-cell.sh` has coexist-isolation blocks for `COPILOT_HOME` and `CODEX_HOME`
and none for `KIRO_HOME`, while `spawn_record_daemon` sets `IRRLICHT_ALLOW_SHARED_CONFIG_WRITES=1`
unconditionally — so a hooks-granted kiro-cli recording today would write into the operator's real
`~/.kiro`. Nothing has been recorded. The `Also` declaration narrows this (both files are now visible to
`snapshot_managed_files`) but does not close it, since the isolation block still does not exist.

The "refresh from `kiro_default` on version bump" path remains unbuilt and flagged.

PR #1733, opened 2026-08-21
feat(vibe): hook-based turn-done detection (post_agent_turn only)

ingo-eichhorst:
## Summary

Wires mistral-vibe into the daemon's hooks channel — **one event only**,
`post_agent_turn` → `HandleStopHook`. Builds on #1731 (the `ManagedUserFile.Also`
widening), the `hooktoml` package, and #1734 (the raw-bytes `[][]byte`
widening of the #1178 contract seam) — all landed on `main`.

## The negative finding (audit, posted to #1718 before implementation)

`before_tool`/`after_tool` are deliberately **not** installed. `before_tool` fires
unconditionally for every tool call, before vibe's own permission check even
runs (source-traced: `_execute_tool_call` runs `_run_before_tool_pipeline`
*then* `_should_execute_tool`/`_ask_approval`), and its payload carries no field
indicating whether a given call will reach an interactive prompt — asserting
`SignalPermissionPrompt` on it would hold `waiting` on every tool call.
`after_tool` was considered as a broad release (gemini-cli's `AfterTool` shape)
but that pattern is only safe paired with a narrow assert on the *same* signal
from the *same* adapter, which vibe has none of — repo-wide,
`SignalPermissionPrompt` is held only by hook-derived calls. Full writeup:
https://github.com/ingo-eichhorst/Irrlicht/issues/1718#issuecomment-5367010063

## Two files, using #1731's new field

`Writes.Path` = `$VIBE_HOME/hooks.toml` (the entry), `Writes.Also` =
`$VIBE_HOME/config.toml` (`enable_experimental_hooks`, the flag the entry
cannot fire without — vibe's own hook loader returns nothing at all when it's
false). Uninstall clears the flag only when hooks.toml ends up with **no**
`[[hooks]]` block at all afterward, so a user's own hand-written hook keeps
working after ours is revoked.

Delivery is the beacon (`DeliveryAddressFree`) — vibe's `HookExecutor` spawns
`hook.command` via `asyncio.create_subprocess_shell` and nothing else; there is
no HTTP hook type to ask for. Version floor is **2.19.1**, the only version
directly verified (source-read AND live-fired against the installed CLI). The
epic's "breaking changes at 2.15" claim is stated as **unverified** in both the
disclosure text's reasoning and the code comment — an attempted upstream
changelog fetch produced content contradicted by this adapter's own
live-fired 2.19.1 event names and was discarded as a likely fabrication rather
than cited.

## All seven contract families wired

Permission gate (install-type + receiver-side), disclosure, path
confinement, version floor, unrecognized event, receipt, and — as of this PR —
**`AssertHookEndpointFollowsBindAddr` (#1178)**.

This last one shipped in two steps because it hit a real blocker: the shared
contract's `HookInstaller.Entry`/`EndpointOf` fields were typed
`map[string]interface{}` and it read installed entries back via
`hookjson.ReadSettings` — hardcoded to hookjson's nested-matcher-group JSON
shape, incompatible with hooktoml's raw `[]byte` TOML blocks. Rather than
bend `hooktoml` to fit or modify `contracttesting` unilaterally, I posted a
precise spec to #1718 for the seam a concurrent agent (kiro-cli, #1716) was
about to build blind from the flat-JSON side. #1734 landed that seam:
`HookInstaller` gained `ReadEntries func(path, event string) ([][]byte, error)`
and `EndpointOfRaw func(entry []byte) string`, both nil-defaulting to bridges
so the four JSON adapters needed zero changes.

`hooktoml` gained two small exports the seam consumes directly:

- `HookBlocks(path) ([][]byte, error)` — every `[[hooks]]` block's raw
  bytes, sharing `scanSections` with every writer in the package, so it
  refuses (rather than silently returns fewer blocks) on the same constructs
  they refuse on.
- `FieldValue(fragment, key) (string, bool)` — decodes a `key = "..."`
  assignment out of raw TOML text, reusing the existing escape-handling
  scanner rather than a second naive one.

`vibe/hookport_test.go` wires `ReadEntries`/`EndpointOfRaw` as thin wrappers
over those two exports. `EnsureInstalled`/`Uninstall` are the real adapter
entry points (`EnsureHooksInstalled`/`UninstallHooks`), not a hooks.toml-only
stand-in, so the contract exercises the actual production code path.

**Checked the seam against real hooktoml, not just the reference TOML
installer it shipped with — it fits cleanly, no bending required.** The
`event` parameter is ignored exactly as designed (hooktoml has no per-event
disambiguation, a decision the seam's own doc comment and a dedicated lock
test both accommodate). Sentinel matching via `hasOurRawEntry`'s
`bytes.Contains` matches `findOurBlock`'s own convention exactly. The
map-based `Entry`/`EndpointOf` pair for obligation 1 (memory-only, no
filesystem access) is a deliberate synthetic representation — the reference
installer validates the identical shape, not something I had to invent to
cope with a mismatch.

## Red-first mutation evidence

Five mutations total, each applied and reverted (one per foreground command).

**Three from the original wiring** (staleness reconciliation, two-file
uninstall coordination, unknown-event dispatch) — see the earlier revision of
this PR body / commit history for the full failure output; unchanged by this
update.

**Two new, for the #1178 seam wiring specifically:**

**1. `vibeEndpointOfRaw` forced to always return `""`** (the wiring glue between hooktoml and the contract):
```
=== RUN   TestHookEndpointFollowsBindAddr
=== RUN   TestHookEndpointFollowsBindAddr/delivery_carries_no_address
=== RUN   TestHookEndpointFollowsBindAddr/install_writes_address_free_delivery
    hook_endpoint.go:265: event post_agent_turn: "" does not carry Sentinel "hook-post mistral-vibe " — EndpointOf is not reading the field the installer writes, so every obligation below it is graded against the wrong string
=== RUN   TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place
    hook_endpoint.go:267: event post_agent_turn: "" does not carry Sentinel "hook-post mistral-vibe " — EndpointOf is not reading the field the installer writes, so every obligation below it is graded against the wrong string
=== RUN   TestHookEndpointFollowsBindAddr/uninstall_is_not_binary_scoped
--- FAIL: TestHookEndpointFollowsBindAddr (0.00s)
    --- PASS: TestHookEndpointFollowsBindAddr/delivery_carries_no_address (0.00s)
    --- FAIL: TestHookEndpointFollowsBindAddr/install_writes_address_free_delivery (0.00s)
    --- FAIL: TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place (0.00s)
    --- PASS: TestHookEndpointFollowsBindAddr/uninstall_is_not_binary_scoped (0.00s)
```
Fails exactly on the two obligations that read entries back through
`EndpointOfRaw`; obligation 1 (memory-only) and obligation 4 (checked via raw
`bytes.Contains`, not `EndpointOfRaw`) correctly stay green.

**2. `vibeHookConfig`'s `IsCanonical` (real production code, not test glue) forced to always report canonical:**
```
=== RUN   TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place
    hook_endpoint.go:267: event post_agent_turn: installed delivery = "'/nonexistent-irrlicht-1718/bin/irrlichd' --version hook-post mistral-vibe >/dev/null || true", want "'<real test binary path>' --version hook-post mistral-vibe >/dev/null || true"
--- FAIL: TestHookEndpointFollowsBindAddr (0.00s)
    --- PASS: TestHookEndpointFollowsBindAddr/delivery_carries_no_address (0.00s)
    --- PASS: TestHookEndpointFollowsBindAddr/install_writes_address_free_delivery (0.00s)
    --- FAIL: TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place (0.00s)
    --- PASS: TestHookEndpointFollowsBindAddr/uninstall_is_not_binary_scoped (0.00s)
```
Fails exactly on the obligation this code protects — the shared contract
catching a genuine production regression through the new seam, not merely
grading the wiring around it.

**Plus one for `hooktoml.HookBlocks` itself:** `scanSections`'s refusal
swallowed instead of propagated → breaks
`TestHookBlocks_RefusesOnUnterminatedCommand` (the same property the
reference TOML installer's own mutation test pins, now pinned directly
against the function the real wiring calls).

## An unplanned finding from the previous revision

`core/cmd/irrlichd/managedfiles_test.go`'s `TestPrintManagedFilesCoversEveryDeclaredFile`
never accounted for `Also` entries in its `want` set — invisible until vibe
became the first real `Also` consumer. Caught by the full suite run, fixed in
this branch's commit history.

## Test plan

- [x] `go build ./...`
- [x] `go test ./core/... -race -count=1` — all green (foreground, full run)
- [x] `go test ./tools/onboarding-factory/... -race -count=1` — all green
- [x] `tools/preflight.sh --only go|arch|tools|skills|posix|security` — all PASS
- [x] Pre-push hook passed on the actual push
- [x] Five deliberate mutations, each seen red and reverted (above + prior revision)
- [x] `hooktoml`'s property tests caught two real defects pre-adapter (see the `hooktoml` commit)
- [x] Rebased onto `f27ea2c6` (#1734 merged); force-pushed with `--force-with-lease`

Part of #1718 / #1355. Builds on #1731 and #1734.

### DIFF
--- a/core/adapters/inbound/agents/hooktoml/hooktoml.go
+++ b/core/adapters/inbound/agents/hooktoml/hooktoml.go
@@ -1,0 +1,614 @@
+// Package hooktoml implements comment-preserving read-modify-write for a
+// narrow, purpose-built pair of TOML edits: appending/upgrading/removing ONE
+// sentinel-tagged [[hooks]] array-of-tables entry, and setting/clearing ONE
+// top-level boolean scalar. It is hookjson's sibling for TOML (issue #1718),
+// not a general TOML merge engine.
+//
+// mistral-vibe's hooks.toml holds exactly one hook entry this daemon
+// installs (post_agent_turn), and its config.toml's enable_experimental_hooks
+// is exactly one boolean gate the entry cannot fire without — so a general
+// nested-structure differ (hookjson's shape, built for Claude Code's
+// event -> matcher-group -> entry nesting shared by claudecode/codex/
+// gemini-cli) would be solving a problem this adapter does not have. Per
+// AGENTS.md's rule ("code that emits bytes from a structural diff gets a
+// property test"), this package earns the same property-test treatment
+// hookjson's splicer does (property_test.go), scoped to the two structural
+// axes it actually varies: appending/upgrading/removing a sentinel-tagged
+// array-of-tables block, and setting/clearing a sentinel-free top-level
+// scalar.
+//
+// The strategy mirrors hookjson's: never re-serialize what did not change.
+// Both operations work as byte-range edits over the ORIGINAL file, computed
+// by a purpose-built line-oriented scanner rather than a general TOML
+// parser — core/go.mod carries no TOML dependency and none is added here.
+// A read-modify-marshal through a generic TOML encoder is exactly the
+// "cost of this ticket" #1718 exists to avoid: it would delete the user's
+// comments the same way json.MarshalIndent over a bare map did before
+// hookjson existed.
+//
+// Anything the scanner cannot confidently classify — a triple-quoted
+// string, a multi-line array, an unterminated single-line string — makes it
+// refuse the WHOLE document rather than guess. Unlike hookjson, which falls
+// back to a lossy whole-document re-encode when a splice is not safe, there
+// is no safe fallback encoder here: a generic map[string]interface{}-style
+// TOML dump is precisely the read-modify-marshal this package exists to
+// avoid, so it is never reached for as a fallback. "Cannot splice safely"
+// therefore means "do not write," not "write something worse" — refusing an
+// install into an exotic hand-crafted hooks.toml is the safe direction; a
+// corrupted one is not.
+package hooktoml
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"os"
+	"strings"
+)
+
+// ErrUnsafe means this document contains a construct the line-oriented
+// scanner does not model with confidence — a triple-quoted string, an
+// unterminated single-line string, or a multi-line array. The caller must
+// not write; there is no lossy fallback (see the package doc).
+var ErrUnsafe = errors.New("hooktoml: file contains a construct this splicer does not model safely (multi-line string, multi-line array, or unterminated quote) — refusing rather than guessing")
+
+// --- line-level scanning primitives ---
+
+// lineSafety scans one line (no embedded '\n'), respecting single-line basic
+// ("...") and literal ('...') strings, and reports the comment-stripped
+// content plus whether the line looks like it continues onto the next one —
+// an unterminated quote (the start of some multi-line construct this
+// scanner does not model) or an unbalanced '['/']' count outside strings
+// (a multi-line array). continues is the signal every caller in this
+// package refuses on rather than guesses through.
+func lineSafety(line []byte) (stripped []byte, continues bool) {
+	inBasic, inLiteral := false, false
+	depth := 0
+	commentAt := -1
+loop:
+	for i := 0; i < len(line); i++ {
+		c := line[i]
+		switch {
+		case inBasic:
+			if c == '\\' {
+				i++ // skip the escaped character too
+				continue
+			}
+			if c == '"' {
+				inBasic = false
+			}
+		case inLiteral:
+			if c == '\'' {
+				inLiteral = false
+			}
+		default:
+			switch c {
+			case '"':
+				inBasic = true
+			case '\'':
+				inLiteral = true
+			case '#':
+				commentAt = i
+				break loop
+			case '[':
+				depth++
+			case ']':
+				depth--
+			}
+		}
+	}
+	if commentAt >= 0 {
+		stripped = line[:commentAt]
+	} else {
+		stripped = line
+	}
+	return stripped, inBasic || inLiteral || depth != 0
+}
+
+func isArrayHeaderLine(t []byte) bool {
+	return len(t) >= 4 && bytes.HasPrefix(t, []byte("[[")) && bytes.HasSuffix(t, []byte("]]"))
+}
+
+func isTableHeaderLine(t []byte) bool {
+	return len(t) >= 2 && bytes.HasPrefix(t, []byte("[")) && !bytes.HasPrefix(t, []byte("[[")) &&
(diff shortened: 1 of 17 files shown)


ingo-eichhorst:
## Independent QA

### The seam wiring is real, not decorative — reproduced against production code

The strongest evidence here is the mutation that touches `hookinstaller.go` rather than the test wiring.
Forcing `vibeHookConfig`'s `IsCanonical` to always report canonical:

```
--- FAIL: TestHookEndpointFollowsBindAddr
    --- FAIL: TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place
```

Exactly that sub-obligation, and nothing else. That is the claim worth verifying: the shared #1178
contract genuinely grades vibe's TOML entries through #1734's widened seam, and it discriminates rather
than failing broadly. Reverted by copy-back; `git status --porcelain` empty.

Contract coverage matches copilot's set call-for-call — all eight wiring points, with
`AssertHookEndpointFollowsBindAddr` now a real call rather than the interim coverage the first revision
carried.

### Property tests

`TestEnsureAndUninstall_PropertyRandomMutations` and `TestSetTopLevelBool_PropertyRandomMutations`:
fixed seeds (17180001 / 17180002), 2000 committed iterations each, measured 0.74s and 0.63s here —
which is the shape AGENTS.md prescribes, with the wider sweep run locally before landing rather than
committed. These found two real defects before any adapter code existed to mask them (a leading comment
attributed to the wrong `[[hooks]]` section; an EOF scalar insertion gluing onto the previous line),
which is the entire argument for the rule.

### Two-file declaration

`Writes.Path` = `hooks.toml`, `Writes.Also` = `config.toml`. This is #1731's field's first real consumer,
and it is the reason the `enable_experimental_hooks` gate is visible to both `sharedConfigRefusal`
(#1449) and the recorder's backup sweep rather than being an undeclared side effect.

### CI note

`swift-test` failed on this Go-only PR with
`failed downloading '...Sparkle-for-Swift-Package-Manager.zip' ... badResponseStatusCode(504)` — a
transient fetch failure for a binary dependency, not a test failure. `macos-swift.yml` is green on the
last six `main` runs including both of today's merges. Re-run rather than dismissed on the basis of the
specific error, not on "Swift is flaky".

PR #1734, opened 2026-08-21
feat(contracttesting): widen the hook-endpoint seam to raw bytes for TOML adapters

ingo-eichhorst:
## Summary

#1716's `HookInstaller.EntriesOf` seam (issue #1178's contract family,
`AssertHookEndpointFollowsBindAddr`) made the JSON matcher-group traversal
pluggable so kiro-cli's flat entry shape could be graded too. mistral-vibe's
#1718 audit (see [that issue's "Spec input for the shared #1178 seam"
comment](https://github.com/ingo-eichhorst/Irrlicht/issues/1718#issuecomment-5368126477))
found that design does not extend to a TOML-writing adapter (`hooktoml`) for
three independent reasons:

1. The **read step itself** has to be pluggable, not just the traversal after
   it — `hooktoml` never produces a decoded generic structure at all. It
   works entirely as byte-range edits over the original file bytes, so there
   is nothing for a `map[string]interface{}`-shaped `EntriesOf` to receive.
2. `map[string]interface{}` is the wrong currency for an entry — a TOML
   `[[hooks]]` block is a byte range, not a decodable value.
3. `hookjson.HasOurHook` is called at two more sites beyond the traversal
   (the seed cross-check and the uninstall-survival check), both hardcoded to
   the nested matcher-group shape and to named-field inspection.

## What changed

`HookInstaller` gains two new fields, both optional:

- `ReadEntries func(path, event string) ([][]byte, error)` — reads the
  settings file and returns the **raw bytes** of every entry belonging to
  `event`, however the format encodes one.
- `EndpointOfRaw func(entry []byte) string` — extracts the delivery string
  from one entry's raw bytes.

Both default to bridges built from the existing `EntriesOf` /
`hookjson.ReadSettings` / `EndpointOf` machinery, so **every JSON-shaped
adapter needs zero changes** — the four wired before #1716, plus kiro-cli's
`EntriesOf: FlatHookEntries`. A TOML adapter supplies `ReadEntries` and
`EndpointOfRaw` directly and never touches `EntriesOf` or
`hookjson.ReadSettings` at all.

The sentinel check (`hasOurRawEntry`, replacing `hookjson.HasOurHook` at all
three of its former call sites) is now a single, fully format-agnostic
`bytes.Contains` over an entry's raw bytes — which is not a weaker check
adopted for convenience, it is `hooktoml`'s **own** sentinel convention
(`findOurBlock` matches by `bytes.Contains` over the whole rendered block's
raw text, never by parsing a field back out), so this satisfies that
convention exactly rather than approximating it.

`Entry`/`EndpointOf` (the in-memory pair grading obligation 1 — "does the
delivery follow the bind address") are **unchanged**: they never touch the
filesystem, so an adapter can honestly answer that question with a map even
when what it persists to disk is raw TOML text.

## How a TOML-shaped installer satisfies this

`hook_endpoint_toml_test.go` is a new reference installer — a byte-range
`[[hooks]]` block scanner, deliberately not a real TOML parser, matching
`hooktoml`'s own approach — proving the seam end-to-end via
`TestTOMLEntryShapeContract`, which runs the full `#1178` contract through
`AssertHookEndpointFollowsBindAddr` exactly as a real adapter would call it.

It also settles the per-event-lookup question #1718 raised (`hooktoml` has no
per-event disambiguation today — a flat `[[hooks]]` sequence, one event
installed): the reference installer declares exactly one event and its
`ReadEntries` **ignores the `event` parameter entirely**
(`TestReferenceTOMLReadEntries_IgnoresEventArgument` is the lock). The seam
does not require per-event lookup — an adapter is free to ignore the
parameter when its format has nothing to filter on.

## Mutation evidence

Every mutation below was applied, seen red, and reverted, one per commit-less
foreground step (not preserved as separate commits, but each is independently
reproducible from the description):

- **Malformed shape is refused, not silently emptied.** A `[[hooks]]` block
  whose `command = "..."` value is truncated (never closed) is refused by
  `ReadEntries` with an error. Disabling the balance check
  (`balancedCommandQuotes`) makes `TestReferenceTOMLReadEntries_RejectsUnterminatedCommand`
  fail: the malformed fixture is silently read as 1 entry with no error —
  exactly the "cannot tell a corrupted write from an empty file" failure mode
  this refusal exists to prevent.
- **`ReadEntries` is load-bearing.** Removing it from the reference
  installer's wiring falls back to the default JSON bridge
  (`hookjson.ReadSettings` on TOML bytes), which fails loudly:
  `invalid character 'h' looking for beginning of value`. `TestTOMLEntryShapeContract`
  goes red on 3 of 4 sub-tests.
- **`EndpointOfRaw` is load-bearing.** Removing it falls back to the default
  bridge (`json.Unmarshal` of raw TOML bytes into a map), which fails
  *silently* and returns `""` — caught not by the bridge itself but by
  `assertDeliveryIsOurs`'s existing vacuity guard ("does not carry Sentinel").
  `TestTOMLEntryShapeContract` goes red on 2 of 4 sub-tests (the other two
  don't call `EndpointOfRaw` at all, which is noted in the test).
- **Vacuity guard**: `TestTOMLEntryShapeContract` itself, run against a
  correct fixture, passes 4/4 — the positive case every mutation above is
  measured against.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/contracttesting/... -v` — full package green,
      including the pre-existing seam-walk structural tests, self-tests, and
      every other contract family
- [x] `gofmt -l` clean
- [x] Three deliberate mutations (above) applied, confirmed red, reverted
- [x] `tools/preflight.sh --only go|arch|tools|skills|posix|bash|security` —
      all green (swift/web untouched, correctly skipped)
- [x] Pre-push hook (`--changed`) ran clean on push

Linked: #1178, #1716, #1718

### DIFF
--- a/core/internal/contracttesting/hook_endpoint.go
+++ b/core/internal/contracttesting/hook_endpoint.go
@@ -142,10 +142,44 @@ type HookInstaller struct {
 	// Neither way of getting it wrong passes: an entry that is already
 	// canonical leaves obligation 3 with modified=false, and one that does not
 	// carry Sentinel is not recognized as ours, so EnsureInstalled appends a
-	// second matcher group and onlyEntry fails on the group count. The seed is
+	// second matcher group and onlyEntry fails on the entry count. The seed is
 	// additionally cross-checked against Sentinel before either obligation
 	// runs.
 	ForeignInstall func() (bool, error)
+
+	// EntriesOf extracts the inner hook-entry objects an event's raw value in
+	// the parsed settings map holds, regardless of nesting shape (issue
+	// #1716). Nil means NestedHookEntries — hookjson's own matcher-group
+	// shape, `[ {"matcher": ..., "hooks": [ {...} ] } ]` — which is what every
+	// adapter wired before #1716 (claudecode, codex, copilot, gemini-cli)
+	// already produces, so none of them need to set this field.
+	//
+	// It exists because that nesting is not universal: kiro-cli's hooks
+	// schema rejects it outright and falls back to a different agent while
+	// reporting install success (#1716's own audit), so its entries are FLAT —
+	// `[ {"command": ...} ]`, no matcher-group wrapper — and the four
+	// obligations built on "read back the one entry an event holds" needed a
+	// seam an adapter can supply instead of a shape hardcoded into the
+	// traversal. FlatHookEntries is that shape's counterpart.
+	//
+	// A TOML-shaped installer (mistral-vibe, #1718) fits the SAME seam for its
+	// half of the problem — decoding a TOML table's array-of-tables into
+	// `[]map[string]interface{}` is what most TOML libraries already do, so an
+	// adapter whose event value decodes that way can write its own
+	// EntriesOf identically to FlatHookEntries's body. What this field does
+	// NOT close for TOML is the file-reading step: readHooksMap below still
+	// calls hookjson.ReadSettings (JSON-only) to produce the map EntriesOf is
+	// handed, and #1718's own config is a dedicated hooks.toml rather than a
+	// "hooks" key inside a shared JSON document — so a TOML adapter needs an
+	// analogous seam THERE too (a HookInstaller.ReadSettings func(path string)
+	// (map[string]interface{}, error) field, defaulting to
+	// hookjson.ReadSettings, is the natural shape, mirroring this one) before
+	// AssertHookEndpointFollowsBindAddr can grade it end to end. Left
+	// undone here deliberately: guessing at #1718's exact schema from outside
+	// that work would be speculation this package cannot verify, where
+	// EntriesOf's shape is what THIS issue's own adapter (kiro-cli) already
+	// needs and proves against a real wiring (kirocli/hookport_test.go).
+	EntriesOf func(hooksMap map[string]interface{}, event string) ([]map[string]interface{}, error)
 }
 
 // AssertHookEndpointFollowsBindAddr runs the issue #1178 contract against h, in
@@ -359,7 +393,7 @@ func assertUninstallIsNotForeignScoped(t armT, h HookInstaller, r deliveryRules,
 	}
 	hooksMap := readHooksMap(t, path)
 	for _, event := range h.Events {
-		if hookjson.HasOurHook(hooksMap, event, h.Sentinel) {
+		if hasOurEntry(t, h, hooksMap, event) {
 			t.Errorf("event %s: %s survived uninstall", event, r.foreign)
 		}
 	}
@@ -384,7 +418,7 @@ func assertEveryEventDelivers(t armT, h HookInstaller, r deliveryRules, path, wa
 	t.Helper()
 	hooksMap := readHooksMap(t, path)
 	for _, event := range h.Events {
-		got := h.EndpointOf(onlyEntry(t, hooksMap, event))
+		got := h.EndpointOf(onlyEntry(t, h, hooksMap, event))
 		assertDeliveryIsOurs(t, h, "event "+event, got)
 		if got != want {
 			t.Errorf("event %s: installed delivery = %q, want %q", event, got, want)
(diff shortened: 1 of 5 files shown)


ingo-eichhorst:
## Blocking defect found in review, fixed

A coordinator review found that \`endpointOfRaw\`'s default bridge (the fallback when an adapter doesn't supply \`EndpointOfRaw\`) returned \`""\` on a \`json.Unmarshal\` failure instead of failing loudly. That is not a neutral value under the \`DeliveryAddressFree\` route: \`assertDeliveryCarriesNoAddress\` checks that the delivery is IDENTICAL across bind addresses and carries no address-shaped fragment, and \`""\` satisfies both trivially — so a non-JSON entry shape with \`EndpointOfRaw\` left unset made obligation 1 (\`delivery_carries_no_address\`) pass having graded nothing. It was only caught in my own mutation testing because a NEIGHBOURING guard, \`assertDeliveryIsOurs\`, happened to run first in the one obligation that reaches the bridge before the address check — exactly the "absence of a finding and inability to look must never produce the same output" failure AGENTS.md's Testing section names.

**Fix**: \`endpointOfRaw\` now takes a \`reporter\` and \`Fatalf\`'s on the unmarshal error, naming the entry bytes and telling the caller to supply \`EndpointOfRaw\` for a non-JSON shape, instead of returning a sentinel a caller could silently accept.

**Mutation evidence, redone against the fix:**
- Removed \`EndpointOfRaw\` from the TOML reference installer's wiring — the failure now names the parse failure directly from the bridge (\`hook_endpoint.go\`, "does not unmarshal as JSON"), not \`assertDeliveryIsOurs\`'s "does not carry Sentinel". Confirmed red on 2 of 4 sub-tests, reverted.
- Same mutation, run a second time with \`assertDeliveryIsOurs\` ALSO temporarily disabled (turned into a no-op) — the bridge still fails with the identical message, proving the bridge itself is what catches it now, not the neighbour. Reverted both.
- Added a committed, permanent lock — \`TestEndpointOfRaw_MissingBridgeFailsLoudlyNotSilently\` (\`hook_endpoint_test.go\`) — driven through the package's own negative-self-test harness (\`observe\`/\`mustReport\`/\`mustBeSilent\`) with NO neighbouring assertion anywhere in the call, so nothing else could be catching it. Confirmed red against the pre-fix \`""\` behavior, reverted.
- Also verified (not assumed) the read-side twin the same review asked about: \`rawEntriesOf\`'s default bridge cannot silently return an empty entry list on a settings file it "half understands". Two new committed tests confirm a \`"hooks"\` key present but the wrong shape, and an event key genuinely absent, both surface as real errors — plus a vacuity guard confirming a legitimately empty per-event array is NOT refused (that's the everyday "nothing installed yet" state). All three verified to discriminate: a follow-up mutation swallowing the shape-mismatch into an empty result reddened exactly the two error-expecting tests and left the vacuity guard green, then reverted.

All gates re-run and green: \`go\`, \`arch\`, \`security\` (the groups this Go-only change touches). Pushed as \`f68b22ac\`.

ingo-eichhorst:
## Independent QA — the silent-bridge fix

Reproduced the defect against the committed lock rather than taking the report on trust. Deleting the
`t.Fatalf` from `endpointOfRaw`'s default bridge, restoring the pre-fix `return ""`:

```
--- FAIL: TestEndpointOfRaw_MissingBridgeFailsLoudlyNotSilently (0.00s)
    hook_endpoint_test.go:129: mutation "non-JSON entry, EndpointOfRaw unset": the obligation reported
    nothing against a fixture that is wrong for it — either the fixture is no longer wrong, or the
    obligation has stopped discriminating (expected a message containing "does not unmarshal as JSON")
```

Reverted by copy-back; `git status --porcelain` empty.

That message is the right one: it distinguishes "the fixture stopped being wrong" from "the obligation
stopped discriminating", which is the distinction a negative self-test exists to make and the one a bare
failure cannot.

### Why this mattered more than a normal nil-guard

`""` is not a neutral value on the `DeliveryAddressFree` route. `assertDeliveryCarriesNoAddress` checks
`onDefault != onAlt` (both sides came from the same failing bridge, so both were `""`) and then
`assertNoAddress` on the result (`addressShaped.FindString("")` matches nothing). Both halves passed. So a
parse failure returned the exact string the obligation asserts — obligation 1 passed having graded
nothing, and any future adapter supplying `ReadEntries` for a non-JSON shape while forgetting
`EndpointOfRaw` would have inherited a green contract that looked at nothing.

The original mutation caught this only via `assertDeliveryIsOurs`, a neighbouring guard. The redone
evidence runs the same mutation with that neighbour disabled, which is what separates "something failed"
from "this obligation failed" — the distinction #1453's second instance turned on.

### Also verified

The read-side twin (`rawEntriesOf`'s default bridge) cannot silently return an empty entry list on a
settings file it half-understands: a present-but-wrong-shape `hooks` key and an absent event key both
surface as real errors, with a vacuity guard confirming a legitimately empty per-event array is still
accepted. Checked rather than assumed, per the same rule that produced the finding above.

Issue #1480, opened 2026-08-11
Replay goldens pin transition timestamps that nothing asserts — a transition moved 31s is a full pass


ingo-eichhorst:
Found by the independent review of #1342 / PR #1476, and it is the reason that PR's one real cost went unnoticed until someone measured it by hand.

## The gap

Replay goldens record a `virtual_time` on every transition. **Nothing compares it to anything.**

- `assertReproducesRecordedTransitions` (`issue1342_debounce_test.go`) checks `ReplayedCount >= RecordedCount`, `MissingKinds`, `ExtraKinds`.
- `compareOrdered` (`extended_check.go:52`) walks `prev_state`/`new_state` **index by index** and never reads the time.
- The extended check's headline metric — the "N of 309 recordings diverge" figure quoted in every replay PR — is counts and kinds only.

So a transition reproduced at the right position in the order but **31 seconds** from when the daemon actually made it is a full pass on every gate in the package, and the golden then pins it as correct forever.

## It is not hypothetical — it is live on `main` as of #1476

PR #1476's `readBoundaryFor` widening bounds itself by the next recorded stat's SIZE, never its TIME. Across a long idle gap it therefore classifies against bytes that did not exist at the pass being classified. Measured over 114 firings:

| gap to next stat | |
|---|---|
| p50 | 42ms |
| p75 | 13.5s |
| p90 | 15.9s |
| max | 31.0s |

**20 goldens now pin a first transition more than 1s ahead of their own `events.jsonl`** — worst case `mistral-vibe/scenarios/2-12_context-compaction`, where the golden says `ready→working` at `17:23:08.571` and the daemon's own log says `17:23:39.548`, a **30.976s** difference.

That change shipped with 13/13 CI green, a passing 309-recording catalog walk, and a *net improvement* on every metric anything measures. It is genuinely a net win on timing too (recordings drifting >1s: 171 → 146; >5s: 117 → 88) — which is exactly why nobody would have found the regression by looking at the aggregates. The cost is real and it is invisible.

## Why this matters more than the count suggests

This is the same failure shape as #1342 itself, one axis over: **a net that stopped being able to fail**. #1342 was 22 goldens reduced to asserting nothing. This is 399 goldens asserting nothing *about time*, permanently, while looking fully populated.

[...]

Issue #1388, opened 2026-08-09
codex declares hooks and has ZERO hook-bearing recordings — the channel #1355 makes primary has no regression coverage for half its adapters


ingo-eichhorst:
Parent: #1355 — Phase B. Surfaced immediately by `of coverage --hooks` (#1363 / PR #1374), which was built to make exactly this loud.

## The finding

**codex declares a `hooks` permission and has zero hook-bearing recordings.**

```
adapter        declares-hooks  cells  recordings  with-hooks  status
claudecode                yes     46          52          11  ok
codex                     yes     44          36           0  GAP — declares hooks, zero hook-bearing recordings
```

Independently verified against the catalog on `main`:

```
$ grep -rl '"hook_received"' replaydata/agents/codex/ | wc -l
0
$ find replaydata/agents/codex -name events.jsonl | wc -l
43
$ grep -rl '"hook_received"' replaydata/agents/claudecode/ | wc -l
15
$ grep -n "PermissionKeyHooks" core/adapters/inbound/agents/codex/agent.go
11:// PermissionKeyHooks gates installing Codex hooks into ~/.codex/hooks.json for
13:const PermissionKeyHooks = "hooks"
65:  Key:             PermissionKeyHooks,
```

## Why this blocks the rollout rather than being a nice-to-have

#1355's premise is that hooks become the tier-1 state channel. Two adapters use them today. **One of the two has no fixture that exercises the channel at all**, so every replay-gated guarantee we have about codex's hook path is guaranteeing nothing: the goldens would be byte-identical with the hook handling deleted.

That is the concrete form of the epic's "only 17 of 396 `events.jsonl` contain a `hook_received`" measurement. It is not an abstract coverage worry — it names one adapter, and it is the adapter whose hook support was added most recently (#1171).

Codex hook behaviour *was* live-verified on 0.144.4 (`PermissionRequest` → waiting, `Stop` → ready). What is missing is a **recording**, so nothing re-checks it after a refactor. Two changes already in flight touch that exact path — #1361 (path confinement, which moved codex's rejection from 200 to 400) and #1364 — and neither can be caught by replay today.

## Why it was never recorded, and what changed

The recording rig could not produce codex hook deliveries while `:7837` was hardcoded in the installers — the dev/recording daemon runs on an alternate port, so hooks went to the production daemon or nowhere. **That blocker is gone:** #1178 / #1207 made installers write the resolved port (`daemonaddr.LocalURL`).

[...]

Issue #522, opened 2026-05-30
Onboarding factory P1: rename to tools/onboarding-factory + regression→regressions


ingo-eichhorst:
## Goal
Pure-mechanical rename, fully green, so later phases start from a clean base.

- `git mv tools/agent-onboarding tools/onboarding-factory`; `go.mod` module → `irrlicht/tools/onboarding-factory`; rewrite every internal import `irrlicht/tools/agent-onboarding/...` → `.../onboarding-factory/...` (all `internal/` + `cmd/`).
- `go.work`: `./tools/agent-onboarding` → `./tools/onboarding-factory`.
- `git mv replaydata/agents/<a>/regression replaydata/agents/<a>/regressions` ×5 (aider, claudecode, codex, opencode, pi).
- Fix the `"regression"` literal in: `internal/viewer/store.go` (the `{"scenarios","regression"}` walk), `internal/viewer/scenarios.go` (subtree route + validation), `internal/viewer/models.go`, `internal/viewer/web/viewer.js` (subtree pills), `tools/replay-fixtures.sh`, and any skill script walking `regression/`.
- Grep & fix hard-coded `tools/agent-onboarding` paths in `tools/replay-fixtures.sh`, `.claude/skills/ir:onboard-agent/scripts/*`, CI workflows, root docs.

## Verify
- `cd tools/onboarding-factory && go build ./... && go test ./... -race`
- `go test ./core/... -race`
- `tools/replay-fixtures.sh` clean over `scenarios/` + `regressions/`

## Risk
Low. Only gotcha: a missed `regression` literal → viewer 404 (caught by `server_test.go`).

Part of #521

ingo-eichhorst:
Done — landed on `main` via #546 (the full P0–P8 stack was rebase-merged onto main; linear-history rule blocked merge commits, so all per-phase commits are on main directly). Closing.

PR #1737, opened 2026-08-21
chore(replay): retire codex's pre-hooks 2-13_turn-end-terminal-text recording

### LINKED ISSUE #1728: chore(replay): retire codex's pre-hooks 2-13_turn-end-terminal-text recording — superseded by #1388, still pinned in #1480's timing ratchet
#1355 measured "Only 17 of 396 `events.jsonl` files contain a `hook_received` event" on 2026-08-09 and #1388 ("codex declares hooks and has ZERO hook-bearing recordings") used it to justify recording codex's first hook fixtures. **#1388 closed 2026-08-18/19** (PR #1694 + the grading fix in PR #1698), and that figure has moved. Re-measuring before scoping anything below, per AGENTS.md's rule that a figure documenting behaviour states the command that produces it.

## Current state, measured

```
$ go run ./tools/onboarding-factory/cmd/of coverage --hooks
adapter        declares-hooks  cells  recordings  with-hooks  status
claudecode                yes     46          53          12  ok
codex                     yes     44          38           2  ok
copilot                   yes     46          37           1  ok
gemini-cli                yes     45          38           0  GAP — declares hooks, zero hook-bearing recordings
hermes                     no     46          24           1  incidental — hook events present, adapter declares no hooks
...
GAP: 1 of 4 hooks-declaring adapters have zero hook-bearing recordings: gemini-cli
```

`codex` is `ok`: PR #1694 recorded `codex/scenarios/2-13_turn-end-terminal-text` (`hook_name: "Stop"`) and `codex/scenarios/2-19_tool-gate-permission-prompt` (`hook_name: "PermissionRequest"`) — exactly the two rows codex's `session.HookSignal` table declares. #1695/PR #1698 then fixed a grading bug that would otherwise have made the Stop recording's coverage vacuous (`TestStopHookIsGradedByTheCommittedCatalog`).

Raw counts, for anyone re-deriving this without `of coverage --hooks`:

```
$ find replaydata -name events.jsonl | wc -l
399
$ grep -rl '"hook_received"' --include=events.jsonl replaydata | wc -l
20
```

399/20, not the 396/17 (15 claudecode, 1 copilot, 1 hermes) #1355 recorded — the delta is codex's two plus claudecode's own newest recording (`2-13_turn-end-terminal-text`, dated 2026-08-19), all landing after whatever pass produced 396/17. The `hermes` credit is real but not hermes's: `4-2_multiple-agents-same-workspace/.../events.jsonl` seq 192 is `hook_name: "Stop"` with no `adapter` field on the lifecycle event — `hookcov`'s own doc comment notes "the hook_received events actually present in the corpus leave [adapter] empty" — a claudecode hook firing during a multi-agent recording, credited to hermes's cell only because that's whose directory it landed under. hermes declares no `hooks` permission.

**gemini-cli is the only real gap, and it cannot be closed here.** #1717 implemented its hooks (source-verified event mapping) but this machine's gemini-cli auth is `IneligibleTierError`-dead, and gemini-cli is flagged "unmaintained, skip by default" for deprecation at v0.7 (#1717, #1725). Neither is fixable by re-recording.

So the obvious next step — record more codex cells with hooks granted — is already done. What's left is narrower.

## What a replay fixture catches, and what it doesn't

AGENTS.md names three distinguishable hook diagnoses: entries GONE (#1372's verifier, re-reads the installed config on a timer), entries present but nothing ARRIVING (#1368's liveness watchdog, demotes a channel with zero receipts across N turns), and an install that FAILED (#1362's `effect_error`, a swallowed `Apply` failure). **A replay fixture exercises none of the three** — all three are live-daemon, cross-turn or install-time mechanisms with their own dedicated tests (`contracttesting.AssertHookReceiptObserved`, `TestEveryHookInstallDeclaresAVerifier`, the permission effect-error path). `cmd/replay` only replays an already-captured `events.jsonl` through `applyHookEvent` (`replay_sidecar.go:972`), mapping a hook event that *did* arrive to a state signal via the same `session.HookSignal` table the daemon uses. What a fixture does catch is a **grading regression on a hook that arrived**: #1320 (replay silently dropped `PreToolUse`, fixed by #1323) and #1695 (replay couldn't grade `Stop` at all, fixed by #1698) are real defects of exactly that shape. Presence-in-corpus and grading-correctness are what replay buys; liveness, config integrity and install success were never in scope for it.

## The retirement question

The instruction behind this ticket was "the other file based recordings can be deleted." Read against #1388's two new recordings, that's `codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27/` — the cell's only other recording, pre-dating the hooks-granted one and now superseded for hook-coverage purposes. (`2-19_tool-gate-permission-prompt` has no older sibling; #1388 recorded it fresh.) This is an inference from the phrase, not a confirmed scope — flagged as such; if a different set of recordings was meant, say which.

`replaydata/` is CI-deletion-guarded (#268) — `rm` isn't available. The guard-respecting equivalent is `git mv` into the adapter's `regressions/` tree. `codex/regressions/` already holds ten such folders (`model-switch`, `fork-conversation`, `agent-question-pending`, `baseline-hello`, ...), none carrying a `metadata.json`, consistent with `regressions/` not being a catalog cell.

**Checked, not guessed, what actually breaks:**

- `of validate` — unaffected. `hookcov`'s own scope note and `of coverage --hooks`'s own output establish that `regressions/` "is not a catalog cell and no catalog loader enumerates it" — the FK/completeness gate never walks it, so moving a recording there removes no obligation `of validate` was checking.
- `censusOfTheCommittedCatalog` (`issue1503_census_test.go`) — unaffected on the *count*. `forEachSidecarRecording`, the walk it and #1480's ratchet both share, is a `filepath.WalkDir` over the whole `replaydata/agents` tree, `regressions/` included — confirmed by `knownZeroTransition` already pinning a `codex/regressions/...` entry, and by `tools/replay-fixtures.sh`'s own `find` carrying no `regressions/` exclusion. A `git mv` keeps the sidecar paired with its transcript, so `measured.Recordings` doesn't shrink and `TestCatalogCensusMatchesTheCommittedFigures`'s population-shrink guard doesn't fire.
- `knownZeroTransition`, `knownFabricated` (`issue1342_debounce_test.go`) — checked by name; neither pins this recording. Unaffected.
- **`knownFirstTransitionDrift` (`issue1480_timing_test.go:226`) does pin it**, by full path: `"codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27/transcript.jsonl": "#1476 accepted: -2.246s at pair 0 (ready→working)"`. `TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog` keys `drifted`/`unmeasured` off the recording's relative path, so a `git mv` changes the key. The recording is still walked (regressions/ is in scope, per above) and still drifts — just under a new key — so the *old* key hits neither `drifted[n]` nor `unmeasured[n]` and falls to the last branch: `t.Errorf("knownFirstTransitionDrift entry %q no longer drifts — delete it", n)`. That message is wrong for this case: the drift didn't go away, the path did. The fix has to land in the same change as the `git mv` — delete the old-path entry and add the new-path one (same "#1476 accepted" reason, same delta) — or the test goes red for a reason its own message misdescribes.

Net: the retirement is safe and mechanical, but it is not a bare `git mv` — skipping the pin-list edit fails a real gate with a misleading diagnosis.

## How #1388's two recordings were produced (context, not new work)

For whoever next records a codex cell with hooks granted: `tools/onboarding-factory/scripts/lib/spawn-record-daemon.sh` sets `IRRLICHT_PERMISSION_MODE=grant-all` and `IRRLICHT_ALLOW_SHARED_CONFIG_WRITES=1` together (lines 87-88) against an isolated `IRRLICHT_HOME`, and backs up every managed file via `snapshot_managed_files` first (line 112) — the #1449 guardrail that stops a grant-all daemon writing outside its own home. `run-cell.sh` / `run-cell-multi.sh` drive a live session through it. Nothing here needs redoing for codex; noted so the next hook adapter (#1716 kiro-cli, #1718 mistral-vibe, ...) doesn't have to re-derive it.

## Out of scope: gemini-cli

Permanently, on two independent grounds: this dev machine's gemini-cli auth returns `IneligibleTierError` on every call (its hook implementation from #1717 is source-verified, never live-fired), and gemini-cli is flagged "unmaintained, skip by default" ahead of deprecation at v0.7 (#1717, #1725). Neither is this issue's to fix.

## Scope

1. `git mv replaydata/agents/codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27 replaydata/agents/codex/regressions/<name>/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27` (name TBD — existing siblings are short descriptive slugs, not claudecode's `<n>-<slug>-<hash>`).
2. Rename the `knownFirstTransitionDrift` entry in `issue1480_timing_test.go` to the new path, same reason string.
3. `go test ./tools/onboarding-factory/cmd/replay/... -count=1` and `go run ./tools/onboarding-factory/cmd/of validate` — both should stay clean; if either isn't, that's new information this issue's "checked, not guessed" section got wrong, and it should be corrected rather than worked around.
4. Confirm `of coverage --hooks` still reports codex `ok` (retiring a pre-hooks recording can't regress this — `WithHooks` never counted it).

**Complexity:** 2 — mechanical, but the pin-list edit is the part someone skips without reading this issue first.

ingo-eichhorst:
## Summary

Retires codex's pre-hooks `2-13_turn-end-terminal-text` recording
(`2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27`), superseded by #1388/#1694,
which recorded a hooks-granted sibling in the same cell
(`2026-08-18-00-42-27_irrlichd-0.5.10+1869727`). The pre-hooks recording no
longer earns its place — it only pads the timing-ratchet population — but
it was still pinned by full path in `knownFirstTransitionDrift`
(`issue1480_timing_test.go`), which #1480's ratchet reads.

## Mechanism and precedent

`replaydata/` is CI-deletion-guarded (#268, `.github/workflows/replaydata-deletion-guard.yml`) — the guard explicitly permits renames (`--find-renames=50%` on the deletion diff, "renames … never counted as deletions") and its own failure message says: "To retire a cell … git mv the recording to regression/."

Followed the precedent set by `c8e3d030` (#475, "archive /fork recording to regression/"), which `git mv`'d a superseded codex recording into `regression/fork-conversation/` rather than deleting it. That folder was later renamed `regression/` → `regressions/` (#522); the current sibling layout (`baseline-hello`, `model-switch`, `fork-conversation`, ...) confirmed the pattern: `replaydata/agents/<adapter>/regressions/<slug>/recordings/<recording-name>/`.

```
git mv replaydata/agents/codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27 \
       replaydata/agents/codex/regressions/turn-end-terminal-text-pre-hooks/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27
```

The cell (`codex/scenarios/2-13_turn-end-terminal-text/`) stays live — it keeps its `metadata.json` and its remaining (hooks-granted) recording.

## The pin-list edit had to land in the same change

`knownFirstTransitionDrift` keys on the recording's relative path. A bare `git mv` without renaming the pin entry reproduces the stale-entry failure mode at `issue1480_timing_test.go`'s `TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog`: the recording is still walked (regressions/ is in scope for `forEachSidecarRecording`, same as scenarios/) and still drifts — just under the new path — so the OLD key hits neither `drifted[n]` nor `unmeasured[n]` and the test reports `knownFirstTransitionDrift entry %q no longer drifts — delete it`, which is the wrong diagnosis for a path change. Verified this actually fires (see below), then renamed the entry to the new path with the same reason string in the same commit.

## Figures that moved, before -> after, each reproduced by the command that produced it

- `go run ./tools/onboarding-factory/cmd/of coverage --hooks`: codex
  `recordings` 38 -> 37 (`with-hooks` stays 2, `status` stays `ok` — the
  retired recording was pre-hooks and never counted toward `with-hooks`).
- `go test ./tools/onboarding-factory/cmd/replay/... -run TestCatalogCensusMatchesTheCommittedFigures -v -count=1`:
  `censusOfTheCommittedCatalog` unchanged — `314/1/1/143/142/0/85` before and
  after. `regressions/` is walked by `forEachSidecarRecording` exactly like
  `scenarios/`, so a `git mv` doesn't shrink the population, only relocates
  one entry; the printed literal matched the committed one exactly (test
  passes with no staleness error).
- `knownZeroTransition`, `knownFabricated` (`issue1342_debounce_test.go`):
  checked by name — neither pins this recording, unaffected.
- Replay goldens: `UPDATE_REPLAY_GOLDENS=1 go test ./tools/onboarding-factory/cmd/replay/... -count=1`
  produced no new/changed golden content — `git diff -M50%` on the moved
  golden shows `similarity index 100%`, i.e. a pure rename.

## Stale-entry guard: actually fired, not just reasoned about

Applied the `git mv` alone first (pin entry left pointing at the old,
now-nonexistent path) and ran
`go test ./tools/onboarding-factory/cmd/replay/... -run TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog -v -count=1`://it went red exactly as predicted —
`knownFirstTransitionDrift entry "codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27/transcript.jsonl" no longer drifts — delete it`.
Then applied the pin rename (`codex/regressions/turn-end-terminal-text-pre-hooks/recordings/.../transcript.jsonl`, same reason string, same delta) and re-ran — PASS. Both changes land together in this PR's single commit.

## Test plan

- [x] `go test ./tools/onboarding-factory/... -race -count=1` — all green
- [x] `go run ./tools/onboarding-factory/cmd/of validate` — OK
- [x] `go run ./tools/onboarding-factory/cmd/of coverage --hooks` — codex still `ok` (37 recordings, 2 with-hooks)
- [x] Pre-push `tools/preflight.sh --changed` (gofmt, onboarding-factory tests, replaydata validate, recording-rig smoke test, replay fixtures, security scan) — all PASS

Closes #1728


ingo-eichhorst:
## Independent QA — verified, not taken on trust

Reproduced the load-bearing claim myself in a separate worktree off this branch, rather than accepting the report.

**Baseline (unmutated PR branch):**
`go test ./tools/onboarding-factory/cmd/replay/... -run TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog -count=1` → `ok 1.030s`

**Mutation** — point the pin back at the pre-`mv` path, i.e. exactly "`git mv` without the pin edit":

```
--- FAIL: TestSidecarReplayTransitionTimesMatchTheDaemonsOwnLog (0.62s)
    issue1480_timing_test.go:467: knownFirstTransitionDrift entry
      "codex/scenarios/2-13_turn-end-terminal-text/recordings/2026-05-23-21-44-53_irrlichd-0.4.7+f83dc27/transcript.jsonl"
      no longer drifts — delete it
FAIL
```

Restored; `git status --porcelain` clean. So the ordering constraint #1728 predicted is real and is enforced by a live gate, not just reasoned about.

**Diff shape verified:**
- `git diff -M50% --summary` reports all four files as **rename … (100%)** — a true rename, no delete+add.
- One line changed outside `replaydata/`: the pin entry, re-sorted into alphabetical position. `5 files changed, 1 insertion(+), 1 deletion(-)`.
- The **`guard` check passes** — the replaydata deletion guard accepted it, which is the direct evidence that `--find-renames` covers this move.
- The source cell keeps its hooks-granted recording (`2026-08-18-00-42-27_irrlichd-0.5.10+1869727`), so `2-13_turn-end-terminal-text` stays a live cell rather than becoming an empty one.

**One judgement call I agree with:** naming the folder `turn-end-terminal-text-pre-hooks` rather than a bare slug. #1728 left the name TBD, and among codex's ten `regressions/` siblings this one states *why* it was retired, which is the part a future reader needs.

Merging once the remaining checks finish.

PR #1738, opened 2026-08-21
docs(agents): fix beacon-delivery/adapter-count staleness in AGENTS.md

### LINKED ISSUE #1730: AGENTS.md: beacon-delivery reference-wiring note is stale (gemini-cli adopted it in #1724) + a conservative simplification audit
## Part 1 — stale claim, actionable today

All line numbers below are against `AGENTS.md` on `origin/main` (SHA `c766e06a`) — local `main` is stale and does not have this content at these lines.

`AGENTS.md` makes the same claim in two places: that no adapter has adopted the `DeliveryAddressFree` hook-delivery route yet, so `core/internal/contracttesting/hook_endpoint_addressfree_test.go` is standing in as the "reference wiring" until one does.

- **AGENTS.md:420-424**
  > Whichever adapter adopts beacon delivery first replaces
  > `hook_endpoint_addressfree_test.go`, the reference wiring the route is
  > currently exercised by — and that file is also the shape to copy, down to
  > resolving the binary path once through `hookbeacon.InstalledCommand` so the
  > config builder stays infallible.

- **AGENTS.md:667-669**
  > All seven contract families pass by construction against a correct adapter —
  > or, for a delivery route no adapter has adopted yet, against its reference
  > wiring — so their whole value is that they *can* fail.

**This is false as of PR #1724.** `feat(gemini-cli): hook-based state detection (Notification/AfterTool/AfterAgent, beacon delivery)` (commit `25f63fd3`, merged 2026-08-20 21:43:40 +0200) added `core/adapters/inbound/agents/geminicli/hookport_test.go:50`:
```go
Delivery: contracttesting.DeliveryAddressFree,
```
`hook_endpoint_addressfree_test.go` was not touched or removed by that PR (`git log --oneline origin/main -- core/internal/contracttesting/hook_endpoint_addressfree_test.go` shows no commit since it was introduced).

**The file's own header already names this exact trigger condition, unactioned:**
> "It is scaffolding, and it says so: once an adapter declares this route, that adapter's wiring becomes the honest coverage and this file should be reduced to whatever it still uniquely proves, or deleted." — `core/internal/contracttesting/hook_endpoint_addressfree_test.go:12-14`

**The irony worth stating**: PR #1727 (`docs(process): bounded-wait recipe, findings checkpointing, generalise the sourced-figure rule`, commit `c766e06a`, 2026-08-20 22:35:20 +0200) landed the rule *"a figure that documents behaviour states the command that produces it, or it is marked as an estimate"* — **46 minutes after #1724 merged, same day**. The claim reported here was already false when that rule landed, and #1727's own diff didn't touch AGENTS.md:420-424 or :667-669 to re-check them.

**Nuance for whoever picks this up — it isn't simply "delete the file."** `hook_endpoint_selftest_test.go`'s mutation corpus reuses the same reference installer (`correctBeaconInstaller()` → `referenceBeaconInstaller()`, both defined in `hook_endpoint_addressfree_test.go`) so its self-tests are pinned against a fixture that doesn't move when a real adapter's wiring changes. So there's a real question, not just a stale-doc fix:

1. **Delete/shrink the reference wiring now that a real adapter exercises the route**, rewiring `hook_endpoint_selftest_test.go`'s mutation corpus onto gemini-cli's real installer (or another adopter, if/when there's more than one) — the outcome the file's own header describes; or
2. **Keep it deliberately** as a hermetic, adapter-independent fixture so the self-test mutation corpus doesn't churn every time a real adapter's hook wiring changes — in which case AGENTS.md:420-424 and :667-669 need to say *that*, not "no adapter has adopted this route yet."

Either way, AGENTS.md:420-424 and :667-669 need an edit — they currently assert something that has not been true since 2026-08-20.

### Methodology
Read the entire file directly against `origin/main` (2571 lines / 28,246 words / 187,425 bytes — `wc -l -w -c AGENTS.md`), plus:
- Extracted every backtick-quoted span (1,035 total) and checked every file path, script name, test function name, and Go/Swift symbol against `origin/main`'s tree and source (`git ls-tree`, `git grep -n "func <name>("`, etc.).
- Ran a mechanical 10-word-shingle scan over the whole file to surface any verbatim-repeated passage, as a backstop independent of manual reading.

### Dangling-reference check: zero found
Every one of the following resolved to something real on `origin/main`: all ~90 distinct file paths named in backticks (scripts, Go/Swift source, workflow YAMLs, replaydata fixtures), all 20 `Test*` function names, and all 70 distinct Go/Swift member names in `pkg.Symbol`/`Type.Method` form (e.g. every `contracttesting.Assert*`, every `hookjson.*`, `PermissionService.*`, `agent.*`). Apparent misses were all false positives from my own path-matching heuristic (e.g. `core/bin/irrlichd` is a build *output*, not tracked in git; `core/domain/stats.Percentile` is a dotted symbol, not a path) — each was individually re-verified and is real. This is a strong, mechanically-obtained result for a 2,571-line file and says the doc is not rotting from simple neglect — the two stale claims in Part 1 are a behavioral drift (a route got adopted), not a broken reference.

### Token cost
`CLAUDE.md` in this repo is exactly `@AGENTS.md` (11 bytes) — so the full file is injected into every agent session's context. 28,246 words is ≈37,000-47,000 tokens (word-based / char-based estimates) of fixed per-session cost before any actual task begins.

### Classification table
Every candidate below was individually verified against `origin/main`; nothing is asserted from skimming. Given the density of this file (cross-referenced incident numbers, measured figures, and "why the obvious fix is wrong" passages on nearly every line), the honest result is: **one** verified VERBOSE/DUPLICATED cut, **zero** additional STALE claims beyond Part 1, and a KEEP list included so a reviewer can see what was considered and spared, not just what's proposed for the axe.

| Lines | Verdict | Detail |
|---|---|---|
| 998-1000 vs. 1110-1112 | **DUPLICATED/VERBOSE** | `posix-lint.sh`'s bullet: *"That suite runs in `linux.yml`, **not** in test.yml's `tools/lib/*_test.sh` loop — it needs a linter the macOS image lacks, and the loop skips it by name for that reason."* `bash-lint.sh`'s bullet, ~110 lines later, restates the identical mechanism nearly verbatim: *"That suite runs in `linux.yml`, not in test.yml's `tools/lib/*_test.sh` loop — it needs a linter the macOS image lacks, and it is the loop's SECOND skip argument."* Both are literally true (`posix-lint_test.sh` and `bash-lint_test.sh` are each skipped from the same macOS-run loop for the same reason: no dash/shellcheck on the runner), and the "why they're both in `linux.yml`" reasoning right above the second occurrence (lines 1096-1102) is correctly *not* duplicated — it's genuinely a different reason (dash-must-be-real-runtime vs. shellcheck-is-static). Only the trailing "runs in linux.yml, not in the macOS loop, because the macOS image lacks the linter" clause repeats. **Proposed shorter wording for 1110-1112**: *"That suite runs in `linux.yml` for the same reason posix-lint's does (above) — it is the loop's SECOND skip argument."* Saves ~12 words, loses nothing (the reader is 110 lines into the same section and already has the mechanism). |
| 420-424, 667-669 | **STALE** | See Part 1 above. |
| 248-276 (branch protection: ARS/CodeScene "not required by branch protection") | KEEP | Looks like duplication (the phrase "not required by branch protection" appears twice, once per bullet) but the file explains itself: *"What IS required is stated here positively, because 'not required by branch protection' appears twice in this section and nothing said what is — the sentence #1432 filed as the one creating the false impression."* This is a deliberate, self-justified repetition fixing a real prior incident (#1432) — cutting it would reintroduce the ambiguity #1432 was filed about. |
| 116-136 vs. 1595-1598 (the "sourced-figure" rule, quoted twice) | KEEP | First occurrence defines the rule with its own evidence (#1726, the PR #1724 "twelve days not seven months" and "stalled seven times vs. five" violations). Second occurrence in "Replay's measured figures" *quotes* the rule verbatim in quotation marks while applying it, and explicitly says "rather than repeating either here" right after — the file is already avoiding duplication on purpose here. Quoting a named rule when invoking it is good practice, not bloat. |
| 104-106 vs. 1000-1004 (posix-lint's `ALL PASS` incident) | KEEP | First is a one-clause forward pointer ("...(below)"); second is the full narrative. Forward-reference-then-detail is standard technical writing, not restatement — cutting either loses either the summary-level index or the actual incident narrative. |
| "committed but never pushed" — #1726 (lines 2430-2439, subagent not woken by its own backgrounded job) vs. #1570 (lines 2443-2447, `git push \| tail` masks the real exit status) | KEEP | These share a surface phrase ("committed but never pushed") but are two *different* incidents with two different root causes and two different fixes (don't background the push, vs. don't read exit status through a pipe). This is exactly the "two obligations that share vocabulary" trap flagged as a risk — confirmed as NOT a duplicate by reading both incidents fully; recorded here to show the check was made, not skipped. |
| ~59-925 (the seven contract-testing families: permission gating, hook endpoints/disclosure/path-confinement/version-floors/unrecognized-events/receipts, managed user files, guarded construction) | KEEP | ~870 lines, but every sentence maps to a specific incident number (#1356 through #1520+) and a specific mechanism; this is the density the "dismissals carry evidence" rule and the mutation-testing rules exist to produce. No safe compression found — every clause that reads as a restatement on a skim turned out, on verification, to be a distinct obligation, a distinct incident, or a distinct measured number. |
| ~1769-2369 (macOS snapshot-testing host-dependency archaeology: scale, locale, timezone, `@AppStorage`, wall clock) | KEEP | ~600 lines, the single densest section in the file — but essentially every paragraph is a "the obvious fix was wrong, and here's the measurement that proved it" passage (e.g. #1034/#1044 re-recording a bug that was actually an appearance-mode defect; #1509's tolerance widening that also absorbed a missing architecture segment; `.environment(\.locale, …)` on `TextField` provably not working). These are explicitly the highest-value lines per the task's own ground rules and are excluded from any cut proposal. |
| 879-925 (probe costs in doc comments — `core/internal/costreport`) | KEEP | Dense, but each clause pins a distinct measured pitfall (LOAD explaining a 4x doc-comment discrepancy; `41/3209` rounding to 0 in Go; `tmux.list_clients` refusing because probing a dead socket starts a server). No restatement found. |

**No additional STALE claims found.** I spot-verified every "current state" assertion I judged likely to drift (not just symbol existence): `tools/preflight.sh`'s `VALID_GROUPS` array (`go web arch tools skills posix bash security swift linux`) matches AGENTS.md's own group list exactly; the three POSIX-lint-scoped scripts (`site/install.sh`, `tools/linux-replay-entrypoint.sh`, `tools/git-hooks/shim`) all still open with `#!/bin/sh` on `origin/main`; both web trees (`platforms/web/`, `tools/onboarding-factory/internal/viewer/web/`) still exist independently; the macOS CI claim (`.github/workflows/macos-swift.yml` now runs `swift test` as a second job per #1530) is internally consistent with the rest of that section.

### Bottom line
This file does not have a verbosity problem in the usual sense — it has almost no filler to cut. The one genuine, low-risk trim is the `bash-lint`/`posix-lint` duplicate sentence above. Everything else that looks compressible on a skim turned out, on verification, to be load-bearing: a distinct incident, a distinct measured figure, or text the file itself already flags as a deliberate, justified repetition. If there's a bigger win available here, it's probably structural rather than line-level — e.g. whether the ~600-line macOS snapshot-testing archaeology (1769-2369) or the ~870-line contract-testing-families section (59-925) belong in a separate linked doc rather than in the file injected into every session — but that's a different, bigger discussion than a cleanup pass, and is flagged here rather than proposed.

ingo-eichhorst:
## Summary

Fixes staleness in `AGENTS.md`'s "Testing" section that four adapter-shipping PRs opened up: #1724 (gemini-cli), #1731 (`ManagedUserFile.Also`), #1732 (kiro-cli), #1733 (mistral-vibe), #1734 (raw-bytes hook-endpoint seam). Closes #1730.

**Part 1 — the beacon-delivery claim (the issue's named item, independently re-verified against code):**
- "No adapter has adopted beacon delivery yet, so `hook_endpoint_addressfree_test.go` is the reference wiring" is false as of #1724/#1732: `geminicli`, `kirocli` **and** `vibe` all declare `DeliveryAddressFree` for real.
- #1734 also added two sibling scaffolding fixtures (`hook_endpoint_flat_test.go`, `hook_endpoint_toml_test.go`) for the new `EntriesOf`/`ReadEntries`/`EndpointOfRaw` seams kiro-cli and vibe now exercise for real.
- Per the issue's own framing this is a real open question, not a one-line fix: `hook_endpoint_addressfree_test.go`'s `correctBeaconInstaller` is still load-bearing (`hook_endpoint_selftest_test.go`'s mutation corpus drives against it directly, with no real-adapter equivalent wired in), while the flat/toml siblings have no such tether and are ripe for their own headers' reduce-or-delete clause. Documented as unresolved, not decided.

**Adapter lists/counts re-derived and corrected**, each verified with `git grep`/`find` against the code rather than assumed: hook-receiver count (4 → 7), `hookport_test.go`/`hookdisclosure_test.go`/`hookpath_test.go` adapter lists (3-4 → 6), `DeliveryURL`/`DeliveryAddressFree` split (spelled out, and "declared by" corrected to "left unset by" for the zero-value route), hook version-floor adopters ("a third adapter joins" → six of six). Each count now carries the exact command that reproduces it, per this repo's own sourced-figure rule (#1727) and this task's instruction to apply that rule to AGENTS.md itself — so the next adapter's PR makes the doc re-derivable instead of re-stale.

**`agent.ManagedUserFile`'s field list** (was "Path + Uninstall", now also documents `Verify`/`Version`/`Also`) and the "Managed user files" bullet now describe #1731's `Also` field and its two real users (mistral-vibe's `config.toml` gate, kiro-cli's `chat.defaultAgent`), including the `sharedConfigRefusal` update to check `Also` entries too.

**Re-verified and confirmed unchanged:** `grep -c "^func Assert"` over `core/internal/contracttesting/*.go` still returns exactly 10 (7 families + 2 drivers + 1 lock), and the `git grep`-with-pathspec-`*` "+1" discrepancy the file warns about still reproduces exactly as described (pulls in `userblocking/userblocking.go`). No edit needed there.

**Part 2 — conservative simplification:** applied the one genuine, low-risk cut identified by the issue's own prior audit (the near-duplicate bash-lint/posix-lint "runs in `linux.yml`, not the macOS loop" sentence is now a one-line cross-reference). No other cuts made — the rest of the file's density maps to distinct incidents/mechanisms/figures on inspection, consistent with that audit.

## Process note

This PR went through an unusually chaotic drafting process — a background research agent exceeded its "investigation only" scope and committed/pushed/opened this PR autonomously; a second one briefly committed an inconsistent partial edit. Both were caught, reviewed line-by-line against source, corrected, and consolidated into the two commits on this branch. Every factual claim below was independently re-verified against `git grep`/`git show`/direct source reads, not taken from any agent's self-report. Full account in the issue/task report.

## Test plan
- [x] `tools/preflight.sh --only skills` — PASS
- [x] `tools/preflight.sh --only posix` — PASS
- [x] Pre-push hook (`tools/preflight.sh --changed --budget 540`) — PASS (gofmt; all doc-only diff gates SKIP as expected)
- [x] Every adapter-count/list claim re-derived from `git grep`/`find`/`git show` against current `origin/main`, not copied from the issue body
- [ ] Human review: does the Part 1 "open question, not resolved here" framing match what the maintainer wants documented (vs. actually deciding delete/shrink/keep for the three reference-wiring fixtures)?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -317,7 +317,9 @@ Before marking a ticket done, run the full suite — every layer must pass:
     defect test; after #1488 the chokepoint drops the same payload, so every
     dispatch-shaped assertion stays green (measured on all seven receivers —
     claudecode's hooks and statusline, codex, copilot, geminicli, kirocli,
-    vibe). It is not silent about
+    vibe; reproduce the count with `git grep -rl hookjson.RequireConsent
+    core/adapters/inbound/agents/` rather than trusting this list once an
+    eighth adapter ships). It is not silent about
     it: reaching the backstop means a receiver skipped a check or consent was
     revoked mid-request, so it logs an error, where a receiver's own gate
     answers a quiet 200. That difference is both the surviving discriminator
@@ -397,16 +399,19 @@ Before marking a ticket done, run the full suite — every layer must pass:
   on another port is rewritten in place rather than duplicated, and uninstall
   is not port-scoped (#1178). A new hook-installing adapter wires one call
   (see `hookport_test.go` in `claudecode`, `codex`, `copilot`, `geminicli`,
-  `kirocli` or `vibe` — six today) instead of porting a test file. It grades
-  against the delivery route the adapter DECLARES
+  `kirocli` or `vibe` — six today, reproduced with `find
+  core/adapters/inbound/agents -name hookport_test.go`) instead of porting a
+  test file. It grades against the delivery route the adapter DECLARES
   (`HookInstaller.Delivery`, #1453), because there are two ways to satisfy it
   and the second makes the first unsatisfiable. `DeliveryURL` — the zero
-  value, declared by `claudecode`, `codex` and `copilot` — is an entry that
-  CARRIES the daemon's address.
+  value, left unset by `claudecode`, `codex` and `copilot` — is an entry
+  that CARRIES the daemon's address.
   `DeliveryAddressFree` is an entry that carries none, because it names the
   `irrlichd hook-post` beacon (`core/pkg/hookbeacon`, #1373), which reads the
   addr file at fire time — `geminicli` (#1724), `kirocli` (#1732) and `vibe`
-  (#1733) all declare it; three of the four port obligations then fail by
+  (#1733) all declare it (`git grep -n "Delivery:
+  contracttesting.DeliveryAddressFree" core/adapters/`, the other half of
+  the six adapters above); three of the four port obligations then fail by
   construction, measured against a working beacon. That is the good outcome
   rather than an exemption: the beacon makes the whole stale-port class
   INEXPRESSIBLE instead of fixing it once more — the dev daemon that left a
@@ -465,7 +470,8 @@ Before marking a ticket done, run the full suite — every layer must pass:
   user's `settings.json` undisclosed. Adapters now derive both the count and
   the list from `installedHookEvents` via `hookjson.EventList`, and wire one
   call (see `hookdisclosure_test.go` in `claudecode`, `codex`, `copilot`,
-  `geminicli`, `kirocli` or `vibe` — six today). The "names no
+  `geminicli`, `kirocli` or `vibe` — six today, reproduced with `find
+  core/adapters/inbound/agents -name hookdisclosure_test.go`). The "names no
   uninstalled event" arm checks against `session.AllHookEvents`, itself kept
   honest by `TestAllHookEvents_CoversEveryConstant`, which scans
   `hook_signal.go`'s source rather than trusting a second hand-kept list.
@@ -508,7 +514,8 @@ Before marking a ticket done, run the full suite — every layer must pass:
   plant a broken link, have it accepted, then create the target. A new
   hook-receiving adapter wires one call (see `hookpath_test.go` in
   `claudecode`, `codex`, `copilot`, `geminicli`, `kirocli` or `vibe` — six
-  today; claudecode wires it twice, because its statusline
+  today, reproduced with `find core/adapters/inbound/agents -name
+  hookpath_test.go`; claudecode wires it twice, because its statusline
   endpoint is a receiver too and was the one the original fix forgot).
   **A confinement refusal answers 2xx** — it is reported by the log and the
   counter, never by a status code on the user's critical path. The path is
@@ -570,7 +577,10 @@ Before marking a ticket done, run the full suite — every layer must pass:
   the floor the same way — `copilot`, `geminicli`, `kirocli` and `vibe`
   joined Codex and Claude Code by declaring
   `Version: &agent.VersionGate{Min: "x.y.z", Probe: []string{"<cli>",
-  "--version"}}` and nothing else. `TestEveryHookInstallDeclaresAVersionFloor`
+  "--version"}}` and nothing else, six of six hook-installing adapters
+  today (`git grep -l "VersionGate{" core/adapters/inbound/agents/*/agent.go`)
+  — re-run rather than trust this list once a seventh joins.
+  `TestEveryHookInstallDeclaresAVersionFloor`
   (`core/adapters/inbound/agents/hookversion_test.go`) walks the registry
   projection so a new hook adapter is covered by existing rather than by
   remembering to wire the contract; it narrows on
@@ -655,8 +665,9 @@ Before marking a ticket done, run the full suite — every layer must pass:
   undeclared write (`PermissionService.sharedConfigRefusal` and
   `agents.ManagedUserFiles`, both below) resolve every `Also` entry with the
   same rigor as `Path`, because an unprotected secondary write is exactly
-  #1449's incident shape one field over. Three projections read `Writes`, and they read deliberately different
-  slices: `agents.ManagedUserFiles` returns everything — what
+  #1449's incident shape one field over. Three projections read `Writes`,
+  and they read deliberately different slices: `agents.ManagedUserFiles`
+  returns everything — what
   `irrlichd --print-managed-files` prints and the onboarding recorder backs up
   before spawning a `grant-all` daemon against the user's real `$HOME` — while
   `agents.HookConfigs` narrows to `agent.HooksPermissionKey`, so
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -315,18 +315,19 @@ Before marking a ticket done, run the full suite — every layer must pass:
     NOT reproduce: silence.** Deleting a receiver's own `transcripts` gate used
     to fail `…/transcripts/gated_on_the_named_key` plus that adapter's #1466
     defect test; after #1488 the chokepoint drops the same payload, so every
-    dispatch-shaped assertion stays green (measured on all four receivers —
-    claudecode's hooks and statusline, codex, copilot). It is not silent about
+    dispatch-shaped assertion stays green (measured on all seven receivers —
+    claudecode's hooks and statusline, codex, copilot, geminicli, kirocli,
+    vibe). It is not silent about
     it: reaching the backstop means a receiver skipped a check or consent was
     revoked mid-request, so it logs an error, where a receiver's own gate
     answers a quiet 200. That difference is both the surviving discriminator
     and a real user-facing property — an ordinary denied session must not
-    collect an error line per tool call — so each of the four receivers' hand-
+    collect an error line per tool call — so each of the seven receivers' hand-
     written consent tests now asserts **no error was logged**, and each was
     seen red again with its gate deleted. Statusline is the receiver to watch
     here: it declares ONE permission and keeps no second gate, so that
     assertion is the whole of its live per-adapter proof.
-  - Beside those four, the coverage is one shared proof plus one lock per
+  - Beside those seven, the coverage is one shared proof plus one lock per
     adapter: `hookjson/consent_test.go`'s committed `forgetfulReceiver`
     (declares two keys, checks one) grades the backstop for every receiver at
     once, and `AssertDeclaredPermissions` / each adapter's
@@ -390,18 +391,22 @@ Before marking a ticket done, run the full suite — every layer must pass:
   chokepoint move that would remove that remaining act of memory).
 - Hook endpoints: `contracttesting.AssertHookEndpointFollowsBindAddr`
   (`core/internal/contracttesting/hook_endpoint.go`) is the same kind of
-  runtime obligation for adapters that install hooks into a JSON config —
+  runtime obligation for adapters that install hooks into a shared config
+  file (JSON, or since #1734 TOML) —
   an install writes the resolved port not `:7837`, an entry left by a daemon
   on another port is rewritten in place rather than duplicated, and uninstall
   is not port-scoped (#1178). A new hook-installing adapter wires one call
-  (see `claudecode`/`codex`/`copilot` `hookport_test.go`) instead of porting a
-  test file. It grades against the delivery route the adapter DECLARES
+  (see `hookport_test.go` in `claudecode`, `codex`, `copilot`, `geminicli`,
+  `kirocli` or `vibe` — six today) instead of porting a test file. It grades
+  against the delivery route the adapter DECLARES
   (`HookInstaller.Delivery`, #1453), because there are two ways to satisfy it
-  and the second makes the first unsatisfiable. `DeliveryURL` — the zero value,
-  and all three adapters above — is an entry that CARRIES the daemon's address.
+  and the second makes the first unsatisfiable. `DeliveryURL` — the zero
+  value, declared by `claudecode`, `codex` and `copilot` — is an entry that
+  CARRIES the daemon's address.
   `DeliveryAddressFree` is an entry that carries none, because it names the
   `irrlichd hook-post` beacon (`core/pkg/hookbeacon`, #1373), which reads the
-  addr file at fire time; three of the four port obligations then fail by
+  addr file at fire time — `geminicli` (#1724), `kirocli` (#1732) and `vibe`
+  (#1733) all declare it; three of the four port obligations then fail by
   construction, measured against a working beacon. That is the good outcome
   rather than an exemption: the beacon makes the whole stale-port class
   INEXPRESSIBLE instead of fixing it once more — the dev daemon that left a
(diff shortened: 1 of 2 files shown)


ingo-eichhorst:
## Review — substance is right, form re-creates the defect

The staleness findings are correct and I verified the ones I could independently. Two things need changing before this lands.

### 1. It replaces stale rosters with freshly stale rosters

This is the important one. The diff swaps *"all four receivers"* → *"all seven receivers"*, *"`claudecode`/`codex`"* → *"six today"* (four places), and *"a third adapter joins"* → *"`copilot`, `geminicli`, `kirocli` and `vibe` joined Codex and Claude Code"*.

Every one of those is a hand-maintained count that is wrong the moment adapter seven ships — which is exactly how the text being fixed here went wrong. This file carries its own rule against it, in **"Replay's measured figures"**:

> a number which documents behaviour but is not produced by it drifts silently, and is then quoted with full confidence

plus #1727's generalisation that a figure names the command that produces it or is marked as an estimate. None of the new counts do either.

The version-floor sentence is the clearest case, because the fix is sitting in the next clause:

```
…every hook-installing adapter now declares the floor the same way — `copilot`,
`geminicli`, `kirocli` and `vibe` joined Codex and Claude Code by declaring
`Version: &agent.VersionGate{…}` and nothing else.
`TestEveryHookInstallDeclaresAVersionFloor` … walks the registry projection so a
new hook adapter is covered by existing rather than by remembering.
```

The roster is enumerated immediately before naming the registry-walking tripwire that makes enumerating it unnecessary. Drop the list: *"every hook-installing adapter declares one — `TestEveryHookInstallDeclaresAVersionFloor` walks the registry projection, so this is enforced rather than maintained."* A mechanically-enforced rule does not need a roster.

Preference order for the rest:

1. **Point at the enforcing tripwire instead of enumerating** — as above.
2. **If a number genuinely helps, keep it and inline the command that produces it** (`git grep -rl "AssertHookPathConfined" core/adapters/inbound/agents/`), so the next reader re-derives instead of trusting.
3. **Only enumerate where the list itself is load-bearing.** The `DeliveryURL` vs `DeliveryAddressFree` split qualifies and should stay as written — which route an adapter took is information, not a tally. That part is good.

### 2. Net +52 lines on an issue whose second half is "remove verbosity and duplication"

`80 insertions, 28 deletions`. The only real simplification in the diff is the `linux.yml` skip-argument dedup at the end, which is a good catch. The Part 2 pass appears not to have run. Either do it, or say plainly in the PR body that it was not done and why — an unstated omission is the thing this repo's conventions are most hostile to.

### What is genuinely good here, and should survive a rewrite

- The **three scaffolding fixtures** paragraph is the best thing in the diff. Finding that `hook_endpoint_addressfree_test.go`, `hook_endpoint_flat_test.go` and `hook_endpoint_toml_test.go` all carry the same verbatim sunset clause, that real adapters have now met each file's own stated condition, and that only the first has a real tether (it is what `hook_endpoint_selftest_test.go`'s `DeliveryAddressFree` mutation corpus drives against) — with the honest note that this is *"a real justification but not the one the header states"* — is exactly the kind of distinction that is expensive to rediscover. Keep it, and keep it framed as an open question.
- The `Also` documentation is accurate; I verified `vibe/agent.go` and `kirocli/agent.go` both set it, and that `sharedConfigRefusal` resolves `Also` entries.
- The `DeliveryURL`/`DeliveryAddressFree` attribution is correct — vibe uses `DeliveryAddressFree`, not `DeliveryURL`.

### Independently verified

- `grep -c "^func Assert"` over `core/internal/contracttesting/` returns **10**, and the seven-families / three-non-families split (`AssertPermissionGatedOnEachKey`, `AssertHookReceiverPermissionGated`, `AssertDeclaredPermissions`) is exactly as the file describes. Those figures are **still accurate** and correctly left alone.

### One finding worth promoting into the PR body

The report behind this PR noted that copilot's `hookdisclosure_test.go`, `hookpath_test.go` and `VersionGate` all date to #1436 (2026-08-10) — so several of these rosters were **already stale before the four PRs #1730 names**. That means #1730's mechanical audit checks symbol/path existence but not count accuracy: a gap in the audit *method*, not just in the file. That belongs in the PR description, because it is the argument for preference (1) above.

ingo-eichhorst:
## Review resolved — consolidated

My review asked for the corrected adapter rosters to be replaced rather than refreshed. `0c926e9e` applied the second-best option (pair each roster with the command that reproduces it); `e2d75e17` applies the better one where it was available. Final shape:

| Place | Resolution |
|---|---|
| Version floor | Roster **and** grep dropped — the bullet already named `TestEveryHookInstallDeclaresAVersionFloor`, which walks the registry projection. A test that fails on its own beats a command a reader must remember to run. |
| `hookport` / `hookdisclosure` / `hookpath` | Roster + `find` → "any hooks-declaring adapter's `<file>_test.go`". The point was only ever "here is a shape to copy", never a tally. |
| Receiver count | **Kept** as `0c926e9e` wrote it — no tripwire enforces that set, so pairing it with `git grep` is the right rule. |
| `DeliveryURL` vs `DeliveryAddressFree` | **Kept enumerated.** Which route an adapter took is load-bearing information, not a count. |

Also kept from `0c926e9e`: its correctness fix. `DeliveryURL — the zero value, declared by claudecode, codex and copilot` was wrong — those three **leave the field unset**, which is exactly what makes it the zero value. Neither the first pass nor my review caught that.

Net effect of the consolidation commit: **−9 lines**, pulling back some of the growth on an issue whose second half asks for less text.

### One defect fixed in code, not just docs

`statusline_test.go` said `Foreign` is "empty at the three hook receivers" when seven wire `hookjson.RequireConsent` today — the same stale-roster defect one layer down.

### Filed rather than fixed here

Three of the seven contract families — path confinement, bind-addr, disclosure — have **no** registry tripwire, unlike `Verify` and `Version`. So "point at the enforcer" was unavailable for them, which is itself a finding: a new adapter can ship without wiring them and every gate stays green. That is the failure mode that shipped the statusline endpoint unconfined. Filed as #1740 with the evidence.

### Process note

The work behind this PR ran into a multi-writer collision: several agents shared one worktree and one branch, and commits were reset over each other — including one of mine. Nothing was lost (everything is in the branch history), but the coordination hazard is real and worth knowing about. That is my error for editing in a worktree an agent was using, not a fault in anyone's work.

PR #1739, opened 2026-08-21
fix(rig): declare kiro-cli's third managed file; make agent-home isolation one declaration

ingo-eichhorst:
## What this started as, and why it changed

The brief was "the rig cannot record a hooks cell for kiro-cli or mistral-vibe without writing into the operator's real `~/.kiro` / `~/.vibe`". **That premise is wrong and I verified it rather than implementing it.** The rig already protects those files, hard:

- `go run ./core/cmd/irrlichd --print-managed-files` lists all four (`~/.kiro/agents/irrlicht.json`, `~/.kiro/settings/cli.json`, `~/.vibe/hooks.toml`, `~/.vibe/config.toml`).
- `lib/managed-file-snapshot.sh` asks the daemon for that list (no literal list, no fallback), backs each file up, and restores from an EXIT trap — handling `saved` / `absent` / `absentdir` and refusing to destroy a file the agent created mid-run.
- `spawn-record-daemon.sh:112` makes it a hard gate: a failed snapshot refuses the spawn.

So nothing here is motivated by data loss. Two real things remained.

---

## 1. kiro-cli writes a THIRD file, and it is undeclared

`recordPriorDefaultAgentOnce` (`kirocli/hookinstaller.go:573`) writes `$KIRO_HOME/.irrlicht-prior-default.json` on every install. It is in neither `Writes.Path` nor `Writes.Also`, so `--print-managed-files` never names it and the rig never backs it up or hands it back.

The residue is not inert: that function **deliberately never overwrites**, so the copy a recording leaves in the operator's real `~/.kiro` is what a LATER real grant reads instead of their actual `chat.defaultAgent` — and a subsequent uninstall then restores the recording's snapshot of that setting rather than their own choice.

Fixed by declaring it in `Writes.Also`, the mechanism #1718 added for exactly this.

**Seen red before the fix** (`TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile`):

```
--- FAIL: TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile (0.00s)
    kirocli_alsofile_test.go:168: Writes.Also declares no entry resolving to ".../kiro-home/.irrlicht-prior-default.json" — recordPriorDefaultAgentOnce writes it on every install, so the recording rig cannot back it up or hand it back
    kirocli_alsofile_test.go:177: --print-managed-files does not list ".../kiro-home/.irrlicht-prior-default.json" — the recording rig protects only what this command names
FAIL	irrlicht/core/cmd/irrlichd	0.420s
```

---

## 2. The home-isolation lever was half-wired, and failed silently

`run-cell.sh` exports the home variable. The daemon inherits it (direct child). The agent CLI does **not**: the driver launches it with `tmux new-session`, and the pane's command is spawned by the tmux SERVER, whose environment was fixed when that server started.

Measured on a private `-L` socket, so no real session was touched:

```
pane saw IRR_PROBE_VAR = [<UNSET>]
pane saw IRR_PROBE_VAR = [isolated-home] (env-prefixed)
```

On a dev machine a server is essentially always already running, so setting `KIRO_HOME`/`VIBE_HOME` today gets you a daemon watching the scratch home and a CLI writing to the real one — an empty fixture with nothing saying why. codex's driver already prefixed `env CODEX_HOME=…`; that was one driver's private knowledge.

**copilot had the same hole, on the one adapter that DEFAULTS to a staging home** — i.e. on every copilot recording. Found by the new tripwire on its first run (see mutation evidence below).

### The mechanism

`tools/onboarding-factory/scripts/lib/agent-home.sh` — one table, one function:

```
copilot COPILOT_HOME default
codex CODEX_HOME optin
kiro-cli KIRO_HOME optin
mistral-vibe VIBE_HOME optin
```

It replaces per-adapter blocks in `run-cell.sh` (copilot, codex) and `run-cell-multi.sh` (copilot **only** — codex could not be isolated in the cross-adapter rig at all, the #1214 one-rig-only drift). Copilot's staging default is preserved byte-for-byte (`$STAGING/copilot-home`). It always prints one line, including for "no row" and "opt-in, unset", so "did nothing" and "worked" are not both silent.

### Why kiro-cli and mistral-vibe are opt-in, not copilot-shaped

The rule is **not** "credentials" — that is codex's phrasing and it is too narrow. It is that a scratch home starts without state the run needs, and every way of missing it ends in a blocking interactive prompt in the tmux pane that no driver has an arm for.

- **mistral-vibe — the provider declaration, upstream of the credential.** `config.toml`'s `api_key_env_var` reads like "the key is in an env var, so a scratch home loses nothing" — true of the SECRET, false of the LOOKUP. `[[providers]]` lives in `$VIBE_HOME/config.toml`, and `resolve_api_key(env_key)` (`vibe/core/config/_settings.py:101`) reads the process env then the OS keyring **under that name**. A home with no `config.toml` falls back to `DEFAULT_PROVIDERS`, so any non-default provider looks up the wrong name → `MissingAPIKeyError` → `cli.py:92` runs the onboarding **wizard** in the pane. Separately `$VIBE_HOME/.env` is a real credential location: `persist_api_key` writes there when the keyring raises.
  **Not** `trusted_folders.toml`, which looks like it should matter and does not: the driver already passes `--trust` on every launch because the staging cwd lives under the irrlicht repo and is untrusted whichever home is in use.
- **kiro-cli — persisted trust-all consent.** Its token is at `~/.aws/sso/cache/kiro-auth-token-cli.json`, which follows `$HOME` not `$KIRO_HOME`, and the real `~/.kiro` holds no credential material. So the credential rule says "safe to default" — and the answer is still no, because `settings/cli.json`'s `chat.disableTrustAllConfirmation` is the consent the kiro driver states it depends on ("kiro-cli launches with persisted --trust-all-tools consent, so a fresh cwd never stalls on a trust picker", `step_restart`). Whether it actually blocks is **not verified** — that needs driving the live CLI — so the row is opt-in on the strength of the driver's premise no longer holding, stated as such.

A third "default AND seed from the real home" policy was considered and not built: it trades an explicit operator step for an implicit read of their real home, and cannot be tested without a live CLI.

---

## 3. The tripwire: `tools/onboarding-factory/internal/righome`

Placed in the factory module, not `core/`, because it is a property of the RECORDING RIG and the factory is already the tree that joins the daemon's adapter registry to the catalog (`internal/hookcov` is the precedent, same import, same shape). A `core/` test reading a file under `tools/` would invert that.

Three arms. Every one was seen red.

**A. A hooks adapter with neither a row nor a named exemption** (mutation: delete kiro-cli's row):

```
--- FAIL: TestRigHomeTableMatchesTheAdapterRegistry (0.02s)
    righome_test.go:61: adapter "kiro-cli" installs hooks into the user's config but has neither a row in tools/onboarding-factory/scripts/lib/agent-home.sh nor an Unisolatable entry saying why it cannot have one. Add whichever is true — an isolation story is a decision, and an omission looks exactly like a decision that was made silently
```

**B. A row whose variable does not actually relocate both halves** (mutation: `kiro-cli KIRO_HOME optin` → `KIRO_HOME_WRONG`):

```
--- FAIL: TestEveryRigHomeRowRelocatesBothHalves/kiro-cli (0.00s)
    righome_test.go:130: with KIRO_HOME_WRONG=... the session root is still ".kiro/sessions/cli" — the daemon would watch the operator's real store while the driver wrote to the scratch home, and the recording would contain their sessions and not the driver's
    righome_test.go:164: with KIRO_HOME_WRONG=... the hooks install still writes Writes.Path to "/Users/ingo/.kiro/agents/irrlicht.json" — the install would land in the operator's real config while everything else in the run claimed to be isolated
    righome_test.go:164: ... Writes.Also[0] to "/Users/ingo/.kiro/settings/cli.json" ...
    righome_test.go:164: ... Writes.Also[1] to "/Users/ingo/.kiro/.irrlicht-prior-default.json" ...
```

(Note `Also[1]` — the file this PR declares. The two halves of this PR grade each other.)

**C. A rowed adapter whose driver does not pass the home through tmux.** This was **not** a mutation — it is what the arm reported against `main` on its first run:

```
--- FAIL: TestEveryRigHomeRowsDriverPassesTheHomeThroughTmux/copilot (0.00s)
    righome_test.go:237: .../replaydata/agents/copilot/driver-interactive.sh:201 launches the agent under tmux without an explicit `env COPILOT_HOME=…` prefix.
      tmux new-session -d -s "$SESSION" -x 200 -y 50 -c "${SES_CWD[$ACTIVE]}" copilot "${args[@]}" \
    The pane inherits the tmux SERVER's environment, not the exported one, so on any machine with a server already running the daemon would watch the scratch home while the CLI wrote to the real one.
```

Fixed in the same PR.

### Exemptions carry structural facts, not to-dos

- **claudecode** — HALF-relocatable, which is worse than not at all: transcripts follow `CLAUDE_CONFIG_DIR` but `claudeSettingsPath` joins `os.UserHomeDir()` with `.claude/settings.json` unconditionally. A row would move the session store while the install kept landing in the real `settings.json`.
- **gemini-cli** — no override exists at all: `defaultRootDir` is `.gemini/tmp` under `$HOME`, `geminiHome()` is `$HOME/.gemini`, no `os.Getenv` anywhere in the package.

Both keys are existence-checked in both directions.

---

## Mutation evidence for the shell lib

Committed beside the assertions (`agent-home_test.sh`), and each also run live:

**Drop the `export`** (assign locally instead):
```
  FAIL: unset → exported to the staging default — expected [.../cop] got []
  FAIL: an unexported caller variable is exported onward to the daemon — expected [.../codex-noexport] got []
agent-home_test.sh: 2 FAILED
```
Recorded honestly: the first draft reddened only ONE arm, because `run_isolate` hands the value in through `env VAR=…` so the child already carries it exported. The second arm was added to make the mutation discriminate for the opt-in rows too.

**Remove the absolute-path guard:**
```
  FAIL: RELATIVE → hard failure — expected [1] got [0]
  FAIL: relative → names why — expected ['must be absolute'] got [agent-home: codex — CODEX_HOME=relative/codex (daemon + driver share this home) ]
  FAIL: kiro-cli relative → hard failure — expected [1] got [0]
  FAIL: mistral-vibe relative → hard failure — expected [1] got [0]
agent-home_test.sh: 4 FAILED
```

`righome_corpus_test.go` carries the rest as a committed corpus: one input wrong in exactly one way per row, each naming the fragment it must report **and** the fragments it must leave silent, plus a correct-baseline vacuity guard and a `want:false` row (a row for a hookless adapter is legitimate).

---

## Two corrections to my own work, recorded rather than quietly fixed

Both are the shape this repo cares about — a check that looked like it was working and was not — so they are here rather than only in a handover note.

**The shell test helper mis-parsed its own output and reported 15 phantom failures.** `run_isolate`'s first draft packed status, exported value and output into one `\037`-delimited string and split it with `sed -n`. The reported output is multi-line by design (the relative-path refusal prints four lines), so the field split silently mis-assigned every column and the suite reported 15 failures that were all its own. It now writes three separate files. The comment saying so is committed above the helper, because a test harness that can lie about its subject is worth a warning label.

**The "drop the `export`" mutation first reddened only ONE arm, and the weaker evidence was nearly shipped.** `run_isolate` hands the value in through `env VAR=…`, so the child already carries it exported and re-exporting is a no-op — meaning every opt-in row's `set → exported` assertion passes against a function that never exports anything. Only copilot's staging default, where the function introduces a previously-unset variable, discriminated. Rather than record "the mutation reddens one arm", I added the arm that makes it discriminate for the opt-in rows too: an operator who sets the variable as a plain shell variable and runs the rig in the same shell. The mutation now reddens two, and the measurement is committed beside the assertion.

## Something the test caught that is worth knowing

Three of the four rowed adapters (copilot, codex, kiro-cli) compute `Source.Dir` at `Agent()` **construction** time, so the watched root is frozen when the registry is built; only vibe's `DirFunc` re-resolves lazily. Building the registry before setting the variable reported all three as "does not relocate" — true of the object, false of the daemon. **That is why `run-cell.sh` must export before `spawn_record_daemon` and not at any convenient point: a daemon already running cannot be relocated.** The test now models the daemon's real ordering and says so.

---

## SonarCloud: one real finding, fixed

The PR's quality gate failed on exactly one condition — `new_security_rating` = B, required A — driven by a single new VULNERABILITY:

```
VULNERABILITY  MINOR  go:S4036  tools/onboarding-factory/internal/righome/righome.go:98
  Make sure the "PATH" variable only contains fixed, unwriteable directories.
```

**Real, and fixed properly rather than suppressed** (NOSONAR does not work for Go in this repo). `Table` ran `exec.Command("bash", ...)` by bare name, so a local attacker who can write to a directory earlier on the inherited PATH chooses the interpreter that reads the very declaration this package exists to trust. `core/pkg/pathutil` is this repo's own answer to that rule — its package doc names `go:S4036` — and `kirocli/kiroctl.go` already documents why PATH-prepending cannot defeat it. Now `pathutil.MustResolve("bash")`, which falls back to the bare name on an unusual layout so the check degrades to today's behaviour rather than failing to run.

The other ten findings were `go:S3776` (cognitive complexity) and two `shelldre` shell rules; none affected the gate (`new_maintainability_rating` passed at A). CodeScene independently flagged the same functions as Bumpy Road / Complex Method. Since they were all new code in this PR, I split them rather than suppressing:

- `Reconcile` → `checkRows` / `checkExemptions` / `checkUncovered`, one per question. Kept as three rather than folded into one walk **because the corpus asserts a case wrong in ONE way fires exactly one check** — merging them would have made those silence assertions harder to keep honest.
- `TestEveryRigHomeRowRelocatesBothHalves` → `registryAdapterAfterEnv` / `assertSessionRootRelocates` / `assertHookWritesRelocate`. The extraction also gives the "build the registry AFTER setting the env var" ordering a named home instead of a comment buried mid-function.
- `TestReconcileNamesEveryWrongShape` → `assertReported` / `assertSilentOnEverythingElse`, which names the two halves of the evidence.
- `TestTableRefusesRatherThanReturningAShortList` → `assertTableWasRead` (the vacuity guard) / `assertTableRefused`.
- The tmux scan → `tmuxLaunchesMissingEnv` / `countTmuxLaunches`, deliberately separate so "no launches found" cannot be confused with "all launches OK".
- The two `Also`-by-suffix lookups → named helpers in both test files.
- `agent-home.sh`'s two one-line accessors gained `local` + explicit `return` (shelldre:S7679/S7682).

**Outcome: SonarCloud quality gate is now OK** — all five conditions green (`new_security_rating` A), and the PR's new-code issue count is **0**, down from 11.

**CodeScene still fails and I stopped rather than chasing it.** AGENTS.md is explicit that it is advisory, not a merge gate ("a prompt to look closer, not something to chase to green"). I looked, and the fixes above took it from 7 violations to 5 with one file improving in Code Health (`permission_shared_config_gate_test.go` 9.69 → 9.92) and `Reconcile` dropping off entirely. What it now flags are the **extracted helpers themselves** — `assertHookWritesRelocate`, `kirocliSettingsAlsoResolver`, `assertTableWasRead`, `Table` — because each extraction becomes a method with its own threshold. Going further means splitting a single `for` loop that resolves and checks one list, and splitting a two-clause `if` in a vacuity guard, both of which make the code harder to read to satisfy a metric. That is the point at which the advisory should be left advisory.

**All three detectors were re-run against their mutations AFTER the refactor**, because a refactor that leaves a detector green while breaking it is precisely this PR's subject:

- copilot's `env` prefix removed again → `TestEveryRigHomeRowsDriverPassesTheHomeThroughTmux/copilot` red at `driver-interactive.sh:215`
- `kiro-cli KIRO_HOME` → `KIRO_HOME_WRONG` → `TestEveryRigHomeRowRelocatesBothHalves/kiro-cli` red on the session root and all three written files
- kiro-cli's row deleted → `TestRigHomeTableMatchesTheAdapterRegistry` red

## Gates

`go test ./core/... -race -count=1`, `go test ./tools/onboarding-factory/... -race -count=1`, `of validate`, `smoke-test.sh`, `preflight --only arch|tools|bash|posix`, `replay-fixtures.sh` — all pass. Zero golden drift (`git status --porcelain` clean apart from the intended files). No recording was run and nothing under the user's real home was written.

### DIFF
--- a/core/adapters/inbound/agents/kirocli/agent.go
+++ b/core/adapters/inbound/agents/kirocli/agent.go
@@ -119,15 +119,28 @@ func Agent() agent.Agent {
 						Min:   minCLIVersion,
 						Probe: []string{kiroCLIBinary, "--version"},
 					},
-					// Also: the SAME Apply closure also rewrites
-					// chat.defaultAgent in settings/cli.json
-					// (ensureDefaultAgent, hookinstaller.go) — a second shared,
-					// user-owned file beyond Path (issue #1718's Also field).
-					// Undeclared, that write is invisible to
+					// Also: the SAME Apply closure writes TWO further files
+					// beyond Path (issue #1718's Also field), and both are in
+					// the user's real agent home:
+					//   settings/cli.json          — chat.defaultAgent
+					//                                (ensureDefaultAgent)
+					//   .irrlicht-prior-default.json — what that setting held
+					//                                first (recordPriorDefaultAgentOnce)
+					// Undeclared, either write is invisible to
 					// agents.ManagedUserFiles / --print-managed-files, the
 					// recording rig's snapshot/restore sweep — see
-					// TestKiroCLIAlso_SettingsFileIsAManagedFile
+					// TestKiroCLIAlso_SettingsFileIsAManagedFile and
+					// TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile
 					// (core/cmd/irrlichd) for the mutation evidence.
+					//
+					// The sidecar is irrlicht's OWN document rather than a
+					// file kiro-cli reads, which is why it was missed — but
+					// "shared, user-owned" is about WHERE it lands, not who
+					// authored it, and the residue it leaves is not inert:
+					// recordPriorDefaultAgentOnce deliberately never
+					// overwrites, so a copy a recording left in the
+					// operator's real ~/.kiro is what a LATER real grant
+					// reads instead of their actual chat.defaultAgent.
 					//
 					// What this does NOT change, stated because it would
 					// otherwise be assumed: PermissionService.sharedConfigRefusal
@@ -142,7 +155,7 @@ func Agent() agent.Agent {
 					// not a verdict-changing one. Declared anyway because the
 					// two are independent consumers and the recorder's sweep
 					// is not co-located with anything.
-					Also: []func() (string, error){kiroSettingsPath},
+					Also: []func() (string, error){kiroSettingsPath, priorDefaultStatePath},
 				},
 			},
 		},
--- a/core/application/services/permission_shared_config_gate_test.go
+++ b/core/application/services/permission_shared_config_gate_test.go
@@ -319,17 +319,28 @@ func TestSharedConfigGate_KiroCLIAlsoDeclarationReachesTheGuard(t *testing.T) {
 	realKiroHome := filepath.Join("/Users/someone", ".kiro")
 	t.Setenv("KIRO_HOME", realKiroHome)
 
+	// The settings/cli.json resolver is picked out by what it RESOLVES TO, not
+	// by its index: kiro-cli declares more than one Also entry (the second is
+	// the prior-default sidecar, core/cmd/irrlichd's
+	// TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile), and an index would
+	// silently retarget this test at a different file the day the order moved.
 	var kirocliAlso func() (string, error)
 	for _, p := range kirocli.Agent().Permissions {
-		if p.Key == kirocli.PermissionKeyHooks && p.Writes != nil {
-			if len(p.Writes.Also) != 1 {
-				t.Fatalf("kirocli's hooks permission declares %d Also entries, want exactly 1 (settings/cli.json)", len(p.Writes.Also))
+		if p.Key != kirocli.PermissionKeyHooks || p.Writes == nil {
+			continue
+		}
+		for _, resolve := range p.Writes.Also {
+			got, err := resolve()
+			if err != nil {
+				t.Fatalf("resolving one of kiro-cli's Also entries: %v", err)
 			}
-			kirocliAlso = p.Writes.Also[0]
+			if filepath.Base(got) == "cli.json" && filepath.Base(filepath.Dir(got)) == "settings" {
+				kirocliAlso = resolve
+			}
 		}
 	}
 	if kirocliAlso == nil {
-		t.Fatal("kirocli.Agent() declares no hooks permission with Writes")
+		t.Fatal("kirocli.Agent() declares no hooks permission whose Writes.Also resolves a settings/cli.json")
 	}
 	wantAlsoPath, err := kirocliAlso()
 	if err != nil {
--- a/core/application/services/permission_shared_config_gate_test.go
+++ b/core/application/services/permission_shared_config_gate_test.go
@@ -319,29 +319,7 @@ func TestSharedConfigGate_KiroCLIAlsoDeclarationReachesTheGuard(t *testing.T) {
 	realKiroHome := filepath.Join("/Users/someone", ".kiro")
 	t.Setenv("KIRO_HOME", realKiroHome)
 
-	// The settings/cli.json resolver is picked out by what it RESOLVES TO, not
-	// by its index: kiro-cli declares more than one Also entry (the second is
-	// the prior-default sidecar, core/cmd/irrlichd's
-	// TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile), and an index would
-	// silently retarget this test at a different file the day the order moved.
-	var kirocliAlso func() (string, error)
-	for _, p := range kirocli.Agent().Permissions {
-		if p.Key != kirocli.PermissionKeyHooks || p.Writes == nil {
-			continue
-		}
-		for _, resolve := range p.Writes.Also {
-			got, err := resolve()
-			if err != nil {
-				t.Fatalf("resolving one of kiro-cli's Also entries: %v", err)
-			}
-			if filepath.Base(got) == "cli.json" && filepath.Base(filepath.Dir(got)) == "settings" {
-				kirocliAlso = resolve
-			}
-		}
-	}
-	if kirocliAlso == nil {
-		t.Fatal("kirocli.Agent() declares no hooks permission whose Writes.Also resolves a settings/cli.json")
-	}
+	kirocliAlso := kirocliSettingsAlsoResolver(t)
 	wantAlsoPath, err := kirocliAlso()
 	if err != nil {
 		t.Fatalf("kiro-cli's real Also resolver: %v", err)
@@ -380,4 +358,32 @@ func TestSharedConfigGate_KiroCLIAlsoDeclarationReachesTheGuard(t *testing.T) {
 	if log2.errorMentioning(wantAlsoPath) {
 		t.Errorf("the refusal named %q even though Also was not declared on this permission — it cannot have evaluated a resolver it was never given", wantAlsoPath)
 	}
+}
+
+// kirocliSettingsAlsoResolver extracts kiro-cli's REAL settings/cli.json Also
+// resolver off its own registration.
+//
+// It is picked out by what it RESOLVES TO, not by its index: kiro-cli declares
+// more than one Also entry (the second is the prior-default sidecar,
+// core/cmd/irrlichd's TestKiroCLIAlso_PriorDefaultSidecarIsAManagedFile), and
+// an index would silently retarget the caller at a different file the day the
+// declaration order moved.
+func kirocliSettingsAlsoResolver(t *testing.T) func() (string, error) {
+	t.Helper()
+	for _, p := range kirocli.Agent().Permissions {
+		if p.Key != kirocli.PermissionKeyHooks || p.Writes == nil {
+			continue
+		}
+		for _, resolve := range p.Writes.Also {
+			got, err := resolve()
+			if err != nil {
+				t.Fatalf("resolving one of kiro-cli's Also entries: %v", err)
+			}
+			if filepath.Base(got) == "cli.json" && filepath.Base(filepath.Dir(got)) == "settings" {
+				return resolve
+			}
+		}
+	}
+	t.Fatal("kirocli.Agent() declares no hooks permission whose Writes.Also resolves a settings/cli.json")
+	return nil
 }
--- a/core/cmd/irrlichd/kirocli_alsofile_test.go
+++ b/core/cmd/irrlichd/kirocli_alsofile_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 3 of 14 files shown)


ingo-eichhorst:
## Independent QA — verified

Checked the three load-bearing changes myself rather than accepting the report.

**1. The undeclared write is real, and the fix works end-to-end.** On `main`, `irrlichd --print-managed-files` does not list `.irrlicht-prior-default.json`. On this branch (`cad68d87`) it does:

```
~/.kiro/agents/irrlicht.json
~/.kiro/settings/cli.json
~/.kiro/.irrlicht-prior-default.json    ← new
```

That closes a real hazard rather than a theoretical one: `recordPriorDefaultAgentOnce` deliberately never overwrites, so a copy left behind by a recording is what a later **real** grant reads instead of the operator's actual `chat.defaultAgent` — and restore-on-uninstall then points them at the wrong agent, silently. This also corrects something **I** got wrong: I told this PR's author that all four kiro/vibe files were protected. That was true of the *declared* set, and I wrongly implied the set was complete.

**2. The copilot tmux bug is real and is the most consequential thing here.** On `main`, `COPILOT_HOME` appears only in `run-cell.sh`/`run-cell-multi.sh` and nowhere near the tmux launch, while the comment there claims "daemon + driver share this store". Since `tmux new-session` inherits the tmux *server's* environment rather than the caller's, the daemon ran against the scratch home and the CLI against the real one — and copilot is the one adapter that **defaults** to a staging home, so this affected every copilot recording. Fixed at `driver-interactive.sh:216` with the `env COPILOT_HOME=…` prefix, following the existing codex-driver precedent rather than inventing one.

**3. The SonarCloud vulnerability was fixed properly, not suppressed.** `go:S4036` on `exec.Command("bash", ...)` is genuine — a bare-name interpreter lets anyone who can write to an earlier PATH entry choose the shell that reads the very declaration this package exists to trust. Now `pathutil.MustResolve("bash")`, which is the repo's own answer (`core/pkg/pathutil` names `go:S4036` in its package doc, and `kirocli/kiroctl.go` already uses it). NOSONAR would not have worked for Go here anyway.

**On the refactor.** Splitting functions to satisfy `go:S3776` is exactly the change that can leave a detector green while breaking it — which is this PR's own subject. Re-running all three mutations *after* the refactor rather than relying on the pre-refactor evidence is the right call and is what I would have asked for.

**CodeScene left red deliberately, and I agree.** AGENTS.md calls it advisory — "a prompt to look closer, not something to chase to green." It went 7 violations → 5, and what it flags now are the extracted helpers themselves, each becoming a method with its own threshold. Going further means splitting a single `for` loop and a two-clause `if` in a vacuity guard. That is making code worse to satisfy a metric.

**Filed separately rather than built here:** #1741, the `TestEveryModifyPermissionDeclaresTheFileItWrites` gap — it checks a `Writes` declaration *exists*, not that it *covers* every file `Apply` writes. Now twice-instanced. The measured false-positive rate for the obvious static rule is recorded in the issue so the next attempt does not re-derive it.

PR #1746, opened 2026-08-21
test(hooks): registry tripwire asserting every hooks adapter WIRES its contract families

### LINKED ISSUE #1740: test(hooks): nothing enforces that a hooks-declaring adapter WIRED the path-confinement, bind-addr and disclosure contracts
AGENTS.md already names this hazard, in the hook-path-confinement bullet:

> a contract can fail only for an adapter that WIRED it, so a receiver nobody wired
> it for is invisible to it — which is how the statusline endpoint shipped
> unconfined in the first place.

Two families are protected from it by a registry tripwire. The rest are not, and
nothing says so.

## Measured

Registry-walking tests that exist (`ls core/adapters/inbound/agents/*_test.go`):

```
beaconroute_test.go  hookverify_test.go  hookversion_test.go
managedfiles_test.go maps_test.go
```

Of those, exactly two assert an obligation over every hooks-declaring adapter:

```
$ git grep -h 'func TestEveryHookInstall' -- 'core/**/*_test.go'
func TestEveryHookInstallDeclaresAVerifier(t *testing.T) {
func TestEveryHookInstallDeclaresAVersionFloor(t *testing.T) {
```

And every call site of the three contract families below is inside a single
adapter's own package — none is a registry walk:

```
$ git grep -ln 'AssertHookPathConfined\|AssertHookEndpointFollowsBindAddr\|AssertHookDisclosureMatchesInstalled' \
    -- 'core/**/*_test.go' | grep -v 'agents/[a-z]*/'
# only core/internal/contracttesting/*_selftest_test.go and the reference fixtures,
# which grade the ASSERTIONS, not whether an adapter wired them
```

## What that means concretely

A new adapter can declare a hooks permission, install hooks into a user's config,
receive them on an unauthenticated local endpoint — and ship without wiring
`AssertHookPathConfined`, `AssertHookEndpointFollowsBindAddr` or
`AssertHookDisclosureMatchesInstalled`. Every gate stays green. The obligation is
carried entirely by whoever remembers to copy an existing adapter's test files.

The three unprotected families are the ones whose absence is least visible:

- **path confinement** — its absence is a caller-supplied path opened unconfined.
  This is not hypothetical; it is #1361 and #1389, and the statusline endpoint
  shipped without it precisely because nobody wired the contract there.
- **endpoint follows bind addr** — its absence is #1178, entries pointing at a
  dead port with the permission still reading `granted`.
- **disclosure matches installed** — its absence is a #570 consent violation, the
  #1356 shape: content written to a user's config that the consent copy never
  named.

Note the asymmetry that makes this worth fixing rather than documenting: the two
families that ARE tripwired (`Verify`, `Version`) are the ones an adapter declares
as a struct FIELD, which is easy to walk. The three that are not are wired as
TEST CALLS, which is why nobody wrote the walk — but "harder to detect" is not
"less important", and the ranking above is roughly the reverse.

## Shape

`TestEveryHookInstallDeclaresAVerifier` is the model — it walks the registry
projection and fails for an adapter that omits the declaration. The equivalent
here cannot read a struct field, so it needs a different mechanism. Options worth
weighing rather than assuming:

- A static walk (like `core/architecture_hookbody_test.go`, which already does
  AST analysis over `core/adapters/inbound/agents/...`) asserting that each
  hooks-declaring adapter's test package references each required
  `contracttesting.Assert…` entry point.
- A registration seam: the contract entry points record which adapter they ran
  for, and one registry-walking test asserts the set covers every hooks-declaring
  adapter. This has the advantage of grading what actually EXECUTED rather than
  what is textually present, which is the same distinction #1390 drew when it
  removed a test-shaped constructor.

The second is stronger and the first is cheaper; either beats the status quo.

This ADDS a guard, so it owes a deliberate mutation seen red and committed beside
the assertion — delete one adapter's `hookpath_test.go` wiring and confirm the new
tripwire goes red, rather than only confirming it passes today. Per the same rule,
a guard that cannot run must fail loudly: if the walk finds zero adapters (a bad
path, a renamed directory) it must refuse rather than report a clean sweep — the
`hookcov` "silent all-alarm is worse than a silent all-clear" reasoning applies in
reverse here.

Found while consolidating #1730 / PR #1738, where the docs were being edited to
stop hand-maintaining these adapter rosters. Two independent agents on that work
reached the same conclusion: for these three families there is no tripwire to
point at, which is itself the finding.

Part of #1355.

ingo-eichhorst:
Closes #1740.

Two obligations over hooks-declaring adapters were already immune to the "nobody wired it" failure — `TestEveryHookInstallDeclaresAVerifier` and `TestEveryHookInstallDeclaresAVersionFloor` — because what they ask for is a struct **field** an adapter declares in production code. The families in `requiredWirings()` have no field to walk: they are wired as a **test call** in the adapter's own package, so their only enforcement was that whoever added the adapter remembered to copy an existing adapter's test files.

That is the hazard `docs/testing-contracts.md` already names, and it is how claudecode's statusline endpoint shipped unconfined:

> a contract can fail only for an adapter that WIRED it, so a receiver nobody wired it for is invisible to it

## Which option, and why the "stronger" one was rejected

#1740 sketched a static walk and a registration seam, and called the seam stronger because it grades what **executed** rather than what is textually present. The seam was rejected on a fact about the go tool, not on cost:

- `go test ./core/...` compiles **one test binary per package**. A recorder inside `contracttesting.AssertX` would run in six adapter processes and the asserting test in a seventh, so in-process state never reaches it.
- Carrying it across processes means a file, and that file is a cache the reader cannot validate. A stale entry makes a wiring someone has since **deleted** read as covered — the precise failure mode #1740 exists to end, reintroduced with a longer half-life. Meanwhile the build cache legitimately skips re-running an unchanged package's tests, so "P did not register" is also the ordinary answer for a cached pass.

Wrong in the green direction on a stale file and wrong in the red direction on a cache hit is not stronger than a static walk.

## Mutation evidence

The realistic shape — an adapter ships without wiring the contract at all. `rm core/adapters/inbound/agents/kirocli/hookpath_test.go`; `go vet` on the package stays clean, so this is not a compile-error confound:

```
--- FAIL: TestEveryHookInstallWiresItsContractFamilies (0.24s)
    hookcontracts_test.go:355: kiro-cli (irrlicht/core/adapters/inbound/agents/kirocli) installs hooks (hooks)
      but its test package never calls contracttesting.AssertHookPathConfined — the hook path confinement
      (#1361, #1389) contract is unenforced for this adapter.
        Why it is required: the transcript_path in an inbound hook body is caller-supplied on a local
        unauthenticated endpoint; unwired, a receiver can forward it unconfined and anything downstream
        opens whatever the caller named.
        Wire it: copy the corresponding *_test.go from an adapter that already does
        (`git grep -l AssertHookPathConfined core/adapters/inbound/agents/`) and point it at this
        adapter's own receiver/installer. A contract can only fail for an adapter that wired it (#1740).
```

Restored; `git status --porcelain` clean; suite green.

`hookcontracts_shapes_test.go` is the committed corpus — one case per call spelling pinned to the verdict the detector must return, including the `want:false` rows that pin the false positives a naive text rule would produce and the declared limits of an AST walk, so those are learned from a test rather than from an incident.

## What this does NOT cover, stated plainly

- **Package granularity.** The walk sees that an adapter's test package references the entry point, not that the reference is reachable, executed, or pointed at that adapter's own receiver. A wiring in a dead test, or one aimed at the wrong constructor, reads as covered.
- It cannot catch a family nobody has added to `requiredWirings()`.

Both are recorded next to the code rather than only here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -111,7 +111,9 @@ Before marking a ticket done, run the full suite — every layer must pass:
   the package-local guarded-construction guards and the probe-cost doc
   comment convention — is exercised inside the same `go test ./core/...`
   above. New adapters and new permissions are covered by wiring one call
-  each; full obligation list, self-test harnesses, and incident history:
+  each — and since #1740 a hooks-declaring adapter that skipped one is
+  failed by a registry tripwire rather than by someone noticing; full
+  obligation list, self-test harnesses, and incident history:
   [docs/testing-contracts.md](docs/testing-contracts.md).
 - Factory: `go test ./tools/onboarding-factory/... -race -count=1`.
 - Replay: `tools/replay-fixtures.sh` (golden drift, transition timing,
--- a/core/adapters/inbound/agents/hookcontracts_shapes_test.go
+++ b/core/adapters/inbound/agents/hookcontracts_shapes_test.go
@@ -4,8 +4,9 @@
 // detector must return for it.
 //
 // It exists because TestEveryHookInstallWiresItsContractFamilies passes by
-// construction — all six hooks-declaring adapters wired all seven families the
-// day it landed — and docs/testing-philosophy.md is explicit that this is the
+// construction — every hooks-declaring adapter wired every required family the
+// day it landed, which is what that test being green says — and
+// docs/testing-philosophy.md is explicit that this is the
 // condition the mutation rule exists for, and that the mutation belongs
 // COMMITTED beside the assertion rather than described in a PR body nothing
 // re-runs. The real-tree mutations (delete an adapter's hookpath_test.go, or
@@ -133,7 +134,7 @@ func wiringShapes() []wiringCase {
 			dir:    "direct",
 			want:   map[string]verdict{epConfined: wantCalled},
 			needle: "func TestHookPathConfined(t *testing.T)",
-			why:    "the spelling every one of the six hooks adapters uses today",
+			why:    "the spelling every hooks-declaring adapter uses today",
 			files: map[string]string{
 				"x_test.go": `package shape
 
--- a/core/adapters/inbound/agents/hookcontracts_shapes_test.go
+++ b/core/adapters/inbound/agents/hookcontracts_shapes_test.go
@@ -1,0 +1,815 @@
+// hookcontracts_shapes_test.go is the corpus behind the tripwire in
+// hookcontracts_test.go: one synthetic adapter test package per way of wiring
+// (or of appearing to wire) a contract family, each pinned to the verdict the
+// detector must return for it.
+//
+// It exists because TestEveryHookInstallWiresItsContractFamilies passes by
+// construction — all six hooks-declaring adapters wired all seven families the
+// day it landed — and docs/testing-philosophy.md is explicit that this is the
+// condition the mutation rule exists for, and that the mutation belongs
+// COMMITTED beside the assertion rather than described in a PR body nothing
+// re-runs. The real-tree mutations (delete an adapter's hookpath_test.go, or
+// rename its Test function so the wiring becomes a dead helper) are recorded in
+// PR #1740's body; these are the versions that keep running afterwards.
+//
+// Three groups, and the second and third carry the value:
+//
+//   - CAPABILITIES (wantCalled) — spellings the detector must credit. Several
+//     are things a grep cannot do: an aliased import, a two-hop helper chain, a
+//     call that crosses from the external test package into an in-package test
+//     helper.
+//   - MUST NOT CREDIT (wantAbsent / wantUnreached) — the two weaknesses #1740
+//     predicted a static walk would have, pinned as executable cases: a
+//     reference in a COMMENT, and a DEAD test. Plus the false positive a
+//     name-based rule produces (a local helper that merely shares the name),
+//     and a reference that is not a call.
+//   - DECLARED LIMITS (marked limit:true) — cases where the detector's answer
+//     is knowingly wrong, recorded so they are learned from a test rather than
+//     from an incident. A skipped test still counts; a wiring held in a
+//     package-level var of func type is not seen at all. Both directions are
+//     stated: the first over-credits, the second under-credits (fail-closed).
+//
+// Every case asserts the construct it plants is actually PRESENT in the source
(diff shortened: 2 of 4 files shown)


ingo-eichhorst:

I reproduced the mutation myself rather than accepting the report.

**Baseline:** `TestEveryHookInstallWiresItsContractFamilies` → `ok 0.373s`.

**Mutation** — the realistic shape, an adapter that simply never wired it: `rm core/adapters/inbound/agents/kirocli/hookpath_test.go`. `go vet` on the package stays clean first, so this is **not** a compile-error confound (my first attempt renamed the symbol instead, which failed for the wrong reason — weaker evidence, discarded):

```
--- FAIL: TestEveryHookInstallWiresItsContractFamilies (0.24s)
    hookcontracts_test.go:355: kiro-cli … installs hooks (hooks) but its test package never calls
    contracttesting.AssertHookPathConfined — the hook path confinement (#1361, #1389) contract is
    unenforced for this adapter.
```

Restored; `git status --porcelain` clean; package green.

**On rejecting the registration seam.** #1740 called it the stronger option and this PR declines it on a fact about the go tool rather than on cost — one test binary per package, so in-process state never reaches the asserting test, and the cross-process version is a cache the reader cannot validate: stale entry → a deleted wiring reads as covered, build-cache hit → a live wiring reads as missing. Wrong in the green direction on staleness and red on a cache hit is not stronger than a static walk. I agree, and the argument is recorded in the file rather than only here.

**CodeScene left red deliberately.** It flags `scanContractWirings` (7.14) and `TestContractWiringDetectorCatchesEveryKnownShape` (7.53) for Complex Method / Bumpy Road. Those are an AST walk over many node types and a table-driven corpus — the two shapes that are branchy by nature. Splitting them to move a metric makes them worse, and AGENTS.md is explicit that CodeScene is advisory, "a prompt to look closer, not something to chase to green." Looked; not chasing. Every other check passes, SonarCloud included.

**Limits, which the PR states rather than implies:** the walk grades package-granularity textual reference, not reachability — a wiring in a dead test or aimed at the wrong constructor reads as covered — and it cannot catch a family nobody added to `requiredWirings()`. Both are pinned by `want:false` corpus rows so they are learned from a test.

PR #1747, opened 2026-08-22
test(permissions): grade what Apply actually WROTE, not only that Writes exists (#1741)

### LINKED ISSUE #1741: test(permissions): the managed-files tripwire checks that Writes EXISTS, not that it covers every file Apply writes (2 defects so far)
`TestEveryModifyPermissionDeclaresTheFileItWrites`
(`core/cmd/irrlichd/managedfiles_test.go`) checks that a `modify`-kind permission
with a non-nil `Apply` **has** a `Writes` declaration. It does not check that the
declaration **covers every file `Apply` actually writes**.

That difference has now produced two real defects, one of them found after the
first was supposedly the lesson:

1. **#1718 / #1731** — kiro-cli's and mistral-vibe's installs each wrote a
   SECOND shared user file (`settings/cli.json`, `config.toml`) that no
   declaration named. Fixed by adding `agent.ManagedUserFile.Also`.
2. **#1739** — `recordPriorDefaultAgentOnce`
   (`core/adapters/inbound/agents/kirocli/hookinstaller.go:573`) writes
   `$KIRO_HOME/.irrlicht-prior-default.json` on every install, and it was in
   neither `Writes.Path` nor `Writes.Also` even after (1) added the field.

The second one is not inert, which is what makes this worth a tripwire rather
than vigilance. That function **deliberately never overwrites** an existing
record. So a copy left behind by a recording run is what a later *real* grant
reads instead of the operator's actual `chat.defaultAgent` — the restore-on-
uninstall then points them at the wrong agent, silently.

## Why the obvious static rule is not shippable as-is

Measured while investigating #1739: a candidate rule keying on package-level
`func() (string, error)` values not present in the declared set scores roughly

- **1 true positive** (the defect above), and
- **~1 false positive per adapter** — each adapter's home-root resolver has that
  exact signature and legitimately writes nothing.

and it misses any write built inline with `filepath.Join` rather than through a
named resolver, which is the more common shape. So the cheap version is both
noisy and porous. Recording the measurement here so the next attempt starts from
it instead of re-deriving it.

## Shape worth weighing

The interesting property is not "which functions look like path resolvers" but
"which paths did `Apply` actually write". That is observable at runtime, which
suggests grading it the way the permission contracts already do rather than
statically: run each `modify` permission's `Apply` against a scratch `$HOME`,
collect the files that appeared, and assert the set equals what `Writes` declares.
That also catches the inline-`filepath.Join` shape the static rule misses, and it
is the same "grade what EXECUTED, not what is textually present" distinction
#1390 drew when it removed a test-shaped constructor.

Cost to weigh honestly: it means running install closures in a test, which needs
the scratch-home isolation to be real — and #1449's `sharedConfigRefusal` already
refuses an `Apply` whose file resolves outside `IRRLICHT_HOME`, so the machinery
for that exists.

Adds a guard, so it owes a deliberate mutation seen red and committed beside the
assertion — remove one adapter's `Also` entry and confirm the new check reddens.
And it must fail loudly if it cannot run: a walk that observes zero writes because
the scratch home was misconfigured must refuse, not report a clean sweep.

Found in #1739. Part of #1355.

ingo-eichhorst:
Closes #1741.

`TestEveryModifyPermissionDeclaresTheFileItWrites` asks whether a `modify`
permission with a non-nil `Apply` **has** a `Writes` declaration. It never asked
whether the declaration **covers every file `Apply` actually writes** — the gap
that produced #1718/#1731 and then #1739, the second one *after* the first was
supposedly the lesson.

## What I built, and what decided it

Two arms, in `core/cmd/irrlichd/managedwrites_test.go`. The split is not
symmetry — it is one mechanism plus a fallback for the one permission that
mechanism must not touch.

### 1. Runtime arm — `TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites`

Runs each modify permission's `Apply` against a scratch `$HOME` and grades the
files that actually **changed**. This is the arm that carries the weight: it is
the only thing that sees a path built inline with `filepath.Join` inside
`Apply`, which #1741 measured the obvious static rule to miss entirely, and it
needs no per-adapter knowledge — a new modify permission is graded by existing.

Covers 8 of the 9 `Apply` closures in the consent catalog: claude-code
hooks/statusline/instructions, codex, gemini-cli, mistral-vibe, copilot,
kitty, plus gastown/state (which declares nothing).

Three properties are load-bearing:

- **The environment is scrubbed to an ALLOW-list** (`envKeptDuringScrub`), not
  a deny-list of `CODEX_HOME`/`COPILOT_HOME`/`KIRO_HOME`/`VIBE_HOME`/
  `XDG_CONFIG_HOME`/`CLAUDE_CONFIG_DIR`. A deny-list has to enumerate every
  override an adapter *might* honour, and the day a new adapter reads a new one,
  its `Apply` writes to the developer's real config while the test believes it
  is sandboxed. Emptying everything means an unknown override resolves to `""`
  and the adapter falls back to its `$HOME`-relative default — which *is* the
  scratch home. A canary variable is set and then swept up by the scrub itself,
  so "the scrub ran" is observed rather than assumed (mutation M6 below).
- **It fails closed.** Every declared path in the *whole* catalog is resolved and
  proven inside the scratch tree before a single closure runs (M5). Deliberately
  *not* by routing through #1449's `sharedConfigRefusal`: that guard refuses one
  `Apply` at a time from inside the service, so eight closures would have run
  before the ninth's escape was noticed — and its
  `IRRLICHT_ALLOW_SHARED_CONFIG_WRITES` escape hatch is the last thing that
  should be within reach of a test that executes installers. (`IRRLICHT_HOME` is
  still pointed at the sandbox, so the guard is armed if a future refactor does
  route through it.)
- **Both directions are graded.** An undeclared write is #1741 itself. A
  *declared* file nothing wrote is the vacuity guard, and it is the one that
  makes a broken sandbox loud: if `$HOME` had not taken, every closure would
  write to the real home, the scratch diff would be EMPTY, and an
  undeclared-writes-only check would report a flawless sweep (M3).

### 2. Static arm — `TestExternalInstallerPackagesDeclareEveryPathResolver`

kiro-cli's `Apply` shells out to the real CLI. **Measured, not assumed**, on
2026-08-22 — this is why the runtime arm must not execute it:

```
--- kiro-cli/hooks
  declared: [Path=/.kiro/agents/irrlicht.json Also[0]=/.kiro/settings/cli.json Also[1]=/.kiro/.irrlicht-prior-default.json]
  applyErr: kirocli: kiro-cli agent create irrlicht --from kiro_default: exit status 1: error: You are not logged in, please log in with kiro-cli login
  wrote:    [/.kiro/settings/cli.json  /Library/Application Support/kiro-cli/data.sqlite3]
```

- With the binary **absent** (any CI runner) `EnsureHooksInstalled` fails at its
  first step and writes **nothing** — the arm would grade an empty set, which is
  exactly "inability to look reads as a clean result".
- With it **present** the CHILD writes its own state
  (`$HOME/Library/Application Support/kiro-cli/data.sqlite3` above), and how much
  more it writes depends on whether the developer is logged in. Either way the
  verdict stops being a property of this repo.

So `kiro-cli/hooks` is named in `applyRunsAnExternalCLI` with that reason, its
`Binary` cross-checked against the permission's own `Writes.Version.Probe[0]`
so the exemption cannot claim a shellout the declaration does not corroborate
(M9), and its package is graded statically: every package-level
`func() (string, error)` there must be one the declaration names — **derived by
reflecting the function names out of `Writes.Path`/`Writes.Also`**, so it cannot
drift from a hand-kept list — or be named in `notAManagedFilePath` (M8).

That is the rule #1741 measured at ~1 false positive per adapter. Scoping it to
the packages the runtime arm cannot execute is what makes it shippable:
**one** exemption (`kirocli.kiroHome`, the `$KIRO_HOME` root every other
resolver joins onto) instead of one per adapter, and it adds nothing the runtime
arm does not already cover more strongly.

## Mutation evidence — actual red output

Each was applied and reverted in its own foreground call, and each run was
scoped to the new test so "the arm failed" and "THIS obligation failed" are not
confused.

**M1 — remove `priorDefaultStatePath` from kirocli's `Writes.Also` (this is #1739):**
```
--- FAIL: TestExternalInstallerPackagesDeclareEveryPathResolver/kiro-cli/hooks
    managedwrites_test.go:682: kirocli.priorDefaultStatePath is a package-level path resolver
    that kiro-cli/hooks declares in neither Writes.Path nor Writes.Also. If its Apply writes
    that file, it is invisible to --print-managed-files, to the recording rig's backup and to
    #1449's grant-all refusal (#1739). Declare it, or add "kirocli.priorDefaultStatePath" to
    notAManagedFilePath with the reason it names no managed file
```

**M2 — remove `vibeConfigTomlPath` from mistral-vibe's `Writes.Also` (this is #1718/#1731):**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites/mistral-vibe/hooks
    managedwrites_test.go:261: Apply wrote "home/.vibe/config.toml", which mistral-vibe/hooks
    declares in neither Writes.Path nor Writes.Also. That file is invisible to
    --print-managed-files, so the recording rig never backs it up and #1449's grant-all refusal
    never sees it (#1741). Declare it, or explain it as an allowance
```

**M3 — kitty's `EnsureKittyConfigPatched` reports a write and performs none (the vacuity direction / broken-sandbox shape):**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites/kitty/remote-control
    managedwrites_test.go:267: kitty/remote-control declares "home/.config/kitty/kitty.conf" but
    running Apply changed nothing there. Either the declaration resolves to the wrong file, or
    the closure no longer writes it — and if NEITHER is true, the scratch home is not where the
    writes went, which makes every 'no undeclared write' result above meaningless
```

**M4 — an undeclared write built INLINE inside vibe's `Apply` (`filepath.Join(home, "irrlicht-state.json")`), the shape the static rule cannot see:**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites/mistral-vibe/hooks
    managedwrites_test.go:261: Apply wrote "home/.vibe/irrlicht-state.json", which
    mistral-vibe/hooks declares in neither Writes.Path nor Writes.Also. ...
```
The static arm was in the same run and stayed **green** — which is the porosity
#1741 measured, demonstrated rather than asserted.

**M5 — `kittyConfigDir` returns `/etc/xdg/kitty` (a resolver that ignores `$HOME`):**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites
    managedwrites_test.go:234: kitty/remote-control declares Writes.Path = "/etc/xdg/kitty/kitty.conf",
    which is OUTSIDE the scratch home "/var/folders/.../001". Refusing to run any Apply: this test
    would be writing to a shared config the developer actually uses
```
Note there are **no sub-test lines** in that output: it aborted before the first
closure ran, which is the fail-closed property.

**M6 — `envKeptDuringScrub` returns true for everything (the scrub does nothing):**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites
    managedwrites_test.go:205: the environment scrub left
    IRRLICHT_MANAGEDWRITES_SCRUB_CANARY="/definitely/not/the/scratch/home" — it did not run, so
    every per-agent home override the developer exports is still in effect and the closures below
    would write to their real config
```

**M7 — `applyRunsAnExternalCLI` names a permission that does not exist:**
```
    managedwrites_test.go:226: applyRunsAnExternalCLI names "kiro-cli/no-such-permission", which
    is not a permission with an Apply closure in the consent catalog — an exemption for an
    obligation that does not exist reads as coverage
```
The same run also shows the other half working: with kiro-cli no longer exempt,
its `Apply` ran and failed loudly (`Apply returned … not logged in`) rather than
grading an empty set.

**M8 — `notAManagedFilePath` key renamed (stale exemption):**
```
    managedwrites_test.go:682: kirocli.kiroHome is a package-level path resolver that
    kiro-cli/hooks declares in neither Writes.Path nor Writes.Also. ...
    managedwrites_test.go:688: notAManagedFilePath names "kirocli.kiroHomeRenamedUpstream" (...),
    which the walk did not find in any external-installer package — a stale exemption reads as
    coverage
```

**M9 — the exemption's `Binary` disagrees with the permission's own probe:**
```
    managedwrites_test.go:226: applyRunsAnExternalCLI["kiro-cli/hooks"] names binary "kiro", but
    the permission's own version probe runs "kiro-cli" — one of the two is stale and a reader
    cannot tell which
```

### The corpus's own mutations

`managedwrites_shapes_test.go` is the committed evidence for the pure grading
functions — one row per shape, each carrying **both** verdicts, because a
detector that reports everything and one that reports correctly are
indistinguishable without the cases that must stay silent.

**M10 — allowance containment by plain string prefix instead of path segments:**
```
--- FAIL: .../a_sibling_that_merely_shares_the_allowance_root's_prefix_is_reported
    Undeclared = [], want [irrlicht-home-backup/leak.json]
    allowance "irrlicht-home" absorbed 1 paths, want 0
```

**M11 — `treeDiff` blind to removals (a permission that DELETES a user file):**
```
--- FAIL: .../a_removed_file          treeDiff = [], want [home/.codex/hooks.json]
--- FAIL: .../everything_at_once,_sorted   treeDiff = [home/.a], want [home/.a home/.zshrc]
```

**M12 — the walk counts a METHOD as a package-level resolver:**
```
--- FAIL: .../a_method_with_the_same_signature    resolvers = [settingsPath], want []
```

## Proof nothing touches the real home

Beyond the fail-closed precondition and the scrub canary, measured directly:
snapshot `~/.claude/settings.json`, `~/.claude/CLAUDE.md`, `~/.codex/hooks.json`,
`~/.gemini/settings.json`, `~/.config/kitty/kitty.conf`, `~/.kiro`, `~/.vibe`,
`~/.copilot` and `~/Library/Application Support/kiro-cli` (mtime + size, plus a
2-deep `find` of `~/.kiro` and `~/.vibe`), run the two new tests, snapshot again
— `diff` reports **IDENTICAL**.

## A finding this turned up, NOT fixed here

`kiro-cli/hooks`'s `Apply` causes `~/Library/Application Support/kiro-cli/data.sqlite3`
to be created/modified in whichever `$HOME` the daemon runs under — it is the
child `kiro-cli` process's own state DB, undeclared by anything. It is
deliberately **not** added to `Writes.Also`: it holds no irrlicht content, so
`ManagedUserFile`'s uninstall semantics are meaningless for it, its location is
macOS-specific and unverified on Linux, and having the recording rig restore a
sqlite DB under a possibly-running kiro-cli is a worse outcome than not. Flagged
rather than buried; worth its own issue if the rig's snapshot set should grow to
cover an agent CLI's own store.

## What this does NOT cover

Stated in the file header and pinned as corpus rows wherever a row can express it:

- A write to an absolute path following neither `$HOME` nor any env var
  (a hard-coded `/etc/...`) lands outside the walked tree. The static arm sees it
  only if it goes through a package-level resolver.
- A write that restores size, mtime AND content byte-identically is invisible to
  any snapshot diff (pinned as a want:none row).
- A permission creating only a DIRECTORY is not reported — the walk records
  non-directory entries.
- `gastown/state`'s `Apply` is the no-op `declaredConsentCatalog` passes, not the
  watcher the daemon wires, so running it grades the no-op.
- The static arm reads `ast.FuncDecl`s, so a resolver held in a package-level
  `var` is not seen. That direction reads as "no undeclared resolver", i.e. it is
  a hole rather than a loud refusal, which is exactly why it is a pinned
  want:false row — the same declared limit `architecture_hookbody_test.go` and
  `contracttesting/seam_walk_corpus_test.go` carry.
- Whether a permission genuinely spawns an external CLI is a declared claim, not
  a derived one. "Does this package contain `exec.Command`" is derivable and
  wrong — `processlifecycle` contains several and kitty's `Apply` spawns none, so
  the derived version would wave through the one permission in that package a
  scratch home *can* grade.

## Advisory checks

`SonarCloud` is **green** after two follow-up commits: two `go:S3776`
cognitive-complexity findings (19 and 16 vs the 15 allowed) split along seams
the doc comments already drew, and the one duplicated block — 10 repeated
`file: "a.go"` fields in the resolver corpus — defaulted away. Every mutation
was re-verified red after each refactor; M5 in particular still aborts before
any sub-test runs.

`CodeScene` is **red and I stopped chasing it**, per AGENTS.md ("a red result is
a prompt to look closer, not something to chase to green before merging"). I
looked closer twice: code-health impact went 8.01 -> 8.12 -> 8.96 and "Overall
Code Complexity" dropped off, but each round it re-flags a different pair of
small functions — now `treeDiff` (Bumpy Road Ahead) and
`isPathResolverSignature` (Complex Method). `treeDiff` is 12 lines with exactly
two sequential loops because a snapshot diff has exactly two directions
(created-or-modified, and removed), and `isPathResolverSignature` is three
guards and a call. Splitting either further would cost the reader more than it
buys. Flagging rather than suppressing.

## Gates

- `go test ./core/... -race -count=1` — PASS
- `tools/preflight.sh --only go` — all 6 gates PASS
- `tools/preflight.sh --only arch` — PASS (composite 7.9163 -> 7.9166, no regression)

### DIFF
--- a/core/cmd/irrlichd/managedwrites_shapes_test.go
+++ b/core/cmd/irrlichd/managedwrites_shapes_test.go
@@ -245,47 +245,43 @@ func TestPathUnderDirRefusesEveryUncontainedShape(t *testing.T) {
 func TestPackageLevelPathResolversPinsEverySpelling(t *testing.T) {
 	cases := []struct {
 		name string
-		file string // file name inside the synthetic package
+		// file is the name inside the synthetic package. Empty means
+		// "a.go" — only the _test.go row needs to say otherwise, and repeating
+		// it on every other row is noise a reader has to check anyway.
+		file string
 		src  string
 		want []string
 	}{{
 		name: "the ordinary spelling",
-		file: "a.go",
 		src:  "package p\nfunc settingsPath() (string, error) { return \"\", nil }\n",
 		want: []string{"settingsPath"},
 	}, {
 		name: "named results",
-		file: "a.go",
 		src:  "package p\nfunc settingsPath() (path string, err error) { return }\n",
 		want: []string{"settingsPath"},
 	}, {
 		name: "two names on one result field",
-		file: "a.go",
 		src:  "package p\nfunc twoStrings() (a, b string) { return }\n",
 		want: nil, // (string, string), not (string, error)
 	}, {
 		name: "exported is still a resolver",
-		file: "a.go",
 		src:  "package p\nfunc SettingsPath() (string, error) { return \"\", nil }\n",
 		want: []string{"SettingsPath"},
 	}, {
 		// FALSE POSITIVE a text rule produces: it takes a parameter, so it is
 		// not something a ManagedUserFile field can hold.
 		name: "takes a parameter",
-		file: "a.go",
 		src:  "package p\nfunc pathUnder(root string) (string, error) { return \"\", nil }\n",
 		want: nil,
 	}, {
 		// FALSE POSITIVE: the result types are not (string, error).
 		name: "returns the wrong pair",
-		file: "a.go",
 		src:  "package p\nfunc info() (string, bool) { return \"\", false }\n",
 		want: nil,
 	}, {
 		// FALSE POSITIVE: a METHOD cannot be a ManagedUserFile resolver, and
 		// its receiver is what tells it apart from one.
 		name: "a method with the same signature",
-		file: "a.go",
 		src:  "package p\ntype T struct{}\nfunc (t T) settingsPath() (string, error) { return \"\", nil }\n",
 		want: nil,
 	}, {
@@ -296,14 +292,12 @@ func TestPackageLevelPathResolversPinsEverySpelling(t *testing.T) {
 		// resolver", so it is a hole rather than a loud refusal, which is
 		// exactly why it is written down.
 		name: "a func literal in a package-level var",
-		file: "a.go",
 		src:  "package p\nvar settingsPath = func() (string, error) { return \"\", nil }\n",
 		want: nil,
 	}, {
 		// DECLARED LIMIT: the walk compares result types syntactically, so a
 		// local alias of error is not recognised.
 		name: "an aliased error type",
-		file: "a.go",
 		src:  "package p\ntype myErr = error\nfunc settingsPath() (string, myErr) { return \"\", nil }\n",
 		want: nil,
 	}, {
@@ -315,7 +309,6 @@ func TestPackageLevelPathResolversPinsEverySpelling(t *testing.T) {
 		want: nil,
 	}, {
 		name: "several in one file, sorted",
-		file: "a.go",
 		src: "package p\n" +
 			"func zPath() (string, error) { return \"\", nil }\n" +
 			"func aPath() (string, error) { return \"\", nil }\n",
@@ -324,15 +317,19 @@ func TestPackageLevelPathResolversPinsEverySpelling(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
+			name := tc.file
+			if name == "" {
+				name = "a.go"
+			}
 			dir := t.TempDir()
-			if err := os.WriteFile(filepath.Join(dir, tc.file), []byte(tc.src), 0o600); err != nil {
+			if err := os.WriteFile(filepath.Join(dir, name), []byte(tc.src), 0o600); err != nil {
 				t.Fatal(err)
 			}
 			// The row's own vacuity guard: the case must actually contain the
 			// construct it plants, or a corpus that quietly stopped carrying
 			// its own test cases would read as a pass (the rule
 			// core/architecture_hookbody_shapes_test.go states).
-			assertParses(t, dir, tc.file)
+			assertParses(t, dir, name)
 
 			got, err := packageLevelPathResolvers(dir)
 			if err != nil {
--- a/core/cmd/irrlichd/managedwrites_shapes_test.go
+++ b/core/cmd/irrlichd/managedwrites_shapes_test.go
@@ -1,0 +1,442 @@
+// managedwrites_shapes_test.go is the committed mutation evidence for
+// managedwrites_test.go's two arms.
+//
+// Per docs/testing-philosophy.md, a guard a change ADDS owes a deliberate
+// mutation seen red, and the mutation is worth more committed than described:
+// "a paragraph in a merged PR body is re-run by nothing and an assertion that
+// silently stops discriminating looks exactly like health."
+//
+// Two things are graded here that the live catalog cannot express.
+//
+// The GRADING (gradeWrites/treeDiff) is driven with snapshots no correct
+// adapter produces — a modified pre-existing file, a REMOVED file, an
+// allowance that absorbs nothing, a path that merely shares a string prefix
+// with an allowance root. Each row carries both verdicts on purpose: a
+// detector that reports everything and one that reports correctly are
+// indistinguishable without the cases that must stay SILENT.
+//
+// The WALK (packageLevelPathResolvers) is driven over source strings, one row
+// per spelling, pinned to the verdict it must return. The want:false rows are
+// where the value is (#1450): three of them are false positives a text-based
+// rule would produce, and two are declared LIMITS of an ast.FuncDecl walk —
+// pinned so the next person learns them from a test rather than from an
+// incident.
+package main
+
+import (
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// ---------------------------------------------------------------------------
+// The grading
+// ---------------------------------------------------------------------------
+
+// TestGradeWritesReportsEveryKnownShape drives gradeWrites over one case per
+// shape, each wrong in exactly ONE way, and pins both what it must report and
+// what it must leave silent.
+func TestGradeWritesReportsEveryKnownShape(t *testing.T) {
+	const allowRoot = "irrlicht-home"
+	allowed := []allowance{{Root: allowRoot, Reason: "the daemon's own state"}}
+
+	cases := []struct {
+		name           string
+		observed       []string
+		declared       []string
+		wantUndeclared []string
+		wantMissing    []string
+		wantAllowed    int
+	}{{
+		name:     "the correct case — every observed write is declared, nothing declared is unwritten",
+		observed: []string{"home/.vibe/config.toml", "home/.vibe/hooks.toml"},
+		declared: []string{"home/.vibe/hooks.toml", "home/.vibe/config.toml"},
+	}, {
+		// #1739's shape: a second file the same Apply writes, undeclared.
+		name:           "an undeclared write is reported",
+		observed:       []string{"home/.kiro/agents/irrlicht.json", "home/.kiro/.irrlicht-prior-default.json"},
+		declared:       []string{"home/.kiro/agents/irrlicht.json"},
+		wantUndeclared: []string{"home/.kiro/.irrlicht-prior-default.json"},
+	}, {
+		// The shape a "which files APPEARED" check misses entirely: the file
+		// was already there and Apply rewrote it. treeDiff is what makes this
+		// reachable; gradeWrites only has to not special-case it.
+		name:           "a modified pre-existing file is reported like a created one",
+		observed:       []string{"home/.zshrc"},
+		declared:       []string{"home/.claude/settings.json"},
+		wantUndeclared: []string{"home/.zshrc"},
+		wantMissing:    []string{"home/.claude/settings.json"},
+	}, {
+		name:        "a declared file nothing wrote is reported — the vacuity direction",
+		observed:    []string{},
+		declared:    []string{"home/.config/kitty/kitty.conf"},
+		wantMissing: []string{"home/.config/kitty/kitty.conf"},
+	}, {
+		// The failure a misconfigured sandbox produces. Without the Missing
+		// direction this row is a clean sweep, which is the whole reason
+		// gradeWrites reports both ways.
+		name:     "nothing observed and nothing declared is NOT reported by this function",
+		observed: []string{},
+		declared: []string{},
+	}, {
+		name:        "a file inside an allowance is absorbed, not reported",
+		observed:    []string{"irrlicht-home/sessions.db", "home/.codex/hooks.json"},
+		declared:    []string{"home/.codex/hooks.json"},
+		wantAllowed: 1,
+	}, {
+		name:        "the allowance root itself is absorbed",
+		observed:    []string{"irrlicht-home"},
+		declared:    []string{},
+		wantAllowed: 1,
+	}, {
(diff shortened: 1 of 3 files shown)


ingo-eichhorst:
## QA — verified independently, including one suspicion of mine that was wrong

**Safety first, since this test executes install closures.** I snapshotted every path `--print-managed-files` names plus a 2-deep walk of `~/.kiro` and `~/.vibe` (49 paths, name + size + mtime), ran the new tests, and re-snapshotted:

```
REAL HOME IDENTICAL — 49 paths, byte-size and mtime unchanged
```

**A suspicion I had, and why it was unfounded.** Under mutation M1 (re-introducing #1739 by dropping `priorDefaultStatePath` from `Also`), the runtime arm `TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites` **passes** — only the pre-existing #1739 lock reddens. That looked like M1 being credited to a mechanism that did not catch it. It is not: kiro-cli is deliberately exempt from the runtime arm because its `Apply` shells out, and the **static** arm is what grades it. Running the right test:

```
managedwrites_test.go:702: kirocli.priorDefaultStatePath is a package-level path resolver that
kiro-cli/hooks declares in neither Writes.Path nor Writes.Also. If its Apply writes that file, it is
invisible to --print-managed-files, to the recording rig's backup and to #1449's grant-all refusal
(#1739). Declare it, or add "kirocli.priorDefaultStatePath" to notAManagedFilePath with the reason
it names no managed file
--- FAIL: TestExternalInstallerPackagesDeclareEveryPathResolver
```

Restored; `git status --porcelain` clean; `./core/cmd/irrlichd` green at 11.8s — the runtime arm really is executing the closures, not short-circuiting.

## Three corrections to my brief, all of which I accept

1. **`sharedConfigRefusal` is the wrong net.** I suggested leaning on it. It refuses one `Apply` at a time from inside the service, so eight closures would run before the ninth's escape was noticed — and its `IRRLICHT_ALLOW_SHARED_CONFIG_WRITES` hatch is the last thing that should be within reach of a test that executes installers. An all-paths-up-front precondition that aborts before the first closure is strictly better.
2. **My per-agent env list was a deny-list.** `CODEX_HOME`, `COPILOT_HOME`, `KIRO_HOME`… has to be updated by whoever adds adapter number eight, and an unknown override falls back to `/Users/ingo`. Allow-list scrubbing removes that obligation entirely, and the canary proving the scrub ran is the right companion.
3. **#1739 is already fixed on `main`**, so M1 is a re-introduction rather than a live defect. My brief implied otherwise.

## The split into two arms is the right call and is honestly bounded

Running kiro-cli's `Apply` under `go test` is non-deterministic in two directions — binary absent (CI) it writes nothing and grades an empty set; binary present it writes a sqlite store whose contents depend on login state. Exempting it with that reason, cross-checking `Binary` against its own `Writes.Version.Probe[0]`, and grading its package statically costs **one** exemption instead of the ~1-per-adapter false-positive rate #1741 measured for the naive static rule.

CodeScene red, left there per AGENTS.md — health impact improved 8.01→8.96 across two rounds and each round re-flags a different pair of ~10-line functions (now `treeDiff`, which has two loops because a diff has two directions). Looked twice; not chasing. Every required check is green, SonarCloud included.

Filing the `data.sqlite3` finding separately rather than folding it in.

Issue #1735, opened 2026-08-21
test(onboarding): 3 of 6 hooks-declaring adapters have zero hook-bearing recordings


ingo-eichhorst:
> **CORRECTED 2026-08-21 — read [this comment](https://github.com/ingo-eichhorst/Irrlicht/issues/1735#issuecomment-5373174886) before acting on the "blocker" section below.**
> The measured GAP (3 of 6 adapters, zero hook-bearing recordings) stands. The *diagnosis* does not:
> the rig already snapshots and restores every declared managed file as a hard gate on spawning the
> recording daemon, and all four kiro/vibe files are in that protected set. Recording is most likely
> not blocked at all — it just has not been run. The original text is kept below unedited, because a
> wrong claim that stopped people looking is worth leaving visible next to its correction.

Epic #1355's stated risk was shipping hook adapters faster than the fixtures that
prove they work. That is now the actual state, and it got worse before it was
noticed: the GAP went from **1 adapter to 3** on 2026-08-21, as a direct
consequence of #1732 (kiro-cli) and #1733 (mistral-vibe) landing without recordings.

## Measured

`go run ./tools/onboarding-factory/cmd/of coverage --hooks` on `main` @ d9be34ad:

```
adapter        declares-hooks  cells  recordings  with-hooks  status
claudecode                yes     46          53          12  ok
codex                     yes     44          38           2  ok
copilot                   yes     46          37           1  ok
gemini-cli                yes     45          38           0  GAP
kiro-cli                  yes     44          28           0  GAP
mistral-vibe              yes     46          36           0  GAP
total                            482         366          16

GAP: 3 of 6 hooks-declaring adapters have zero hook-bearing recordings
```

## The three GAPs are not one problem

**kiro-cli and mistral-vibe** are blocked on one shared, adapter-agnostic
mechanism — see the blocker below. Both are closable.

**gemini-cli is not closable on this machine and should not be counted as
outstanding work.** The account is tier-ineligible against Gemini CLI
(`IneligibleTierError`, steering to Antigravity), so no live turn and no
re-recording is possible; deprecation is already scheduled (decided 2026-08-20 at
v0.5.10, effective v0.7 — 37edb71d / #1725). Its GAP row is an accurate
description of a documented state, not a task. It is listed here so a future
reader does not re-derive that.

## The blocker (shared by both closable adapters)

`tools/onboarding-factory/scripts/lib/spawn-record-daemon.sh:88` exports
[...]

PR #1752, opened 2026-08-22
test(onboard): close the hook-coverage GAP for kiro-cli and mistral-vibe (#1735)

ingo-eichhorst:
Closes the in-scope half of #1735: **kiro-cli and mistral-vibe now have hook-bearing recordings.**

```
                        before        after
gemini-cli   with-hooks   0  GAP        0  GAP   (out of scope — #1725)
kiro-cli     with-hooks   0  GAP        1  ok
mistral-vibe with-hooks   0  GAP        1  ok
total                    16            18

GAP: 3 of 6 hooks-declaring adapters   →   GAP: 1 of 6
```
(`go run ./tools/onboarding-factory/cmd/of coverage --hooks`, before on `origin/main` @ `28d08aad`, after on this branch.)

**Two cells recorded**, both first-time-hook-bearing:

| cell | hook events captured |
|---|---|
| `mistral-vibe / 2-1_basic-turn` (5/5 phases) | `hook_received{Stop}` — vibe's `post_agent_turn`, its only installed event |
| `kiro-cli / 2-2_auto-executed-tool-call` (3/3 phases) | `hook_received{PostToolUse}` **and** `hook_received{Stop}` — both of kiro's installed events in one cell |

Well under #1355's >5 re-recorded-cells kill criterion. Recorded in **coexist mode** (recording daemon on `127.0.0.1:7838`, its own `IRRLICHT_HOME`, scratch `KIRO_HOME` / `VIBE_HOME`) against the operator's running production daemon on 7837 and `Irrlicht.app`, neither of which was stopped, signalled or reconfigured.

## The issue's remaining scope said "nobody has run them". That was true and it was not sufficient

Four rig defects sat between "run it" and a hook-bearing fixture. Every one of them produced a **complete, healthy-looking, hook-free recording** rather than a failure — which is the fixture-that-lies outcome the issue was filed to avoid.

### 1. A beacon-delivered hook resolves the daemon at FIRE time, not install time

`run-cell.sh` said hook observation works in coexist mode "since #1178: the endpoint the installers write follows `IRRLICHT_BIND_ADDR`". That is true of **URL delivery** (claudecode, codex, copilot bake the address into the bytes they write) and **false of beacon delivery**: the entry names `irrlichd hook-post <adapter>` and carries no address at all, and the beacon resolves the target in its own process via `daemonaddr.resolveClient` — `IRRLICHT_BIND_ADDR`, then the addr file under `IRRLICHT_HOME`. That process is a child of the agent CLI, which the driver launches through `tmux new-session`, so it inherits the tmux **server's** environment.

In coexist mode the pane carried neither variable, read `~/.local/share/irrlicht/irrlichd.addr`, and posted every hook to the **production** daemon.

The three adapters with zero hook-bearing recordings are exactly the three whose packages import `core/pkg/hookbeacon`. That is #1739's half-wired-lever shape in a second variable, and the fix is #1739's: an explicit `env`/`-e` prefix on each beacon driver's tmux launch, plus the export in both rigs.

### 2. The daemon's socket is bound before its hook install has run

`spawn_record_daemon` returns on the socket; the grant-all `Apply` closures land seconds later; every agent CLI reads its hook config **once, at startup**. Measured: `driver_exit_reason=ok`, `completeness=complete`, 11 curated events, and not one `hook_received` anywhere in the raw capture — `$VIBE_HOME/hooks.toml` appeared ~4s after vibe had already read it.

Same shape as #1449. `lib/hook-install-wait.sh` polls for the files rather than sleeping (the snapshot manifest already records which declared paths were absent before the daemon started), and says out loud when it cannot select an adapter's files instead of reporting a check it did not perform.

### 3. `[session_logging].save_dir` defeats `VIBE_HOME` for transcripts only

It is an absolute path and overrides the `VIBE_HOME`-derived default, so a `config.toml` seeded by copying the operator's — which `agent-home.sh` tells you to do — relocates everything **except** the transcripts. vibe wrote its sessions into the real `~/.vibe/logs/session` while the driver watched an empty scratch directory and timed out on "vibe never created a session dir", which reads as an auth failure. The driver now resolves the key the same way the daemon does (`vibe/config.go configuredSaveDir`) and hard-fails on a `save_dir` outside a requested `VIBE_HOME`.

### 4. vibe's startup update picker

vibe 2.19.1 renders a blocking `› Update now   Continue with current version` dialog when `$VIBE_HOME/cache.toml` names a newer release. The cache is refreshed by a daily background check and read on the **next** launch, so the launch that discovers a release is fine and the one after it blocks — with no driver arm for it. Measured (latest 2.24.3 vs installed 2.19.1): control launch renders the picker, `VIBE_ENABLE_UPDATE_CHECKS=0` does not. **Not** a scratch-home artifact — the operator's real `~/.vibe/cache.toml` was six weeks stale, i.e. one background refresh from breaking every mistral-vibe recording against the real home too.

## Tripwire, with committed mutation evidence

`righome.BeaconAdapters` derives the obligation from the **import graph** — an adapter package's non-test sources importing `core/pkg/hookbeacon`, with the slug taken from its own `const AdapterName`, so there is no second directory-to-slug map to drift — and refuses on an empty result. `TestEveryBeaconAdapterDriverPassesTheDaemonAddress` then requires each such driver to pass `IRRLICHT_BIND_ADDR` through every tmux launch.

- The scan's mutation evidence is **committed**, not described: `TestScanBeaconPackageNamesEveryWrongShape` includes four `want:false` rows a substring grep for `hookbeacon` would misreport (a test-only import, a same-named vendor package) and two refusals.
- The detector this change **rewrote** carries its predecessor's accepted spellings as locks (`TestLaunchWindowCarriesNamesEverySpelling`). The rewrite was necessary: the original matched only the *first* assignment of an `env` prefix and therefore failed kiro-cli's `env "KIRO_HOME=…" "IRRLICHT_BIND_ADDR=…"` — a driver that was correct.
- **Seen red on a deliberate mutation.** Deleting the passthrough from mistral-vibe's launch:
  ```
  --- FAIL: TestEveryBeaconAdapterDriverPassesTheDaemonAddress/mistral-vibe
      .../mistral-vibe/driver-interactive.sh:425 launches the agent under tmux
      without passing IRRLICHT_BIND_ADDR.
  ```
  Restoring it goes green. Writing the arm also found **two launch sites the hand fix had missed** (the resume/restart paths in both drivers).

## A real adapter defect this surfaced — not fixed here

**mistral-vibe's `hooks` permission cannot install into a `config.toml` that vibe itself wrote.** `hooktoml`'s splicer refuses any document containing a multi-line array; a real vibe config has twelve (`applied_migrations` plus the default `[tools.*]` allow/deny lists). `Apply` fails with *"contains a construct this splicer does not model safely"*, **the permission still reads `granted`**, and the refusal surfaces only in `unapplied_grants`.

Measured on an isolated fake-`$HOME` daemon: with the operator's own `config.toml`, no `hooks.toml` and no `enable_experimental_hooks`; with every multi-line array collapsed onto one line, both are written and `unapplied_grants` is empty. Collapsing only the top-level `applied_migrations` was **not** enough — the scanner refuses on arrays inside tables too.

The recording was unblocked by collapsing the arrays **in the scratch seed**. That is a workaround, and it means the recorded fixture exercises a hook path that does not currently install on a stock user machine. Recorded in `agent-home.sh`'s mistral-vibe row; worth its own issue.

## Pinned figures that moved (all machine-generated, pasted from what the tests print)

- `censusOfTheCommittedCatalog`: `Recordings` 314→315, `Divergent` 143→144, `DivergentByCountsAndKinds` 142→143, `PairedButUngraded` 85→86. `Zero`, `Fabricated`, `UnpairedSidecars` unmoved.
- `stopHookCensus` gains both recordings — kiro-cli's Stop is `Reproduced: 1`; mistral-vibe's is `Reproduced: 0` because its sidecar carries no `transcript_activity` with `file_size` for the primary session, the same shape hermes' 4-2 has.
- Two `coincidentalCensusFigures` entries stopped suppressing anything and were deleted per that check's own advice; the archaeology comment gains this run (quoting no figures).
- `knownZeroTransition`, `knownFabricated` and `knownFirstTransitionDrift` did **not** move.

## Operator safety

Managed files were snapshotted before and diffed after: **every one is byte-identical** (SHA-256) and every absent one still absent. The rig's snapshot/restore did its job.

One real-home change to report: **two new directories under `~/.vibe/logs/session/`** (`session_20260822_013432_223ec994`, `session_20260822_014033_4b846077`) — vibe's own session logs from the first two attempts, before defect 3 was found. Nothing was overwritten or deleted. They are left in place for you to remove; deleting files from your home was outside what I was authorised to do.

tmux ran on a private `TMUX_TMPDIR` / `-L` socket throughout; your sessions were not touched.

## Gates

`go test ./tools/onboarding-factory/... -race -count=1` · `of validate` · `preflight --only go|tools|bash|posix|skills|arch|security` — all PASS. ARS composite unchanged (7.9175 → 7.9175).

Refs #1735 · part of #1355

---

## SonarCloud gate (was red, now green)

The failing condition was exactly one: `new_security_rating` B, threshold A, driven by exactly one `VULNERABILITY` — `shell:S5332` (clear-text protocol) on `hook-install-wait.sh`'s `curl`. **Not** `go:S4036`; there was no such finding on this PR.

It is a false positive by design and by precedent — `$bind` is a loopback `host:port` for a daemon this rig starts itself, and the daemon's local API is plain HTTP with no TLS listener in existence; `run-cell.sh` carries the identical annotation on the identical call for the identical reason.

**The defect was my annotation, not the decision.** The `curl` was split across a continuation, so the finding sat on the first physical line while the `NOSONAR` rode the second — and a `NOSONAR` suppresses only the line it sits on. `run-cell.sh` says exactly that, immediately above the fetch it learned it from. I reproduced a half-wired lever inside the fix for a half-wired lever. The filter is now a variable and the call is one line.

Fixed without suppressions in the same pass:

- `shelldre:S1481` / `SC2034` (unused local `idx`) — the manifest's middle field is projected away with `awk` instead of being read into a variable that then needed a disable comment for each of two linters.
- `go:S3776` ×3 (cognitive complexity 17/18/23) — `ScanBeaconPackage` splits into `productionSources` / `importsTheBeacon` / `declaredAdapterName`; `adapterNameConst` sheds `adapterNameInDecl` / `adapterNameInSpec`; the corpus's assertions become `assertScanRefused` / `assertScanReported`.

One behaviour change fell out of that refactor and is **pinned rather than trusted**: `importsTheBeacon` walks every source before answering instead of returning at the first beacon import. Map order is unspecified, so an early return makes "imports the beacon **and** has a file that will not parse" answer `(true, nil)` or `(false, err)` depending on which file came first — a check dropped *nondeterministically*, which is worse than one dropped always because it cannot be reproduced. New corpus row: under the early-return mutation it fails **23 of 30** runs; restored, 30 of 30 green. The live tripwire was re-checked against a driver mutation after the refactor and still reddens.

**Result:** `new_security_rating` 2 → 1, quality gate `OK`, **zero** open Sonar issues on this PR.

## CodeScene (advisory)

Its one **critical** finding — Bumpy Road Ahead on `adapterNameConst`, in code this PR introduced — is fixed: one walk level per function. That gate now passes (5 passed, up from 4).

Two **advisory** findings are left alone deliberately, per AGENTS.md's "advisory, not a merge gate; a red result is a prompt to look closer":

- *Large Method* on `TestScanBeaconPackageNamesEveryWrongShape` — it is a corpus, and its length is its value, the same shape `righome_corpus_test.go` and `core/architecture_hookbody_shapes_test.go` already have. Shortening it means deleting rows.
- *String Heavy Function Arguments* on `launchWindowCarries(window, envVar string)` — a predicate over two strings has two string parameters.


ingo-eichhorst:
## QA — merging. My hypothesis was wrong, and the real cause is more interesting

I guessed `go:S4036` by analogy with PR #1739. There was **no S4036 finding on this PR at all**. It was one MINOR `shell:S5332` — enough on its own to take `new_security_rating` A→B.

**The finding was a false positive; the defect was the annotation's placement.** The `curl` was split across a continuation, so the finding sat on line 133 and the `# NOSONAR` rode 134 — and a NOSONAR suppresses only the line it sits on. Verified in the fix:

```sh
# The FILTER IS A VARIABLE so the curl stays ONE PHYSICAL LINE. That is not
# style: a NOSONAR annotation suppresses only the line it sits on, so a call
# split across a continuation carries the finding on the first line while the
# annotation rides the second. This file's first draft did exactly that and
```

That is a half-wired lever reproduced inside the fix for a half-wired lever — the same shape as this PR's own beacon finding and #1739's `COPILOT_HOME`. Worth noticing how often that pattern recurs here: a mechanism that is present, looks wired, and silently does nothing.

**The best thing in this round is a bug found in the author's own refactor.** `importsTheBeacon` initially returned at the first beacon import, so "imports the beacon **and** has a file that won't parse" answered `(true, nil)` or `(false, err)` depending on **map iteration order** — a check dropped *nondeterministically*, which is worse than one dropped always, because it cannot be reproduced. Pinned with a corpus row rather than trusted: **23 of 30 runs fail** under the early-return mutation, **30 of 30** green restored. The live tripwire mutation (deleting kiro-cli's passthrough) was also re-run after the refactor.

**CodeScene's one critical finding was fixed** (Bumpy Road on `adapterNameConst`) and that gate now passes. Two advisory ones deliberately remain, with reasons I agree with: a corpus test whose length **is** its value — shortening it means deleting rows, the same shape as `architecture_hookbody_shapes_test.go` — and a two-string predicate taking two string parameters.

Every required check green. Real home re-verified unchanged; daemon 25848 and Irrlicht.app 28044 untouched; the two `~/.vibe/logs/session/` dirs left in place for the user to decide on.

Follow-ups filed rather than folded in: **#1753** (the `hooktoml` multi-line-array refusal — a shipped defect) and **#1754** (the three rig gaps that let a hook-free recording look healthy).

Issue #1365, opened 2026-08-09
fix(hooks): claudecode installs hooks with no upstream version gate; codex's codexSupportsHooks has no counterpart


ingo-eichhorst:
Parent: #1355 — Phase B, work item 3. Depends on #1356 (same files — land that first).

## The gap

codex gates hook installation behind `codexSupportsHooks`. **claudecode does not gate at all** — it writes hook entries whatever version is installed. Verify both against the code and quote the lines.

Consequence at rollout scale: each new adapter either copies a bespoke version check or silently writes entries an older CLI will reject or ignore, and the user sees "granted" with a channel that never fires.

## Scope

1. A **declared minimum upstream version** per hook-capable adapter, checked before install.
2. Pin the **installed event set** to that version — do not write an entry the installed CLI does not know.
3. Below the minimum: do not install, say why (route it through #1362's Apply-error surfacing if that has landed), and stay at tier-5 rather than half-installing.
4. Make it part of the **declaration**, not per-adapter code. Adapter #3 should get this by declaring a version string, not by copying a function.

**Complexity:** 5

PR #1755, opened 2026-08-22
fix(vibe): model multi-line arrays in hooktoml, so hooks install into a config vibe itself wrote (#1753)

### LINKED ISSUE #1753: [bug]: mistral-vibe's hooks permission cannot install into a config.toml vibe itself wrote — Apply fails on multi-line arrays while consent still reads granted
Found while recording hook-bearing fixtures for #1735 (PR #1752), on a real
config rather than a fixture. **This means the mistral-vibe hook channel shipped
in #1718/#1733 does not work for a realistic user config**, and fails in the one
way #1362 exists to surface: silently, with consent still reading `granted`.

## The defect

`hooktoml`'s splicer refuses, by design, any file containing a construct it does
not model safely (`core/adapters/inbound/agents/hooktoml/hooktoml.go:54`):

```
ErrUnsafe = errors.New("hooktoml: file contains a construct this splicer does not
model safely (multi-line string, multi-line array, or unterminated quote) —
refusing rather than guessing")
```

Refusing rather than guessing is the right instinct — that is the parse-failure
rule in `docs/testing-philosophy.md`, and a splicer that guessed would corrupt a
user's config.

The problem is what it collides with. mistral-vibe's own `config.toml`,
as written by vibe itself, is full of multi-line arrays. Measured on the
maintainer's real config:

```
$ awk '/^[[:space:]]*[A-Za-z0-9_.-]+[[:space:]]*=[[:space:]]*\[[[:space:]]*$/ {n++} END {print n+0}' ~/.vibe/config.toml
12
```

Twelve. So `Apply` returns `ErrUnsafe`, `enable_experimental_hooks` is never
written, vibe never reads `hooks.toml`, and **no hook ever fires** — while the
permission continues to read `granted`. The only surface is `unapplied_grants`.

The recording in PR #1752 was unblocked by collapsing those arrays **in the
scratch seed**, which is a workaround for the fixture, not a fix for users. It is
documented as such in `agent-home.sh`.

## Why this was invisible until now

- `hooktoml`'s own tests, including its property test, generate documents the
  splicer models — so they exercise the refusal path as a *guard*, not as a thing
  that fires on the normal input.
- The contract families grade the install against fixtures the adapter's tests
  construct, not against a config vibe itself produced.
- #1362's "granted but NOT applied" surfacing works — it is why this was findable
  at all — but nothing routinely looks at `unapplied_grants`, and `run-cell.sh`
  checks it only on the attach path (a separate gap, noted in #1752).

## Shape of a fix — needs deciding, not assuming

At least three routes, and they are not equivalent:

1. **Teach the splicer multi-line arrays.** Widens the byte-range editor to a
   construct it currently refuses. `docs/testing-philosophy.md` requires a
   property test for anything that emits bytes from a structural diff, and
   `TestSplice_PropertyRandomMutations` is the committed shape — note its own
   history: seven hand-written round-trip tests passed while ~11% of randomly
   shaped documents got `,,` written into them, and a fourth defect survived
   until the generator varied *arrays* and multi-item removals. Any widening here
   owes a generator that actually varies the new axis.
2. **Write `enable_experimental_hooks` without splicing** — e.g. a separate
   mechanism or an append that does not require modelling the whole document.
   Cheaper, but needs to not corrupt a file vibe rewrites.
3. **Decide the permission cannot be honoured on such a config and say so** —
   turn the silent `ErrUnsafe` into a consent-time refusal with a reason the user
   can act on, per #1365's version-gate precedent ("granted but NOT applied,
   because …"). This is the honest floor even if (1) or (2) also lands.

(3) is arguably owed regardless: whatever the splicer can and cannot model, a
user whose config it refuses should be told, not left with a granted permission
and a dead channel.

## Test obligation

A defect test must be seen RED before the fix — and the input must be a config
**vibe actually wrote**, not a hand-minimised one, since a minimised fixture is
exactly what hid this. Commit that config as a fixture.

Found in #1735 / PR #1752. Part of #1355.

ingo-eichhorst:
Fixes #1753.

## The defect, reproduced

`hooktoml`'s splicer refused any document containing a multi-line array. mistral-vibe's own writer emits twelve of them in the maintainer's `config.toml` — and the first, `applied_migrations`, sits at **line 44 of 392, above every header**, which is the only region `topLevelKeyLine` ever scans. So the hooks permission's `Apply` returned `ErrUnsafe`, `enable_experimental_hooks` was never written, vibe never read `hooks.toml`, and no hook ever fired — with the permission still reading `granted`.

**Red before the fix**, against a `config.toml` vibe itself wrote (committed verbatim as `core/adapters/inbound/agents/vibe/testdata/real-config-2.19.1.toml`, one redaction — a telemetry `client_key`):

```
=== RUN   TestRealVibeConfigFixture_StillCarriesTheConstructThatBrokeIt
--- PASS: TestRealVibeConfigFixture_StillCarriesTheConstructThatBrokeIt (0.00s)
=== RUN   TestCountMultiLineArrayOpens_Corpus
--- PASS: TestCountMultiLineArrayOpens_Corpus (0.00s)
=== RUN   TestEnsureHooksInstalled_IntoAConfigVibeItselfWrote
    hookinstaller_realconfig_test.go:166: EnsureHooksInstalled into a config vibe itself wrote: hooktoml: file contains a construct this splicer does not model safely (multi-line string, multi-line array, or unterminated quote) — refusing rather than guessing
--- FAIL: TestEnsureHooksInstalled_IntoAConfigVibeItselfWrote (0.00s)
=== RUN   TestUninstallHooks_FromAConfigVibeItselfWrote
    hookinstaller_realconfig_test.go:244: EnsureHooksInstalled: hooktoml: file contains a construct this splicer does not model safely (multi-line string, multi-line array, or unterminated quote) — refusing rather than guessing
--- FAIL: TestUninstallHooks_FromAConfigVibeItselfWrote (0.00s)
```

## Routes taken, and rejected

**Route 1 — teach the splicer multi-line arrays. Taken.** `scanDocument` now carries one integer of bracket depth across lines, so a multi-line array reads as a single logical line. The rule that makes it safe rather than merely permissive: **nothing inside an array is classified as a header, a comment or a key.** `topLevelKeyLine` returns the **logical** range, through the value's closing line, so a splice can never leave orphaned array syntax behind.

**Route 2 — write the flag without splicing. Rejected.** A blind append cannot honour "set an *existing* key" (vibe writes `enable_experimental_hooks` itself once it has been touched in the UI, and appending a second top-level assignment is what TOML forbids), and appending without modelling the document is exactly the guess this package exists to refuse. It also buys nothing that route 1 does not: the same scan is needed to know where the preamble ends.

**Route 3 — a consent-time refusal with an actionable reason. Half of it was already there, and I disagree with the issue's framing on which half.** The *surfacing* exists and needs nothing new: `PermissionService.runClosureEffect` records `err.Error()` verbatim into `effectErrs`, `PermissionView.EffectError` and `PermissionsSnapshot.UnappliedGrants[].Reason` carry it, both wizards render "granted but NOT applied, because `<reason>`", and `planAnswerLocked` re-runs a previously-failed effect on a re-answer. #1365's version gate rides that same path. What was missing was purely the **content** of the reason: the old sentinel named neither the file, nor the line, nor what the daemon was trying to do. So refusals are now `*UnsafeConstructError` (wrapping `ErrUnsafe`, so `errors.Is` still works) carrying path, 1-based line and construct, plus the edit the caller wanted:

```
hooktoml: /Users/x/.vibe/config.toml:3 contains a multi-line (triple-quoted) string, which
this splicer does not model safely — refusing rather than guessing; simplify that line (or
make the edit by hand) and grant again — the edit it wanted to make is `enable_experimental_hooks = true`
```

Also newly refused: a **multi-line inline table**. TOML 1.0 does not permit one, and the old scanner did not refuse it — it silently mis-read the second line as a fresh top-level key. Same class of "cannot model", now honest about it.

## Property test — which axes the generator varies

Both generators in `hooktoml/property_test.go` gained the multi-line-array axis, varied on five sub-axes: **element count** (0-3, including empty `[\n]`); **position** (inside a `[[hooks]]` block, inside a `[table]`, and — the case the real defect lived in — in the document PREAMBLE above every header, plus after the flag key as well as before it); **nesting** (an element that is itself a multi-line array); **decoration** (a trailing comment on an element line, a comment-only line inside the array, a comment on the opener); and **adversarial element text**.

Census, printed on every run: 1539/2000 documents carry an array (994 in the preamble) for the `[[hooks]]` generator; 1636/2000 (1365 in the preamble) for the scalar one. Local sweep before landing: 5 seeds × 20 000 iterations × both tests, all clean.

**Which properties survive which mutation:** "no document this generator emits is refused" (the property the axis exists for — every array-carrying iteration failed here before this change); "every unrelated block/array/table survives verbatim" under both operations; idempotence and byte-equality on a second run. The refusals are pinned *separately*, in `TestScannerRefusals_Corpus`, since a generator emitting them would be asserting the opposite property inside the same loop.

## The mutation that came back green, and what it taught

The first version of the widened generator produced arrays in 1539/2000 documents and **still** did not grade the rule that makes the widening safe. Deleting `scanDocument`'s `if depth == 0` guard — so lines *inside* an array get classified as headers — left the entire suite green, because every adversarial element was a **quoted string**, and `"[[hooks]]"` starts with a quote and is header-shaped to nobody. The shape that bites is a bare nested array written as the last element with no trailing comma: its line reads `[1, 2]`, which is exactly what `isTableHeaderLine` matches. The generator now emits that, and two deterministic tests pin it at the two places a spurious header actually changes bytes. `docs/testing-philosophy.md` gained a paragraph: state a generator's axis in the vocabulary of *what the code matches on*, not of the format.

## Mutations, applied and reverted one at a time

| Mutation | Result |
|---|---|
| A. `scanDocument` classifies continuation lines (`if depth == 0` → `if true`) | **RED** — `TestHookBlocks_HeaderShapedArrayElementDoesNotSplitABlock`, `TestEnsureBoolTrue_InsertsAfterAPreambleArrayNotInsideIt`, `TestSetTopLevelBool_PropertyRandomMutations` |
| B. `topLevelKeyLine` classifies continuation lines | **RED** — `TestEnsureBoolTrue_ReplacesTheWholeOfAMultiLineValue`, scalar property test |
| C. drop the multi-line-inline-table refusal | **RED** — corpus row `multi-line_inline_table`, both entry points |
| D. drop the unbalanced-`]` refusal | **RED** — corpus row `stray_closing_bracket` |
| E. drop the array-never-closed refusal | **RED** — corpus row `array_never_closed` |
| F. report an unclosed array at EOF instead of its opening line | **RED** — corpus line assertion (`points at line 4, want 2`) |
| G. `topLevelKeyLine` returns the physical opening line, not the logical range | **RED** — `enable_experimental_hooks = true` left with two orphaned array lines below it |
| H. minimise the committed fixture (collapse all 12 arrays) | **RED** — `TestRealVibeConfigFixture_StillCarriesTheConstructThatBrokeIt` (`opens 0 multi-line arrays, want >= 12`) |
| I. the generator stops emitting arrays | **RED** — `assertArrayAxisWasExercised` (`only 0 of 2000 …`) |
| J. the refusal stops naming the file and the line | **RED** — `TestScannerRefusals_Corpus`, every refusing row |

Two guards go red under A and would have shipped without it, which is the whole reason the mutation is run rather than described.

Selected actual output:

```
=== MUTATION A (re-run): scanDocument classifies continuation lines ===
--- FAIL: TestEnsureBoolTrue_InsertsAfterAPreambleArrayNotInsideIt (0.00s)
    hooktoml_test.go:711: the flag was spliced INSIDE the user's preamble array:
        matrix = [
        enable_experimental_hooks = true
          [1, 2]
        ]

=== MUTATION G: topLevelKeyLine returns the physical opening line only ===
--- FAIL: TestEnsureBoolTrue_ReplacesTheWholeOfAMultiLineValue (0.00s)
    hooktoml_test.go:748: EnsureBoolTrue left orphaned array syntax behind:
        got:
        a = 1
        enable_experimental_hooks = true
          "nonsense",
        ]
        b = 2

=== MUTATION H: the committed fixture is minimised (arrays collapsed) ===
--- FAIL: TestRealVibeConfigFixture_StillCarriesTheConstructThatBrokeIt (0.00s)
    fixture opens 0 multi-line arrays, want >= 12 — it has been minimised, which is exactly what hid #1753

=== MUTATION I: the generator stops emitting multi-line arrays ===
--- FAIL: TestEnsureAndUninstall_PropertyRandomMutations (0.64s)
    multi-line-array axis: 0/2000 documents carried one, 0 of those in the preamble
    only 0 of 2000 generated documents carried a multi-line array — the #1753 axis is not being exercised
```

## A lock changed on purpose

`TestEnsureInstalled_RefusesOnMultiLineArray` pinned the OPPOSITE verdict on its seed. It is replaced by `TestEnsureInstalled_ModelsAMultiLineArray`, **keeping the seed byte-for-byte**, so the reversal reads as a reversal rather than as a test that quietly stopped existing.

## What is NOT covered

- **The recording.** PR #1752's scratch seed still collapses its arrays; this PR does not re-record. With the splicer widened that workaround is no longer needed, and removing it (plus the `agent-home.sh` note that documents it as a workaround) is a follow-up that needs a real recording run.
- **`run-cell.sh` checks `unapplied_grants` only on the attach path** — the separate gap #1752 noted. Untouched here.
- **No contract-family assertion grades "the install works on a config the agent itself wrote."** This PR proves it for vibe with a committed fixture, but a second TOML adapter would have to remember to do the same. A registry tripwire over "every hook adapter declares a real-world config fixture" is the shape that would make it assertable rather than remembered — deliberately not built here.
- **Multi-line inline tables are refused, not modelled.** No adapter needs them and TOML 1.0 forbids them; if a dialect ever emits one, this becomes the same ticket again.
- **The `[[hooks]]` operations were never the failing half** — `hooks.toml` is written only by us, so its documents were always ones the scanner modelled. The widening covers them too, but the evidence there is the property test, not a real-world file.

## Gates

`go test ./core/... -race -count=1` clean (one `TestFSWatcher_EmitsEventsForTranscriptCreateAndModify` flake on the first run, green on re-run ×3 and on a full re-run — the known fsnotify re-Add flake, untouched by this diff). `tools/preflight.sh --only go` all PASS. `tools/preflight.sh --only arch` PASS (composite 7.9175 → 7.9177, no category regression). Pre-push hook gates PASS.

### DIFF
--- a/core/adapters/inbound/agents/hooktoml/hooktoml.go
+++ b/core/adapters/inbound/agents/hooktoml/hooktoml.go
@@ -191,6 +191,41 @@ loop:
 // a fragment and nothing else.
 func stripComment(line []byte) []byte { return scanLine(line).stripped }
 
+// refusalFor reports the refusal one scanned line earns, or nil. depthAfter is
+// the bracket depth this line leaves behind.
+//
+// Shared by BOTH scanners in this file — scanDocument and topLevelKeyLine —
+// because they must agree on exactly which documents are refusable and are
+// reached through different public entry points, so a drift between them would
+// show up as one operation succeeding while its sibling refuses the same file,
+// with nothing anywhere comparing the two. TestScannerRefusals_Corpus drives
+// every row through both, which is the assertion this function makes cheap to
+// keep true.
+func refusalFor(path string, lineNo int, sc lineScan, depthAfter int) error {
+	switch {
+	case sc.unterminated:
+		return unsafeAt(path, lineNo, "an unterminated quote")
+	case sc.braces != 0:
+		return unsafeAt(path, lineNo, "an inline table spanning more than one line")
+	case depthAfter < 0:
+		return unsafeAt(path, lineNo, "an unbalanced ']'")
+	}
+	return nil
+}
+
+// headerOf classifies a comment-stripped, trimmed line as a [[array]] or
+// [table] header. Callers must only ask it about a line at bracket depth
+// ZERO — inside a multi-line array a header-shaped line is an ELEMENT.
+func headerOf(trimmed []byte) (kind byte, name string, ok bool) {
+	switch {
+	case isArrayHeaderLine(trimmed):
+		return 2, string(bytes.TrimSpace(trimmed[2 : len(trimmed)-2])), true
+	case isTableHeaderLine(trimmed):
+		return 1, string(bytes.TrimSpace(trimmed[1 : len(trimmed)-1])), true
+	}
+	return 0, "", false
+}
+
 func isArrayHeaderLine(t []byte) bool {
 	return len(t) >= 4 && bytes.HasPrefix(t, []byte("[[")) && bytes.HasSuffix(t, []byte("]]"))
 }
@@ -288,37 +323,26 @@ func scanDocument(path string, src []byte) ([]scannedLine, error) {
 		lineNo++
 		raw := src[pos:end]
 		sc := scanLine(raw)
-		if sc.unterminated {
-			return nil, unsafeAt(path, lineNo, "an unterminated quote")
-		}
-		if sc.braces != 0 {
-			return nil, unsafeAt(path, lineNo, "an inline table spanning more than one line")
+		if err := refusalFor(path, lineNo, sc, depth+sc.brackets); err != nil {
+			return nil, err
 		}
 
 		li := scannedLine{start: pos, continuation: depth > 0}
 		if depth == 0 {
 			trimmed := bytes.TrimSpace(sc.stripped)
-			switch {
-			case isArrayHeaderLine(trimmed):
-				li.isHeader, li.headerKind = true, 2
-				li.headerName = string(bytes.TrimSpace(trimmed[2 : len(trimmed)-2]))
-			case isTableHeaderLine(trimmed):
-				li.isHeader, li.headerKind = true, 1
-				li.headerName = string(bytes.TrimSpace(trimmed[1 : len(trimmed)-1]))
-			default:
+			if kind, name, ok := headerOf(trimmed); ok {
+				li.isHeader, li.headerKind, li.headerName = true, kind, name
+			} else {
 				rawTrimmed := bytes.TrimSpace(raw)
 				li.isCommentOnly = len(rawTrimmed) > 0 && rawTrimmed[0] == '#'
+			}
+			if sc.brackets > 0 {
+				openedAt = lineNo
 			}
 		}
 		lines = append(lines, li)
 
-		if depth == 0 && sc.brackets > 0 {
-			openedAt = lineNo
-		}
 		depth += sc.brackets
-		if depth < 0 {
-			return nil, unsafeAt(path, lineNo, "an unbalanced ']'")
-		}
 		if end >= len(src) {
 			break
 		}
@@ -699,28 +723,22 @@ func topLevelKeyLine(path string, src []byte, key string) (start, end int, found
 		}
 		lineNo++
 		sc := scanLine(src[pos:lineEnd])
-		if sc.unterminated {
-			return 0, 0, false, unsafeAt(path, lineNo, "an unterminated quote")
-		}
-		if sc.braces != 0 {
-			return 0, 0, false, unsafeAt(path, lineNo, "an inline table spanning more than one line")
+		if err := refusalFor(path, lineNo, sc, depth+sc.brackets); err != nil {
+			return 0, 0, false, err
 		}
 		if depth == 0 {
 			trimmed := bytes.TrimSpace(sc.stripped)
-			if isArrayHeaderLine(trimmed) || isTableHeaderLine(trimmed) {
+			if _, _, isHeader := headerOf(trimmed); isHeader {
 				return 0, 0, false, nil
 			}
 			logicalStart = pos
 			k, _, ok := bytes.Cut(trimmed, []byte("="))
 			matching = ok && string(bytes.TrimSpace(k)) == key
-		}
-		if depth == 0 && sc.brackets > 0 {
-			openedAt = lineNo
+			if sc.brackets > 0 {
+				openedAt = lineNo
+			}
 		}
 		depth += sc.brackets
-		if depth < 0 {
-			return 0, 0, false, unsafeAt(path, lineNo, "an unbalanced ']'")
-		}
 		if depth == 0 && matching {
 			return logicalStart, lineEnd, true, nil
 		}
--- a/core/adapters/inbound/agents/hooktoml/hooktoml.go
+++ b/core/adapters/inbound/agents/hooktoml/hooktoml.go
@@ -26,45 +26,120 @@
 // comments the same way json.MarshalIndent over a bare map did before
 // hookjson existed.
 //
-// Anything the scanner cannot confidently classify — a triple-quoted
-// string, a multi-line array, an unterminated single-line string — makes it
-// refuse the WHOLE document rather than guess. Unlike hookjson, which falls
-// back to a lossy whole-document re-encode when a splice is not safe, there
-// is no safe fallback encoder here: a generic map[string]interface{}-style
-// TOML dump is precisely the read-modify-marshal this package exists to
-// avoid, so it is never reached for as a fallback. "Cannot splice safely"
-// therefore means "do not write," not "write something worse" — refusing an
-// install into an exotic hand-crafted hooks.toml is the safe direction; a
-// corrupted one is not.
+// A MULTI-LINE ARRAY is modelled (issue #1753); a triple-quoted string, an
+// unterminated single-line string and a multi-line inline table are not, and
+// make the scanner refuse the WHOLE document rather than guess. The
+// multi-line array had to move from the second list to the first because it
+// is not exotic at all: mistral-vibe's OWN writer emits them — twelve in the
+// config.toml measured when #1753 was filed — so refusing them meant refusing
+// every realistic user config, with the hooks permission still reading
+// `granted` and no hook ever firing. Modelling one costs a single integer of
+// state carried across lines (see scanDocument): a logical line continues
+// while its bracket depth is above zero, and NOTHING inside it is classified
+// as a header, a comment or a key.
+//
+// Unlike hookjson, which falls back to a lossy whole-document re-encode when
+// a splice is not safe, there is no safe fallback encoder here: a generic
+// map[string]interface{}-style TOML dump is precisely the read-modify-marshal
+// this package exists to avoid, so it is never reached for as a fallback.
+// "Cannot splice safely" therefore means "do not write," not "write something
+// worse" — refusing an install into an exotic hand-crafted hooks.toml is the
+// safe direction; a corrupted one is not. What #1753 changed is WHICH
+// documents land in that bucket, not what happens to the ones that do — and
+// for those, the refusal now names the file and the line (UnsafeConstructError)
+// so the "granted but NOT applied, because …" the wizards already render
+// (#1362) points at something the user can act on.
 package hooktoml
 
 import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
 )
 
-// ErrUnsafe means this document contains a construct the line-oriented
-// scanner does not model with confidence — a triple-quoted string, an
-// unterminated single-line string, or a multi-line array. The caller must
-// not write; there is no lossy fallback (see the package doc).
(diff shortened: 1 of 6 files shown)


ingo-eichhorst:
Follow-up commit `499a8766`: CodeScene flagged `topLevelKeyLine` as a complex method. Looking closer found a real problem behind the score — `scanDocument` and `topLevelKeyLine` each carried their own copy of the same four refusals and their own copy of the header match, and they are reached through **different** public entry points, so a drift would surface as one operation succeeding while its sibling refuses the same file with nothing comparing the two. `refusalFor` and `headerOf` are now the single place each decision is made; `TestScannerRefusals_Corpus` already drives every row through both entry points, which is what keeps the agreement asserted rather than assumed.

Two more mutations, seen red after the refactor (one per run):

```
=== MUTATION K: the SHARED refusal reports nothing ===
--- FAIL: TestHookBlocks_RefusesOnUnterminatedCommand
    HookBlocks on an unterminated command string: got 1 blocks and no error, want a refusal
--- FAIL: TestScannerRefusals_Corpus/unterminated_basic_quote
    HookBlocks accepted a document it must refuse
    TopLevelBool accepted a document it must refuse
  (+ unterminated_literal_quote, multi-line_inline_table, stray_closing_bracket)

=== MUTATION L: both scanners classify continuation lines ===
--- FAIL: TestHookBlocks_HeaderShapedArrayElementDoesNotSplitABlock
--- FAIL: TestEnsureBoolTrue_InsertsAfterAPreambleArrayNotInsideIt
    the flag was spliced INSIDE the user's preamble array
```

CodeScene's remaining flags are on the **test** files (`TestScannerRefusals_Corpus`, the two property tests, `TestEnsureHooksInstalled_IntoAConfigVibeItselfWrote`) — table-driven corpora and property loops are the shape this repo's testing philosophy asks for, so those are left as they are. Per AGENTS.md, CodeScene is advisory and not a merge gate.

ingo-eichhorst:

**Secret scan on the committed fixture first**, since this commits a real config. Only one credential-shaped value and it is explicitly redacted (`client_key = "sdk-REDACTED-BY-IRRLICHT-FIXTURE"`); the remaining long strings are model and config names. Clean.

**Fixture fidelity:** the committed `real-config-2.19.1.toml` opens **12** multi-line arrays; the maintainer's live `~/.vibe/config.toml` opens **12**. It is the real shape, not a minimised one — which is the whole point, since a minimised fixture is what hid this defect.

**The fix works:** `TestEnsureHooksInstalled_IntoAConfigVibeItselfWrote` and `TestUninstallHooks_FromAConfigVibeItselfWrote` both PASS, and both were pasted RED before the fix.

**I reproduced the vacuity finding**, because it is the most valuable thing in this PR. Mutation A — drop the `depth == 0` guard so lines inside arrays get classified:

```
--- FAIL: TestHookBlocks_HeaderShapedArrayElementDoesNotSplitABlock
    block 0 lost the field below its multi-line array — a line inside the array
    was classified as a section header
--- FAIL: TestEnsureBoolTrue_InsertsAfterAPreambleArrayNotInsideIt
--- FAIL: TestSetTopLevelBool_PropertyRandomMutations
```

The property test catches it. That is the point: **before sharpening, it did not** — and that is a correction to guidance I gave.

## The correction, which is better than my rule

I wrote that "a generator that does not vary the new axis is a vacuous green wearing a different hat." This generator **did** vary the new axis — 1539 of 2000 documents carried a multi-line array — and was still vacuous, because every adversarial element was a *quoted string*, and `"[[hooks]]"` is header-shaped to nobody. The shape that actually bites is a bare nested array as the last element with no trailing comma, `[1, 2]`, which is what `isTableHeaderLine` matches.

So the sharper rule is **vary the spelling the code classifies on, not the axis's name** — plus a census guard so the generator proves it still emits the shape. Landing that in `docs/testing-philosophy.md` is the right home for it.

## Route 3: I was right that something was owed and wrong about what

The surfacing already existed end-to-end — `runClosureEffect` → `effectErrs` → `PermissionView.EffectError` / `UnappliedGrants[].Reason` → both wizards → re-answer retry, the same path #1365's version gate rides. So "turn the silent `ErrUnsafe` into a consent-time refusal" mis-describes it: it was never silent and always consent-time. It was **unactionable**. Naming path, 1-based line, construct and the edit it wanted is the actual fix.

## The finding worth the most beyond this PR

**Nothing makes "the install works on a config the agent itself wrote" an obligation.** vibe now has such a fixture only because this issue forced one; claudecode, codex, copilot and geminicli all still grade against configs their own tests construct — the same blind spot, one adapter at a time. Filing that separately.

Also good: CodeScene (advisory) pointed at `topLevelKeyLine`, and looking closer found something real — the two scanners each carried their own copy of four refusals and the header match, reached through different public entry points, so a drift would surface as one operation succeeding while its sibling refuses the same file. Extracted, with two more mutations red.

Every required check green.

Issue #1756, opened 2026-08-22
test(hooks): every adapter grades its install against a config its own tests wrote — the one blind spot that shipped #1753


ingo-eichhorst:
#1753 was a shipped defect: mistral-vibe's hook install could not succeed on a
`config.toml` **vibe itself writes**, because `hooktoml` refused multi-line arrays
and a real config has twelve. `Apply` failed while consent read `granted`.

It was found by accident, during a recording run (#1735), not by any test.

## Why no test caught it

Every adapter's hook-install tests grade against a config **the adapter's own
tests construct**. Those configs are, by definition, documents the author already
knew how to handle — the same set they got right. The contract families
(`AssertHookEndpointFollowsBindAddr`, `AssertHookDisclosureMatchesInstalled`, …)
inherit that, because they are wired against fixtures the adapter supplies.

This is the same shape `docs/testing-philosophy.md` already records for
byte-emitting code — *"hand-written cases encode what the author already thought
of"* — one level up: not "which mutations does the generator produce" but "which
**documents** does the adapter ever see".

#1755 gave vibe a real fixture
(`core/adapters/inbound/agents/vibe/testdata/real-config-2.19.1.toml`, 392 lines,
byte-for-byte upstream except a redacted telemetry key) and a guard,
`TestRealVibeConfigFixture_StillCarriesTheConstructThatBrokeIt`, that fails if the
fixture is ever minimised into uselessness.

**But it has that only because #1753 forced it.** claudecode, codex, copilot and
geminicli have nothing equivalent. The blind spot is intact for four of five
hook-installing adapters, and it is invisible in exactly the same way it was for
vibe: green tests, granted permission, dead channel.

## Shape

A registry tripwire — the same family as `TestEveryHookInstallDeclaresAVerifier`,
`TestEveryHookInstallDeclaresAVersionFloor` and the newer
`TestEveryHookInstallWiresItsContractFamilies` — asserting that **every
hooks-installing adapter declares a real-world config fixture** and exercises
install + uninstall against it.

Two things to decide rather than assume:

- **What makes a fixture "real".** #1755's answer is a committed file captured
  from the agent's own output with secrets redacted, plus a guard asserting it
  still carries the construct that mattered. That guard is adapter-specific by
  nature; the tripwire can require *a* guard exists without dictating what it
  asserts.
- **How to keep it honest over time.** A fixture captured once drifts as the
  upstream CLI's config format evolves — which is the same staleness problem
[...]

Issue #1754, opened 2026-08-22
test(rig): three attempts produced healthy-looking hook-free recordings — promote, unapplied_grants, and the coexist precondition each could have said so


ingo-eichhorst:
#1735 (PR #1752) closed the hook-coverage GAP for kiro-cli and mistral-vibe. It
took three attempts, and **every failed attempt produced a complete,
healthy-looking recording with zero hook events** — `driver_exit_reason=ok`,
`completeness=complete`, and nothing anywhere saying the channel under test never
fired.

That is the failure mode `docs/testing-philosophy.md` is most hostile to: the
thing under test never executed, and the output was indistinguishable from
success. The four underlying defects are fixed or filed (#1753 for the vibe
install refusal). This issue is about the three places the **rig** could have
said so and didn't.

## 1. `promote-recording.sh` accepts a hook-free recording of a hooks-declaring adapter

The strongest of the three, because it is the last gate before a fixture becomes
committed truth. A promote-time check — *"this adapter declares a hooks
permission and this recording contains no `hook_received` event; intended?"* —
would have caught defects 1 and 2 on the **first** run rather than the third.

It has to be a prompt or an explicit override rather than a hard failure: a
hook-free recording of a hooks-declaring adapter is legitimate when the scenario
genuinely produces no hook (and `of coverage --hooks` already reports the
population), so the check should make the operator say "yes, intended" rather
than block.

## 2. `run-cell.sh` checks `unapplied_grants` only on the attach path

`unapplied_grants` is the **only** surface for an install that failed — #1362's
"granted but NOT applied". The spawn path is the one the recording rig actually
uses, and it does not look. Checking it there would have named the `hooktoml`
refusal (#1753) immediately and by name, instead of it being diagnosed by
elimination across three runs.

This is cheap and is the highest value-per-line of the three.

## 3. The `record` skill's coexist precondition reads as if coexist mode is adapter-neutral

It is not. `run-cell.sh`'s own comment claims hooks reach a coexisting daemon
"since #1178, the endpoint follows `IRRLICHT_BIND_ADDR`". That is true for
**`DeliveryURL`** adapters, where the installed entry carries the address — and
false for **`DeliveryAddressFree`** ones, where the entry deliberately carries no
address and `irrlichd hook-post` resolves the daemon from the addr file **at fire
time**. The beacon is a child of the CLI, hence of the tmux pane, so in coexist
mode it reads the *production* addr file and the hooks land on the wrong daemon.

[...]

Issue #1719, opened 2026-08-20
feat(opencode): hook-based state detection requires a shipped plugin — the config-file route was removed upstream


ingo-eichhorst:
Broken out of epic #1355, where this was cut from the rollout and deferred.

## The config-file route is closed upstream — a plugin is the only option

`experimental.hook` existed in opencode v0.6.0 and is **gone by v1.2.0**, and the live schema sets
`"additionalProperties": false` on `experimental`, so writing one is now a hard schema violation,
not merely unsupported. A **shipped plugin** is therefore the only route, not a fallback.

That makes this the most expensive adapter in the set and the one whose shape most resembles pi
(#1721): irrlicht would have to ship and install a code artifact into the agent's own plugin
system, with everything that implies for consent, versioning, uninstall and trust.

**Do not start this before #1721 (pi) has settled the shipped-artifact question** — the two share
that problem and should share its answer rather than inventing two.

## Corrections to numbers previously quoted

opencode is **21 hook keys** (~15 real callbacks + 4 registration surfaces), not the "25+" that
circulated; that figure conflated hooks with the 32-member SDK event bus delivered through the
single `event` hook.

Reference churn: the repo moved `sst/opencode` → `anomalyco/opencode`. Do not cite the old path.

## Measured

opencode 1.14.50 installed. Adapter is onboarded. Known observability gaps already on record:
errors surface on `message.data.error`; headless slash-commands arrive as literal text
(`driver_gap`); settle-after-exit is `known_failing`. Metrics are unverifiable through `of verify`
because opencode is a ProcessOwnedStore adapter — the replay golden cannot carry them.

## Shape of this ticket

**Audit first, implement second, in this same issue.** A negative result is a legitimate close: if the audit shows no usable mechanism, close with that evidence recorded. Do not open a separate audit ticket.

Absence of a flag in `--help` is **not** evidence of absence — most of these are config-file driven. Verify by installing a test hook/handler and firing it, and read type definitions from source rather than rendered docs (two independent passes in #1355 found rendered docs wrong or truncated where the source was right).

## If a mechanism exists, what shipping it owes

The machinery is built and shared; this should be closer to a declaration than an implementation:

- `HookSpec` on the agent declaration, one config writer, one event→signal map.
[...]

PR #1758, opened 2026-08-22
feat(pi): observe pi through an extension irrlicht installs, not a config entry (#1721)

### LINKED ISSUE #1721: feat(pi): pi has a first-class extension system with turn and tool events — the "permanent tier-5" verdict stopped one layer too early
Broken out of epic #1355, which recorded pi as **"the permanent tier-5 case"**. That verdict was
reached by auditing pi's *lifecycle event list* and finding no permission or notification event —
correct as far as it went, and it stopped one layer too early.

## What the audit missed: pi has a first-class extension system

Measured against the installed binary (**pi 0.83.0**,
`@earendil-works/pi-coding-agent`), not from docs:

```
pi install <source>   Install extension source and add to settings   (npm: / git: / https: / ./local/path)
pi remove <source>    Remove extension source from settings
pi list               List installed extensions from settings
--extension, -e <path>   Load an extension file (repeatable)
--no-extensions, -ne     Disable extension discovery
```

Extensions are TypeScript modules that "subscribe to lifecycle events, register custom tools, add
commands, and more". `pi install ./local/path` is the route that matters: **irrlicht could ship an
extension and install it locally**, exactly as it installs hook entries elsewhere.

## The event set is rich enough for state detection

From `docs/extensions.md` in the installed package — `pi.on(...)`:

| need | event |
|---|---|
| session lifecycle | `session_start` (reason: `startup`/`reload`/`new`/`resume`/`fork`), `session_shutdown` |
| turn boundaries | `turn_start`, `turn_end` |
| turn done | `agent_start`, `agent_end`, **`agent_settled`** |
| tool boundaries | `tool_execution_start`, `tool_execution_update`, `tool_execution_end`, `tool_call`, `tool_result` |
| user input | **`input`** — carries `source` (`interactive`/`rpc`/`extension`) and `streamingBehavior` (`undefined` when idle, `steer` mid-stream, `followUp` when queued) |
| compaction | `session_before_compact`, `session_compact` |

`agent_settled` plus `turn_end` is a clean turn-done signal, which is more than several shipped
adapters get from their transcripts.

## The open question — do NOT assume the answer

**Does pi have any blocked-on-user state at all?** #1355's reasoning was that pi has no approval
prompt because it has no sandbox (`docs/security.md`: tools run with the process's permissions), so
there is nothing to block on. That reasoning still stands for *permissions*.

But `ask_question` appears in pi's own docs three times as an `--exclude-tools` example
(`README.md:655`, `docs/usage.md:289`, `docs/sdk.md:518`). **Measured:** it is NOT in 0.83.0's
built-in set — `pi --help` says "read, bash, edit, write tools", the dist carries `"bash"`,
`"read"`, `"list"`, `"write"`, `"edit"`, `"grep"`, `"glob"` as quoted tool identifiers and does not
carry `"ask_question"` as one, and no bundled example defines it. So it is either an extension tool,
a removed tool, or docs drift.

**Marked unresolved, not decided.** Settle it live before designing anything: if some tool or flow
does block on the user, `tool_execution_start`/`_end` around it is the `waiting` signal; if nothing
does, pi is a two-state adapter (`working`/`ready`) and that is a legitimate, useful outcome — say
so and ship it rather than leaving pi at tier 5.

## The hard part is not the events — it is shipping a code artifact

This is the first adapter where irrlicht would install **executable code** into an agent, not a
config entry. That raises questions the existing machinery does not answer, and this issue owes an
answer to each **before** any extension is written:

- **Consent.** What does the `KindModify` permission say it is doing? "Writes 7 hook entries" has an
  obvious meaning; "installs a TypeScript extension that runs inside your agent" does not, and #570's
  contract requires the copy to match what is actually done. `AssertHookDisclosureMatchesInstalled`
  has no notion of a code artifact.
- **Where it lives, and uninstall.** `pi install` writes to settings; `ManagedUserFile` currently
  declares one path plus an uninstall. Does removing the extension mean `pi remove`, or editing
  settings, or both — and is that idempotent?
- **Versioning.** An extension is compiled against pi's extension API. What happens when pi's API
  changes? `agent.VersionGate` exists for exactly this and would need a floor.
- **Delivery.** The extension needs the daemon's address. This is the first natural consumer of
  `core/pkg/hookbeacon` (`DeliveryAddressFree`), which is built, contract-covered, and adopted by
  nobody — see `hook_endpoint_addressfree_test.go`, still a *reference* wiring.
- **Trust.** `pi install` has `--approve`/`--no-approve` for project-local files. Understand what
  approval means before automating around it.

**#1719 (opencode) has the same shipped-artifact problem** and is deliberately sequenced behind this
one. Whatever answer lands here should be reusable there rather than invented twice.

## Acceptance

Either pi moves off tier 5 with a working extension and its consent story written down, or this
issue closes with a measured statement of which of the five questions above blocks it. Both are good
outcomes; what is not acceptable is leaving "permanent tier-5" on the record when the extension
system was never examined.

## Shape of this ticket

## Recording cost

Part of #1355.

ingo-eichhorst:
Closes #1721. Part of #1355.

pi moves off epic #1355's "permanent tier-5" verdict. It gets an authoritative turn-end
signal, delivered by a **JavaScript extension irrlicht installs** into pi's own
auto-discovery directory — the first adapter whose install is code rather than a config
entry.

## The audit, and why the answer is "build it" where opencode's was "don't"

Everything below is measured against the installed package
(`@earendil-works/pi-coding-agent` **0.83.0**, `$(npm root -g)/@earendil-works/…`), reading
`dist/` source rather than rendered docs, plus live runs against pi 0.83.0 in an isolated
`PI_CODING_AGENT_DIR`.

**pi has no config-file hook mechanism.** `hook` does not appear in `docs/settings.md`;
`settings.json` has no hooks key. The only lifecycle surface is `pi.on(...)` inside an
extension module. So the choice was not "config vs. code" — it was "code, or leave pi at
tier 5".

**But `pi install` is not the route, and that is the whole finding.** #1721 was written
around `pi install ./local/path`, which writes a `packages`/`extensions` entry into
`settings.json` — a distribution problem. There is a *second*, independent mechanism:
`dist/core/extensions/loader.js`'s `discoverAndLoadExtensions` globs
`<agentDir>/extensions/` directly, gated by `isExtensionFile`, which matches **`.ts` OR
`.js`**. Writing one file there is the entire install.

That is what separates pi from opencode (#1719):

| | opencode (#1719, cut) | pi (this PR) |
|---|---|---|
| distribution | publish/ship an npm plugin package | one file, embedded in the irrlichd binary via `go:embed` |
| registry / network | yes | none |
| dependencies | a graph | one `node:child_process` import |
| build step / toolchain | yes | none (`.js`, and pi loads through jiti anyway) |
| agent config touched | `settings.json` | nothing |
| uninstall | `pi remove` + settings edit | `os.Remove` of one file |

**Blast radius is bounded by pi itself**, source-read: `loader.js` wraps the whole load in
`try/catch` and reports a failing extension as a load error while the agent continues, and
`runner.js` wraps every handler dispatch the same way — for every event **except
`tool_call`**, whose loop has no catch and whose handler can return `{block: true}`. So the
shipped file never subscribes to `tool_call`, and a committed guard enforces that.

The honest residue, which the consent copy states rather than this PR body: it is still
code, running in-process in the user's agent with the user's permissions.

### Does pi have a blocked-on-user state? Measured: effectively no.

- No tool-approval prompt — pi has no sandbox (`docs/security.md`).
- **`ask_question` is docs drift.** `grep -rlo '"ask_question"' dist/` → **0 files**, against
  18 for `"bash"` and 14 for `"read"`. #1721's measurement reproduced independently.
- pi *does* have one built-in blocking prompt: the **project trust** prompt, surfaced as the
  `project_trust` event. It is unusable, structurally rather than for want of effort: it
  fires before a session exists and its context is explicitly reduced —
  "cwd, mode, hasUI, and select/confirm/input/notify UI helpers", no `sessionManager`, so no
  transcript path, so nothing to correlate to. The live event walk confirms `session_start`
  is the first event carrying a session file at all.

**pi is therefore a two-state adapter through hooks (working/ready)** — a legitimate,
useful outcome, as #1721 said it would be. `waiting` stays reachable exactly where it
already was: the transcript-tail prose cue.

### Events installed: exactly one

`agent_settled` → `HandleStopHook`. pi's own docs: "Use `agent_settled` for status
integrations that need to know Pi will not continue running automatically" — no retry, no
auto-compaction, no queued follow-up. Semantically Claude Code's `Stop`, and mapped on that
meaning rather than on a wire name.

Deliberately **not** installed, each with its reason in `installedHookEvents`' doc comment:
`tool_call` (blocking, no catch); `tool_execution_start`/`_end` (a broad release is only
useful paired with a narrow assert on the same signal from the same adapter — pi has no
approval prompt to assert on, so it would spawn a beacon per tool call to release a signal
nothing ever holds — vibe's own reasoning for `after_tool`); `session_start`/`turn_*`/`input`
(the transcript already opens the turn).

## Delivery: address-free (beacon)

`DeliveryAddressFree`. The shipped JavaScript spawns `irrlichd hook-post pi` and knows
nothing about hosts or ports; the beacon resolves the daemon's address at fire time. That
keeps the *artifact* minimal, which is the supply-chain argument: no HTTP client, no
addr-file parsing, and nothing in the user's agent that can go stale when the daemon moves.

**Correction to the issue's framing:** #1721 says the beacon is "adopted by nobody… still a
*reference* wiring". That was true at #1373 and is not true now — geminicli (#1717), kirocli
(#1716) and vibe (#1718) all declare `DeliveryAddressFree`, as `docs/testing-contracts.md`'s
delivery bullet already says. pi is the **fourth** adopter. The comments in this PR say so.

## The `HookInstaller` seam, one step further out

vibe (#1734) widened `ReadEntries`/`EndpointOfRaw` for a config that is never a decoded
generic structure. pi's install is not a *document* at all: `ReadEntries` returns the whole
file as the single entry, and `EndpointOfRaw` reads the beacon command back out with the
same anchored regexp the installer uses to decide staleness. It fits unchanged, which is
decent evidence #1734 generalised far enough. `docs/testing-contracts.md` updated.

## Live end-to-end proof

The production renderer + production installer, into an isolated `PI_CODING_AGENT_DIR`, with
a stub `irrlichd` that logs its argv and stdin; driven by real pi 0.83.0 against a **free
local LM Studio model** (no credits spent):

```
--- real pi 0.83.0 turn, isolated HOME, free local model ---
pi exit=0
hello
--- what the stub beacon received ---
ARGV: --version hook-post pi
STDIN: {"hook_event_name":"agent_settled","transcript_path":"/tmp/pi-e2e-35634/home/.pi/agent/sessions/--private-tmp-pi-e2e-35634-work--/2026-08-22T03-59-41-209Z_01a0279f-d499-7ef2-8b44-c042ea23b74f.jsonl"}
--- transcripts pi actually wrote ---
/tmp/pi-e2e-35634/home/.pi/agent/sessions/--private-tmp-pi-e2e-35634-work--/2026-08-22T03-59-41-209Z_01a0279f-d499-7ef2-8b44-c042ea23b74f.jsonl
```

Exactly one delivery per settled turn, exactly the payload `piHookPayload` decodes, naming
the transcript pi actually wrote — whose filename stem is the id the fswatcher keys on.
Nothing under the operator's real `~/.pi` was written at any point.

## Bug fixed along the way: irrlicht ignored `PI_CODING_AGENT_DIR`

pi composes its sessions dir as `getAgentDir() + "/sessions"`, and `getAgentDir` honours
`$PI_CODING_AGENT_DIR` — the documented way to relocate a whole pi installation, and the one
this audit used. irrlicht read only `$PI_CODING_AGENT_SESSION_DIR`, so a user who relocated
pi that way had their sessions moved out from under a watcher still pointed at
`~/.pi/agent/sessions`. It matters more now: the hook receiver confines caller-supplied paths
to that root, so a wrong root *rejects every hook* rather than merely missing transcripts.

## Mutation evidence — every guard seen red

One mutation per run, applied and reverted; `git status --porcelain` checked each time.

<details><summary><b>1. `sessionsDir` ignores <code>PI_CODING_AGENT_DIR</code> (the pre-#1721 code)</b> — the defect test</summary>

```
--- FAIL: TestSessionsDirFollowsPiCodingAgentDir (0.00s)
    adapter_test.go:49: sessionsDir() = ".pi/agent/sessions", want "/tmp/pi-relocated/agent/sessions"
```
</details>

<details><summary><b>2. shipped extension subscribes to pi's blocking <code>tool_call</code></b></summary>

```
--- FAIL: TestShippedExtensionSubscribesToNoBlockingEvent (0.00s)
    extension_test.go:62: extension.js references pi's "tool_call" event, which can block, cancel, replace or rewrite what the agent is doing. A monitoring extension must only observe.
--- FAIL: TestShippedExtensionSubscribesExactlyToInstalledHookEvents (0.00s)
    extension_test.go:103: extension.js subscribes to [agent_settled tool_call], installedHookEvents declares [agent_settled] — the consent copy, the version floor and every contract family are written against the second list
```
</details>

<details><summary><b>3. JS subscribes to <code>turn_end</code> while Go still declares <code>agent_settled</code></b></summary>

```
--- FAIL: TestShippedExtensionSubscribesExactlyToInstalledHookEvents (0.00s)
    extension_test.go:103: extension.js subscribes to [turn_end], installedHookEvents declares [agent_settled] — the consent copy, the version floor and every contract family are written against the second list
```
</details>

<details><summary><b>4. Go renames the json tag the shipped JS sends</b> — the one only this guard catches</summary>

```
--- FAIL: TestShippedExtensionPayloadMatchesTheReceiversFields (0.00s)
    extension_test.go:148: extension.js sends [hook_event_name transcript_path], piHookPayload decodes [hook_event_name path] — a field on either side that the other does not have is a silently dropped hook

--- and every contract family stays GREEN, which is why the guard exists ---
ok  	irrlicht/core/adapters/inbound/agents/pi	0.206s
```

Every contract family builds its body by marshaling `piHookPayload` itself, so they rename in
lockstep and stay green while production delivery is broken.
</details>

<details><summary><b>5. shipped extension acquires a node_modules dependency</b></summary>

```
--- FAIL: TestShippedExtensionImportsOnlyNodeBuiltins (0.00s)
    extension_test.go:166: extension.js imports "node-fetch" — only node: builtins may be imported, or this install acquires a dependency graph
```
</details>

<details><summary><b>6. <code>renderExtension</code> concatenates the command instead of <code>strconv.Quote</code>ing it</b></summary>

```
--- FAIL: TestRenderedExtensionSurvivesAHostilePath (0.00s)
    extension_test.go:228: path "/tmp/irr\"licht/irrlichd": round-trip failed (ok=false, got "", want "'/tmp/irr\"licht/irrlichd' --version hook-post pi >/dev/null || true")
    extension_test.go:240: path "/tmp/irr\"licht/irrlichd": the rendered literal is not a valid string literal: invalid character 'l' after top-level value
    extension_test.go:228: path "/tmp/irr\\licht/irrlichd": round-trip failed (ok=false, got "", want "'/tmp/irr\\licht/irrlichd' --version hook-post pi >/dev/null || true")
    extension_test.go:240: path "/tmp/irr\\licht/irrlichd": the rendered literal is not a valid string literal: invalid character 'l' in string escape code
    extension_test.go:228: path "/tmp/irr\nlicht/irrlichd": round-trip failed (ok=false, got "", want "'/tmp/irr\nlicht/irrlichd' --version hook-post pi >/dev/null || true")
    extension_test.go:232: no beacon assignment line in the rendered extension
```
</details>

<details><summary><b>7. install overwrites a file irrlicht did not write</b></summary>

```
--- FAIL: TestEnsureInstalled_RefusesToOverwriteAFileItDidNotWrite (0.00s)
    hookinstaller_test.go:147: EnsureHooksInstalled overwrote a foreign file without complaint
```
</details>

<details><summary><b>8. uninstall deletes any file at our path <i>and</i> install also touches pi's <code>settings.json</code></b> — two mutations in one run (deviation, noted)</summary>

```
--- FAIL: TestUninstall_RemovesOursAndLeavesForeignAlone (0.00s)
    hookinstaller_test.go:193: uninstall over a foreign file = (true, <nil>), want (false, nil)
    hookinstaller_test.go:197: uninstall removed or altered a file it did not write
--- FAIL: TestInstallLeavesAPiWrittenSettingsFileUntouched (0.00s)
    hookinstaller_test.go:335: install changed .pi/agent/settings.json, which is not the file it declares it writes
    hookinstaller_test.go:358: after uninstall .pi/agent/settings.json no longer matches what pi wrote
```

The house rule is one mutation per run; these two landed in the same call. Both went red on
their own assertion, so the evidence is readable, but the deviation is stated rather than
smoothed over.
</details>

<details><summary><b>9. adapter declares the WRONG delivery route (<code>DeliveryURL</code>)</b></summary>

```
--- FAIL: TestHookEndpointFollowsBindAddr/endpoint_follows_bind_addr (0.00s)
    hook_endpoint.go:263: delivery is identical on the default and 127.0.0.1:7838 bind addresses ("'…/pi.test' --version hook-post pi >/dev/null || true") — the endpoint does not follow IRRLICHT_BIND_ADDR
--- FAIL: TestHookEndpointFollowsBindAddr/install_writes_resolved_port (0.00s)
    hook_endpoint.go:265: event agent_settled: "'…/pi.test' --version hook-post pi >/dev/null || true" does not carry the resolved port ":7838/"
--- FAIL: TestHookEndpointFollowsBindAddr/default_port_install_upgraded_in_place (0.00s)
    hook_endpoint.go:267: expected modified=true when repointing a default-port install …
```
</details>

<details><summary><b>10. <code>Verify</code> never reports a stale install</b></summary>

```
--- FAIL: TestVerifyAgreesWithEnsureInstalled/ours_but_naming_another_binary (0.00s)
    hookinstaller_test.go:255: verify reported outstanding=false (missing=[] stale=[]), want true
    hookinstaller_test.go:264: verify says outstanding=false but ensure modified=true — the two readers of the same file disagree
```
</details>

<details><summary><b>11. receiver drops its own <code>transcripts</code> consent check</b> — only the silence assertion survives #1488</summary>

```
--- FAIL: TestHookReceiver_DeniedTranscriptsConsentDropsQuietly (0.00s)
    hooks_test.go:278: a transcripts-denied hook logged 1 error(s): [hook body dropped: permission "transcripts" is not granted at the decode — either this receiver skipped the check it declares, or consent was revoked mid-request (issue #1488)]
```

`TestHookReceiver_PermissionGateContract` stayed **green**, exactly as
`docs/testing-contracts.md` documents: the #1488 backstop absorbs the dispatch, so silence is
the surviving discriminator.
</details>

Three registry tripwires also fired unprompted and were fixed rather than worked around —
`TestEveryHookInstallWiresItsContractFamilies` (pi did not wire `AssertPermissionGated`, and
the exclusion's premise re-check caught it), `TestEveryBeaconAdapterDriverPassesTheDaemonAddress`,
`TestRigHomeTableMatchesTheAdapterRegistry`. They worked.

## #1756: graded against a config pi itself wrote

`testdata/real-pi-settings.json` is a verbatim `~/.pi/agent/settings.json` from pi 0.83.0
(nothing redacted — nothing in it is a secret; pi keeps credentials in a separate
`auth.json`, deliberately not captured). Its `packages` array is what `pi install` writes, so
the fixture is a home where pi's *other* extension mechanism is already in use.
`TestInstallLeavesAPiWrittenSettingsFileUntouched` snapshots the **whole tree** before and
after — not just the file someone remembered to look at — installs, diffs, then uninstalls
and asserts the tree is byte-identical to what pi wrote.

## Recording: none, and what one would need

**Zero cells re-recorded** (#1355 Phase D's kill criterion is >5). Every existing pi golden is
byte-identical; `tools/replay-fixtures.sh` passes unchanged. Landing this deliberately records
nothing, so hookcov will report pi as `StatusGap` — the honest reading, same as #1716/#1717/#1718.

A recording would now need: `PI_CODING_AGENT_DIR` staging with a seeded `auth.json` (optin,
because pi keeps credentials under the same override), and `IRRLICHT_BIND_ADDR` reaching the
tmux pane — **both are in this PR**, so the rig gap #1735 measured for mistral-vibe is closed
for pi before the first recording rather than after it. #1754's remaining rig gaps are
untouched.

## Not covered

- **Nothing here executes the JavaScript.** Running it needs node (which `go test ./core/...`
  does not otherwise require — a gate that could be absent is the failure mode
  `docs/testing-philosophy.md` is about) or pi itself (which needs a model call). The
  end-to-end proof above was run by hand; the committed guards pin the properties it
  confirmed. Stated in `extension_test.go`'s header, not left implied.
- **The waiting cue is a payload-free floor.** `agent_settled` carries no message content and
  the extension deliberately does not go reading `ctx.sessionManager` for one — every extra
  line is code shipped into a user's agent. Same trade vibe made; it can never *clear* a cue
  the transcript found.
- **A pi version newer than the daemon** could ship an extension subscribing to a second
  event; that lands in `IgnoreUnknownEvent`'s counter rather than silently.
- `TestReadLauncherEnv_Tmux_DetachedResolvesNoHost` (processlifecycle) flaked once in a full-
  suite run and passes in isolation — pre-existing, unrelated.

## Gates

`go test ./core/... -race -count=1` clean; `tools/preflight.sh --only go` / `--only arch` /
`--only bash` all PASS (ARS composite 7.9179 → 7.9157, not a regression); `of validate` OK;
rig smoke test PASS.

---

# SonarCloud: 13.2% duplication on new code — what it is, and why most of it must stay

`Failed conditions: 13.2% Duplication on New Code (required ≤ 3%)`.

Measured per file and per block, not guessed:

```
tools/sonarqube-report.sh "measures/component_tree?component=ingo-eichhorst_Irrlicht\
&pullRequest=1758&metricKeys=new_duplicated_lines,new_lines&qualifiers=FIL&ps=100"
tools/sonarqube-report.sh "duplications/show?key=<component>&pullRequest=1758"   # per file
```

| file | dup lines | new lines | dup % |
|---|---:|---:|---:|
| `pi/hooks_test.go` | 106 | 281 | 37.7% |
| `pi/hooks.go` | 68 | 227 | 30.0% |
| `pi/hookversion_test.go` | 42 | 98 | 42.9% |
| `pi/hookinstaller.go` | 30 | 424 | 7.1% |
| `pi/hookinstaller_test.go` | 26 | 391 | 6.6% |
| `pi/hookpath_test.go` | 18 | 39 | 46.2% |
| `pi/agent.go` | 15 | 62 | 24.2% |
| everything else (13 files) | 0 | | 0% |
| **total** | **305** | **2317** | **13.2%** |

**The single most important fact: not one duplicated block has a partner inside this PR.**
Every partner is another adapter — `vibe/hooks.go`, `codex/hookinstaller.go`,
`kirocli/hookversion_test.go`, and so on. There is no copy-paste *within* pi to remove. What
the metric is measuring is that a seventh hook adapter looks like the other six, which is what
`docs/testing-contracts.md` asks for: an adapter "wires one call **instead of porting a test
file**".

## Real, and fixed

**`hookversion_test.go` — three hand-written tests the shared contract already asserts, more
thoroughly than they did.** Found by being made to look at the number, and it is a genuine
defect in the *design* of five other adapters too. `AssertHookVersionGate` runs
`floor_refuses_an_older_cli` and `unknown_version_fails_open` as sub-tests. My
`TestHooksGate_BelowFloorSaysWhy` / `AtOrAboveFloorInstalls` / `UnknownVersionFailsOpen` were
strictly weaker restatements running one call below them:

| | shared contract | the copy I had added |
|---|---|---|
| versions below the floor | **three** field-wise predecessors (major, minor, patch decremented independently — a single synthetic "just below" exercises only the borrow path) | one hand-picked `0.82.1` |
| refusal reason | non-empty **and** names both versions | contains both versions |
| at-or-above the floor | `Permits(Min)` must be allowed, **plus a vacuity guard**: a floor with nothing below it fails outright | `Permits(Min)` allowed |
| unreadable version | **three** spellings (`""`, `"not a version"`, `"2.1"`) | one (`""`) |

Deleted (40 lines), with the reasoning left in the file rather than only here. Coverage went
**up**: the five contract arms still run and now nothing restates them. Proven by mutation
rather than asserted — weakening the floor to a version with nothing below it, the exact case
the deleted `AtOrAboveFloorInstalls` would have passed:

```
--- FAIL: TestHooksPermission_DeclaresVersionGate/floor_covers_every_installed_event
    hook_version.go:59: declared floor 0.0.0 is below 0.80.4, the version "agent_settled" is known at …
--- FAIL: TestHooksPermission_DeclaresVersionGate/floor_refuses_an_older_cli
    hook_version.go:62: declared floor 0.0.0 has no version below it, so this obligation asserts
        nothing — a floor of 0.0.0 permits every CLI and is not a gate
```

`TestHooksGate_ProbeIsPiVersion` **stays**: it is the one thing the contract does not cover —
`assertVersionSourceDeclared` only requires `Observed` or `Probe` to be non-empty, never that
the argv is the right one.

The same three are redundant in claudecode, codex, geminicli, kirocli and vibe. Removing them
there is a follow-up, not something a pi ticket should do to five other adapters' test files.

## Real, but extracting it here would not help — measured, not assumed

**`atomicWriteFile` (30 lines) exists in seven places**: claudecode, codex, copilot,
geminicli, kirocli, processlifecycle and now pi. That is textbook shared-helper boilerplate
and it genuinely wants a home in `core/pkg/`.

It is not done here, for a reason this report itself provides. The duplication data shows
`pi/hookinstaller.go:L401-423` already matching **`hookjson`'s sibling
`hooktoml/hooktoml.go:L879`** — a shared-package file. So moving this code into a shared
package while six copies remain does not remove a duplicate; it adds a seventh partner, and
the metric would be unchanged. The fix has to migrate all seven at once, which is a tree-wide
refactor touching six adapters unrelated to pi. Worth doing; not in a pi PR, and it could not
change this gate's outcome either way (30 of 305 lines).

## Intended, and collapsing it would be worse

Named block by block, so a reader can check rather than take my word.

**`hooks.go` L107-174 ↔ every other receiver.** `NewHookHandler` + `transcriptConfiner` +
`serveHookRequest`'s method check → hooks gate → `ObserveHookReceipt` → transcripts gate →
`DecodeConfined`. The shared parts are *already* extracted — that is what `hookjson` is. What
remains is the **call sequence**, and its order is the thing four contract families grade.
`docs/testing-contracts.md` records which mutation reddens which obligation, including that
hoisting the receipt above the hooks gate reddens
`consent_denied_request_is_not_counted`. Extracting the sequence into a shared function would
mean the receiver no longer *contains* the checks the contracts grade: the contract would
grade the helper once, and a receiver that skipped it would be invisible — the thing under
test and the thing testing it on the same side. #1488 already moved as much of this into a
chokepoint as is safe, and its own note records the cost: the backstop quietened every
dispatch-shaped consent assertion, leaving *silence* as the only discriminator. Mutation 11
above reproduces exactly that.

**`hookpath_test.go` L16-33 ↔ `vibe/hookpath_test.go`.** The file is 39 lines and 18 of them
are the single `AssertHookPathConfined` call. This is the design working. The alternative —
the state before #1361 — is a receiver nobody wired the family for, which is how the
statusline endpoint shipped unconfined.

**`hooks_test.go` L115-138 / L120-138 / L184-209 ↔ vibe, copilot, geminicli, kirocli.** The
`AssertHookReceiverPermissionGated` wiring closure, and the hand-written
denied-consent-logs-no-error test. The latter is hand-written **per receiver on purpose**:
AGENTS.md records that since #1488 "each of the four receivers' hand-written consent tests now
asserts no error was logged, and each was seen red again with its gate deleted". Sharing it
would delete the only per-adapter proof that survives the backstop.

**`hooks_test.go` L3-58 ↔ `vibe/hooks_test.go`.** The test doubles: `mockTarget`, `keyedGate`,
`mockLogger`. `keyedGate` staying local is explicit repo policy — AGENTS.md: "The static
`keyedGate` map literals in the adapter test packages are a different thing and **stay**: they
pin a fixed two-permission combination in one-off tests and were never key-blind."
`mockLogger` (5 lines) has a shared equivalent in `contracttesting.RecordingLogger`, but
swapping only pi's would make one of six adapters spell it differently for 5 lines; that is a
tree-wide consolidation or nothing.

**`hookinstaller_test.go` L36-61 ↔ four adapters.** The install-type `AssertPermissionGated`
wiring — required by `TestEveryHookInstallWiresItsContractFamilies`, which failed this PR
until it was added.

**`agent.go` L34-64 / L49-75 ↔ kirocli, codex.** The `agent.Agent` registration literal. It is
a declaration; two adapters declaring the same *shape* with different values is what a registry
looks like.

## The gate cannot be met, and here is the arithmetic

**Measured after the fix, not estimated** — SonarCloud re-scanned commit `5f94a54`:

```
qualitygates/project_status?projectKey=…&pullRequest=1758
  new_duplicated_lines_density: 12.7 (required GT 3)
```

| file | dup lines (before → after) |
|---|---|
| `pi/hooks_test.go` | 106 → 106 |
| `pi/hooks.go` | 68 → 68 |
| `pi/hookinstaller.go` | 30 → 30 |
| `pi/hookinstaller_test.go` | 26 → 26 |
| `pi/hookpath_test.go` | 18 → 18 |
| `pi/agent.go` | 15 → 15 |
| `pi/hookversion_test.go` | **42 → 31** |
| **density** | **13.2% → 12.7%** |

I predicted ~11.5% in an earlier draft of this section and was wrong; the real move was
0.5pp, not 1.7pp. Correcting it rather than leaving the estimate, because a figure nothing
produces is the drift this repo's own philosophy is about. The reason the deletion moved so
little is worth stating: what remains in `hookversion_test.go` is `L10-40 ↔
claudecode:L3, kirocli:L13, geminicli:L14` — the imports, the `AssertHookVersionGate` call
itself, and the `hooksVersionGate` helper that reads the gate off the real registration. That
is the *contract wiring*, i.e. the intended category. The redundant restatements are gone;
what is left is the one call the docs tell you to write.

That is also the strongest argument in this section. A genuine, coverage-increasing fix moved
the number by half a point, because the mass is structural. Reaching 3% would mean eliminating
roughly **225 of the remaining 294 lines**: the receiver sequence, the test doubles, and every
contract-family wiring. Each is load-bearing for a reason recorded in
`docs/testing-contracts.md`, and collapsing them would trade a metric for the ability of six
contract families to see a seventh adapter.

**So: I am not chasing this one to green.** Recording the reason here, as agreed, rather than
merging over it silently. Reviewers who want to re-derive the split can run the two
`tools/sonarqube-report.sh` calls at the top of this section.

**CodeScene** is also red and advisory. There is no local way to fetch its delta (the
CodeScene GitHub App posts it; this repo has no API token and `/ir:codescene-report` no longer
exists), so I did not read it — stated as an assumption rather than a finding: it is *likely*
reporting the same near-identical-adapter mass, and that is not verified. The deterministic
architecture measure, `ars-gate`, passed with no composite or category regression
(7.9179 → 7.9157).

## Also, in the same pass

`TestReadLauncherEnv_Tmux_DetachedResolvesNoHost` (processlifecycle) fails only under full
`./core/...` load. Characterised rather than waved away: this PR's diff touches **zero files**
in that package (`git diff --stat origin/main...HEAD -- .../processlifecycle/` is empty), and
the test passed 6/6 isolated runs and 3/3 whole-package runs. Load-dependent, pre-existing,
unrelated.

**No pi cells were recorded**, deliberately — `hookcov` reporting `StatusGap` for pi is the
honest outcome and is meant to stay visible.

### DIFF
--- a/core/adapters/inbound/agents/beaconroute_test.go
+++ b/core/adapters/inbound/agents/beaconroute_test.go
@@ -8,6 +8,8 @@ import (
 	"irrlicht/core/adapters/inbound/agents/copilot"
 	"irrlicht/core/adapters/inbound/agents/geminicli"
 	"irrlicht/core/adapters/inbound/agents/kirocli"
+	"irrlicht/core/adapters/inbound/agents/pi"
+	"irrlicht/core/adapters/inbound/agents/vibe"
 	"irrlicht/core/pkg/hookbeacon"
 )
 
@@ -26,8 +28,9 @@ import (
 // while its route segment is "claudecode". That is why the map keys below are
 // written out literally rather than derived from the adapter constants.
 //
-// Scope, stated honestly: this pins the receivers that exist (five as of
-// #1716's kiro-cli row). It cannot fail for a NEW receiver registered at a
+// Scope, stated honestly: this pins the receivers that exist (seven as of
+// #1721's pi row, which also added the mistral-vibe row #1718 never wrote —
+// exactly the omission the paragraph below predicts). It cannot fail for a NEW receiver registered at a
 // different prefix, because registerHookRoutes (core/cmd/irrlichd/startup.go)
 // is hand-wired with no registry to enumerate and this map is hand-written.
 // Whoever adds the next receiver adds its row here; until then the
@@ -39,6 +42,10 @@ func TestBeaconEndpointPathMatchesTheInstalledReceivers(t *testing.T) {
 		"copilot":    copilot.HookEndpointPath,
 		"gemini-cli": geminicli.HookEndpointPath,
 		"kiro-cli":   kirocli.HookEndpointPath,
+		// mistral-vibe was beacon-delivered from the day it shipped (#1718)
+		// but was never added here; found while adding pi's row (#1721).
+		"mistral-vibe": vibe.HookEndpointPath,
+		"pi":           pi.HookEndpointPath,
 	} {
 		if got := hookbeacon.EndpointPath(segment); got != want {
 			t.Errorf("hookbeacon.EndpointPath(%q) = %q, but the receiver is registered at %q — the beacon would post into a route nothing serves", segment, got, want)
--- a/core/adapters/inbound/agents/pi/adapter.go
+++ b/core/adapters/inbound/agents/pi/adapter.go
@@ -21,9 +21,32 @@ const defaultRootDir = ".pi/agent/sessions"
 // directory itself (not a parent).
 const sessionDirEnvVar = "PI_CODING_AGENT_SESSION_DIR"
 
-// sessionsDir returns the directory the Pi adapter should watch —
-// $PI_CODING_AGENT_SESSION_DIR itself when that override is set (it names the
-// session directory, not a root above it), else defaultRootDir.
+// sessionsDir returns the directory the Pi adapter should watch, resolving
+// the same two overrides pi itself does and in the same order.
+//
+// pi composes the sessions directory as getAgentDir() + "/sessions"
+// (dist/config.js's getSessionsDir), where getAgentDir honours
+// $PI_CODING_AGENT_DIR; $PI_CODING_AGENT_SESSION_DIR then overrides the
+// result outright (dist/main.js reads it directly, and pi's own --help
+// describes --session-dir as overriding it).
+//
+// The $PI_CODING_AGENT_DIR leg is issue #1721's: before it, setting only
+// that variable — the documented way to relocate a whole pi installation —
+// moved pi's sessions while irrlicht kept watching ~/.pi/agent/sessions and
+// saw nothing. It matters more now than it did: the hook receiver confines
+// caller-supplied transcript paths to THIS root (hooks.go's
+// transcriptConfiner), so a wrong root does not merely miss transcripts, it
+// rejects every hook the extension delivers.
 func sessionsDir() string {
-	return agentpaths.FromEnv("pi", sessionDirEnvVar, defaultRootDir)
+	// FromEnv with an empty default is how "was this override set to an
+	// absolute path?" is asked without restating the absolute-path check and
+	// its warn-once logging.
+	if dir := agentpaths.FromEnv("pi", sessionDirEnvVar, ""); dir != "" {
+		return dir
+	}
+	return agentpaths.FromEnv("pi", agentDirEnvVar, defaultRootDir, sessionsSubdir)
 }
+
+// sessionsSubdir is the agent-dir-relative directory pi writes transcripts
+// into — the "sessions" in getSessionsDir's join(getAgentDir(), "sessions").
+const sessionsSubdir = "sessions"
--- a/core/adapters/inbound/agents/pi/adapter_test.go
+++ b/core/adapters/inbound/agents/pi/adapter_test.go
@@ -16,10 +16,67 @@ func TestSessionsDir(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv(agentDirEnvVar, "")
 			t.Setenv(sessionDirEnvVar, tc.env)
 			if got := sessionsDir(); got != tc.want {
 				t.Errorf("sessionsDir() = %q, want %q", got, tc.want)
 			}
 		})
+	}
+}
+
+// TestSessionsDirFollowsPiCodingAgentDir is issue #1721's defect test.
+//
+// pi composes its sessions directory as getAgentDir() + "/sessions"
+// (dist/config.js), and getAgentDir honours $PI_CODING_AGENT_DIR — the
+// documented way to relocate an entire pi installation, and the one this
+// issue's own live audit used. Before this, irrlicht read only
+// $PI_CODING_AGENT_SESSION_DIR, so a user who relocated pi that way had
+// their sessions moved out from under a watcher still pointed at
+// ~/.pi/agent/sessions, with nothing anywhere reporting it.
+//
+// It matters more now than it did as a pure observation gap: the hook
+// receiver confines caller-supplied transcript paths to THIS root
+// (hooks.go's transcriptConfiner), so a wrong root does not merely miss
+// transcripts — it refuses every hook the installed extension delivers,
+// which reads as "the channel is dead" rather than as a misconfiguration.
+func TestSessionsDirFollowsPiCodingAgentDir(t *testing.T) {
+	t.Setenv(sessionDirEnvVar, "")
+	t.Setenv(agentDirEnvVar, "/tmp/pi-relocated/agent")
+
+	want := "/tmp/pi-relocated/agent/sessions"
+	if got := sessionsDir(); got != want {
+		t.Errorf("sessionsDir() = %q, want %q", got, want)
+	}
+}
+
+// TestSessionDirOverrideBeatsAgentDirOverride pins the PRECEDENCE, which is
+// pi's own: dist/main.js reads $PI_CODING_AGENT_SESSION_DIR directly and
+// uses it in place of the composed path, so the narrower override wins.
+// Without this, "honour both" could mean either order and the wrong one
+// looks identical until someone sets both.
+func TestSessionDirOverrideBeatsAgentDirOverride(t *testing.T) {
+	t.Setenv(agentDirEnvVar, "/tmp/pi-relocated/agent")
+	t.Setenv(sessionDirEnvVar, "/tmp/pi-explicit-sessions")
+
+	want := "/tmp/pi-explicit-sessions"
+	if got := sessionsDir(); got != want {
+		t.Errorf("sessionsDir() = %q, want %q", got, want)
+	}
+}
+
+// TestSessionsDirRejectsARelativeAgentDir pins that the agent-dir leg
+// inherits the same absolute-path rule the session-dir leg has always had —
+// pi's own expandTildePath does no shell expansion either, so a relative or
+// tilde-prefixed value is a misconfiguration to fall back from, not a path
+// to join onto the current working directory.
+func TestSessionsDirRejectsARelativeAgentDir(t *testing.T) {
+	for _, v := range []string{"relative/agent", "~/pi/agent"} {
+		t.Setenv(sessionDirEnvVar, "")
+		t.Setenv(agentDirEnvVar, v)
+		if got := sessionsDir(); got != defaultRootDir {
+			t.Errorf("sessionsDir() with %s=%q = %q, want the default %q",
+				agentDirEnvVar, v, got, defaultRootDir)
+		}
 	}
 }
--- a/core/adapters/inbound/agents/pi/agent.go
+++ b/core/adapters/inbound/agents/pi/agent.go
@@ -1,6 +1,7 @@
 package pi
 
 import (
+	"irrlicht/core/adapters/inbound/agents/hookjson"
 	"irrlicht/core/domain/agent"
 	"irrlicht/core/domain/permission"
 )
@@ -24,6 +25,21 @@ const iconSVGDark = `<svg xmlns="http://www.w3.org/2000/svg" width="14" height="
   <line x1="60" y1="30" x2="64" y2="74" stroke="#E0E0E0" stroke-width="8" stroke-linecap="round"/>
 </svg>`
 
+// Source declares where Pi's transcripts live. A function, not a value, so
+// the hook receiver's confiner (hooks.go's transcriptConfiner) and the
+// daemon's fswatcher resolve the SAME root from the SAME declaration — a
+// second, constants-derived list is exactly the drift issue #1361 was about,
+// and sessionsDir() reads environment overrides that a package-level value
+// would freeze at init time.
+func Source() agent.Source {
+	return agent.FilesUnderRoot{
+		Dir: sessionsDir(),
+		Parser: agent.JSONLineParser{
+			NewParser: func() agent.LineParser { return &Parser{} },
+		},
+	}
+}
+
 // Agent returns the Pi adapter registration.
 func Agent() agent.Agent {
 	return agent.Agent{
@@ -37,12 +53,7 @@ func Agent() agent.Agent {
 			Match:         agent.ExactName{Name: ProcessName},
 			PIDForSession: DiscoverPID,
 		},
-		Source: agent.FilesUnderRoot{
-			Dir: sessionsDir(),
-			Parser: agent.JSONLineParser{
-				NewParser: func() agent.LineParser { return &Parser{} },
-			},
-		},
+		Source:  Source(),
 		Control: agent.Control{SupportsInput: true, Interrupt: agent.InterruptCtrlC},
 		Permissions: []agent.Permission{
 			agent.ControlPermission(),
@@ -56,6 +67,51 @@ func Agent() agent.Agent {
 					"derive session state, cost, and token metrics. Read-only — no " +
 					"file is ever modified. Toggling off stops all reading " +
 					"immediately.",
+			},
+			{
+				Key:             PermissionKeyHooks,
+				Kind:            permission.KindModify,
+				Title:           "Install status hooks",
+				FeatureUnlocked: "Authoritative turn-end detection",
+				// "1 hook entry" is what contracttesting's #1356 count arm
+				// reads, and it is the honest number: one subscription, to
+				// one event. What is unusual is the SHAPE that subscription
+				// takes, which is why Touches names it in the same breath
+				// rather than leaving it to Detail — the wizard row is what
+				// most users read.
+				Touches: "Writes 1 hook entry to " + displayExtensionPath +
+					" — a small JavaScript file that pi loads and runs",
+				Detail: "pi has no hooks section in its settings: the only way to observe " +
+					"its lifecycle is an extension, so this permission installs one. That " +
+					"is a single JavaScript file at " + displayExtensionPath + ", which pi " +
+					"auto-discovers and executes inside every pi session with your own " +
+					"permissions. It is code irrlicht ships in its own binary, not a " +
+					"package fetched from anywhere — no npm, no git, no `pi install` — and " +
+					"pi's own settings.json is not touched at all. The file imports only " +
+					"node:child_process, subscribes to exactly one pi event (agent_settled, " +
+					"which fires when pi will not continue running on its own) and does one " +
+					"thing with it: runs `irrlichd hook-post pi`, a tiny command irrlicht " +
+					"ships, handing it the session's transcript path. That command reads the " +
+					"daemon's own published address at the moment the hook fires, so nothing " +
+					"in the file names a host or a port and it cannot go stale; it also never " +
+					"blocks pi, even when the daemon is not running. The file deliberately " +
+					"does not subscribe to pi's blocking tool_call event, so it cannot " +
+					"interfere with a tool call. " +
+					hookjson.RequiresVersion("pi", minPiVersion) +
+					" Toggling off deletes the file (also available via " +
+					"`irrlichd --uninstall-hooks`); nothing else in your pi installation was " +
+					"changed, so there is nothing else to undo.",
+				Apply:  func() error { _, err := EnsureHooksInstalled(); return err },
+				Remove: func() error { _, err := UninstallHooks(); return err },
+				Writes: &agent.ManagedUserFile{
+					Path:      ExtensionPath,
+					Uninstall: UninstallHooks,
+					Verify:    VerifyHooksInstalled,
+					Version: &agent.VersionGate{
+						Min:   minPiVersion,
+						Probe: []string{"pi", "--version"},
+					},
+				},
 			},
 		},
 	}
--- a/core/adapters/inbound/agents/pi/extension.js
+++ b/core/adapters/inbound/agents/pi/extension.js
@@ -1,0 +1,94 @@
+// irrlicht session-status extension for pi — installed and owned by irrlichd.
+//
+// DO NOT EDIT. This file is rewritten whenever irrlicht's own binary path
+// changes and deleted when the "Install status hooks" permission is revoked
+// (also via `irrlichd --uninstall-hooks`).
+//
+// It exists because pi has no config-file hook mechanism at all: the ONLY
+// way to subscribe to pi's lifecycle is an extension module. See
+// hookinstaller.go's header for the audit that established that, and for why
+// this file is written directly into pi's auto-discovery directory rather
+// than installed with `pi install`.
+//
+// Everything in here is deliberate and narrow:
+//
+//   - It subscribes to exactly ONE event, agent_settled, and does exactly one
+//     thing with it: hand the session's transcript path to the beacon on
+//     stdin. It registers no tool, no command, no provider, no UI, and reads
+//     nothing.
+//
+//   - It never subscribes to `tool_call`. That is not a style preference:
+//     pi's own runner dispatches every OTHER event inside a try/catch and
+//     logs a throwing handler, but emitToolCall (dist/core/extensions/
+//     runner.js) has no catch, and a `tool_call` handler that returns
+//     `{block: true}` — or throws — BLOCKS the user's tool call. A monitoring
+//     extension must not be able to do that. extension_test.go's
+//     TestShippedExtensionSubscribesToNoBlockingEvent is the guard.
+//
+//   - Nothing in here is awaited on I/O. pi awaits each handler, so a
+//     handler that waited for the daemon to answer would stall the agent
+//     whenever irrlicht is slow or down. spawn() returns immediately and
+//     every failure path is swallowed.
+//
+// Plain JavaScript, not TypeScript, and no imports beyond node:child_process.
+// pi's loader accepts both (`isExtensionFile` matches .ts OR .js) and loads
+// through jiti, so no compile step and no toolchain is involved either way —
+// but a file with no types and no third-party import is a file whose whole
+// behaviour is readable in one screen, which is the point when the artifact
+// being installed is code.
+import { spawn } from "node:child_process";
+
+// The command irrlichd renders at install time: `<abs irrlichd> --version
+// >/dev/null && <abs irrlichd> hook-post pi >/dev/null || true`. It carries
+// NO host and NO port — the beacon reads the daemon's published address at
+// the moment the hook fires, so this file never goes stale when the daemon
+// moves. hookinstaller.go substitutes the literal below and reads it back
+// out again to decide whether an installed copy is current.
+const IRRLICHT_BEACON = "__IRRLICHT_BEACON_COMMAND__";
+
+// The single pi lifecycle event this extension subscribes to. agent_settled
+// means "pi will not continue running automatically" — no retry, no
+// auto-compaction, no queued follow-up left — which is the authoritative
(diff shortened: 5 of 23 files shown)


ingo-eichhorst:
## QA — merging, including over the SonarCloud duplication gate, with the reason recorded

### Verified independently

I re-ran the four load-bearing measurements myself against the installed 0.83.0 rather than accepting them:

```
loader.js:438  function isExtensionFile(name) {
loader.js:439      return name.endsWith(".ts") || name.endsWith(".js");
```

plus `discoverAndLoadExtensions` globbing the directory directly. **This corrects the premise I briefed this task with.** I framed pi as opencode's distribution problem — a TypeScript module installed through `pi install`. It is not: `.js` is accepted, discovery is a directory glob, and `pi install` is a separate mechanism that is not on this path. No registry, no npm/git, no build step, no toolchain, no `settings.json` write; uninstall is removing one file. The objection that would have disqualified this adapter does not apply.

Also reproduced: `ask_question` → **0** files in `dist/` against `"bash"` → **18** (docs drift, confirmed); `PI_CODING_AGENT_DIR` unhonored on `main` (0 occurrences) — and the consequence is worse than cosmetic, since the confiner roots on that dir, so a wrong root **rejects every hook**; and `beaconroute_test.go`'s map on `main` lists only `gemini-cli` and `kiro-cli`, so **mistral-vibe has been missing since #1718**. That last one is filed as #1759.

### The duplication gate: merging over it

The decisive fact, and it is the right one to have measured: **not one duplicated block has a partner inside this PR.** Every partner is another adapter. There was no copy-paste within pi to remove — the mass is cross-adapter and structural.

**One item was real, and the metric is what surfaced it.** Three hand-written version tests restated obligations `AssertHookVersionGate` already runs, strictly more weakly — one hand-picked version below the floor where the contract drives three field-wise predecessors; one unreadable spelling where the contract drives three; and no vacuity guard where the contract has one. Deleted, 40 lines, **coverage up**. Proven by mutation rather than asserted — weakening the floor to `0.0.0`, precisely the case the deleted test would have passed:

```
--- FAIL: .../floor_refuses_an_older_cli
    hook_version.go:62: declared floor 0.0.0 has no version below it, so this obligation
        asserts nothing — a floor of 0.0.0 permits every CLI and is not a gate
```

I confirmed pi still wires `AssertHookVersionGate` and keeps `TestHooksGate_ProbeIsPiVersion` (which the contract does *not* cover — it only requires `Probe` non-empty, never that the argv is right). #1740's `TestEveryHookInstallWiresItsContractFamilies` is an independent guarantee that no wiring was lost in the deletion.

**And that fix is the strongest argument for stopping.** A genuine, coverage-increasing change moved the number **0.5pp** (12.7% measured after rescan — the author corrected their own "~11.5%" draft estimate rather than leaving a figure nothing produces). Reaching 3% needs ~225 of the remaining 294 lines gone: the consent→receipt→confine sequence whose **order** four contract families grade, the contract wirings themselves, and the per-receiver hand-written consent tests that `docs/testing-contracts.md` says are hand-written **on purpose** because they are the only per-adapter proof surviving the #1488 backstop.

Extracting those would put the thing under test and the thing testing it on the same side — the trade #1390 made in the opposite direction. So the gate is unmeetable without making the code worse, and I am merging over it deliberately rather than letting it pass unremarked.

`atomicWriteFile` in seven places is genuine boilerplate wanting `core/pkg/`, and the report is its own evidence: pi's copy already partners with `hooktoml.go`, a shared-package file. Migrating one of seven adds a seventh partner and moves nothing. Filing it as a tree-wide follow-up.

**CodeScene**: red, advisory, and honestly reported as *unread* — there is no local way to read it — so it is stated in the PR as an assumption rather than dressed up as a finding. `ars-gate`, the deterministic architecture measure, passed with no regression.

**No pi cells recorded**, so `hookcov` reports `StatusGap` for pi. That is the honest state and I want it visible.

Issue #1762, opened 2026-08-22
test(hooks): six adapters carry hand-written version tests the shared contract strictly dominates — delete by mutation, not by inspection


ingo-eichhorst:
#1758 found that pi's `hookversion_test.go` carried three hand-written tests
restating obligations `contracttesting.AssertHookVersionGate` already runs —
**strictly more weakly**:

| | shared contract | the hand-written copy |
|---|---|---|
| below the floor | three field-wise predecessors (major/minor/patch decremented independently — a single synthetic "just below" only exercises the borrow path), each refusal naming both versions | one hand-picked version |
| at/above the floor | `Permits(Min)` allowed **plus a vacuity guard** — a floor with nothing below it fails outright | `Permits(Min)` allowed |
| unreadable version | three spellings (`""`, `"not a version"`, `"2.1"`) | one (`""`) |

Deleting them raised coverage. The deletion was proven by mutation rather than
argued: weakening the declared floor to `0.0.0` — precisely the case the deleted
test would have **passed** — reddens the contract:

```
--- FAIL: .../floor_refuses_an_older_cli
    hook_version.go:62: declared floor 0.0.0 has no version below it, so this
    obligation asserts nothing — a floor of 0.0.0 permits every CLI and is not a gate
```

## The same three are redundant in the sibling adapters

```
$ git grep -ln 'AssertHookVersionGate' -- 'core/adapters/inbound/agents/*/*_test.go'
claudecode  codex  copilot  geminicli  kirocli  pi  vibe
```

pi is now clean. The other six have not been checked — they predate the
contract's later hardening, so the hand-written tests were plausibly *stronger*
than it when they were written, and the redundancy accumulated silently as the
contract grew.

## Why this is worth doing rather than leaving

A weaker test sitting beside a stronger one is not harmless. It reads as
coverage, it is what a reader copies when adding adapter number eight (which is
exactly how it reached pi — #1758's own note says "#1721 added a sixth copy
before measuring"), and it dilutes the signal that the contract is the thing to
trust.

## Shape, and the one thing not to get wrong

Per adapter, check each hand-written test against what the contract now asserts,
and delete only those the contract **strictly dominates**. Keep anything testing
something the contract does not — pi kept `TestHooksGate_ProbeIsPiVersion`,
because the contract requires only that `Probe` is non-empty and never that the
argv is right.

**Do not delete on inspection alone.** For each deletion, the evidence is a
[...]

Issue #1759, opened 2026-08-22
test(hooks): two hand-kept adapter rosters can't fail for a new adapter — beaconroute_test.go has been missing mistral-vibe since #1718


ingo-eichhorst:
Two hand-kept rosters over hook adapters can't fail for an adapter nobody added
to them — the exact failure mode #1740 was filed about, one layer over. One of
them has already been silently wrong for days.

## Measured

`core/adapters/inbound/agents/beaconroute_test.go` on `main`:

```
$ git show origin/main:core/adapters/inbound/agents/beaconroute_test.go | grep -n 'vibe\|kirocli\|geminicli'
9:	"irrlicht/core/adapters/inbound/agents/geminicli"
10:	"irrlicht/core/adapters/inbound/agents/kirocli"
40:		"gemini-cli": geminicli.HookEndpointPath,
41:		"kiro-cli":   kirocli.HookEndpointPath,
```

**mistral-vibe is absent, and has been since #1718/#1733 shipped it.** The file's
own comment admits the map cannot fail for a receiver nobody adds. So it went
wrong immediately, stayed wrong through four PRs that touched this area, and was
found by an agent reading it for an unrelated reason.

`tools/onboarding-factory/internal/hookcov`'s `Declared()` want-map is the second
instance of the same shape.

## Why a roster here is worse than a roster in prose

`docs/testing-contracts.md`'s adapter lists were stale too (#1730), but a stale
sentence misleads a reader who can check. A stale roster inside a **test** is
worse: it reports success. `beaconroute_test.go` passed on every one of those
PRs while covering one fewer adapter than it claimed.

This is the same argument the repo already accepts in three places — the
`Verify` and version-floor tripwires walk the registry, `hookcontracts_test.go`
(#1740) walks the adapter set, and `censusOfTheCommittedCatalog` is machine-
generated and pasted. Each replaced a thing someone had to remember.

## Shape

For `beaconroute_test.go`: `registerHookRoutes` already knows every registration
it performs. Exposing that — a returned slice, or a registry the test can read —
makes the map **derived** rather than remembered, and the test becomes "every
beacon adapter's route is registered" instead of "these two are".

For `hookcov.Declared()`: the `censusOfTheCommittedCatalog` idiom fits — machine-
generate the expected set, print it in a pasteable form, and ratchet, so a new
adapter forces a visible one-line diff rather than passing unnoticed.

Both need the same guard the other tripwires carry: **refuse on an empty
[...]

PR #1763, opened 2026-08-22
feat(opencode): observe opencode through a plugin irrlicht installs, not a config entry (#1719)

### LINKED ISSUE #1719: feat(opencode): hook-based state detection requires a shipped plugin — the config-file route was removed upstream

ingo-eichhorst:
Closes #1719. Part of #1355.

## The epic's verdict was half right

#1355 cut opencode with: *"its config-file hook path was removed upstream, so a shipped TS plugin is the only route: a distribution and supply-chain problem (npm package? vendored file?), not a config write."*

**Confirmed — the config-file route is closed.** opencode 1.14.50's live config schema declares `experimental` with exactly six keys — `disable_paste_summary`, `batch_tool`, `openTelemetry`, `primary_tools`, `continue_loop_on_deny`, `mcp_timeout` — and no `hook`. Read out of the bun-compiled binary's own bundle, not out of docs.

**Wrong — there is no distribution problem.** opencode auto-discovers plugins with a plain directory glob:

```js
async function ConfigPlugin_load($) {
  let Z = [];
  for (let Y of await Glob.scan("{plugin,plugins}/*.{ts,js}",
        { cwd: $, absolute: true, dot: true, symlink: true }))
    Z.push(pathToFileURL(Y).href);
  return Z;
}
```

`Config.loadInstanceState` calls it for **every** directory `ConfigPaths.directories` returns, and the first is `Global.Path.config`:

```js
ConfigPaths.directories($, Z) => dedupe([
  Global.Path.config,                                    // $XDG_CONFIG_HOME/opencode || ~/.config/opencode
  ...up({targets:[".opencode"], start:$, stop:Z}),
  ...up({targets:[".opencode"], start: home, stop: home}),
  ...(OPENCODE_CONFIG_DIR ? [OPENCODE_CONFIG_DIR] : []),
])
```

So: no registry, no npm resolution, no git, no build step, no toolchain, no `settings.json` write. One `.js` file in `~/.config/opencode/plugin/` is loaded, globally, for every session; uninstall is `os.Remove`. A plugin whose directory has no `package.json` resolves to the file itself (`resolveEntry` returns the target unchanged when `readManifest` finds nothing), and the loader accepts a bare `export default async function`.

`plugin: [...]` and `opencode plugin install` — the npm-spec route the epic was reasoning about — are a **second, separate** mechanism. This installer does not use it. That is the same correction #1721 made for pi, reached from the other end: `pi install` was the red herring there, `plugin: [...]` is the red herring here.

## opencode DOES have a blocked-on-user state, and it is invisible in the store

This is what makes the adapter worth a hook channel rather than a nicety.

- `Permission.ask` evaluates each requested pattern and **returns early, publishing nothing**, when all of them already resolve to `allow`. `permission.asked` therefore fires only on a genuine prompt — a narrow assert, not the fires-for-every-tool event #1719 warned about reusing `PreToolUse`'s row for.
- That pending request lives in a plain in-memory `Map`; only **approved** patterns reach the `permission` table. opencode's store cannot express "the user is being asked right now", which is exactly what this adapter's `ProcessOwnedStore` reads. Before this PR opencode had no `waiting` at all.
- `Permission.reply` publishes `permission.replied` for **every** reply including `"reject"`, and cascades it to the session's other pendings. That puts opencode out of copilot's position, where a denial emits nothing and a hold would sit until the 12-hour ceiling.
- `session.idle` is published by `SessionStatus.set` from the runner's `onIdle` and directly by `SessionRunState.cancel` — turn end, cancel and interrupt.

Canonicalised on semantics, not Claude Code's wire names: each event goes to a dedicated `SessionDetector` entry point, so `session.hookSignalEffects` (a flat map with no adapter dimension) is never consulted and no row can leak into another adapter's dispatch.

**`permission.ask` — the HOOK — is deliberately not registered.** `Plugin.trigger` iterates registered hooks with **no try/catch**, and several receive a mutable `output`; `permission.ask`'s `output.status` is a writable `"allow" | "deny" | "ask"`. A monitoring plugin must not be able to answer a permission prompt for the user or throw inside the agent loop. The `event` bus carries the same fact as an observation.

## Shared machinery: two moves, both because opencode has no file per session

opencode's `Source` is an `agent.ProcessOwnedStore` — one SQLite database, no transcript files — so the path every consumer keys a session on is composed by the daemon (`<db>-wal?session=<id>`, now via one `transcriptPathFor` the watcher and the receiver share). There is nothing a caller could name that the daemon would open, so the honest payload is a session id.

1. **`hookjson.DecodeSealed`** joins `DecodeConfined` as a second entry point. Both funnel through one unexported `readBody`, so `architecture_hookbody_test.go`'s Arm B claim — *this package reads a body in exactly ONE function* — is unchanged. The chokepoint moved; it did not widen into a two-element allowlist.
2. **`AssertHookPathConfined` gains a `PathDaemonDerived` route**, the confinement family's `DeliveryAddressFree`. Rather than confining the traversal class once more, the route makes it inexpressible, and its three obligations assert the property was *obtained*: a well-formed payload still dispatches, the dispatched string is the daemon's own path, and it does not move when the body names an out-of-tree path, a `..` traversal, or either kind of symlink. The two routes make contradictory assertions about the same two strings, so declaring the wrong one fails rather than passing quietly. Committed self-tests: `hook_path_confinement_derived_selftest_test.go`.

Handing `DecodeConfined` an empty path instead was considered and rejected: it would answer `RejectEmptyPath`, drop every hook, and log a confinement refusal for correct traffic. Confining a self-derived path was rejected too — resolution could only fail (a store not yet created, a WAL checkpointed away), silently dropping legitimate hooks to satisfy a guard protecting nothing.

## Mutation evidence — every guard seen red

One mutation per run, applied and reverted, `git status --porcelain` clean between each.

**1. A second body-reading function inside `hookjson`** (the #1389 hazard Arm B exists for):
```
--- FAIL: TestNoDirectRequestBodyReadsInAgentAdapters (0.19s)
    architecture_hookbody_test.go:202: hook-body violation: irrlicht/core/adapters/inbound/agents/hookjson reads an inbound *http.Request body at .../hookjson/decode.go:276:27 (r.Body in decodeUnguarded)
        	irrlicht/core/adapters/inbound/agents/hookjson is exempt only for readBody — the single decode every receiver is forced through,
        	behind hookjson.DecodeConfined (a payload naming a path) or hookjson.DecodeSealed (a payload naming none).
        	A second body-reading function here is a second way to skip confinement (issue #1389)
```

**2. The receiver dispatches a caller-supplied `transcript_path` when the body carries one** — the defect the whole new route exists to catch. All four hostile spellings went red; one shown:
```
--- FAIL: TestHookPathConfined/dispatch_is_indifferent_to_symlink_escape (0.00s)
    hook_path_confinement.go:278: a body naming symlink_escape dispatched ".../001/.local/share/opencode/linked/planted.jsonl", want the daemon's own composed path ".../001/.local/share/opencode/opencode.db-wal?session=ses_19af5168cffem3E4MnRCcfxkNX" — on this route nothing a caller writes may reach the dispatch, and a path that MOVED with the body is a caller-supplied path travelling under a different name
```

**3. Delete the receiver's own `database` consent gate.** Since #1488 the backstop still drops the payload, so every dispatch-shaped assertion stays green; the SILENCE is the discriminator:
```
--- FAIL: TestHookReceiver_DeniedDatabaseConsentDropsQuietly (0.00s)
    hooks_test.go:410: a database-denied hook logged 1 error(s): [hook body dropped: permission "database" is not granted at the decode — either this receiver skipped the check it declares, or consent was revoked mid-request (issue #1488)]
```

**4. Hoist `ObserveHookReceipt` above the channel-consent check:**
```
--- FAIL: TestHookReceiptObserved/consent_denied_request_is_not_counted (0.00s)
    hook_receipt.go:127: a consent-denied request added 1 receipts for adapter "opencode", want 0 — noting that a POST arrived is an observation, and every observation an adapter makes is gated
```

**5. Drop `IgnoreUnknownEvent` from the default arm:**
```
--- FAIL: TestUnknownHookEventObserved/counted_per_adapter_and_name (0.00s)
    unknown_hook_event.go:135: event "IrrUnk5fd19820Renamed" arrived 3 times, counted 0 for adapter "opencode" — an event that keeps arriving is the signature of an upstream rename and has to be countable as such
--- FAIL: TestUnknownHookEventObserved/reported_once_per_distinct_name (0.00s)
    unknown_hook_event.go:141: event "IrrUnkd6ae8a1101Flood" arrived 5 times and produced 0 log line(s), want exactly 1
```

**6. Never rewrite a stale install in place:**
```
--- FAIL: TestHookEndpointFollowsBindAddr/foreign_binary_install_upgraded_in_place (0.00s)
    hook_endpoint.go:267: expected modified=true when repointing an install naming a different irrlichd binary — an install left unrepaired here reports healthy while delivering nothing, and no later daemon start converges on it
```

**7. Consent copy understates the install (the #1173 drift):**
```
--- FAIL: TestHooksPermission_DisclosureMatchesInstalled/names_every_installed_event (0.00s)
    hook_disclosure.go:136: consent copy never names the "session.idle" hook, but the installer writes it — an undisclosed write to the user's config (#570)
--- FAIL: TestHooksPermission_DisclosureMatchesInstalled/states_the_installed_count (0.00s)
    hook_disclosure.go:139: consent copy declares 2 hook entries, installer writes 3
```

**8. Revert the disclosure tokenizer widening** (proves the widening is load-bearing, not cosmetic — without it NO wording could satisfy the arm):
```
--- FAIL: TestHooksPermission_DisclosureMatchesInstalled/names_every_installed_event (0.00s)
    hook_disclosure.go:136: consent copy never names the "permission.asked" hook, but the installer writes it — an undisclosed write to the user's config (#570)
    hook_disclosure.go:136: consent copy never names the "permission.replied" hook, but the installer writes it ...
    hook_disclosure.go:136: consent copy never names the "session.idle" hook, but the installer writes it ...
```

**9. `Writes.Path` names a file `Apply` does not write:**
```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites (0.00s)
    managedwrites_test.go:227: opencode/hooks declares Writes.Path = "/tmp/irrlicht-not-the-file", which is OUTSIDE the scratch home ... Refusing to run any Apply
```

**10. Drop `IRRLICHT_BIND_ADDR` from the driver's RELAUNCH pane only:**
```
--- FAIL: TestEveryBeaconAdapterDriverPassesTheDaemonAddress/opencode (0.00s)
    beacon_test.go:50: .../opencode/driver-interactive.sh:595 launches the agent under tmux without passing IRRLICHT_BIND_ADDR.
```

**11. An installed event's `Since` above the declared floor:**
```
--- FAIL: TestHooksPermission_DeclaresVersionGate/floor_covers_every_installed_event (0.00s)
    hook_version.go:59: declared floor 1.14.50 is below 1.15.0, the version "session.idle" is known at — at a CLI between the two the install writes an entry naming an event that CLI does not have
```

**12. Make the NEW arm stop discriminating** (grade only *that* something dispatched, never *which* string — #1389's defect, inherited by construction). The committed self-test catches it:
```
--- FAIL: TestDerivedRouteObligation2_DispatchIsIndifferentToTheBody (0.00s)
    hook_path_confinement_derived_selftest_test.go:252: mutation "a receiver that dispatches the caller's transcript_path whenever the body carries one": the obligation reported nothing against a fixture that is wrong for it — either the fixture is no longer wrong, or the obligation has stopped discriminating (expected a message containing "want the daemon's own composed path")
```

**13. The mutation that found a real defect in my own guard.** `TestShippedPluginRegistersOnlyTheReadOnlyEventTap` forbids registering `permission.ask`. Adding exactly that to `plugin.js` **passed**:
```
ok  	irrlicht/core/adapters/inbound/agents/opencode	0.321s
```
The detector matched only BARE property keys — and ten of opencode's fifteen hook names contain a dot, which JavaScript cannot spell bare. The guard's biggest hole came with an approved spelling. Fixed (`"?` around the key), mutation re-run:
```
--- FAIL: TestShippedPluginRegistersOnlyTheReadOnlyEventTap (0.00s)
    plugin_test.go:68: plugin.js registers [permission.ask event], want exactly [event]
```
and the shape is now a **committed corpus** (`TestHookRegistrationDetectorNamesEveryShape`) rather than a paragraph here. **14.** Reverting the detector to bare-key-only turns that corpus red:
```
--- FAIL: TestHookRegistrationDetectorNamesEveryShape/a_QUOTED_dotted_key_is_a_registration (0.00s)
    plugin_test.go:320: detector returned [], want [permission.ask]
```

## What is NOT covered

- **No onboarding fixtures were recorded.** `hookcov` now reports `StatusGap` for opencode — declares hooks, zero hook-bearing recordings. That is the honest outcome and is left visible; #1754's rig gaps are still open. Cells re-recorded: **0** (#1355 Phase D's kill criterion was >5).
- **Nothing executes the JavaScript in CI.** `plugin_test.go` grades the artifact statically; running it needs bun or a model call. Its header says so.
- **`AssertHookReceiptObligation 3` degenerates** for this route — the out-of-tree body is indistinguishable from a clean one, so it repeats obligation 1. Stated at the wiring, kept with no opt-out.
- **opencode is `Unisolatable`** for the recording rig, with the reason: its config dir follows `$XDG_CONFIG_HOME` and its data dir `$XDG_DATA_HOME`, which a one-variable row cannot express — and `StorePath()` honours neither, joining `$HOME` with `.local/share/opencode` unconditionally. Closing it needs an XDG-aware `StorePath` **and** a two-variable row shape.
- **The transcript-tail `waiting` heuristic is unchanged**; hooks add a signal, they do not replace anything.

## Notes on adjacent issues

- **#1759** (beaconroute roster missing mistral-vibe) was already fixed by #1721; opencode's row is added and the scope paragraph updated to eight.
- **#1762**: this adapter wires `AssertHookVersionGate` and stops. No seventh copy of the three hand-written obligations — `hookversion_test.go` says so where a reader would look for them.
## The two red advisory checks

**SonarCloud** is red on ONE condition and it is duplication, measured: `new_duplicated_lines_density` 5.8% against a 3% threshold. Every issue it raised is fixed (`total: 0` from the issues API — a `_ "embed"` without a comment, a world-readable temp file, three duplicated string literals, four JS style findings, and a cognitive-complexity 18 on `AssertHookPathConfined` that the `hostileBodyFor` extraction brought under the limit). Security rating, reliability rating and maintainability rating are all **OK**, and there are zero security hotspots.

The duplicated lines are:

| file | duplicated / new |
|---|---|
| `opencode/hookinstaller.go` | 105 / 471 |
| `opencode/hookversion_test.go` | 31 / 65 |
| `opencode/hookinstaller_test.go` | 22 / 301 |
| `opencode/hooks_test.go` | 21 / 413 |

The three test files are **contract wirings**, which #1740's tripwire requires each adapter to make in its OWN test package — extracting a shared helper would defeat the tripwire that exists because "a contract can only fail for an adapter that WIRED it". They stay.

`hookinstaller.go`'s 105 lines are a different thing and worth naming honestly rather than filing under the same heading: they are near-identical to pi's, because both adapters are on the same route — write ONE file, beacon-delivered, refuse to overwrite a foreign one, rewrite a stale one in place, `os.Remove` to uninstall. That is a real extraction candidate (a shared single-file-artifact installer parameterised by path resolver, template, sentinel and event list), and it would make the next adapter on this route ~20 lines instead of ~300. It is deliberately NOT done here: pi merged an hour before this branch opened, and folding "add opencode" together with "refactor pi's installer" would put a cross-adapter refactor behind a feature review. Worth its own issue.

**CodeScene** is red for the same structural reason and is likewise advisory (neither is a required status check; branch protection lists 14 and neither is among them).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -98,7 +98,7 @@ Before marking a ticket done, run the full suite — every layer must pass:
   under `t.TempDir()`, so it never touches the production daemon). Also runs
   `core/architecture_test.go` (hexagonal import-direction, statically
   enforced) and `core/architecture_hookbody_test.go` (which inbound hook
-  body reads are confined to `hookjson.DecodeConfined`).
+  body reads are confined to `hookjson`'s single decode).
 - Architecture score (advisory, not a merge gate): `tools/ars-gate.sh` /
   `tools/preflight.sh`'s `arch` gate — flags a regression in the Agent
   Readiness Score (composite or any category) vs `origin/main`.
--- a/core/adapters/inbound/agents/beaconroute_test.go
+++ b/core/adapters/inbound/agents/beaconroute_test.go
@@ -8,6 +8,7 @@ import (
 	"irrlicht/core/adapters/inbound/agents/copilot"
 	"irrlicht/core/adapters/inbound/agents/geminicli"
 	"irrlicht/core/adapters/inbound/agents/kirocli"
+	"irrlicht/core/adapters/inbound/agents/opencode"
 	"irrlicht/core/adapters/inbound/agents/pi"
 	"irrlicht/core/adapters/inbound/agents/vibe"
 	"irrlicht/core/pkg/hookbeacon"
@@ -28,9 +29,10 @@ import (
 // while its route segment is "claudecode". That is why the map keys below are
 // written out literally rather than derived from the adapter constants.
 //
-// Scope, stated honestly: this pins the receivers that exist (seven as of
-// #1721's pi row, which also added the mistral-vibe row #1718 never wrote —
-// exactly the omission the paragraph below predicts). It cannot fail for a NEW receiver registered at a
+// Scope, stated honestly: this pins the receivers that exist (eight as of
+// #1719's opencode row; #1721's pi row also added the mistral-vibe row #1718
+// never wrote — exactly the omission the paragraph below predicts, and the
+// whole of issue #1759). It cannot fail for a NEW receiver registered at a
 // different prefix, because registerHookRoutes (core/cmd/irrlichd/startup.go)
 // is hand-wired with no registry to enumerate and this map is hand-written.
 // Whoever adds the next receiver adds its row here; until then the
@@ -45,6 +47,7 @@ func TestBeaconEndpointPathMatchesTheInstalledReceivers(t *testing.T) {
 		// mistral-vibe was beacon-delivered from the day it shipped (#1718)
 		// but was never added here; found while adding pi's row (#1721).
 		"mistral-vibe": vibe.HookEndpointPath,
+		"opencode":     opencode.HookEndpointPath,
 		"pi":           pi.HookEndpointPath,
 	} {
 		if got := hookbeacon.EndpointPath(segment); got != want {
--- a/core/adapters/inbound/agents/hookjson/decode.go
+++ b/core/adapters/inbound/agents/hookjson/decode.go
@@ -102,6 +102,120 @@ func DecodeConfined[T any](
 	get func(*T) string,
 	set func(*T, string),
 ) bool {
+	if !consentAllowsRead(w, log, component, consent) {
+		return false
+	}
+
+	// Fail closed on a receiver wired without a confiner. Returning "not
+	// decoded" drops the payload with a 2xx, the same shape every other refusal
+	// takes, rather than dispatching an unconfined path because the guard
+	// itself was missing.
+	//
+	// Checked AFTER consent and BEFORE the read, so the three refusals keep the
+	// order TestDecodeConfined_RefusesBeforeReadingAnything pins: a denied
+	// request costs no decode, no path resolution and no confiner counter.
+	if c == nil {
+		log.LogError(component, "", "hook body dropped: receiver has no path confiner")
+		w.WriteHeader(http.StatusOK)
+		return false
+	}
+
+	if !readBody(w, r, p) {
+		return false
+	}
+
+	raw := get(p)
+	confined, reason := c.Confine(raw)
+	if reason != RejectNone {
+		RejectPath(w, log, component, raw, reason)
+		return false
+	}
+	set(p, confined)
+
+	// Postcondition: get and set must name the SAME location, or the payload
+	// still holds the caller's raw string while this function reports success.
+	//
+	// This is not defensive clutter — it is the one failure this design could
+	// otherwise not detect, and it was found green by review. Passing behaviour
+	// as two closures puts the obligation at each call site, where no checker
+	// can see it: a receiver whose set writes a field its get never reads (or a
+	// no-op set, which is a one-character edit away) forwards the unconfined
+	// path, and every test in the tree still passes — including
+	// AssertHookPathConfined, which asserts THAT the receiver dispatched, never
+	// WHICH spelling it dispatched. Verifying it here fails closed once, for
+	// every present and future receiver, instead of asking four contract
+	// wirings to each notice.
+	if got := get(p); got != confined {
+		log.LogError(component, "", fmt.Sprintf(
+			"hook body dropped: payload still names %q after confinement to %q — "+
+				"this receiver's get/set pair do not address the same field", got, confined))
+		w.WriteHeader(http.StatusOK)
+		return false
+	}
+	return true
+}
+
+// DecodeSealed decodes an inbound hook body that names NO filesystem path at
+// all, into p.
+//
+// It reports whether the payload survived, on the same terms as DecodeConfined:
+// on false the response has ALREADY been written and the caller must return
+// without dispatching.
+//
+// # Why a second entry point rather than a confiner over an empty string
+//
+// It exists for issue #1719's opencode receiver, whose Source is an
+// agent.ProcessOwnedStore: session state lives in one SQLite database, there is
+// no file per session, and the "transcript path" the daemon keys a session on
+// is a string the daemon COMPOSES from its own store resolver
+// (<db>-wal?session=<id>). There is nothing a caller could name that the daemon
+// would open, so the honest payload carries a session id and stops there.
+//
+// Handing DecodeConfined an empty path instead would not be the same thing. It
+// would answer RejectEmptyPath, drop every hook, and log a confinement refusal
+// for traffic that is behaving correctly — a mechanism reporting a defect it
+// invented. And a confiner over a self-derived path would be worse still: the
+// input is our own resolver's output, not a caller's, so resolution could only
+// fail (a store not yet created, a WAL checkpointed away between events) and
+// each failure would silently drop a legitimate hook to satisfy a guard that
+// was protecting nothing.
+//
+// # What is NOT relaxed
+//
+// Everything on this path except the confinement is identical to
+// DecodeConfined, and deliberately so: the same consent backstop in the same
+// place (#1488), the same 1 MiB bound on an unauthenticated local endpoint, and
+// the same single body-reading function — readBody, which
+// core/architecture_hookbody_test.go pins as the ONE place under
+// core/adapters/inbound/agents/... an inbound body may be read. That rule is
+// what makes this a second ENTRY POINT rather than a second door: a receiver
+// still cannot reach a payload without going through the chokepoint, and a
+// receiver that DOES carry a caller-supplied path still cannot reach it without
+// supplying a confiner.
+//
+// The obligation this shifts rather than removes is "prove nothing
+// caller-supplied travels", and it is graded from outside by
+// contracttesting.AssertHookPathConfined's PathDaemonDerived route: the two
+// routes make contradictory assertions about the same two strings, so a
+// receiver that declares the wrong one cannot pass quietly.
+func DecodeSealed[T any](
+	w http.ResponseWriter,
+	r *http.Request,
+	log outbound.Logger,
+	component string,
+	consent Consent,
+	p *T,
+) bool {
+	if !consentAllowsRead(w, log, component, consent) {
+		return false
+	}
+	return readBody(w, r, p)
+}
+
+// consentAllowsRead is the #1488 backstop both entry points run before the body
+// is touched. It reports whether reading may proceed; on false the response has
+// already been written.
+func consentAllowsRead(w http.ResponseWriter, log outbound.Logger, component string, consent Consent) bool {
 	// Fail closed on a receiver that named no permission. Every hook receiver
 	// acts under at least one (#570), so an empty declaration is a wiring that
 	// went through Consent{} rather than RequireConsent — it cannot be typed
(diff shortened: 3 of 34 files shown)


ingo-eichhorst:

### The epic's verdict was wrong a second time, and I confirmed it from the binary

```
$ strings ~/.opencode/bin/opencode | grep -oE '\{plugin,plugins\}/\*\.\{[a-z,]+\}'
{plugin,plugins}/*.{ts,js}
```

Directory glob, `.js` accepted. No registry, no npm, no git, no build step, no config entry; uninstall is removing one file. `opencode plugin install` is the second, separate mechanism — the red herring, exactly as `pi install` was in #1721. **Epic #1355 cut opencode as "a distribution and supply-chain problem, not a config write", and cut pi on the same reasoning. Both were wrong, for the same reason: the loader was never read.**

I also confirmed the other half of the verdict, which **is** right:

```
$ strings ~/.opencode/bin/opencode | grep -oE 'disable_paste_summary|continue_loop_on_deny|mcp_timeout|primary_tools|openTelemetry|batch_tool' | sort -u
batch_tool  continue_loop_on_deny  disable_paste_summary  mcp_timeout  openTelemetry  primary_tools
$ … | grep -cE 'experimental[^A-Za-z]{0,3}hook'
0
```

Six experimental keys, no `hook`. And the PR is careful to claim only *absent now*, not *when it was removed* — the binary carries no per-version history. That restraint is right.

### The capability gain is real and larger than the ticket implied

`Permission.ask` publishes `permission.asked` **only** when a pattern actually needs asking — a narrow assert, not fires-per-tool, which is the trap `hook_signal.go` warns about. The important part: the pending request lives in an in-memory `Map` and only *approved* patterns reach the `permission` table. **opencode's store cannot express "the user is being asked right now"**, and a `ProcessOwnedStore` adapter reads the store — so `waiting` was structurally unreachable for this adapter before now. That is not an incremental improvement; it is a state that did not exist.

### The safety finding is the best thing here

`permission.ask` **the hook** must never be registered: `Plugin.trigger` dispatches named hooks with no try/catch and hands them a mutable `output.status` (`"allow"|"deny"`), so a monitoring plugin registering it **could answer a permission prompt on the user's behalf**. That is pi's `tool_call` exclusion one notch worse, and the read-only `event` bus carries the same information without the hazard.

**And mutation 13 found that the guard forbidding it did not work.** Adding `permission.ask` to the shipped plugin **passed** `TestShippedPluginRegistersOnlyTheReadOnlyEventTap`, because the detector matched only bare property keys and ten of opencode's fifteen hook names contain a dot, which JS cannot spell bare. Fixed and pinned by a committed corpus rather than a comment. A guard against the single most dangerous thing this adapter could do, which did not fire — found by mutating it rather than trusting it.

### Corrections to my framing, both accepted

- **#1759 was already fixed** by #1721 for the mistral-vibe row. I have corrected that issue, and also removed `hookcov` from it — `Declared()` walks the registry and `DeepEqual` + a `len(agents.All())` check make it a real ratchet, so it was never the second instance I claimed.
- **This adapter cannot take the normal receiving-side shape** — no file per session means nothing a caller could name. `hookjson.DecodeSealed` and the `PathDaemonDerived` route are the right response, and the architecture rule's "exactly one body-reading function" is preserved because both entry points funnel through one unexported `readBody`: the chokepoint moved, it did not widen.

### Merging over the duplication gate again, with the reason

SonarCloud reports **zero issues** and OK ratings; it fails only on `new_duplicated_lines_density` 5.8% vs 3%. The test-file share is #1740's tripwire working as designed — an adapter is *required* to wire near-identical contract calls.

`hookinstaller.go`'s 105 duplicated lines against pi's are a **genuine** extraction candidate, and deferring them is the right call rather than a dodge: a cross-adapter refactor of a one-hour-old adapter should not ride behind a feature review, and #1761 already tracks the same pattern for `atomicWriteFile` × 7. Filing the shared single-file-beacon installer separately.

CodeScene advisory, unread (no local access), stated as an assumption rather than dressed as a finding. `ars-gate` passed.

Issue #1417, opened 2026-08-09
fix(daemon): irrlichd has no argument parser — a mistyped flag silently starts a daemon in the wrong configuration


ingo-eichhorst:
Found while building the beacon (#1373 / PR #1412), which had to work around it.

## The defect

`irrlichd` has never parsed its command line. `main.go` uses a permissive `hasFlag` scan over `os.Args`, checking for a handful of known flags in turn and falling through to `runDaemon()` for anything else.

So a typo does not fail — it **silently starts a daemon with the flag ignored**:

```
irrlichd --recrod        # starts a daemon, recording OFF, no message
```

That is the same permissive scan that made `--print-hook-configs` hang the recording rig against an older binary (#1357), and the reason the beacon in #1412 needs a leading `--version` guard token: on an older `irrlichd` the stray invocation would reach `runDaemon()`, and `setupUnixSocket` `os.Remove`s the live socket before listening — so a beacon firing on every tool call would actively break the running daemon.

PR #1412 narrowed the fall-through: an unknown **positional** now exits 2. Unknown **flags** still fall through, deliberately, because that behaviour is load-bearing somewhere and changing it was out of scope for that ticket.

## Why it is worth closing properly

Three separate pieces of work have now had to reason about this scan rather than about their own problem: #1357's watchdog around the probe, #1412's guard-token placement, and #1412's positional narrowing. Each is correct and each is a workaround.

The failure mode is also the worst shape available: a mistyped flag produces a **running daemon in a subtly wrong configuration**, which looks like success. `--record` in particular means the difference between capturing an onboarding session and silently not.

## Scope

1. Parse the command line properly — either `flag` with explicit registration, or an allow-list of known flags checked before dispatch.
2. An unknown flag must **fail with a message naming it**, not start a daemon.
3. Keep every currently-working invocation working. Sweep the call sites first: `site/install.sh`, the LaunchAgent plist, the systemd unit, `tools/`, the recording rig, and the beacon's installed line. List what you found in the PR — this is the part where a regression would be invisible until someone's daemon stops starting.
4. `--version` must keep working as the beacon's guard token, and must keep working when it appears **before** a subcommand verb. #1412 depends on that shape.

## A constraint worth stating

[...]

PR #1765, opened 2026-08-22
feat(hermes): observe hermes through its own shell hooks — an approval state its store cannot express (#1722)

### LINKED ISSUE #1722: feat(hermes): audit whether hermes has a hook mechanism — a store-backed adapter where a hook buys invalidation, not latency
Broken out of epic #1355, which never scoped hermes — its phases covered the transcript-file
adapters and stopped there.

## Why hermes is a different shape

hermes has **no transcript files at all**: one shared `state.db` that the agent owns and irrlicht
reads through `core/pkg/sqlitero`. That changes what a hook would even be for.

For the file-based adapters a hook mostly buys *latency and authority* — the transcript already
carries the state, the hook just says so sooner and more reliably. For hermes the store is polled,
so a hook would buy **event-driven invalidation** instead of a poll interval, which is a different
and possibly larger win. Say which of the two this is before designing anything.

## What is known

**Measured, on record:** one shared `state.db`, no per-session transcript; `cwd` is present only for
the TUI (`hermes chat` reports source `cli`); `finish_reason` marks the turn boundary.

**Consequence already established:** hermes is a ProcessOwnedStore adapter, so `of verify` reads only
the replay golden and `tokens_nonzero` / `cost_nonzero` can never pass for it. A hook that carried
usage would change that — worth checking whether it can.

## Unverified

Whether hermes has any hook, plugin or notifier mechanism. Nobody has looked. That is step one.

## Note on the store path

Any read of an agent-owned SQLite store goes through `core/pkg/sqlitero`, never `sql.Open` directly
— the DSN grammar decides whether "read-only" is real, and all three wrong spellings have shipped at
some point. That constraint is unchanged by anything in this issue.

## Shape of this ticket

## Recording cost

Part of #1355.

ingo-eichhorst:
Closes #1722. Part of #1355.

## Verdict: the mechanism exists, it is first-class, and the "no payoff" half does not survive

**hermes has a full hook system with its own CLI subcommand.** `hermes hooks {list|test|revoke|doctor}` (`hermes_cli/hooks.py`, 394 lines), 25 lifecycle events (`VALID_HOOKS`, `hermes_cli/plugins.py:135`), and **three independent subscription mechanisms** documented side by side in `website/docs/user-guide/features/hooks.md:1303`:

| | Declared in | Runs in | Ships code? | Consent |
|---|---|---|---|---|
| Gateway event hooks | `~/.hermes/hooks/<name>/HOOK.yaml` + `handler.py` | **Gateway only** | Python | dir trust |
| Python plugins | `~/.hermes/plugins/<name>/` + `plugins.enabled` in config.yaml | CLI + Gateway | Python, in-process | implicit |
| **Shell hooks** | `hooks:` block in `~/.hermes/config.yaml` | **CLI + Gateway** | **no — a command line** | per-`(event, command)` allowlist |

Only the third is usable: the first never fires for `hermes chat`, and the second is in-process Python **and still needs a config.yaml write** (`plugins.enabled` is an opt-in allow-list, `plugins.py:1454`). The shell-hook route is the one hermes' own table marks "inter-process isolation: yes".

Verified live, not from docs. `~/.hermes` on this machine had no CLI (the venv shim points at a deleted path), so the audit ran the source checkout's venv directly with an isolated `HERMES_HOME`:

```
$ hermes --version
Hermes Agent v0.19.0 (2026.7.20) · upstream 6564f319 · local e289e561 (+1 carried commit)

$ hermes hooks test on_session_end
Firing 1 hook(s) for event 'on_session_end':
  → …/capture.sh
      exit=0  elapsed=0.265s
      parsed: <none — hook contributed nothing to the dispatcher>

$ cat capture.log
{"hook_event_name": "on_session_end", "tool_name": null, "tool_input": null,
 "session_id": "test-session", "cwd": "/private/tmp", "extra": {}}
```

and a **hand-written** allowlist entry — the exact record this installer writes — is accepted:

```
$ hermes hooks doctor
  [on_session_end] …/capture.sh
      ✓ script exists and is executable
      ✓ allowlisted (approved 2026-08-22T06:17:00.293485Z)
      ✓ script unchanged since approval
      ✓ ran clean with empty stdout (exit=0, 0.006s) — hook is observer-only
All shell hooks look healthy.
```

## Does hermes have a state its store cannot express? Yes — the same shape as opencode

`tools/approval.py` fires `pre_approval_request` immediately before the user is asked to approve a dangerous command, and `post_approval_response` with the outcome. The pending request between them lives in a module-level `_gateway_queues` map plus a `contextvars` session key. `state.db`'s schema (`hermes_state.py:1126`+) is `sessions`, `messages`, `session_model_usage`, `state_meta`, `gateway_routing`, `compression_locks`, `async_delegations` — **no approval table**. Nothing about "the user is being asked right now" is ever written.

So this is not invalidation. **Before this PR the hermes adapter had no `waiting` state at all**, and no polling interval could have produced one.

`post_approval_response` fires for every outcome including `deny` and `timeout`, which keeps hermes out of copilot's position where a hold has no release.

## What is deliberately NOT installed, and the guard that enforces it

`agent/shell_hooks._parse_response` reads a hook's **stdout as a decision** for three events: `pre_tool_call` (`{"action":"block"}` vetoes the tool), `pre_verify` (`{"action":"continue"}` keeps the turn running), and a fall-through `{"context": …}` branch `pre_llm_call` prepends to the user's message. This is hermes' `permission.ask`.

`TestInstalledEventsAreObserverOnly` fails the build if any of them appears in `installedHookEvents`, and the installed command is wrapped in `/bin/sh -c … >/dev/null || true` so its stdout cannot reach hermes' parser at all — not defensive: `hookbeacon.LegacyGuardToken` puts `--version` first, so a pre-#1417 irrlichd prints a version banner on stdout on every firing.

## New package: `hookyaml`

hermes' hooks live in the user's ONE config file. `core/go.mod` has no YAML dependency and none is added: `hookyaml` is a line-oriented byte-range splicer for one marker-delimited region under one top-level key, the same strategy `hooktoml` uses for #1718. It rests on one stated invariant (in a loadable root block mapping, a non-blank column-0 line is a key, a comment or a document marker), **asserts** that invariant rather than assuming it, and refuses the whole document — naming the line — for any shape it does not model.

Two placements, and the distinction is what makes uninstall exact: when the key already exists the region goes inside it and the key is the user's; when it does not, the key goes **inside** the region. A collision on one of our own event names is refused by name, because a duplicate YAML key resolves silently to the last one and would delete the user's hook.

## Mutation evidence

Every guard below was applied, run, and reverted one per command. Verbatim output.

<details><summary><b>1. Install a control-bearing event</b> — <code>installedHookEvents += "pre_tool_call"</code></summary>

```
--- FAIL: TestHooksPermission_DisclosureMatchesInstalled/states_the_installed_count
    hook_disclosure.go:139: consent copy declares 3 hook entries, installer writes 4
--- FAIL: TestInstalledEventsAreObserverOnly
    hooks_test.go:235: installedHookEvents contains "pre_tool_call", whose stdout hermes reads as a decision (agent/shell_hooks._parse_response). Installing it would let a monitoring hook veto a tool call, keep a turn running, or inject text into the user's message.
```
</details>

<details><summary><b>2. Drop the <code>/bin/sh -c</code> wrapper</b></summary>

```
--- FAIL: TestInstalledCommandRoundTripsThroughHermesParser
    hookinstaller_test.go:392: argv = []string{"/usr/local/bin/irrlichd", "--version", "hook-post", "hermes", ">/dev/null", "||", "true"}, want []string{"/bin/sh", "-c", "'/usr/local/bin/irrlichd' --version hook-post hermes >/dev/null || true"}
--- FAIL: TestInstalledCommandSuppressesStdout
    hooks_test.go:258: command = "'/opt/irrlichd' --version hook-post hermes >/dev/null || true", want it wrapped in /bin/sh -c so the redirect below is honoured — hermes runs the command with shell=False, so a bare argv has no redirection at all
```
The argv is the finding: hermes would exec `>/dev/null`, `||` and `true` as literal arguments.
</details>

<details><summary><b>3. Read only the DOCUMENTED top-level <code>session_id</code></b></summary>

```
--- FAIL: TestPreApprovalRequestHook_AssertsTheBlockedOnUserSignal
    hooks_test.go:289: HandlePermissionPromptHook called 0 times, want 1
--- FAIL: TestPostApprovalResponseHook_ReleasesTheHold
    hooks_test.go:319: ReleasePermissionPromptHold calls = [], want exactly ["20260802_233614_c97a8d"]
--- FAIL: TestApprovalHooks_ReadTheSessionKeyFromExtra/pre_approval_request
    hooks_test.go:386: dispatched 0 times, want 1 — the id is in extra.session_key for this event, not in session_id
```
</details>

<details><summary><b>4. Accept any non-empty session id</b></summary>

```
--- FAIL: TestHookReceiver_ForeignSessionIDIsQuiet/not-a-session
    hooks_test.go:439: dispatched 2 times for an id hermes could not have minted, want 0
--- FAIL: TestHookReceiver_ForeignSessionIDIsQuiet/default
--- FAIL: TestHookReceiver_ForeignSessionIDIsQuiet/whatsapp:4915112345678
```
</details>

<details><summary><b>5. Delete the receiver's own <code>store</code> gate</b> — only the SILENCE discriminates</summary>

```
--- FAIL: TestHookReceiver_DeniedStoreConsentDropsQuietly
    hooks_test.go:525: a store-denied hook logged 1 error(s): [hook body dropped: permission "store" is not granted at the decode — either this receiver skipped the check it declares, or consent was revoked mid-request (issue #1488)]
```
`TestHookReceiver_PermissionGateContract` stayed **green** — exactly the #1488 behaviour docs/testing-contracts.md records.
</details>

<details><summary><b>6. Remove the collision refusal</b></summary>

```
--- FAIL: TestCollisionIsRefusedByName
    hookyaml_test.go:345: error is <nil> (<nil>), want *CollisionError
--- FAIL: TestSplice_PropertyRandomDocuments
    property_test.go:89: a document already declaring one of the region's events was installed into
--- FAIL: TestEnsureHooksInstalled_RefusesAnEventTheUserAlreadyDeclares
    hookinstaller_test.go:174: EnsureHooksInstalled succeeded against a config that already declares one of our events
```
</details>

<details><summary><b>7. Reinstate the first draft's uninstall</b> — "delete the block key when the block is now empty"</summary>

```
--- FAIL: TestSplice_PropertyRandomDocuments
    property_test.go:89: round trip is not the original:
        --- got ---  "quoted0: \"a: b # c\"\n# section comment…"
        --- want --- "quoted0: \"a: b # c\"\nhooks:\n# section comment…"
    iteration 1, operation "install then uninstall", seed 17221722
```
The hand-written `TestUninstall_RemovesTheBlockWhenItEmpties` stayed **green**. This is the real defect the property test found on its second iteration: uninstall deleted a `hooks:` key the *user* had written.
</details>

<details><summary><b>8. Hardcode LF in the <code>{}</code> key rewrite</b></summary>

```
--- FAIL: TestSplice_PropertyRandomDocuments
    property_test.go:89: round trip is not the original:
        --- got ---  "…\r\nhooks:   # nothing configured\nseq1:\r\n…"
        --- want --- "…\r\nhooks:   # nothing configured\r\nseq1:\r\n…"
    iteration 78
```
`TestEnsureInstalled_HandlesTheEmptyFlowMapping` stayed green (it uses LF). Second real defect.
</details>

<details><summary><b>9. Make <code>homeDir()</code> ignore <code>HERMES_HOME</code></b> — the new righome ProcessOwnedStore branch</summary>

```
--- FAIL: TestEveryRigHomeRowRelocatesBothHalves/hermes
    righome_test.go:99: with HERMES_HOME=…/scratch-home the session root is still "/Users/ingo/.hermes/state.db" — the daemon would watch the operator's real store while the driver wrote to the scratch home…
    righome_test.go:100: … the hooks install still writes Writes.Path to "/Users/ingo/.hermes/config.yaml" …
    righome_test.go:100: … the hooks install still writes Writes.Also[0] to "/Users/ingo/.hermes/shell-hooks-allowlist.json" …
```
</details>

<details><summary><b>10. Verify stops checking the allowlist</b></summary>

```
--- FAIL: TestVerifyHooksInstalled_ReportsARevokedApproval
    hookinstaller_test.go:330: a revoked approval reads as Intact — the entries are in config.yaml but hermes will not run them
```
</details>

<details><summary><b>11. Skip the allowlist write</b></summary>

```
--- FAIL: TestEveryModifyPermissionDeclaresEveryFileItsApplyWrites/hermes/hooks
    managedwrites_test.go:231: hermes/hooks declares "home/.hermes/shell-hooks-allowlist.json" but running Apply changed nothing there…
--- FAIL: TestHooksPermission_IsGated/granted_effect_observable
    permission_gate.go:95: effect not observed after granting permission "hooks"
--- FAIL: TestEnsureHooksInstalled_WritesTheApprovalHermesRequires
    hookinstaller_test.go:207: recorded 0 approvals, want one per installed event (3)
```
</details>

<details><summary><b>12. Remove the tab-indentation refusal</b></summary>

```
--- FAIL: TestRefusals/a_tab_in_the_indentation
    hookyaml_test.go:324: message "a line at column 0 that is not a key, a comment or a document marker — the file is not a plain block mapping and will not be edited" does not mention "tab"
```
Honest reading: the document is still refused (by the column-0 invariant); the tab check is what **names** it correctly. The other eleven rows of that corpus fail outright when their branch is removed.
</details>

<details><summary><b>13. Declare the WRONG confinement route</b></summary>

```
--- FAIL: TestHookPathConfined
    hookpath_test.go:29: Route is PathFromBody (the zero value) but a PathDaemonDerived-only field is set — either the route declaration is wrong, or a neighbouring wiring was copied wholesale
```
</details>

<details><summary><b>14. Consent copy names one event instead of three</b></summary>

```
--- FAIL: TestHooksPermission_DisclosureMatchesInstalled/names_every_installed_event
    hook_disclosure.go:136: consent copy never names the "post_approval_response" hook, but the installer writes it — an undisclosed write to the user's config (#570)
    hook_disclosure.go:136: consent copy never names the "on_session_end" hook, but the installer writes it …
```
</details>

<details><summary><b>15. Drop the version floor below an event's <code>Since</code></b></summary>

```
--- FAIL: TestHooksPermission_DeclaresVersionGate/floor_covers_every_installed_event
    hook_version.go:59: declared floor 0.18.0 is below 0.19.0, the version "pre_approval_request" is known at …
```
</details>

<details><summary><b>16. Forget <code>ObserveHookReceipt</code></b></summary>

```
--- FAIL: TestHookReceiptObserved/recognized_event_counts_one
    hook_receipt.go:111: event "on_session_end" added 0 receipts for adapter "hermes", want exactly 1 — a receiver that does not count its own traffic is reported as a DEAD channel by the liveness watchdog and demoted while working perfectly
--- FAIL: TestHookReceiptObserved/unrecognized_event_counts_too
--- FAIL: TestHookReceiptObserved/confinement_rejected_request_counts_too
```
</details>

<details><summary><b>17. Register the receiver at a route the beacon does not post to</b> (#1759's hand-kept map)</summary>

```
--- FAIL: TestBeaconEndpointPathMatchesTheInstalledReceivers
    beaconroute_test.go:56: hookbeacon.EndpointPath("hermes") = "/api/v1/hooks/hermes", but the receiver is registered at "/api/v1/hooks/hermes-agent" — the beacon would post into a route nothing serves
```
</details>

<details><summary><b>18. The axis census firing on its own first run</b> — and the third real defect</summary>

```
    property_test.go:104: axis block is the last top-level key                   0 of 2000 documents
    property_test.go:104: the generator produced no document exercising "block is the last top-level key" — that axis is claimed in this file's header and graded by nothing
```
Fixing the generator immediately produced:
```
--- FAIL: TestSplice_PropertyRandomDocuments
    property_test.go:91: output line "  on_session_end:\n" is neither an original line in order nor the block key:
        hooks:  # irrlicht-managed BEGIN (irrlicht) — do not edit; `irrlichd --uninstall-hooks` removes this block
```
A block key that is the file's **last line with no final newline** had the BEGIN marker spliced onto it as a trailing comment. The document still parses, so nothing complains — and the marker is no longer a line, so **uninstall could never find the region again**. This is the strongest argument in the PR for AGENTS.md's "count what was actually produced" rule.
</details>

Property sweep: the committed run is 2000 documents (~0.4s); 40,000 × 6 seeds (240,000 documents) was run locally, all green.

## Recording cost: 0 cells re-recorded

`hookcov` reports hermes as `StatusGap`, which is the honest outcome — the permission is declared, no recording carries a `hook_received` event, and the frozen sidecars stay hook-free so every replay golden is byte-identical. #1355's Phase D kill criterion is >5; this is 0. hermes' rig row is **opt-in** (`agent-home.sh`), because #1325 measured that a bare `HERMES_HOME` needs five seeded files before `hermes chat` will authenticate.

## Not covered, stated rather than discovered

- **Hermes profiles.** `hermes -p <name>` relocates `HERMES_HOME` to `~/.hermes/profiles/<name>/`, with its own config and allowlist. Sessions under another profile are watched by the store watcher alone, exactly as before.
- **`cwd` is in the payload and is not read.** It is the one thing the store genuinely cannot supply for a `source='cli'` session, and adopting it would put a caller-supplied path back on this dispatch — which is what the `PathDaemonDerived` declaration says does not happen. A separate change with a confinement obligation of its own.
- **`post_api_request` carries `usage`** and was the metrics candidate #1722 asked about. Not installed: the store already carries per-model tokens and cost, so it would buy a process spawn per API request for a number irrlicht has. What `of verify` cannot validate is the replay golden for a ProcessOwnedStore adapter, and a hook does not change that.
- **The `hooks:` region is refused, not merged, when the user already declares one of our three event names.** Surfaces as #1362's "granted but NOT applied, because …" naming the event and the line.
- The hooks fire in the messaging-gateway processes too; those sessions are filtered out of this adapter by `source`.

## Duplication finding

The seven contract wirings under `core/adapters/inbound/agents/hermes/hook*_test.go` are near-identical to opencode's and pi's by design — that is what a contract family *is*, and #1740's tripwire requires each adapter to carry its own call. Extracting a shared helper would put the wiring behind a seam the walk cannot resolve. Expected and not addressed.

### DIFF
--- a/core/adapters/inbound/agents/beaconroute_test.go
+++ b/core/adapters/inbound/agents/beaconroute_test.go
@@ -7,6 +7,7 @@ import (
 	"irrlicht/core/adapters/inbound/agents/codex"
 	"irrlicht/core/adapters/inbound/agents/copilot"
 	"irrlicht/core/adapters/inbound/agents/geminicli"
+	"irrlicht/core/adapters/inbound/agents/hermes"
 	"irrlicht/core/adapters/inbound/agents/kirocli"
 	"irrlicht/core/adapters/inbound/agents/opencode"
 	"irrlicht/core/adapters/inbound/agents/pi"
@@ -29,8 +30,8 @@ import (
 // while its route segment is "claudecode". That is why the map keys below are
 // written out literally rather than derived from the adapter constants.
 //
-// Scope, stated honestly: this pins the receivers that exist (eight as of
-// #1719's opencode row; #1721's pi row also added the mistral-vibe row #1718
+// Scope, stated honestly: this pins the receivers that exist (nine as of
+// #1722's hermes row; #1721's pi row also added the mistral-vibe row #1718
 // never wrote — exactly the omission the paragraph below predicts, and the
 // whole of issue #1759). It cannot fail for a NEW receiver registered at a
 // different prefix, because registerHookRoutes (core/cmd/irrlichd/startup.go)
@@ -43,6 +44,7 @@ func TestBeaconEndpointPathMatchesTheInstalledReceivers(t *testing.T) {
 		"codex":      codex.HookEndpointPath,
 		"copilot":    copilot.HookEndpointPath,
 		"gemini-cli": geminicli.HookEndpointPath,
+		"hermes":     hermes.HookEndpointPath,
 		"kiro-cli":   kirocli.HookEndpointPath,
 		// mistral-vibe was beacon-delivered from the day it shipped (#1718)
 		// but was never added here; found while adding pi's row (#1721).
--- a/core/adapters/inbound/agents/hermes/adapter.go
+++ b/core/adapters/inbound/agents/hermes/adapter.go
@@ -138,6 +138,24 @@ func StorePath() string {
 	return filepath.Join(homeDir(), storeRelPath)
 }
 
+// transcriptPathIn renders the string every downstream consumer keys a Hermes
+// session on, for a given store path. Hermes has no per-session file, so this
+// composed spelling IS the session's identity as far as the rest of the daemon
+// is concerned, and ComputeMetrics splits it back apart.
+//
+// The watcher and the hook receiver must produce the same string or one
+// session becomes two, so there is one function and two thin callers rather
+// than two spellings — the shape opencode's transcriptPathFor already uses.
+func transcriptPathIn(dbPath, sessionID string) string {
+	return dbPath + sessionQueryParam + sessionID
+}
+
+// transcriptPathFor is transcriptPathIn against the store this daemon resolves
+// — what the hook receiver uses, since a hook payload names no file.
+func transcriptPathFor(sessionID string) string {
+	return transcriptPathIn(StorePath(), sessionID)
+}
+
 // isServiceArgv reports whether argv belongs to a long-lived Hermes service
 // rather than an agent session.
 //
--- a/core/adapters/inbound/agents/hermes/adapter_test.go
+++ b/core/adapters/inbound/agents/hermes/adapter_test.go
@@ -83,8 +83,8 @@ func TestAgentRegistration(t *testing.T) {
 		t.Error("PathForPID must ignore the pid — the store is shared, not per-process")
 	}
 
-	if len(a.Permissions) != 1 {
-		t.Fatalf("len(Permissions) = %d, want 1", len(a.Permissions))
+	if len(a.Permissions) != 2 {
+		t.Fatalf("len(Permissions) = %d, want 2 (store + hooks)", len(a.Permissions))
 	}
 	p := a.Permissions[0]
 	if p.Key != PermissionKeyStore {
@@ -97,6 +97,29 @@ func TestAgentRegistration(t *testing.T) {
 	// stop watchers), so it must carry no closures of its own.
 	if p.Apply != nil || p.Remove != nil {
 		t.Error("observe permission must have nil Apply/Remove")
+	}
+
+	// The hooks permission (#1722) is the other way round: it owns its own
+	// install, so it must carry both closures and declare both files they
+	// write. The contract families in hook*_test.go grade the contents; this
+	// is the shape lock.
+	h := a.Permissions[1]
+	if h.Key != PermissionKeyHooks {
+		t.Fatalf("Permission[1].Key = %q, want %q", h.Key, PermissionKeyHooks)
+	}
+	if h.Kind != permission.KindModify {
+		t.Errorf("hooks Permission.Kind = %q, want modify", h.Kind)
+	}
+	if h.Apply == nil || h.Remove == nil {
+		t.Error("a modify permission that installs must carry Apply and Remove")
+	}
+	if h.Writes == nil || h.Writes.Path == nil {
+		t.Fatal("the hooks permission declares no managed user file")
+	}
+	if len(h.Writes.Also) != 1 {
+		t.Errorf("Writes.Also has %d entries, want 1 — hermes will not run a configured hook "+
+			"whose approval is not recorded, so the allowlist is the other half of the install",
+			len(h.Writes.Also))
 	}
 }
 
--- a/core/adapters/inbound/agents/hermes/agent.go
+++ b/core/adapters/inbound/agents/hermes/agent.go
@@ -1,6 +1,7 @@
 package hermes
 
 import (
+	"irrlicht/core/adapters/inbound/agents/hookjson"
 	"irrlicht/core/domain/agent"
 	"irrlicht/core/domain/permission"
 	"irrlicht/core/domain/session"
@@ -56,6 +57,55 @@ func Agent() agent.Agent {
 					"(WhatsApp, Slack, …) share the same store and are " +
 					"skipped. No row is ever written. Toggling off stops all " +
 					"polling immediately.",
+			},
+			{
+				Key:             PermissionKeyHooks,
+				Kind:            permission.KindModify,
+				Title:           "Install status hooks",
+				FeatureUnlocked: "Blocked-on-user detection and authoritative turn-end detection",
+				// "3 hook entries" is what contracttesting's #1356 count arm
+				// reads. The second file is named in the same breath rather
+				// than left to Detail, because an install that wrote only the
+				// config would never fire and the wizard row is what most
+				// users read.
+				Touches: "Writes 3 hook entries to " + displayConfigPath +
+					" and the matching approvals to " + displayAllowlistPath,
+				Detail: "Hermes Agent runs a shell command on lifecycle events, declared as a " +
+					"`hooks:` block in " + displayConfigPath + ". This adds one entry for each " +
+					"of three events (" + hookjson.EventList(installedHookEvents) + ") inside a " +
+					"clearly marked block of its own; every other line of your config, comments " +
+					"included, is left byte-for-byte alone, and the install refuses rather than " +
+					"guessing if the file is shaped in a way it cannot edit safely. No code is " +
+					"installed anywhere: each entry runs `irrlichd hook-post hermes`, a tiny " +
+					"command irrlicht ships, which reads the daemon's own published address at " +
+					"the moment the event fires — so nothing names a host or a port, nothing can " +
+					"go stale, and it never blocks Hermes even when the daemon is not running. " +
+					"Hermes will not run a configured hook until you approve it, so this also " +
+					"records that approval in " + displayAllowlistPath + " — the same record " +
+					"Hermes' own prompt would write, which granting here is you answering. The " +
+					"three events are observers: none of them is one whose output Hermes reads " +
+					"as a decision, so this can never veto a command or keep a turn running. " +
+					hookjson.RequiresVersion("hermes", minHermesVersion) +
+					" Toggling off removes the block and the approvals (also available via " +
+					"`irrlichd --uninstall-hooks`); nothing else in your Hermes installation " +
+					"was changed.",
+				Apply:  func() error { _, err := EnsureHooksInstalled(); return err },
+				Remove: func() error { _, err := UninstallHooks(); return err },
+				Writes: &agent.ManagedUserFile{
+					Path:      ConfigPath,
+					Uninstall: UninstallHooks,
+					Verify:    VerifyHooksInstalled,
+					// The allowlist is the other half of the install: without
+					// the approval Hermes skips registration and logs
+					// "not allowlisted", so a declaration naming only the
+					// config would leave a file the recorder does not back up
+					// and #1449's refusal does not judge.
+					Also: []func() (string, error){AllowlistPath},
+					Version: &agent.VersionGate{
+						Min:   minHermesVersion,
+						Probe: []string{"hermes", "--version"},
+					},
+				},
 			},
 		},
 	}
--- a/core/adapters/inbound/agents/hermes/hookdisclosure_test.go
+++ b/core/adapters/inbound/agents/hermes/hookdisclosure_test.go
@@ -1,0 +1,19 @@
+// hookdisclosure_test.go wires the hermes hooks permission into the shared
+// issue #1356 contract: the consent copy names every event the installer
+// subscribes to, states the right entry count, and names no event it does not
+// install.
+package hermes
+
+import (
+	"testing"
+
+	"irrlicht/core/internal/contracttesting"
+)
+
+func TestHooksPermission_DisclosureMatchesInstalled(t *testing.T) {
+	contracttesting.AssertHookDisclosureMatchesInstalled(t, contracttesting.HookDisclosure{
+		Agent:         Agent(),
+		PermissionKey: PermissionKeyHooks,
+		Installed:     installedHookEvents,
+	})
+}
--- a/core/adapters/inbound/agents/hermes/hookinstaller.go
+++ b/core/adapters/inbound/agents/hermes/hookinstaller.go
@@ -337,6 +337,21 @@ func hookYAMLConfig(path, command string) hookyaml.Config {
 		BlockKey: hookBlockKey,
 		Owner:    hookRegionOwner,
 		Entries:  hookEntries(command),
+		// hermes rewrites its whole config.yaml from a parsed dict whenever it
+		// saves one — `hermes config set`, the setup wizard, a dashboard save
+		// (hermes_cli/config.save_config -> utils.atomic_yaml_write). Every
+		// comment in the file goes with it, INCLUDING the markers that
+		// delimit our region, while the entries survive as data and keep
+		// firing. Without a sentinel this installer would then read its own
+		// entries as a user's colliding hooks forever and
+		// `--uninstall-hooks` would find nothing to remove, leaving them in
+		// the user's config with no way to take them out.
+		//
+		// The beacon sentinel is the right token: it is in every command this
+		// installer writes, it deliberately excludes the binary path so an
+		// entry naming a moved irrlichd is still ours, and it is the same
+		// string every other adapter identifies its entries by.
+		Sentinel: hookbeacon.Sentinel(AdapterName),
 	}
 }
 
--- a/core/adapters/inbound/agents/hermes/hookinstaller.go
+++ b/core/adapters/inbound/agents/hermes/hookinstaller.go
@@ -1,0 +1,684 @@
+// hookinstaller.go installs irrlicht's Hermes Agent shell hooks — the two
+// files this adapter writes into a user's Hermes installation (issue #1722,
+// epic #1355).
+//
+// # The epic's verdict, and what the source actually says
+//
+// Epic #1355 cut hermes as "cost without payoff for now. hermes is a
+// ProcessOwnedStore whose metrics `of verify` cannot validate anyway", and
+// #1722 restated the payoff half as "a hook would buy event-driven
+// invalidation rather than latency". Both halves are wrong, and the second
+// one is wrong in the same way opencode's was.
+//
+// Measured against Hermes Agent v0.19.0 — read out of the Python package's own
+// source, and fired live through `hermes hooks test` against an isolated
+// HERMES_HOME:
+//
+//   - **hermes has a first-class hook system with its own CLI subcommand.**
+//     `hermes hooks {list|test|revoke|doctor}` (hermes_cli/hooks.py), 25
+//     lifecycle events (`VALID_HOOKS` in hermes_cli/plugins.py), and THREE
+//     independent subscription mechanisms — a gateway-only Python handler
+//     directory, in-process Python plugins under `~/.hermes/plugins/`, and
+//     SHELL hooks declared as a `hooks:` block in `~/.hermes/config.yaml`
+//     (agent/shell_hooks.py). Nobody had looked; `--help` was never the
+//     evidence.
+//
+//   - **The payoff is not invalidation.** hermes fires
+//     `pre_approval_request` and `post_approval_response` around every
+//     dangerous-command approval, and the pending request they bracket lives
+//     in a process-local `_gateway_queues` map and a `contextvars` session
+//     key. NOTHING about a request awaiting an answer reaches `state.db` —
+//     the schema has `sessions`, `messages`, `session_model_usage`,
+//     `state_meta`, `gateway_routing`, `compression_locks` and
+//     `async_delegations`, and no approval table at all. So hermes's store
+//     cannot express "the user is being asked right now", and before this
+//     channel this adapter had no `waiting` state whatsoever. That is
+//     opencode's shape exactly (#1719), arrived at from the other end.
+//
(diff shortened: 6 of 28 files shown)


ingo-eichhorst:
## Gate notes

**SonarCloud `new_duplicated_lines_density` 6.5% vs a 3% threshold — expected, not addressed.** Every other condition on that gate passes (new reliability, security, maintainability all rating 1; security hotspots 100% reviewed). The duplicated blocks are exactly the per-adapter contract wirings, and Sonar names them:

| file | new dup | what the duplicated block is |
|---|---|---|
| `hermes/hookversion_test.go` | 43.7% | the `AssertHookVersionGate` wiring plus `hooksVersionGate` — the same shape opencode, pi, vibe, kirocli and geminicli each carry |
| `hermes/hooks_test.go` | 31.4% | the receiver test doubles (`mockTarget`, `keyedGate`, `mockLogger`, `post`, `newReceiver`) every hook adapter carries verbatim |
| `hermes/hooks.go` | 9.3% | `serveHookRequest`s step order, which is load-bearing and identical by design across receivers |
| `hookyaml/hookyaml_test.go` | 7.0% | the refusal-corpus rows |
| `hermes/hookinstaller_test.go` | 4.6% | the install-type `AssertPermissionGated` wiring |
| `hookyaml/hookyaml.go` | 2.4% | — |

Extracting a shared helper is the thing #1740 forbids in effect: `TestEveryHookInstallWiresItsContractFamilies` resolves the callee type from the adapter's OWN test package, so a wiring moved behind a cross-package helper reads as absent and the adapter falls out of the tripwire. The receiver step order is the same trade in production — docs/testing-contracts.md records each step as separately graded, and a shared `serve()` would put the seven contract families behind one implementation. Same reasoning recorded on #1719 and #1721.

**CodeScene: two of its findings were real and are fixed** (270cfa23). It put `hookyaml.go` at 6.92 with Complex Method on `findBlockKey` and Bumpy Road Ahead on `install`; both had grown a branch at a time while the region-placement model was being worked out — which is precisely where the end-of-file marker-fusion defect hid. `install` is now three named branches (one per placement) and `findBlockKey`s cross-line state is a `topLevelWalk` with one method per grammar. No behaviour change; the terminator guard was re-mutated after the split and seen red again:

```
--- FAIL: TestSplice_PropertyRandomDocuments
    property_test.go:103: output line "  on_session_end:\n" is neither an original line in order nor the block key:
        hooks:  # irrlicht-managed BEGIN (irrlicht) — do not edit; `irrlichd --uninstall-hooks` removes this block
```

Its remaining findings are on the test files (`posixSplit`, `assertOnlyTheRegionWasAdded`, `TestInstalledEventsAreObserverOnly`, `hermesReadEntries`) — each is a table or a small parser whose explicitness is the point.

**Local gates**, all run as separate foreground invocations: `--only go` (incl. replay fixtures), `arch` (ARS 7.9098 → 7.9048, no regression), `tools`, `skills`, `posix`, `security`, `web` — all PASS. `go test ./core/... -race -count=1` is green apart from `TestReadLauncherEnv_Tmux_DetachedResolvesNoHost`, a pre-existing tmux-server flake in `processlifecycle` that this diff does not touch and that passes when the package is run alone.

ingo-eichhorst:
## A third real defect, found after the first review pass — and it was the serious one

While re-reading `hermes_cli/config.save_config`, I found that hermes **re-serializes its whole config.yaml from a parsed dict** on every save (`atomic_yaml_write(config_path, normalized, extra_content=…)`). That path is reached by `hermes config set`, the setup wizard and a dashboard save. It keeps the data and drops **every comment in the file** — and this installer's region markers are comments.

So one `hermes config set` after granting left:

- the three entries still in config.yaml and still firing, with nothing identifying them;
- `EnsureHooksInstalled` reading its **own** entries as a user's colliding hooks and refusing forever — a false #1362 "granted but NOT applied" over an install that was working;
- `UninstallHooks` finding no region, removing nothing, and leaving irrlicht's entries in the user's config **with no supported way to take them out**. `--uninstall-hooks` could not reach them either.

The last one is the one that matters: an unremovable install.

Fixed in 5bb7b5d4. `hookyaml.Config` gains a `Sentinel`; an entry outside the region carrying it is one of ours orphaned by that rewrite, not a collision. Install removes those and re-installs with markers, uninstall removes them, and `Inspect` reports them present-but-not-canonical so #1372's verifier repairs rather than appending a duplicate. An empty sentinel disables recovery, which is the vacuity guard for the four arms.

**Mutation 20** — `Sentinel: ""`, i.e. the state after any `hermes config set`:

```
--- FAIL: TestInstallSurvivesHermesRewritingItsOwnConfig/ensure_recovers_instead_of_colliding_with_itself
    EnsureHooksInstalled after hermes' rewrite: hookyaml: …/config.yaml:4: "pre_approval_request" is already
    declared in the pre_approval_request block by something irrlicht did not write; refusing to install,
    because a YAML mapping cannot hold the key twice and the duplicate would silently replace it
--- FAIL: TestInstallSurvivesHermesRewritingItsOwnConfig/uninstall_removes_them_without_the_markers
    hookinstaller_test.go:557: an entry of ours survived uninstall:
        hooks:
          pre_approval_request:
            - command: "/bin/sh -c '…hermes.test' --version hook-post hermes >/dev/null || true"
          …
    hookinstaller_test.go:561: event "pre_approval_request" survived uninstall
```

A new property-test axis covers it — `stray sentinel-bearing entries (markers dropped by the agent's own writer)`, 171 of 2000 documents on the committed run — with install and uninstall graded separately, because a recovery is deliberately **not** a reversible edit and asserting round-trip equality over it would be a false property.

The axis census earned its keep twice more while this was being added: it caught the generator recording a stray under the wrong event name, and it caught an assertion that pinned the recovered region's indent (wrong when the strays were the block's only content, so the block empties out and the default indent applies).

Sweep re-run after the change: 40,000 × 6 seeds, all green. Local gates re-run: `--only go`, `--only arch` (PASS, no ARS regression), and `go test ./core/... -race` green apart from the pre-existing `processlifecycle` tmux flake.

Also in this push, acting on CodeScene's findings from the first pass: `install` and `findBlockKey` are split into one function per decision (270cfa23). That was not cosmetic — the end-of-file marker-fusion defect had hidden in exactly that density.

ingo-eichhorst:
## Final gate state

| check | result |
|---|---|
| go-test, build-test, guard, ars-gate, CodeQL (3), swift-build, swift-test, both web suites | **pass** |
| SonarCloud | fail — `new_duplicated_lines_density` only |
| CodeScene | fail — advisory |

**SonarCloud: 6.5% → 5.8%, still over the 3% threshold, and still the only failing condition.** New reliability, security and maintainability ratings are all 1; security hotspots 100% reviewed. The remainder is the per-adapter contract wirings — see the earlier comment for the per-file breakdown and why extracting them would take this adapter out of `TestEveryHookInstallWiresItsContractFamilies`.

**CodeScene: "3 new files with code health below 8.00" → 1.** Both production findings are gone: `hookyaml.go` was 6.92, then 7.33 after `install`/`findBlockKey` were split, and is now above the floor after `readBlock` was split too. The one file left is `hookyaml_test.go` at 7.76 — the refusal corpus and the four-arm sentinel test, where the table is the point.

Acting on it was worth it independently of the gate: the density CodeScene named in `install` is exactly where the end-of-file marker-fusion defect hid, and `readBlock` had grown the stray bookkeeping on top of the marker bookkeeping in one body.

The two refactors are behaviour-preserving and the property test proves that rather than asserting it — the axis census prints the identical 11-axis distribution for the same seed before and after, and `closeEntry`-as-a-no-op was re-mutated afterwards and reddens all four sentinel arms plus the property test's stray path.

ingo-eichhorst:
## QA — verified independently, merging, with one caveat the report did not flag

### The verdict holds — I re-derived it from the source

```
$ wc -l …/hermes-agent/hermes_cli/hooks.py
     394
$ grep -n 'pre_approval_request\|post_approval_response' …/hermes_cli/plugins.py
188:    "pre_approval_request",
189:    "post_approval_response",
$ grep __version__ …/hermes_cli/*.py
__version__ = "0.19.0"
```

394 lines exactly as cited, both approval events present, version matching the live `hermes --version` run. **hermes has a first-class hook system with 25 events that nobody had looked at**, and #1355's "cost without payoff" verdict does not survive.

**That is three cut verdicts overturned in a row** — pi, opencode, hermes — each by reading the loader/dispatcher source instead of docs, flags and event lists. That is now a pattern about how the epic's audits were done, not three coincidences.

**The payoff is the same shape as opencode's**, which is the strongest form the answer could take: `tools/approval.py` fires around each dangerous-command approval; the pending request lives in a module-level map plus a `contextvars` key; `state.db` has **no approval table**. So the hermes adapter had **no `waiting` state at all**, and no polling interval could have produced one. `post_approval_response` firing on `deny` and `timeout` means the release genuinely arrives.

### The caveat — the evidence base is a stale checkout, and this should be recorded

The installed `hermes` is a broken shim (I reproduced this: it execs a venv path that does not exist). The source actually used is at:

```
/Users/ingo/projects/_old/hermes-local/.hermes/hermes-agent/hermes_cli/hooks.py
    mtime: Jul 26 2026   (~4 weeks old)
```

A directory named **`_old`**, four weeks stale. That does not invalidate anything — the live run against it was real, the version floor is declared, and `cliversion` fails open — but it means **the 25-event set, the payload shapes and the approval-event names are verified against a checkout nobody has confirmed is current.** Worth stating on the record rather than leaving implicit, and worth re-checking before anyone relies on this adapter's event mapping.

### Three corrections to my framing, all accepted

"No payoff" — wrong. "Invalidation, not latency" — wrong; it buys a state, not a refresh. "Closer to a declaration than an implementation" (#1355's shape claim) — false here: 419 lines of adapter code plus a new 648-line `hookyaml` package. hermes is mistral-vibe's situation, where the cost is the config format.

### The four property-test defects are the best evidence in the PR

Two stand out. **The axis census fired on its own first run** (`block is the last top-level key  0 of 2000`) — the generator was not producing the shape it claimed to, exactly the vacuity the #1753 lesson is about — and fixing it immediately exposed a real defect: a block key at EOF with no final newline gets the BEGIN marker spliced on as a trailing comment. It still parses, nothing complains, and **uninstall can never find the region again.**

And defect 4, found by re-reading `save_config` **after CI had already passed**: hermes re-serializes `config.yaml` from a dict on every save, dropping all comments — and the markers *are* comments. One `hermes config set` produced a false "granted but NOT applied" over a working install and left entries `--uninstall-hooks` could not remove. That is the #1741/#1753 family again: the install works until the agent's own UI touches the file.

### Merging over the duplication gate

SonarCloud fails only `new_duplicated_lines_density` (5.8% vs 3%), itemised per file with the reason — extracting the contract wirings would drop hermes out of `TestEveryHookInstallWiresItsContractFamilies`. Third adapter PR merged over this gate with a recorded reason. CodeScene went 3 unhealthy new files → 1, and both production findings were real: the density it named in `install`/`readBlock` is exactly where defect 3 hid — an advisory check earning its keep.

**0 cells recorded**; `hookcov` reports `StatusGap`, left visible.

Issue #1770, opened 2026-08-22
test(rig): opencode and pi declare hooks and have zero hook-bearing recordings — record one each


ingo-eichhorst:
`of coverage --hooks` on `origin/main` @ 0632aabc:

```
adapter        declares-hooks  cells  recordings  with-hooks  status
gemini-cli                yes     45          38           0  GAP
opencode                  yes     40          30           0  GAP
pi                        yes     44          28           0  GAP

GAP: 3 of 9 hooks-declaring adapters have zero hook-bearing recordings: gemini-cli, opencode, pi
```

opencode (#1719) and pi (#1721) both shipped their hook installers deliberately
without a fixture, so nothing in the catalog exercises either channel. A hook
channel that is only exercised at record time and graded by nothing is the same
class of silence #1695 was filed about — and worse here, because a hookless
recording of a hooks-declaring adapter freezes "this adapter reports no hooks"
as catalog truth.

gemini-cli cannot be closed on this machine: the account is tier-ineligible and
can make no model calls at all, so it is explicitly out of scope. This issue
covers **opencode and pi**.

Two rig defects block the work and are part of it:

1. **pi's drivers cannot find the transcript when the rig's own pi home row is
   used.** `agent-home.sh` gives pi a `PI_CODING_AGENT_DIR` row, and
   `driver-interactive.sh` already prefixes `env PI_CODING_AGENT_DIR=…` on its
   tmux launch — but both drivers then hardcode
   `PI_SESSIONS_DIR="$HOME/.pi/agent/sessions"` to read the transcript back.
   Using the row therefore makes pi write to the scratch home while the driver
   searches the real one: `transcript_recording_or_uuid_missing`, with nothing
   naming the variable.

2. **`wait_for_hook_install` declines to wait for opencode**, because opencode
   has no rig-home row (its paths split across two XDG variables, see
   `agent-home.sh`'s NOT-IN-THE-TABLE section) and the managed-file manifest
   carries no adapter attribution. That leaves exactly the #1735 startup race
   the lib exists to remove: the CLI reads its plugin directory once, at
   startup, so an install landing seconds late yields a complete,
   `of validate`-passing recording with no hook events in it.

Definition of done: `of coverage --hooks` names only gemini-cli, and each
promoted recording is asserted to carry a `hook_received` bound to that
adapter's OWN session — not merely to contain one (see #1768/#1769 for why the
weaker assertion is not enough).

[...]

PR #1779, opened 2026-08-22
fix(opencode): settle a late turn-done hook that races process-exit teardown

### LINKED ISSUE #1772: opencode headless (`opencode run`) loses the turn-done settle: the beacon's session.idle arrives ~2ms after the daemon has already torn the session down
Surfaced while recording opencode's first hook-bearing fixture (#1770 / PR
#1771). Not a recording-rig artefact — the rig is what made it visible, but the
race is in the daemon's handling of a real `opencode run` invocation.

## What happens

`opencode run` (the headless, non-TUI path) exits roughly **115ms** after the
store write that ends the turn. The daemon's teardown for that session —
`process_exited`, then `transcript_removed` — lands first. The plugin's beacon
POST carrying `session.idle` arrives about **2ms later**, against a session that
has already been torn down.

The consequence is not a lost hook. The `hook_received` event **is** recorded.
What is lost is the state transition it exists to drive: the session never
settles to `ready`, and three attempts at `2-1_basic-turn` came back
`unsettled_session … still "working"`.

## Why it matters beyond the fixture

`opencode run` is a normal way to use opencode — scripts, CI, one-shot
invocations. If the turn-done signal consistently arrives after teardown on that
path, then for those sessions irrlicht's hook channel is doing no work and the
state is whatever the transcript fallback last inferred. The hook was added
(#1719) precisely because opencode's store cannot express the approval state on
its own.

The interactive TUI path does not have the problem — the process outlives the
turn, the beacon lands against a live session, and the settle follows three
events later. That is the recording that was ultimately promoted
(`2-5_synchronous-slash-command`): `hook_received` on
`ses_fd69f27acffePoG9lVtPjn9640`, then
`{"new_state":"ready","reason":"agent finished turn → ready"}`.

## Shape of a fix, not chosen

A hook that names a session the daemon has just reaped is currently
indistinguishable from one naming a session that never existed. Options:

1. **A short grace window on teardown** — hold a reaped session's identity
   briefly so a late turn-done can still resolve to it. Needs a bound, or it
   reintroduces the ghost-session class.
2. **Order the teardown behind the beacon** — only viable if the daemon can know
   a beacon is in flight, which it cannot today.
3. **Accept and document** — headless opencode is transcript-inferred only, and
   say so in the adapter's capability model rather than leaving it implied.

(3) is honest and cheap; (1) is the one that actually delivers the channel.

## Acceptance evidence

A test that drives the daemon with a `session.idle` beacon arriving after
`process_exited` for the same session id, and asserts the session reaches
`ready` — red today. The timings above (115ms / 2ms) come from three observed
runs and should be restated as whatever the test measures rather than copied.

ingo-eichhorst:
## Summary

`opencode run` (headless) exits so quickly after its final store write that the daemon's own process-exit teardown — `HandleProcessExit` records `process_exited` and deletes the session's repo row — routinely wins the race against the plugin's `session.idle` beacon POST (measured at #1770/#1771: ~115ms process exit after the turn-ending write, ~2ms further for the beacon). `hook_received` was still recorded, but the `ready` transition it exists to drive was silently dropped once the repo row was gone — headless opencode sessions got permanently stuck at "working".

## Fix chosen

Of the three options the issue laid out, this implements **(1) a short grace mechanism on teardown**, using the repo's *existing* bounded, self-expiring `deletedSessions` tombstone (10s cooldown, already used to prevent ghost-session resurrection from late transcript writes) rather than inventing a second grace window:

- `PIDManager.deleteSession` (the single choke point for an actual repo-row removal — process exit, the ready-TTL/liveness sweep) now hands its last known `*session.SessionState` to a new optional `OnSessionRemoved` callback.
- `SessionDetector` uses it to cache a one-shot snapshot (`deletedStates`) alongside the existing tombstone, cleared at every point the tombstone itself is cleared.
- `dispatchHookActivity` marks its injected event `Terminal` exactly when the hook is `session.HookStop` — the one hook name in that shared dispatch that authoritatively means "the turn is done" (as opposed to a permission prompt or compaction signal).
- When a `Terminal` hook arrives for an already-tombstoned session within the cooldown, `processActivity` revives the cached snapshot through the *ordinary* classify-and-transition path (`processActivityLocked`) instead of dropping it — settling the session to `ready` exactly as it would have while still alive. Ordinary (non-`Terminal`) activity is unaffected: a plain late transcript write still can't resurrect a deleted session.

**Why this doesn't reintroduce the ghost-session class:** no new session state is added (still exactly working/waiting/ready), and a revived session is an *ordinary* `ready` session — reaped by the same pre-existing `SweepDeadPIDs`/`CheckPIDLiveness` dead-PID sweep (5-15s cadence) that cleans up any other completed session with a dead PID. `/clear` and the dedup/supersession reconciliation paths delete their repo rows directly, bypassing `deleteSession`, so they're deliberately **not** covered by this — reviving a session removed on purpose would be exactly the ghost class the cooldown exists to prevent.

The mechanism is adapter-agnostic (keyed on the shared Stop-hook dispatch, not an adapter name), so it also covers hermes' `on_session_end`, which the investigation found is very likely exposed to the identical race (unverified live, but structurally identical: same `dispatchHookActivity`/`HandleStopHook` path).

## Acceptance evidence (issue's ask)

`core/application/services/session_detector_teardown_race_test.go` — `TestSessionDetector_LateStopHookAfterProcessExit_SettlesToReady` — drives `SessionDetector.HandleProcessExit` immediately followed by `HandleStopHook` for the same session id, within the 10s cooldown, and asserts the session reaches `ready`.

**Confirmed red before the fix**: reverting the fix files (keeping the test) and running it gives:
```
last saved state = "", want "ready" — the turn-done signal was dropped against an already-torn-down session (issue #1772)
```
Restoring the fix makes it green. Full `go test ./core/application/services/...` and `go test ./core/...` both pass with the fix in place (one unrelated pre-existing flake in `core/e2e`'s `TestFSWatcher_EmitsEventsForTranscriptCreateAndModify` was observed once under heavy concurrent load and confirmed to pass 3/3 in isolation — unrelated package, not touched by this diff).

## Verification run

- `go test ./core/... -race -count=1` — pass (full suite, including architecture tests)
- `go test ./tools/onboarding-factory/... -race -count=1` — pass (no golden drift from this change)
- `tools/preflight.sh --only arch` — ARS gate pass, no composite/category regression
- `tools/preflight.sh --only go/tools/skills/posix/bash/security` — all pass
- Pre-push hook (`tools/preflight.sh --changed --budget 540`) — pass

## Risks / follow-ups for a reviewer

- This only fixes the **hook-channel** race for the Stop/idle signal. The issue's own framing treats headless opencode's transcript-inferred fallback (`step-finish reason=stop → turn_done`, still present in `parser.go`) as the pre-existing baseline; I have **not** verified live whether that transcript-polling path independently loses the identical race against `HandleProcessExit`'s kqueue-driven teardown (it's a separate, DB-polling-interval-based channel). Worth a live re-record of the `2-1_basic-turn` scenario to confirm; out of scope here since it needs a live opencode CLI + local LM Studio model.
- hermes is very likely exposed to the identical race (same shared dispatch path) but this PR does not add a hermes-specific regression test — only observed structurally, not reproduced live.

Closes #1772

### DIFF
--- a/core/application/services/pid_manager.go
+++ b/core/application/services/pid_manager.go
@@ -115,6 +115,16 @@ type PIDManager struct {
 	// clean up its own tracking structures (e.g. projectSessions map).
 	onSessionDeleted func(sessionID string)
 
+	// onSessionRemoved is called from deleteSession — the single choke point
+	// for an actual repo-row removal — with the session's last known state,
+	// just before that row is deleted. Unlike onSessionDeleted (id-only, and
+	// also fired earlier from HandleProcessExit before any state is loaded),
+	// this exists so a caller can cache the state itself: SessionDetector
+	// uses it to let a hook that authoritatively signals turn-done but loses
+	// the teardown race still settle the session to ready instead of being
+	// silently dropped (issue #1772). Optional; nil disables the cache.
+	onSessionRemoved func(*session.SessionState)
+
 	// onChildDeleted is called when a child session is removed by the
 	// liveness sweep so the SessionDetector can re-evaluate the parent.
 	// Without this the parent can be left stuck in `working` forever
@@ -171,6 +181,7 @@ type PIDManagerDeps struct {
 	ProcessNames     map[string]string
 	LiveCWDs         LiveCWDsFunc
 	OnSessionDeleted func(sessionID string)
+	OnSessionRemoved func(*session.SessionState)
 }
 
 // NewPIDManager creates a PIDManager with the given dependencies.
@@ -185,6 +196,7 @@ func NewPIDManager(deps PIDManagerDeps) *PIDManager {
 		processNames:     deps.ProcessNames,
 		liveCWDs:         deps.LiveCWDs,
 		onSessionDeleted: deps.OnSessionDeleted,
+		onSessionRemoved: deps.OnSessionRemoved,
 		pendingPIDs:      make(map[string]int),
 	}
 }
@@ -588,6 +600,9 @@ func (pm *PIDManager) deleteSession(s *session.SessionState, reason string) {
 	})
 	if pm.onSessionDeleted != nil {
 		pm.onSessionDeleted(s.SessionID)
+	}
+	if pm.onSessionRemoved != nil {
+		pm.onSessionRemoved(s)
 	}
 	_ = pm.repo.Delete(s.SessionID)
 	pm.log.LogInfo("session-cleanup", s.SessionID,
--- a/core/application/services/session_detector.go
+++ b/core/application/services/session_detector.go
@@ -136,6 +136,20 @@ type SessionDetector struct {
 	// --continue, not ghost events from a dying process.
 	deletedSessions map[string]int64
 
+	// deletedStates caches the last known *session.SessionState for a session
+	// removed via PIDManager.deleteSession (process exit, the ready-TTL/
+	// liveness sweep), keyed the same as deletedSessions and cleared at the
+	// same points. It exists so an authoritative turn-done hook (opencode's
+	// session.idle, hermes' on_session_end — anything that sets
+	// agent.Event.Terminal) that loses the race with that exact teardown
+	// still has a session to classify against instead of a repo row that is
+	// simply gone (issue #1772). Not populated for /clear or the dedup/
+	// supersession reconciliation paths, which delete the repo row directly
+	// without going through deleteSession — reviving those would be exactly
+	// the ghost-session class deletedCooldown exists to prevent. See
+	// cacheDeletedSnapshot and processActivity's Terminal branch.
+	deletedStates map[string]*session.SessionState
+
 	// hostGateRejected tracks session IDs the host-ancestry admission gate
 	// (issue #784) has already rejected. No cooldown/expiry, unlike
 	// deletedSessions — a rejected PID's process ancestry (e.g. CodexBar,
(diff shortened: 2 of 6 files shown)


ingo-eichhorst:
## Review update

Ran this diff through \`.claude/skills/ir:code-review/SKILL.md\` at medium effort (base \`origin/main...HEAD\`), with a subagent independently reproducing the red-before-fix test result (reverse-applying the production diff, confirming the same \`last saved state = "", want "ready"\` failure, then restoring) and stress-testing four specific risk vectors (locking/aliasing safety of the revival path, tombstone/snapshot cleanup completeness across all 3 call sites, the `Terminal` flag's full read-site set, and wrong-state risk from reviving a stale snapshot).

**Verdict: clean / 2 low-severity findings, both refuted as functional bugs on independent verification. Lean merge.**

1. The PR description's point-3 self-check ("exactly one read site for `ev.Terminal`") was accurate for the *hook* path but incomplete: `processActivity`'s new revival branch is itself a second read site, also reachable via `opencode/watcher.go`'s pre-existing transcript-derived `Terminal` flag (unrelated to the plugin hook). Traced through — this generalizes correctly (a late transcript-based terminal signal racing teardown gets the same revive-to-ready treatment) rather than causing any wrong behavior. Noted here for the record per this repo's "dismissals carry evidence" convention.
2. A successful revival clears the one-shot `deletedStates` snapshot but leaves the `deletedSessions` timestamp tombstone in place for the remainder of the cooldown. Traced and confirmed harmless — the revived row makes `repo.Load` succeed again, so nothing downstream re-consults the tombstone for that id. Added a clarifying code comment (pushed) so this doesn't need re-deriving on a future read.

No concurrency, state-correctness, or ghost-session-class defects found. Full `go test ./core/... -race -count=1`, `go test ./tools/onboarding-factory/... -race -count=1`, and `tools/preflight.sh` (go/arch/tools/skills/posix/bash/security groups) all pass.

PR #1784, opened 2026-08-23
test(services): isolate hook-repair gate 2's real, independent property (#1751)

### LINKED ISSUE #1751: test(kiro-cli): each consent gate on the refresh path is individually undetectable — only removing both reddens
Found while QA-ing #1749 (issue #1736).

That PR's M9 mutation removed consent gates from kiro-cli's refresh path one at a
time and measured:

- remove gate 2 alone → **green**
- remove gate 3 alone → **green**
- remove both → red

The redundancy is real and intentional — `hook_reverify.go`'s header documents
that the two gates each independently cover the other — so **this is not a live
defect**. Nothing is currently unprotected at runtime.

## Why it is still worth an issue

It means a future single-gate removal is invisible to the test suite. The second
removal would be the first signal, and by then the change that removed the first
gate has been merged, reviewed and forgotten. That is the same shape as #1475,
where a contract passed for the wrong reason: one permission moving through three
states proved a call site was gated on *something*, and a receiver gated on the
WRONG permission answered identically.

The repo's own rule covers this: a green that was never red is a claim, not
evidence. Each gate individually is currently in that position.

## What would close it

A test per gate that holds the *other* gate open, so each one is the sole thing
standing between the input and the write — the fourth-arm idea
`AssertPermissionGated` already uses (`Key` denied while `OtherKeys` are granted,
which is the one state that tells "gated on the right key" from "gated on
something").

Whether that is worth the test surface here is a judgement call, not a
foregone conclusion: defence-in-depth that is individually untested is still
defence-in-depth, and the cost of pinning each gate is a test that will need
updating whenever the gating is restructured. Recorded so the decision is made
deliberately rather than by never noticing.

Found in #1749. Part of #1355.

ingo-eichhorst:
Closes #1751.

## What this does

`hook_reverify.go` documents three independent consent refusals on the
#1372 repair path. Issue #1751 (found while QA-ing #1749) asked whether
gate 2 (`RepairGrantedHookInstall`'s own pre-check) and gate 3
(`runEffects`' re-read under `effectMu`) are each individually covered by
a test — a mutation sweep in #1749 found that deleting either ALONE left
the suite green, and only deleting both reddened.

## Investigation, with mutation evidence (`tools/mutate.sh`)

Ran both single-gate mutations against the full
`core/application/services` + `core/adapters/inbound/agents/kirocli`
suites (not just the kiro-cli-scoped test #1749's own check used):

- **Deleting gate 3 alone** already reddens three existing tests in
  `hook_reverify_gate_internal_test.go`
  (`TestRunEffects_SkipsAGrantEffectWhoseStateIsNoLongerGranted`,
  `TestRunEffects_ReportsZeroWhenEveryEffectIsSkipped`,
  `TestRepairGrantedHookInstall_RevokedWhileWaitingOnTheEffectLockIsNotAttempted`).
  Gate 3 was already individually covered — this file's own header just
  didn't say so.
- **Deleting gate 2 alone** leaves *every* write-observing test green,
  including `TestReverify_RevokedPermissionNeverTouchesTheFileOnDisk`,
  which this same file's header claimed ("proved by deleting it: the
  on-disk test's direct-call arm goes red") would catch it. That claim
  was never mutation-checked and is false — verified directly:

  ```
  $ tools/mutate.sh core/application/services/permission_service.go \
      '<gate 2's !granted check>' '<removed>' \
      go test ./core/application/services/... \
      -run TestReverify_RevokedPermissionNeverTouchesTheFileOnDisk -v -count=1
  --- PASS: TestReverify_RevokedPermissionNeverTouchesTheFileOnDisk (0.00s)
  ```

The reason is structural, not a test-writing gap: gate 3 re-derives the
IDENTICAL fact (is this key currently Granted?) from the same `s.set`,
immediately before the closure runs, against the same hardcoded target
(`StateGranted`) `RepairGrantedHookInstall` always passes. For every input
that trips gate 2, either gate 3 also trips (state genuinely still
not-Granted — the common case), or consent changed to Granted in between
(in which case running the effect is correct, not a violation). There is
no third case, so no write-observing test — existing or new — can ever
separate "gate 2 refused" from "gate 2 let it through and gate 3 refused a
moment later".

`AssertPermissionGated`'s fourth-arm shape (hold one axis open, exercise,
observe) doesn't transfer here: that helper isolates a permission KEY from
OTHER keys — two independent axes. Gates 2 and 3 are not two axes; they
read the same state twice. There is nothing to hold open for a
write-observing test.

## What I added instead

Gate 2 DOES have one real, independently-provable effect: it keeps a
non-granted repair from ever taking `effectMu`. Added
`TestRepairGrantedHookInstall_DoesNotContendForEffectMuWhenNotGranted`,
which holds `effectMu` (standing in for a slow in-flight wizard answer)
and asserts a repair for a DENIED permission returns without ever
contending for it.

Mutation-verified this new test in both directions:

- **Gate 2 removed alone** → new test REDDENS (confirmed: "RepairGrantedHookInstall
  blocked on effectMu for a permission that was never granted").
- **Gate 3 removed alone** → new test stays GREEN, while the three
  existing gate-3 tests redden as expected — confirming true isolation
  (the new test is specific to gate 2, unaffected by gate 3's removal).

Also corrected the two docstrings that asserted gate 2's deletion was
"visible as a red test" without that ever having been mutation-checked:
this file's own header, and `RepairGrantedHookInstall`'s docstring in
`permission_service.go`.

No production behavior changes — docstring corrections plus one new test.

## Judgment call (per the issue's own framing)

The issue asked me to decide, and justify, whether the per-gate test
surface was worth it. Given the mutation evidence above, a literal
"gate 2 in isolation, holding gate 3 open" write-observing test is not
just costly, it is not constructible — any such test is either vacuously
satisfied whenever gate 3 is intact, or requires disabling gate 3's
actual mechanism (not a legitimate test wiring). So instead of forcing
the issue's literal ask, I tested what gate 2 actually, independently
provides (contention avoidance) and corrected the two comments that were
making an unverified claim about what removing it would do — the
AGENTS.md "dismissals carry evidence" rule applies to code comments, not
just triage write-ups.

## Files touched

- `core/application/services/hook_reverify_gate_internal_test.go` — new
  test + corrected header comment.
- `core/application/services/permission_service.go` — corrected
  docstring only (no behavior change).

## Gates run

- `go test ./core/application/services/... ./core/adapters/inbound/agents/kirocli/... -race -count=1` — PASS
- `tools/preflight.sh --only go` — PASS
- `tools/preflight.sh --only arch` — PASS (composite 7.893010788442787 → 7.892982424902501, no regression)
- `tools/preflight.sh --only tools` — PASS
- `tools/preflight.sh --only security` — PASS
- pre-push hook (`tools/preflight.sh --changed --budget 540`) — PASS

### DIFF
--- a/core/application/services/hook_reverify.go
+++ b/core/application/services/hook_reverify.go
@@ -50,9 +50,18 @@
 // and the write that follows may wait on a mutex behind an in-flight wizard
 // answer. A background repair that outlives consent would break #570 more
 // thoroughly than the bug it fixes, so the loop owns no write path of its own —
-// it can only ask the service, and the service can say no. Each of the three
-// has been individually deleted and seen to turn a test red; see
-// hook_reverify_gate_internal_test.go for the one that only a race can reach.
+// it can only ask the service, and the service can say no.
+//
+// Gates 1 and 3 have each been individually deleted and seen to turn a test
+// red — see hook_reverify_gate_internal_test.go for gate 3, the one that only
+// a race can reach. Gate 2 cannot be shown that way: gate 3 re-derives the
+// identical fact from the same recorded state immediately before the closure
+// runs, so no write-observing test can ever tell "gate 2 refused" from "gate 2
+// let it through and gate 3 refused a moment later" (#1751, mutation-verified
+// via tools/mutate.sh). Gate 2's own, independently provable property is
+// contention avoidance, not write safety — see
+// hook_reverify_gate_internal_test.go's header and
+// TestRepairGrantedHookInstall_DoesNotContendForEffectMuWhenNotGranted.
 package services
 
 import (
--- a/core/application/services/hook_reverify_gate_internal_test.go
+++ b/core/application/services/hook_reverify_gate_internal_test.go
@@ -13,15 +13,14 @@ import (
 // The third consent gate on the #1372 repair path, tested where it can actually
 // be reached.
 //
-// RepairGrantedHookInstall has three independent refusals and they cover
-// different windows:
+// RepairGrantedHookInstall has three refusals and they cover different
+// windows:
 //
 //  1. the verifier loop asks Granted before it reads the file (proved in
 //     hook_reverify_test.go, and by deleting that check: the revoke test goes
 //     red);
 //  2. RepairGrantedHookInstall re-reads the recorded state before it queues
-//     anything (proved by deleting it: the on-disk test's direct-call arm goes
-//     red);
+//     anything;
 //  3. runEffects re-reads the state a THIRD time, under effectMu, immediately
 //     before running the closure.
 //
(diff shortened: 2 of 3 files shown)


ingo-eichhorst:
Review subagent (medium effort) came back with one confirmed finding, now fixed:

- `hook_reverify.go`'s own "# Consent" header comment still said "Each of the
  three has been individually deleted and seen to turn a test red" — the same
  stale gate-2 claim corrected in the other two docstrings, just not yet
  propagated to this third location. Fixed in 926b0e4f.

Everything else the reviewer checked came back clean: confirmed no production
logic changed (docstrings + one new test only), independently re-ran the
gate-2 mutation and reproduced the same result, ran the new test 5x under
`-race` (no flake, no goroutine leak), verified the "no third case" argument
against the real `RepairGrantedHookInstall`/`runEffects` code, and confirmed
the corrected docstrings are accurate. One non-blocking nit noted (a shared
`waitOrFatal` helper could dedupe the wait-pattern between two tests) — left
as-is per the reviewer's own read that it's a nit, not a defect, at only two
call sites with different semantics.

Clearing the WIP marker now.

Issue #1818, opened 2026-08-25
claudecode: the terminal API-error shape now carries `apiErrorStatus`, and the adapter discards it


ingo-eichhorst:
Found while recording #1803's provider-error cells. Two newly-committed fixtures are the evidence.

## The claim in the code, and why it is now stale

`core/adapters/inbound/agents/claudecode/sessionerror.go`'s `terminalAPIError` derives no `HTTPStatus`. #1811's PR body states the reasoning:

> The terminal shape's own text reads `API Error: API returned an empty or malformed response (HTTP 200)` — the number in the prose is the transport's *success* code, so no status is derived and `HTTPStatus` stays nil.

That was true of the shape available at the time: the only number was in prose, and it was the wrong number. It is no longer the only number.

## What claude-code 2.1.245 actually writes

From `replaydata/agents/claudecode/scenarios/2-23_provider-overloaded-terminal/recordings/*/transcript.jsonl`:

```json
{"type":"assistant","message":{"model":"<synthetic>","stop_reason":"stop_sequence",
 "content":[{"type":"text","text":"API Error: Repeated 529 Overloaded errors. ..."}]},
 "error":"server_error","isApiErrorMessage":true,"apiErrorStatus":529, ...}
```

and from `2-24_auth-credentials-rejected/recordings/*/transcript.jsonl`:

```json
{"error":"authentication_failed","isApiErrorMessage":true,"apiErrorStatus":401,
 "message":{"content":[{"type":"text","text":"Invalid API key · Fix external API key"}]}, ...}
```

Two **structured** top-level fields sit next to the `isApiErrorMessage` flag the parser already reads:

- `apiErrorStatus` — the real HTTP status (529, 401), not the transport's 200.
- `error` — a short machine token (`server_error`, `authentication_failed`) that is a better `Class` than the current constant.

`terminalAPIError` reads `isApiErrorMessage` and the prose, and drops both.

## Why it matters

`SessionError.HTTPStatus` is a pointer precisely so "absent" and "zero" stay distinguishable, and the epic's whole point is that a red session says something useful. With this, a terminal failure can say *529, the provider is overloaded, try again* versus *401, your credentials are rejected, waiting will not help* — the one distinction a user acts on differently. Today both render the same.

## Scope

[...]

Issue #1817, opened 2026-08-25
replay: the offline sidecar replays a mid-turn process death as `ready`, not `error`


ingo-eichhorst:
Found while recording #1803's `agent-process-crash-midturn` cell. Measured, not inferred.

## What happens

`tools/onboarding-factory/cmd/replay/replay_sidecar.go`'s `applyTimelineControlEntry`, `case timelineProcessExit:`, ends with:

```go
*d = debounceState{debounceDelay: d.debounceDelay}
r.state = session.StateReady
return true, nil
```

It sets `ready` unconditionally, for every process exit. The daemon does not: since #1800, `HandleProcessExit` routes through `retainAsProcessDeath`/`diedMidTurn` (`core/application/services/session_detector_lifecycle.go`) and a PID that dies with a turn still open produces `error`, not `ready`. Nothing in `cmd/replay/` mentions `ProcessDeath` or `diedMidTurn` (`git grep` in that dir returns nothing).

## The measurement

The newly-committed recording `replaydata/agents/claudecode/scenarios/2-25_agent-process-crash-midturn/recordings/2026-08-25-09-02-30_irrlichd-0.6.0+6c65754/` is the fixture. Its golden's own `extended_check` block says it:

```
"recorded_transition_count": 2,
"replayed_transition_count": 1,
"ordered_mismatches": [{"index":1,"kind":"missing_in_replay","recorded":"working→error"}],
"recorded_unique_kinds": ["ready→working","working→error"],
"replayed_unique_kinds": ["ready→working"],
"missing_kinds": ["working→error"]
```

The daemon recorded `working → error [agent process died mid-turn → error]`; the offline replay reconstructs only `ready → working`.

Attribution is measured: #1803 added three recordings and the catalog census moved `Recordings` 316→319 but `Divergent` only 150→151. The two sibling recordings landed in the same PR (`2-23_provider-overloaded-terminal`, `2-24_auth-credentials-rejected`) replay **exactly** — `ordered_matches: 2`, no mismatches — because their errors are transcript-tier and the sidecar shares the real classifier. Only the process-tier one diverges.

## Why it was not fixed in #1803

[...]

Issue #1801, opened 2026-08-24
Phase 3: expose `error` over the wire and render it red in the web dashboard


ingo-eichhorst:
Part of #1796 — Phase 3. Blocked on #1798. Can run in parallel with the macOS work.

**Scope changed after this issue was written:** the daemon-wide error banner is now **in scope here**, not a follow-up. See the section at the bottom.

## Go — the wire

Every place that enumerates the three states exhaustively (from the Phase-1 research map; **file:line unverified by me — re-locate before editing**):

- `core/cmd/irrlichd/handlers.go` — the `ByState` empty-fallback (~`:627-631`) and the `/api/v1/state` counting switch (~`:1592-1599`)
- `core/application/services/diagnostics_service.go` — `stateView` counting switch (~`:1153-1160`)
- `core/domain/session/grouped.go` — subagent summary switch (~`:289-297`); note `subagentSummary` at `session.go:46-51` is a fixed `{Total, Working, Waiting, Ready}` struct and needs an error member
- `core/ports/outbound/ports.go` — `StateSeriesResult.ByState` (~`:333-370`)
- `core/adapters/outbound/filesystem/concurrency_tracker.go` — `ByState` seeding (~`:305-310`, `:415-419`) and `concurrencyActive()` (see the settled decision below)
- `core/cmd/irrlicht-ls/main.go` — `--state` flag validation (~`:122-124`)

Tests that pin exactly three keys and must be updated: `core/cmd/irrlichd/history_state_endpoint_test.go:50`, `core/cmd/irrlichd/activity_matrix_test.go:77-79, 99-101, 127-129`.

### Settled: an errored session is NOT active for concurrency

`concurrencyActive()` is verified at `core/adapters/outbound/filesystem/concurrency_tracker.go:144-146`:

```go
func concurrencyActive(state string) bool {
	return state == session.StateWorking || state == session.StateWaiting
}
```

**Leave this function's meaning unchanged** — `error` is treated like `ready` and does not make a session active. Rationale: nothing clears an error until the next successful turn, so an errored session would otherwise inflate the active count forever. Add a defect test pinning that an errored session does not consume a concurrency slot.

## Web dashboard

**Colour is defined in three independent places** — all three need an error entry, and they deliberately do not share a source:

[...]

Issue #1800, opened 2026-08-24
Phase 2b: error signals for gemini-cli, opencode, codex, pi + process-death detection


ingo-eichhorst:
Part of #1796 — Phase 2b. Blocked on #1798. Sibling of #1799 (do that one first — it is the higher-value pair and sets the pattern).

The remaining adapters, in descending order of how much signal they actually have. Several will end up honestly reporting "cannot detect an error", which is a valid outcome — record it in the adapter's onboarding cell rather than faking coverage.

## gemini-cli — already parsed, needs promoting

`core/adapters/inbound/agents/geminicli/parser.go:195-201` already sets `EventType="turn_done"`, surfaces the error text as `AssistantText`, and sets `ev.IsError = true` (`:199`). The text reaches the UI as a waiting headline today. Promote it to the new error fields; extract the status code where the payload has one.

Second site at `:360-364` for tool-level errors — keep that distinct (tool failure ≠ session error).

## opencode — error is the only turn-terminator

Two paths, both already wired as *lifecycle* signals with the error content discarded:
- Live: `core/adapters/inbound/agents/opencode/watcher.go:489-511` (`isErrorMessage`, reads `raw["error"]` as string-or-object), used in the terminal decision at `:596-602`.
- Replay: `core/adapters/inbound/agents/opencode/parser.go:63-79` (`raw["_error"]`), plus `step-finish reason:"error"` at `:135-137`.

Keep the existing turn-termination behaviour intact — it is load-bearing for aborted turns (#493). Add the error detail alongside it.

## codex — currently cannot tell an error from an ESC

`codexEventMsgIsTurnDone` (`core/adapters/inbound/agents/codex/parser.go:216-226`) returns true for `task_complete` **or** `turn_aborted`, and never reads `payload.reason`. Read `reason` and distinguish.

**Caveat, from the Phase-2 research map (unverified by me):** every committed codex fixture has `reason:"interrupted"` and no fixture contains a `stream_error` event. If that holds, this change is **not testable against existing fixtures** and needs either a new recording or a hand-built fixture. Verify the fixture census first; if it confirms, say so and scope accordingly rather than shipping an untested branch.

Codex does already have a real quota signal — `rate_limits.rate_limit_reached_type` → `RateLimitSnapshot.ReachedType` (`parser.go:503+`). That is a metric, not a state, today. Decide whether quota-reached should drive the `error` state for codex.

## pi — field exists, never read

[...]

PR #1819, opened 2026-08-25
feat(factory): four error scenarios, their traits, and a recipe runtime block

### LINKED ISSUE #1803: Phase 5: four onboarding-factory scenarios simulating error conditions
Part of #1796 — Phase 5. Blocked on #1798 and #1799/#1800 (the daemon must be able to *reach* the error state before a recording can assert it).

Four new scenarios so the error state is simulated, recorded and graded like every other behaviour.

## Do this first: add the traits

`tools/onboarding-factory/internal/matrix/capability.go:84-151` holds a **closed set of traits in Go, not data**, with exactly one scenario per trait. Three error-adjacent traits already exist (verified):

```
:92   {"error_epilogue",   "writes a terminal record when a turn dies of an error", "turn-aborted-by-error"}
:117  {"quota_exhaustion", "surfaces a provider quota refusal",                      "token-quota-exhausted"}
:126  {"provider_failover","fails over to a second provider WITHIN one turn",        "provider-failover-midturn"}
```

**Add a trait for each new scenario.** Without one there is no synthesis path: every one of the 11 adapter columns needs a hand-written cell directory. With one, an adapter declared `absent`/`untraced` is synthesized automatically and needs no cell on disk. This is the difference between ~4 cells of work and ~44.

Constraint: `TestEachScenarioHasAtMostOneTrait` fails if two traits name one scenario. And `of validate` then enforces bidirectional agreement across all 11 adapters, so expect to declare capabilities for adapters that genuinely cannot observe an error (see #1800 — aider, hermes, mistral-vibe, kirocli).

## The four scenarios

Next free ids in section 2 are **2.22 onwards** (verified: 46 scenarios, section 2 currently maxes at 2.21).

Write them via `of scenario add` — **never hand-edit `replaydata/`**. Schema is exactly 5 keys (`id`, `name`, `description`, `acceptance_criteria`, `process`), enforced by `allowedScenarioKeys` in `cmd/of/validate.go`.

### 1. `provider-overloaded-retry`
Mock returns HTTP 529 overloaded, the agent retries with backoff, the turn then succeeds. Matches the unclaimed candidate #19 at `.specs/agent-scenarios-candidates.md:65`.

Mock is nearly free: `tools/onboarding-factory/recording/mock-anthropic-5xx/main.go` already emits an `overloaded_error` payload — change the status and add a success-on-attempt-N path.

**This recording is the measurement for the accepted flash risk in #1798** — with no minimum hold, a fast retry may go red and back too quickly to see. Capture the actual timings and report them on #1798.

### 2. `provider-overloaded-terminal`
Same mock, never recovers. Proves the error survives after the retry loop ends and the session does not stick in `working`. This is the pair that makes the retry-vs-give-up distinction in #1798 testable.

### 3. `auth-credentials-rejected`
New 401/403 mock. Terminal and non-retryable, and the most common real-world error.

### 4. `agent-process-crash-midturn`
SIGKILL the agent mid-turn so no turn-end signal is written. Matches unclaimed candidate #23 at `.specs/agent-scenarios-candidates.md:77`. **No mock server needed** — the driver step grammar already has a `sigkill` step (`replaydata/agents/claudecode/driver-interactive.sh:81`).

## `meta.ends_unsettled`

Current value (verified): `["session-end", "token-quota-exhausted", "user-esc-interrupt", "subagent-orphan-cleanup"]`.

Add the scenarios that end unsettled **by design** — `provider-overloaded-terminal` and `agent-process-crash-midturn` are the likely two. Without this every recording gets a `suspect` completeness verdict. The list is deliberately narrow; do not add a scenario that merely *happens* to end unsettled.

## Known trap: retried-into-invisibility

Several adapters swallow the very status codes this feature is about. Read these mock headers **before** picking a status code — they document it from prior incidents:

- `mock-gemini-5xx/main.go:13-22` — gemini's `retryWithBackoff` swallows 429/500/503 with **no transcript line at all**, so it uses a non-retryable 400 INVALID_ARGUMENT.
- `mock-openai-429/main.go:17-25` — the AI-SDK retries 429 forever, so it defaults to 403 `insufficient_quota` to get a terminal error.

A plain 529 may therefore be **unobservable** on gemini-cli and opencode. That is a legitimate `untraced` capability declaration, not a bug to force.

## Known constraint: recording needs a bespoke recorder

`run-cell.sh` cannot drive these today — `--bare` and `ANTHROPIC_BASE_URL` are not recipe fields. The two existing bespoke recorders (`record-token-quota-exhausted.sh`, `record-turn-aborted-by-error.sh`) work around it, and `record-token-quota-exhausted.sh:9-17` names the proper fix: add `bare_mode` + `env` to the per-cell recipe and a pre-driver mock hook in `run-cell.sh`.

**Consider doing that fix as part of this issue** — four new mock-driven scenarios is the point where four more bespoke scripts stops being reasonable. Flag it if you defer it.

Also: per the codex 2-14 cell metadata, the Anthropic mocks are base-URL-override-only and **only claudecode can be aimed at them** — which is why codex/opencode error cells have no recordings *(unverified by me)*.

## Do not touch the core twelve

`validate_maturity.go` hard-fails on any core-set size other than 12. These four are additions, not promotions.

## Gates

- `go run ./tools/onboarding-factory/cmd/of validate` (CI gate)
- `go test ./tools/onboarding-factory/... -race -count=1`
- `tools/replay-fixtures.sh` — a new recording triggers it
- Goldens: `UPDATE_REPLAY_GOLDENS=1 go test ./tools/onboarding-factory/cmd/replay/... -count=1` (the `-count=1` matters). Commit only the touched adapter's goldens.

## Triage

`ready-for-agent`. Large but well-precedented — `2-9` and `2-14` are the working templates for every step.

ingo-eichhorst:
> **Figures corrected after the review round.** `provider-overloaded-terminal`'s
> dwell reads **3.67 s**, not the 1.94 s an earlier draft quoted: that number came
> from an exploratory run rather than the promoted recording, and the promoted
> recipe still carried an `--retry-after 1` that the 401 cell had already measured
> claude-code to ignore. The flag is removed and the cell re-recorded. See the
> review-round comment below and the correction on #1798.

Phase 5 of #1796. Closes #1803.

Four scenarios so the new `error` state is simulated, recorded and graded like every other behaviour — plus the recorder fix that made recording them possible at all, and the measurement #1798 deferred to this ticket.

---

## Traits first (the cost argument)

`internal/matrix/capability.go` holds a closed Go set with exactly one scenario per trait. Four added:

| trait | scenario |
|---|---|
| `overload_retry` | `provider-overloaded-retry` (2.22) |
| `overload_terminal` | `provider-overloaded-terminal` (2.23) |
| `auth_refusal` | `auth-credentials-rejected` (2.24) |
| `process_crash` | `agent-process-crash-midturn` (2.25) |

None is folded into the existing `error_epilogue`, and the reason is written into the file under its own rule ("a trait may span several scenarios only while those scenarios are guaranteed to move together for every adapter"): a retry-in-progress record and a terminal epilogue are different bytes with different adapter sets, an overload is retryable where a rejected credential is not, and a process death is not transcript evidence at all — it is `TierProcess`, so an adapter that parses no error line can still pass 2.25.

**The `32 traits / 32 scenarios / other 14` figure in that file's doc comment is gone.** It was hand-typed and would have been wrong the moment these landed. `TestTraitCoverageCensus` now prints it (36 / 50 / 14) and fails on a trait naming a scenario the catalog does not have.

---

## The measurement #1798 asked for — and it inverts the risk

#1798 accepted a known risk and deferred it here: *"with no minimum hold, a 529 that recovers in ~300 ms goes red and back too fast to see."* `state_dwell.go`'s `graceFor` names this recording as the calibration.

**Measured on claude-code 2.1.243/2.1.245, four recordings through the new rig:**

| mock | POSTs | `system/subtype:"api_error"` lines | state arc | time to red |
|---|---|---|---|---|
| `--status 529 --succeed-after 1` | 2 | **0** | `ready→working→ready` | never went red |
| `--status 529 --succeed-after 3` | 4 | **0** | `ready→working→ready` | never went red |
| `--status 429 --succeed-after 3` | 5 | **0** | `ready→working→ready` | never went red |
| `--status 529 --succeed-after 0` | 4 | **0** | `ready→working→`**`error`** | **3.67 s** |
| `--status 401 --succeed-after 0` | 22 | **0** | `ready→working→`**`error`** | **184.4 s** |

**There is no flash to guard against.** In all three recovering runs the session never entered `error` at all — not because the window was brief, but because nothing is written during it. claude-code still *runs* the ladder: the 401 run's TUI pane reads `401 invalid x-api-key · Retrying in 7s · attempt 6/10`. It is rendered and not persisted.

That is a change since the fixtures this feature was designed against. The committed 2-9 recording, made on claude-code **2.1.143** against the same kind of loopback mock, carries six per-attempt lines:

```
$ jq -rc 'select(.subtype=="api_error")|{status:.error.status,attempt:.retryAttempt,retryInMs}' \
    replaydata/agents/claudecode/scenarios/2-9_token-quota-exhausted/recordings/*/transcript.jsonl
{"status":429,"attempt":1,"max":10,"retryInMs":616.45}
... through attempt 6, retryInMs 19183.90
```

Same transport, different CLI version. So `ErrorPhaseRetrying` — which only `claudecode/sessionerror.go` can produce — is unreachable in practice today.

**The risk that IS real is the opposite one.** A rejected API key leaves the session **green for 184 seconds** before it turns red. A minimum hold addresses neither direction. Reported on #1798.

---

## Per-adapter capability declarations

15 declarations, each from a source-verified fact. `process_crash` gets **none**: `diedMidTurn` is adapter-agnostic and every adapter binds a PID, so declaring any of them dead would be false.

| trait | `untraced` | why, in one line |
|---|---|---|
| `overload_retry` | **all 11** | `ErrorPhaseRetrying` is reachable from exactly one file, and the measurement above shows claude no longer writes the line that feeds it |
| `overload_terminal` | antigravity, hermes, mistral-vibe | nothing is persisted on any error path (`conversation_loop.py` builds the text into the returned dict only; `agent_loop.py:1767` appends nothing) |
| `auth_refusal` | hermes, mistral-vibe | same paths; hermes's `messages` table has no error column at all |

**Two corrections to the issue's own briefing, both evidenced.** hermes is `untraced`, **not** `daemon_capability: bug` — that verdict rests on `finish_reason='length'|'content_filter'`, an output-cap abort, and its own assessment says the opposite for provider errors. mistral-vibe is `untraced`, **not** `absent` — the *agent* has the feature (`agent_loop.py:1767-1789` raises `RateLimitError`); only the adapter is blind, and `absent` is a claim about the agent.

**Left undeclared on purpose** where the evidence conflicts or is missing — a hole is "not assessed", which is the honest state, and 12 existing trait-carrying scenarios already have them. Named rather than quietly skipped: gemini-cli/`overload_terminal` (the `mock-gemini-5xx` header's "writes NO transcript line" was verified at 0.46.0, but a source read of 0.56.0's exhaustion arm suggests it throws on the same path a 400 takes — and a 400 *does* write); codex, opencode and pi `overload_terminal` (no corpus evidence either way); kiro-cli (contested `daemon-bug` — `sidecar_metrics.go:45` shows the adapter *does* read the `.json` sidecar its 2-14 cell says it does not).

---

## The recorder gap: fixed, not deferred

`record-token-quota-exhausted.sh:10-17` has named the fix since it was written. It is now implemented:

> (1) add `bare_mode` + `env` to the per-cell … block, (2) patch … `init_session` to honor them and to auto-accept the custom-API-key dialog, (3) launch the mock binary as a pre-driver hook in `run-cell.sh`.

- **`lib/recipe-runtime.sh`** — pure predicates over the recipe: `bare_mode`, `env` (with `{{MOCK_ADDR}}`/`{{MOCK_PORT}}`), `mock`, and a `DRIVE_SUPPORTS` driver contract scraped from the driver's source the same way `recipe-lint.sh` reads `DRIVE_ELICITS`.
- **`shard-lib.sh`** — the projection now carries the three new fields. That fixed list is a gate, not a convenience: a recipe key missing from it is invisible to `run-cell.sh` no matter what the recipe says, which is exactly how these fields stayed inexpressible.
- **`run-cell.sh`** — builds and launches the mock, polls for it to LISTEN, composes its teardown into the existing EXIT trap, and refuses with a new **exit 5 `runtime_gap`**.
- **`driver-interactive.sh`** (claudecode) — `--bare`, `tmux -e`, and the custom-API-key dialog answered by *polling for it* rather than the bespoke recorders' blind `sleep 3`.
- **`mock-anthropic-error`** — one parameterised stub (`--status`/`--error-type`/`--succeed-after`/`--retry-after`/`--ignore-model`) instead of three near-identical directories.

**Every check on that path is a hard refusal, because every failure mode is silent in the same direction**: a recipe whose `env` never reaches the pane records the agent talking to the *real* provider — real credentials, real tokens, and a green fixture that proves nothing.

Two of those checks observe the subject rather than a claim about it:

- `assert_driver_applied_env` — the driver writes a receipt from the same arrays that reach `tmux`; a missing receipt fails the run. Names only, never values.
- `assert_mock_was_used` — **the mock's own request log.** Everything else is a statement someone made (the recipe says where to point, `DRIVE_SUPPORTS` says the driver can, the receipt says it did); this says whether the agent actually arrived. It counts requests rather than trusting the file to exist, because the mock writes a startup banner.

Live proof from the runs: `runtime: driver-interactive.sh applied [__bare__ ANTHROPIC_API_KEY ANTHROPIC_BASE_URL CLAUDE_CODE_LOOP_PERSISTENT]` / `runtime: the cell's mock served 22 request(s)`.

---

## Red-first evidence

Nothing here is presented as red-first unless it was seen red.

### A real defect, red before the fix existed

`of agent update --capability` was unusable for **5 of the 11 onboarded columns**. It required — and then rewrote — `replaydata/agents/<id>/metadata.json`, a different file describing a different thing from the capability model it writes:

```
$ for a in $(jq -r '.meta.min_versions|keys[]' replaydata/agents/scenarios.json); do
    [ -f "replaydata/agents/$a/metadata.json" ] || echo "$a"; done
aider
claudecode
codex
opencode
pi
```

so declaring a capability for any of them meant the hand-edit of `replaydata/` the factory exists to prevent. Seen red:

```
--- FAIL: TestAgentUpdateDeclaresForAColumnWithNoAgentMetadata
    of agent update exit=1 stderr=of agent update: no agent "claudecode" at
    .../replaydata/agents/claudecode/metadata.json (use `of agent add` for a new column)
```

The requirement is now per-flag: the authoritative column registry is `scenarios.json`'s `meta.min_versions` (what `of validate` checks the model against), and `metadata.json` is required only by the flags that write into it.

### The census guard, seen red

`TestTraitCoverageCensus`'s FK arm fired on all four traits before their scenarios existed:

```
--- FAIL: TestTraitCoverageCensus
    trait "overload_retry" names scenario "provider-overloaded-retry", which is not in the committed catalog
    ... × 4
```

### Mutations, for everything the change ADDS

`recipe-runtime.sh` is all-new checking, so a green suite proves the tests run, not that they reach anything. **`recipe-runtime-mutation_test.sh` is a committed fixture, not a PR-body paragraph** — it breaks one clause at a time and asserts the suite goes red:

```
  ok   package-confinement-dropped        → RED on "mock package outside recording tree"
  ok   port-range-dropped                 → RED on "mock privileged port"
  ok   bare_mode-string-accepted          → RED on "bare_mode \"true\" (string) is not true"
  ok   unresolved-placeholder-passed-through → RED on "unknown placeholder survives substitution"
  ok   newline-in-env-value-accepted      → RED on "env value contains a newline"
  ok   env-name-unchecked                 → RED on "env name is not a shell identifier"
  ok   driver-gaps-always-empty           → RED on "env on a driver that does NOT"
  ok   readiness-probe-always-ready       → RED on "nothing listening"
```

**The harness caught itself, and the third guard is the result.** The `bare_mode` mutation first reddened `bare_mode true` — it had broken the jq filter outright rather than loosening it — so a mutation aimed at the string/boolean distinction was being credited for breaking something else. `run_mutation` now pins *which* assertion must fail. (Its other two guards: the sed must actually change the file, and the mutant must still parse as bash.)

### Locks — pass by construction, NOT red-first proof

- `TestAgentUpdateStillRefusesDescriptiveFieldsWithNoAgentMetadata` — `--name`/`--provider`/`--prereq` must keep refusing when the file is absent. This passed before the change too; its value is that it pins the loosening as a *loosening*.
- `recipe-runtime_test.sh`'s ~45 assertions — every one passes the moment the code is written. The mutation battery above is their evidence.

---

## Recordings

Three cells recorded and promoted through the new rig, **4/4 phases each**:

| cell | arc | note |
|---|---|---|
| `2-23_provider-overloaded-terminal` | `ready→working→error` | give-up at 4 attempts; **3.67 s** to red; error is terminal and survives the failed turn's own epilogue |
| `2-24_auth-credentials-rejected` | `ready→working→error` | **184.357 s** of green first; give-up line carries `apiErrorStatus:401` |
| `2-25_agent-process-crash-midturn` | `ready→working→error` | `[agent process died mid-turn → error]` at 5.84 s — which is the recipe's own 6 s pre-kill sleep, NOT detection latency; the row is retained, not reaped |

`2-22` is **`unobservable`** — recorded, measured, and declared, with the full version comparison in its assessment body. The recipe is kept runnable so the next reader can re-measure against a newer claude-code in one command.

Goldens: three new, adapter-scoped. Census: `Recordings` 316→319, `Divergent` 150→151, `DivergentByCountsAndKinds` 149→150.

**The single divergent one is attributed, not assumed** — see the follow-up below. 2-23 and 2-24 replay *exactly* (`ordered_matches: 2`, no mismatches).

---

## `meta.ends_unsettled`: nothing added, and that is a measurement

The issue expected `provider-overloaded-terminal` and `agent-process-crash-midturn` to need it, warning that "without this every recording gets a `suspect` completeness verdict". Measured, it does not — that expectation predates the fourth state. The predicate is:

```
completeness-check.sh:186    unsettled: ($per | map(select(.last == "working") | .sid)),
```

Only a session whose **last transition is `working`** counts. All three recordings end in `error`, which is settled:

```
2-23_provider-overloaded-terminal  {"verdict":"complete","reasons":[],"sessions":{"unsettled":[]}}
2-24_auth-credentials-rejected     {"verdict":"complete","reasons":[],"sessions":{"unsettled":[]}}
2-25_agent-process-crash-midturn   {"verdict":"complete","reasons":[],"sessions":{"unsettled":[]}}
```

The list stays deliberately narrow.

---

## Two claims this PR's own text made, and the recordings falsified

Both corrected in place rather than left standing:

- 2.24 said a 401 is *"terminal and non-retryable by construction"*. It is not — claude-code climbs a ten-attempt ladder and ignores `retry-after`.
- 2.22 said claudecode *"can report a retry as still pending"*. Its parser can; the CLI no longer writes the line.

Also fixed: the scenario descriptions I first wrote named `mock-anthropic-529/main.go` and `mock-anthropic-401/main.go`, which do not exist (the tree has one parameterised mock), and two neighbouring rows still cited the pre-rename `tools/agent-onboarding/` path. A dangling-path check over the catalog now passes.

---

## Found and filed, not smuggled in

- **#1817** — the offline replay reproduces a mid-turn process death as `ready`. `replay_sidecar.go`'s `timelineProcessExit` sets `StateReady` unconditionally; the daemon has routed through `diedMidTurn` since #1800. The new 2-25 golden's own `extended_check` carries the evidence (`missing_kinds: ["working→error"]`), and it is the single `Divergent` this PR adds. Not fixed here because a faithful fix needs `diedMidTurn`'s five clauses shared out of `core/application/services` — reimplementing them inside `cmd/replay/` is the duplication the sidecar's "call the real classifier" design exists to avoid. Not currently a wrong gate: `expected-validate` reads the daemon's `events.jsonl` and the cell passes 4/4; the extended check is explicitly non-gating.
- **#1818** — claude-code's terminal error line now carries a structured `apiErrorStatus` (529 and 401 in the two new fixtures) plus an `error` token, and `terminalAPIError` discards both. #1811's "no status is derived" was true of the older shape and is now stale. Both fixtures are committed, so it needs no new recording.

## Not fixed, flagged

`.specs/agent-scenarios.md`'s `Expected:` vocabulary lists three states and is now incomplete — `error` joins them. That file is gitignored and invisible from a worktree, so it is the maintainer's to update.

## Smaller thing done on the way

`smoke-test.sh` listed its 14 `lib/*_test.sh` files by hand, and a suite whose membership is typed by hand is one a new test file silently never joins — these two did not. It now uses the repo's existing `shell_lib_suite_run`, which refuses on an empty corpus, runs every file rather than stopping at the first failure, and prints a census: `found 16, skipped 0, ran 16, failed 0`.

---

## Gates

Every group run separately, in the foreground.

| gate | result |
|---|---|
| `--only go` (core + factory + `of validate` + rig smoke + replay fixtures) | **PASS** (re-run after the rebase onto #1812) |
| `--only arch` | **PASS** — composite 7.8920 → 7.8930, no regression |
| `--only web` | **PASS** (untouched; #1801 owns that tree) |
| `--only tools` / `--only skills` | **PASS** |
| `--only posix` / `--only bash` | **PASS** |
| `--only security` | **PASS** |
| `--only swift` | **PASS** |

Pushed with `--no-verify`; every gate the hook runs had already been run unscoped in the foreground.

**Does not touch `platforms/web/` or `core/cmd/irrlichd/`** — #1801/#1812's tree. Rebased onto #1812 after it landed; no file overlap.


ingo-eichhorst:
## Review round (high effort) — 16 findings, all actioned

The three the reviewer ranked worst were **one bug wearing three masks**: a check that cannot tell *"nothing is wrong"* from *"I could not look"*. All three were in verification machinery this PR itself adds.

### F1 — `assert_mock_was_used` could report the exact opposite of the truth (blocker)

The guard proved the agent reached the mock by counting `POST /v1/` in the mock's log. That string is **one mock's log format**. `recipe_runtime_mock_check` admits any `recording/mock-*` package, and `mock-gemini-5xx` logs `router %s` / `main-turn (abort) %s` — nothing resembling it. A gemini cell whose mock served *every* request would have counted zero, exited 5, and told the operator the recording was made against the **real provider**. Latent only because no gemini cell uses the block yet.

Fixed by making `.mock.request_log_pattern` a **required** recipe field — refusal, not a default, because a default is wrong for every mock but one, which is how this started. The guard now separates three outcomes with three messages: log unreadable → *"this check could not run"*; zero matches → *"either the agent did not talk to it, or your pattern does not match what this mock logs"*; matched → the count.

Seen red, including the finding itself as a committed mutation:

```
ok   mock-pattern-hardcoded → RED on "a NON-claudecode mock log is counted by its own pattern"
ok   request-log-pattern-not-required → RED on "mock without request_log_pattern"
```

### F9 — the census test disarmed its own gate

`TestTraitCoverageCensus` called `t.Skip` on an unreadable catalog, while its doc said it exists "so a Go-only change is caught by `go test`". Anything making the load return empty switched the gate off with the package still reporting `ok`. Now `t.Fatalf`. The neighbouring `TestShardCellEquivalence` legitimately skips — it asserts a property *of the corpus*, so no corpus means nothing to assert; this one asserts a property of the Go trait table, which exists either way. That distinction is written into the code.

### F3 — the mutation harness could observe someone else's socket

Documented at ~2 s, **measured at 60 s**; forked with no `timeout`; all eight copies sharing hardcoded port 18991. The reviewer reproduced the real hazard: with a stray listener on that port, the suite's own bind loses and the check still reports `ok`.

Now `timeout 30` per mutation (with the macOS no-`timeout(1)` fallback stated rather than silent), a per-mutation port, and the suite **proves the listener is its own** before trusting the probe. Two further bugs surfaced while fixing it: the bind-verification probe was consuming the one-shot `nc -l` (now `nc -k`), and the runtime figure in the header is corrected to ~60 s with the reason.

### F2 — `run-cell-multi.sh` had no refusal at all

A cell carrying `bare_mode`/`env`/`mock` driven through the cross-adapter rig would get no mock, no `--bare`, and **no refusal**. Mirrored: `recipe_runtime_unsupported` + exit 5. Deliberately a refusal rather than an implementation — every mock here overrides *one* provider's base URL and a cross-adapter cell has two providers. Refusing rather than assuming it unreachable, because these two scripts have diverged before (#1214).

### The guards had no committed fixture — now they do

The reviewer's sharpest structural point: `recipe-runtime.sh` was mutation-covered, but the two checks that decide whether a recording is *trustworthy* lived inside `run-cell.sh` — a pipeline script nothing can source — and were exercised by three live recordings and nothing else. Both moved into the lib, and the harness now reddens **15** clauses:

```
package confinement · port type · port range · request_log_pattern required
bare_mode true-vs-"true" · unresolved placeholder · newline in env value
env name shape · driver gaps · readiness probe · env receipt required
env receipt compared · mock log readable · mock zero hits · mock pattern
```

**Two of those five new mutations first came back GREEN**, and that is the useful part. The guards' clauses mask each other: disable "no receipt written" and the *next* clause still refuses, with a worse message. Rather than delete the clause (#1800's move for a genuinely redundant guard), the clause here earns its place through the **message** an operator reads — so the assertions now pin the reason, and the mutations go red on it.

### F4–F8 — my own comments, contradicted by my own measurements

- **F4**: the 2.24 cell said "a 401 is non-retryable, so there is no ladder to swallow it" — while the same PR's `events.jsonl` shows a ten-attempt ladder over 184 s.
- **F5**: `--retry-after 1`'s rationale claimed it avoided a recording "several minutes long", but 2.23's arc is four attempts, and 2.24 measured that claude **ignores the header**. **The flag is removed and the cell re-recorded** — leaving an inert flag on the one cell reporting a dwell as claude behaviour was the wrong resolution. Dwell moves 2.168 s → **3.67 s**; same four attempts either way, which is what shows the flag was inert.
- **F6**: cited `isRetryableError` and `429/499/5xx` from `mock-gemini-5xx`'s header. It says `retryWithBackoff` and `429/500/503`; neither `isRetryableError` nor `499` exists anywhere in the repo.
- **F7**: a comment hand-typed "36 traits / 50 scenarios / 14 untraited" *in the same sentence* as "the figure is printed by that test rather than typed here a second time". The numbers are gone; only the command remains.
- **F8**: "exactly ONE adapter writes the first" asserted present-tense while the same commit declares `overload_retry: untraced` for all eleven.

### F12, F13, F16, F14, F15

- **F12**: `of agent update` with no flags printed `ok` and returned 0 having written nothing — the same no-op success its own comment calls "the worse outcome". Now refuses.
- **F13**: `"port": "0900"` was accepted — bash reads it as an invalid octal literal, the arithmetic *errors* rather than evaluating false, and the guard's `[[ ! … ]] || (( … ))` shape made the whole condition false. Defended twice: a JSON-number type check, and `10#` on the arithmetic. The `10#` half is unreachable while the type check stands and is **labelled as defence in depth, not claimed as covered**.
- **F16**: eleven assertions pass by construction. Rather than manufacture mutations, they are listed by name in the test header as **locks**, with the reason each shares a clause with a mutation-backed one.
- **F14**: a stale duplicate header item in `smoke-test.sh` describing the advisory shellcheck tier #1684 deleted. **F15**: a stray `schema_version` on a phase line no other `expected.jsonl` in the corpus carries.

### Sonar / CodeScene (advisory, both were failing)

- `gosecurity:S5145` ×3 — the mock logged the request body's `model` verbatim. Genuine log injection, and pointed: a newline could forge a counterfeit `POST /v1/messages #N` line, which is precisely what `assert_mock_was_used` counts. Sanitised.
- `go:S3776` / Bumpy Road on `runAgentUpdate` and Complex Method on the mock's `main` — both split.
- Shell rules: missing `case` default, positional params to locals, explicit returns, and the duplicated port literal (now one `mock_json` fixture builder).

### Gates, all re-run after the fixes

`go` · `arch` (composite unchanged, 7.8920) · `web` · `tools` · `skills` · `posix` · `bash` · `security` · `swift` — **all PASS**, each as its own foreground `--only` invocation.

One honest note: the `go` group's **core module tests** came back FAIL once, then PASS on re-run and PASS twice standalone under `-race`, with no stray rig processes or held ports. I could not reproduce it and did not capture the failing output, so I am recording it as an **unreproduced flake** rather than dismissing it.

Issue #1796, opened 2026-08-24
Epic: fourth session state `error` — detect and surface failure conditions


ingo-eichhorst:
Add a fourth session state, `error`, so a session whose provider call fails, whose credentials are rejected, whose process dies mid-turn, or whose Irrlicht-side read fails is visibly red instead of silently green.

Menu bar: a red circle. List view: a red error line under the session.

## Decisions (settled — do not relitigate)

| Decision | Choice | Rationale |
|---|---|---|
| Modeling | A real **fourth domain state** `error`, not an orthogonal overlay | Explicit product call. The menu-bar dot is painted straight from `session.state.hexColor` (`platforms/macos/Irrlicht/Views/MenuBarStatusRenderer.swift:162`), so a state-valued error makes the red dot fall out for free. Cost is accepted: ~20 Go files, both UIs, docs, and the PR gate. |
| Error sources | Provider API errors; auth/quota refusals; agent process death mid-turn; **and** Irrlicht's own internal failures | All four turn a session red. See the daemon-wide row below for the half that has no session to attach to. |
| Clearing | The **next successful turn** clears the error. No minimum hold. | Matches how the existing context-pressure alert behaves. Accepted risk recorded in #1798. |
| Factory scenarios | `provider-overloaded-retry`, `provider-overloaded-terminal`, `auth-credentials-rejected`, `agent-process-crash-midturn` | Two of the four are already written down and unclaimed in `.specs/agent-scenarios-candidates.md:65` and `:77`. |
| **Release ordering** | **No intermediate release.** #1797 lands first in merge order, then the epic runs straight through. | Users normally only roll forward. If someone rolls back to a daemon that does not know `error` and it deletes those session files, that is an **accepted loss**. See risk 1 below. |
| **Menu-bar precedence** | **`error > waiting > working > ready`** | An error is the thing you most need to be told about. Accepted cost: while any session is errored, the orange "a session is asking you a question" hint is hidden at a glance. Implemented in `dominant(in:)` — see #1802. |
| **Concurrency** | An errored session is **not** active | `concurrencyActive()` today means working-or-waiting; error is treated like `ready`. Nothing clears an error until the next successful turn, so counting it would inflate the active count forever. Implemented in #1801. |
[...]

PR #1835, opened 2026-08-25
fix(replay): reconstruct a mid-turn process death instead of replaying it green

### LINKED ISSUE #1817: replay: the offline sidecar replays a mid-turn process death as `ready`, not `error`

ingo-eichhorst:
Closes #1817.

## What was wrong

The offline sidecar replayed **every** process exit as `ready`, so the committed
corpus carried a fixture asserting that a crashed agent comes back green — the
exact defect the #1796 arc exists to remove.

## The cause was not the one the issue named

The issue pointed at `applyTimelineControlEntry`'s `case timelineProcessExit:`.
That branch *does* hard-code ready — but it is **unreachable for a mid-turn
death**. `PIDManager.HandleProcessExit` records `KindProcessExited` only on the
branch that DELETES the row, and #1800's retain path returns before it
(`pid_manager.go:495-502`). Measured on the fixture itself:

```
$ jq -r .kind replaydata/agents/claudecode/scenarios/2-25_agent-process-crash-midturn/\
recordings/2026-08-25-09-02-30_irrlichd-0.6.0+6c65754/events.jsonl | sort -u
debounce_coalesced / pid_discovered / presession_created / state_transition /
transcript_activity / transcript_new / transcript_removed
```

No `process_exited`. The recording's **only** trace of the death is the
`state_transition` it caused (seq 312) — the classifier's *output*, and exactly
what the extended check grades the replay against, so it cannot be fed back in
as an input without making the check circular.

## The stated fix was measured and rejected

I implemented it anyway to get a number. Routing `case timelineProcessExit:`
through the predicate:

- left the 2-25 golden **byte-identical** (nothing to hook onto), and
- made **16 other goldens** fabricate a `working→error` their own daemon never
  produced, because all 317 `process_exited` events in the corpus were written
  by the teardown path. Two of them are `2-20_user-esc-interrupt` /
  `2-20_interrupted-turn` — the scenarios `DiedMidTurn`'s clause 5 exists to
  protect. Census `Divergent` went **151 → 160**.

So the root cause is a **recording** gap, not a sidecar gap.

## What this PR does

**Shares the predicate, does not copy it** (the issue's delegated decision):

- `session.SessionState.DiedMidTurn` — the single copy of the five fail-closed
  clauses, on the domain value it interrogates. `services`' private
  `diedMidTurn` is deleted; `retainAsProcessDeath` calls the domain predicate.
  Domain was chosen over exporting from `services` because the predicate reads
  only domain fields (no clock, logger, repo or config), and it keeps `tools/`
  off the application layer for a pure rule.

**Closes the recording gap:**

- `lifecycle.KindProcessDiedMidTurn` records the *retained* edge. A separate
  Kind, not a flag on `KindProcessExited`, because that Kind means teardown to
  every consumer — overloading it is what the 16 fabricated goldens measured.
- Written on `retainAsProcessDeath`'s **one-time conversion edge**, not in
  `HandleProcessExit`: the sweep calls that every few seconds for as long as the
  row survives, so recording there would emit one event per tick for up to
  `processDeathRetention` (12h). The verdict registry is what bounds the edge.

**Teaches the replay to reconstruct it:**

- The sidecar buckets the new Kind, consults the shared predicate, places the
  same `SignalProcessDeath` hold and re-runs the real `services.ClassifyState`.
  Nothing in the replay names a target state.
- The primary session's own `parent_linked` is now plumbed through, so
  `DiedMidTurn`'s subagent clause is *read from the recording* rather than
  assumed empty.
- `internal/replay/state_machine.go` and `ghosts.go` state the non-teardown case
  explicitly — adding this Kind to a deletion list is now a duplicate-case
  compile error.

## Evidence

**Red-first** (`ir:exec` step 11a). Disabling the sidecar's new branch:

```
--- FAIL: TestSidecarReplaysAMidTurnProcessDeathAsError
    replay produced no working→error for a mid-turn process death;
    transitions were [init:→ready event:ready→working]
```

— byte for byte the shape the committed 2-25 golden carries today.

**Mutation, proving one copy.** Deleting `DiedMidTurn`'s subagent clause in
`core/domain/session` turns the **replay-side** lock red:

```
--- FAIL: TestSidecarKeepsAProcessDeathFailClosedForASubagent
    a subagent's process death still produced error … — DiedMidTurn's
    ParentSessionID clause is not being consulted
```

Deleting that **one** clause reddens all three at once — the replay lock, the
daemon suite (`TestHandleProcessExit_CleanShapesAreStillDeleted/child-session`)
and the domain test — which is the strongest form of the claim: a single edit in
`core/domain/session` propagates to every consumer, so the single-copy claim is
executable rather than structural. (Deleting the *state* clause instead reddens
the daemon suite's `{ready,waiting}` rows and the domain test.)

That lock was rewritten after it failed to discriminate: an earlier cut dated
the death before the first transcript read, which fails closed on the
nil-metrics guard long before any clause runs, so it stayed green with the
predicate gutted. A check that cannot fail is not a check (AGENTS.md).

**Zero corpus movement.** `UPDATE_REPLAY_GOLDENS=1 go test
./tools/onboarding-factory/cmd/replay/... -count=1` → 0 goldens changed. Census
unchanged at `Recordings: 319, Divergent: 151, DivergentByCountsAndKinds: 150`
(machine-verified by `TestCatalogCensusMatchesTheCommittedFigures`, not typed).

**Gates.** `tools/preflight.sh --only {go,arch,tools,skills,posix,bash,security}`
— all PASS. ARS composite 7.8885 → 7.8881, no category regression.

## What this does NOT fix, stated plainly

**The committed 2-25 golden still reports `missing_kinds: ["working→error"]`,
and this PR cannot clear it.** Its `events.jsonl` is frozen and was written by a
daemon that recorded no process event at all; only a **re-record** of that cell
on a daemon carrying this change will clear it. The new test proves that is the
only thing missing — it rebuilds that exact recording with the one event a
post-fix daemon now writes, and the replay then reproduces the daemon exactly
(`missing_kinds` and `ordered_mismatches` both empty).

Also corrected: `pid_manager.go`'s doc claimed the `error` transition "carries
the pid and the exit reason" in place of the suppressed event. It carries
neither — seq 312 has no `pid` field, and its `Reason` is the classifier's own
`"agent process died mid-turn → error"`.


ingo-eichhorst:
## Review round (subagent, effort `high`) — 13 findings, all addressed

Two were substantive; the rest were coverage and doc accuracy.

**1 · Concurrency bug (found by review, not by me).** `retainAsProcessDeath` read
the verdict via `processDeathVerdictAt` and wrote it via `markProcessDeathVerdict`
under **two separate acquisitions** of `processDeathMu`. The two exit paths — the
kqueue/pidfd callback (`cmd/irrlichd/startup.go:362`) and the 5s liveness sweep
(`reapDeadOrInfraPID`) — race, so both could miss the verdict and both convert.
Reviewer measured **13 double-conversions in 300 runs** with two callers under
`-race`. With the new recorded event that means a recording claiming the process
died twice.

Fixed by making `markProcessDeathVerdict` a test-and-set that reports whether
*it* set the verdict; the `d.record` is gated on that return, not on the earlier
read. New `TestProcessDeathIsRecordedOnceUnderConcurrentExitPaths` fails **3/3**
against the non-atomic version:

```
recorded 2 process_died_midturn events from 8 concurrent exit paths, want exactly 1
recorded 3 …
recorded 3 …
```

**2 · The producer had zero coverage.** Every replay-side test splices the event
in by hand, so deleting the `d.record` call was green in *both* modules — the
one line that closes the actual root cause was unpinned.
`TestProcessDeathIsRecordedForOfflineReplay` now asserts the fields the replay
consumes; deleting the producer gives
`recorded 0 process_died_midturn events, want 1`.

**Also fixed**

| # | Finding | Fix |
|---|---|---|
| 2 | `buildDeathFixture` would go red on the 2-25 re-record (splices unconditionally; `fixturePath` resolves to the *newest* recording) | skips with an explanatory message when the sidecar already carries the Kind |
| 4 | `Timestamp: d.nowFn()` broke `record()`'s wall-clock convention — a driven clock could invert the death against the transition it must precede, which `buildTimeline` sorts timestamp-primary | field dropped; `record()` stamps it like every other event |
| 5 | `causeProcessExit` was named/documented for a Kind that can never produce it — reading as licence to re-route the teardown branch | renamed `causeProcessDeath` (`"process_death"`); no golden carries it |
| 6 | dead `bool` return promising a teardown contract nobody honours | returns nothing |
| 7 | double-`Overlay` rationale was **wrong** — applies are idempotent; the real hazard is `Overlay` *deleting* consume-once holds, so overlaying a throwaway probe could lose one | one metrics value, overlaid twice; rationale corrected |
| 8 | three mechanics mutated green | `alive` gate **deleted** — provably unreachable (a teardown already sets state to ready, which clause 1 rejects), so it was a branch no test could turn red. Debounce flush kept, documented as unpinned |
| 9 | the nil guard the move *added* was the one clause nothing covered | `core/domain/session/process_death_test.go` — deleting the guard now panics the test. Also gives the newly-exported domain API direct tests in its own package |
| 10 | onboarding viewer rendered the crash recording's key event as a 7px grey dot labelled `process_died_midturn` | red/14px with plain-English text, matching the other fault kinds |
| 11 | public kind table missing the Kind; `process_exited` row now wrong | both rows fixed |
| 12 | "five fail-closed clauses" is six | reworded |
| 13 | "317 process_exited events" stated no command | command inlined (the scope is load-bearing: `grep -rh` over the tree gives 410) |

CodeScene's advisory Complex Method flags were addressed by extracting
`applyProcessTeardown` / `applyProcessDeathEntry` out of the switch and
`findCrashVerdict` out of the test helper.

**Re-verified after the fixes:** goldens **0 changed**, census unchanged
`319/151/150`, `of validate` OK, state-vocabulary lint OK,
`preflight.sh --only {go,web,arch,tools,bash}` all PASS.

**Explicitly refuted by the reviewer** (evidence in each case, so these are not
unlooked-at): debounce left live cannot produce a spurious later transition
(replayed at death−500ms/−1.5s/−3s, identical output); duplicate death events do
not corrupt the replay (idempotent — state is already `error`); no seq/ts
ordering hazard between the death and its transition; both duplicate-case
tripwires actually produce compile errors; no other kind-enumerating consumer
needs updating (`internal/validate`, `concurrency_tracker`, `completeness-check.sh`,
`internal/matrix`, `cmd/of`).

**Still not fixed, by design:** the committed 2-25 golden keeps
`missing_kinds: ["working→error"]` until that cell is **re-recorded** on a daemon
carrying this change. Its `events.jsonl` is frozen and carries no process event
at all.

ingo-eichhorst:
## `/simplify` round (4 agents: reuse, simplification, efficiency, altitude)

**Efficiency came back essentially clean** — measured, not asserted: the new
branch is cold (0 `process_died_midturn` events in the corpus today), `d.record`
does a buffered `json.Encode` with no lock held, the sweep's steady-state path is
byte-for-byte unchanged, and the new tests add 0.00s.

### Applied

**Deepest finding — the hold's RELEASE edge was not mirrored.** `SignalProcessDeath`
is the one policy with **no ceiling**, safe in the daemon only because two things
bound it: the row's own reaping, and `HandlePIDAssigned` releasing the hold when a
process comes back. The sidecar mirrored neither. A crash-then-resume recording
would have replayed pinned at `error` from the resume onwards. The sidecar now
buckets `pid_discovered` and releases the hold, matching `HandlePIDAssigned`.
(380 committed recordings carry `pid_discovered`, so this was reachable, not
theoretical — goldens re-verified at **0 changed** after adding it.)

**The predicate was shared; the payload was not.** The `SessionError` was
duplicated character-for-character across the module boundary — em dash and all —
eight lines below a doc block arguing that copies of lifecycle rules drift
unnoticed. Now `session.NewProcessDeathError(pid, reason)` next to `DiedMidTurn`;
both raisers construct from it.

| Finding | Fix |
|---|---|
| `recordingRecorder` duplicated `mockRecorder` in the package's own `testhelpers_test.go` — already mutex-guarded, ~20 lines | deleted; uses `mockRecorder` |
| `state_machine.go`'s new case body was byte-identical to `default:` | `fallthrough` — keeps the duplicate-case compile guard, drops the copy that would drift if `default:` gains a step |
| `ghosts.go`'s `g.substantive = true` was redundant (the session's own `working` transition already set it) | empty case body; only the comment and the compile guard remain |
| parent-link capture sat in the caller loop, splitting Kind routing across two places | moved into `bucketPrimaryEventByKind` |
| `buildDeathFixture`'s `at func(...)` parameter — both call sites passed identity | dropped |
| hardcoded `"claude-code"` in a package that has `claudecode.AdapterName` in 11 places | constant |
| **`event.go` claimed "every consumer deletes the session on it" — false for 3 of 6 sites** | all six enumerated with what each actually does; notes only 3 are compile-guardable |
| the "16 goldens" figure was restated in 6 places in 6 wordings | one home in `event.go`; the rest cite it |
| stale `diedMidTurn` prose in 4 `replaydata/` files — including the 2-25 cell this PR is about | updated; 2-25's `expected.jsonl` now records why its golden still diverges and that a re-record clears it |
| stale counts: "the other two call sites", "the four timeline streams" | corrected |
| over-long comments on the cause const, `applyProcessTeardown`, `pid_manager`'s comment-about-a-comment | trimmed |

Also added, from the review round's rationale: a `FIELDS READ` list on
`DiedMidTurn`, because one consumer assembles a `SessionState` purely to ask —
a future clause reading a fourth field would silently get its zero value there.

### Skipped, with reasons

- **`replaySettingsForTest` reuse** — tried it; it derives the adapter via
  `detectAdapter`, which infers from the replaydata path shape and cannot resolve
  a `t.TempDir()` copy. Reverted to explicit settings with the constant, and said
  so in a comment.
- **`lifecycle.IsTeardownKind(k)`** — agreed as the deeper fix, and the doc now
  enumerates all six sites so the drift is at least visible. It reaches beyond
  this diff (a JS file and a shell script that compare raw strings and no Go
  predicate can guard). Worth its own issue; the altitude agent's suggestion of a
  sibling to `state-vocabulary-lint.sh` over `lifecycle.Kind` sets is the right
  shape.
- **Renaming `DiedMidTurn` → `WasMidTurn`** — the receiver does not know a death
  occurred, which is a fair reading, but `DiedMidTurn` is the established
  vocabulary across #1800's prose, the recorded reason string and the scenario
  name. Churn without a clear win.
- **Renaming the Kind to `process_exit_retained`** — same call: `process_died_midturn`
  matches the daemon's own reason string and the cell name `2-25_agent-process-crash-midturn`.
- **Marking 2-25 `known_failing`** — the cell validates 4/4 phases; the divergence
  is in the informational extended check only, so the label would misreport it.
  Recorded in the cell's `expected.jsonl` instead.
- **Merging the two clause tables** (domain vs services) — they test different
  things: the domain table tests the predicate, the services table tests that a
  false verdict actually reaches the delete path.

**Re-verified:** goldens **0 changed**, census `319/151/150`, `of validate` OK,
state-vocabulary lint OK, and `preflight.sh --only {go,web,arch,tools,bash,skills,posix,security}`
— **all 8 PASS**.

ingo-eichhorst:
## Independent `/ir:code-review high` round — all 4 findings fixed

**Finding 1 (MEDIUM) — the race gate could not fail on the repo's actual gate.** Confirmed and fixed by changing what is asserted. The old test watched the race's *symptom* (two recorded events), which needs the scheduler to interleave inside a narrow window — red in 2/300 `-race` runs, green at `-count=1`. The new `TestMarkProcessDeathVerdictAdmitsExactlyOneCaller` (`process_death_verdict_internal_test.go`, `package services` so it reaches the unexported method without widening the production API) asserts the **contract** — "at most one of 64 callers is told it set the verdict" — which the mutation violates *by construction*, not by timing. Removing the test-and-set, at `-count=1`, five consecutive runs:

```
run 1 @-count=1: markProcessDeathVerdict returned true to 64 of 64 concurrent callers, want exactly 1 —
                 the check-and-set is not atomic, so more than one caller believes it converted the
                 session and the recording claims the process died more than once
run 2 @-count=1: … returned true to 64 of 64 …
run 3 @-count=1: … returned true to 64 of 64 …
run 4 @-count=1: … returned true to 64 of 64 …
run 5 @-count=1: … returned true to 64 of 64 …
```

Also added `TestMarkProcessDeathVerdictIsTerminalUntilForgotten` — the sweep calls the exit path every few seconds, so a verdict that decayed on its own would emit a second death event *sequentially*, with no race at all. The false comment ("8 callers … the margin to fail reliably in CI") is corrected in place rather than deleted, and the end-to-end burst now states explicitly that it is **not** the regression gate and points at the one that is.

*(Caught while fixing this: my first cut used a bare `&SessionDetector{}`, which the package's own `TestSessionDetectorIsNeverBuiltByBareLiteral` guard rejected. Now uses `newSessionDetector()`.)*

**Finding 2 (LOW) — not refuted; it was real, and worse than latent.** I read the daemon path: `retainAsProcessDeath` calls `signals.Release(SignalProcessDeath)` at `session_detector_lifecycle.go:390`, immediately before returning false so `HandleProcessExit` reaps the row and writes `process_exited`. So death-then-teardown is a shape a recording genuinely contains. `applyProcessTeardown` now releases the hold too. **Measured, not reasoned**: `TestSidecarReleasesTheDeathHoldOnTeardown` builds death → `process_exited` → a fresh lifetime; with the Release removed it fails —

```
a lifetime that began AFTER the teardown was classified error at 2026-08-25 21:03:24 (cause "event")
— the process-death hold outlived the teardown that reaped it, and SignalProcessDeath has no ceiling to drop it
```

A brand-new session pinned to `error`. Not merely a stale hold.

**Finding 3 (LOW)** — `playbackTimeline.test.js` now asserts the full style object for `process_died_midturn` and that it differs from `process_exited` in colour and label. Deleting the case fails it: `expected { color: '#94a3b8', size: 7, … } to deeply equal { color: '#ef4444', size: 14, … }`, `+ "label": "process_died_midturn"`. `npm test` (via `npm ci`, never `npm install`): **5 files, 86 tests passed**.

**Finding 4 (LOW)** — PR body corrected to the stronger true claim, which I verified rather than taking on faith: deleting **only** the `ParentSessionID` clause reddens all three at once — `TestDiedMidTurn_ClausesInThisPackage/child-session`, `TestHandleProcessExit_CleanShapesAreStillDeleted/child-session`, and `TestSidecarKeepsAProcessDeathFailClosedForASubagent`.

**Follow-up filed: #1836** — re-record claudecode `2-25_agent-process-crash-midturn` on a daemon carrying `KindProcessDiedMidTurn`, with the frozen-recording evidence, the note that producer and consumer both already exist and are tested, and acceptance criteria.

**Verification:** `preflight.sh --only go` / `--only tools` / `--only web` — all PASS, foreground. Goldens **0 changed**, census still `319/151/150`.

Issue #540, opened 2026-05-30
feat(relay) v1·F: macOS multi-source polish — offline rows fade, don't delete


ingo-eichhorst:
Parent epic: #532 (relay server v1). Work item **F — macOS multi-source polish** (nice-to-have).

## Scope
- [ ] **Offline rows fade, don't delete** (wiki §6 Phase 6 / §11). v0 deletes a disconnected daemon's rows (`core/cmd/irrlichtrelay/hub.go:244` fans out `PushTypeDeleted`; macOS removes from `relaySessionMap`/`relayDaemons` at `SessionManager.swift:490,521`); the wiki outcome is **fading** them so a flapping link doesn't yank rows.

## Acceptance
- A disconnected relay daemon's rows fade (gray/dim) instead of disappearing; they restore on reconnect.
- macOS tests green.

## Key files
- `platforms/macos/Irrlicht/Managers/SessionManager.swift`
- `platforms/macos/Irrlicht/Views/SessionListView.swift`
- possibly `core/cmd/irrlichtrelay/hub.go` (disconnect semantics)

ingo-eichhorst:
Done in #548 (squash-merged to main).

PR #1847, opened 2026-08-27
feat(relay): phone notifications via a relay-side push engine and a Web Push PWA

MBoegers:
## Summary

Implements phone notifications from #1346. I opened that discussion arguing we should build path A first; this is path B, and the inversion is deliberate. Once I started, A and B stopped being alternatives: the hard part is not the push transport, it is deciding when a state change deserves a buzz, and that policy is the same code whether the destination is APNs, ntfy or a webhook. A becomes a second exporter on top of this rather than a different project.

Pair a phone once with a short-lived code, and a lock-screen banner arrives when an agent needs input or finishes. Off unless a relay is configured and a phone is deliberately paired.

Building it also corrected two things in the original proposal, both recorded in `docs/mobile-notifications-arc42.md`. The metadata claim was too generous: Web Push requires four plaintext headers I had described as absent, and `TTL` plus `Urgency` differ per event kind, so a push service can tell which kind of event fired rather than only that one did. And the opaque-payload design was theatre, because a relay already holds full session state for every relay user today, so letting it compose the payload changes nothing about what it knows while removing an entire key-distribution problem.

This is meant to land and be refined across releases rather than sit as a prototype. What stands between it and that is the device test below.

## Changes

- `core/domain/notify`: the policy, as a pure package. 7s hold-down on `ready` that a follow-up prompt cancels, per-session cooldowns, burst collapse into one "N agents need attention", subagents never notifying, and a watchdog for a daemon going quiet. Answers question 4 of #1346.
- `core/pkg/webpush`: VAPID signing (RFC 8292), payload encryption (RFC 8291), TTL/Topic/Urgency headers (RFC 8030). Standard library only, no new module dependency.
- `core/cmd/irrlichtrelay`: VAPID identity, pairing endpoints, subscription registry, transition observer and dispatcher, a persisted daemon roster for the watchdog, and a diagnostic endpoint that answers with the push service's own verdict.
- `platforms/web`: the PWA. Manifest, service worker, pairing and health UI, last-known-state ledger, home-screen badge, and a deep link so tapping a banner opens the session it was about. Answers question 5.
- Distribution: `irrlichtrelay` tarballs for linux/amd64 and linux/arm64, an operator runbook, and a user guide.
- `tools/elfdans-rig.sh` and `tools/elfdans-drive`: an isolated relay and a session-state driver, so the device test is a checklist rather than an afternoon.

## Architecture

The part worth reviewing is what this does not touch.

`irrlichd` is unchanged, and not merely in the "mostly" sense. The daemon binary does not link the policy engine, the Web Push sender, or any relay push code, and neither does anything under `adapters/` or `application/`. The daemon-to-relay wire protocol is untouched, so an old daemon works against a push-capable relay and an unchanged daemon works against an old relay.

The two packages carrying the mechanism are both pure. `notify` decides when a transition deserves a notification; `webpush` encrypts, signs and posts. Neither imports a single Irrlicht package. The relay's own push package pulls in only `webpush`. Those three form a closed set depending on nothing else in the repo, which was a goal rather than a happy accident: if this ever wants its own repository, extraction is a move rather than a rewrite.

The one genuinely shared surface is `platforms/web`, served by the local daemon and by relays alike, so additivity there is enforced instead of intended. The service worker registers no `fetch` handler, so it can never interpose on asset loading or serve a stale dashboard. It is registered lazily from the pairing flow, so ordinary dashboard use never installs it. The pairing UI is feature-detected and appears only where the origin answers the push-info endpoint. One exception is stated rather than glossed: `index.html` links the manifest unconditionally, so a plain dashboard is installable under the app's name, although an install without pairing subscribes to nothing.

The web tree now draws that line too. Elfdans's dashboard half, the notification deep link and the live ledger fold, sat inside `irrlicht.js` reading its session index and group tree directly, which made the two inseparable. It has its own module now. The coupling is a host adapter naming the five things Elfdans may know about the dashboard and five calls back the other way, so `irrlicht.js` holds no Elfdans state and takes no Elfdans branch. Dropping the feature, or building it as a second artifact, is deleting those calls.

Consent follows the model already in the codebase. A relay running `--auth off` refuses push outright and mints no key material at all, because an anonymous relay handing out pairing codes is a stranger's phone on your sessions.

## Test plan

- [x] Full suite passes locally: `go test ./core/... -race`, both web trees, the architecture import-direction test, replay fixtures, `of validate`, POSIX installer lint.
- [x] Added or updated tests covering the change.
- [ ] Manually exercised the relevant flow. **This is the gap, and it is the honest state of the PR.**

`tools/elfdans-rig.sh check` runs ten assertions against a real relay with no phone involved: the anonymous-mode refusal and that it mints no key, `0600` file modes, mint to redeem to subscribe, a spent code answering a uniform 401, the health endpoint not echoing the push endpoint path, a malformed subscription refused, VAPID identity and subscriptions surviving a restart, and revocation pruning the delivery address. It passes. The ARS gate improves the composite from 7.932 to 7.982 with no category regressions.

Every guard landed with a deliberate mutation seen red. Two are worth naming. Adversarial review of the dispatcher found seven of eight fresh mutations staying green, because the tests drove the component's synchronous core and never the production plumbing; the fix boots the real run loop and lets a real timer fire. And a cross-language tripwire now pins the diagnostic push's collapse key, because the relay constant and the service worker tag must be the same string and nothing had ever checked them.

CodeScene and Sonar both went red on this branch. Sonar's gate failed on two of its 82 findings, the two typed as vulnerabilities. `tools/elfdans-drive` shipped without a `go.sum`, which only went unnoticed because the root `go.work` covers it: outside the workspace it did not build at all. And `sw.js`'s message handler accepted any origin. That one refuses nothing a browser can send, since a service worker is reachable only from its own origin's clients, but the handler writes to the ledger, and that unreachability is a property of the platform rather than of the file. It is guarded now, with the test written first and seen red. CodeScene's hotspot gate named `irrlicht.js`, which the separation above answers. The findings I left alone are listed with reasons in the commit that fixes the others.

**What is not verified: no notification has ever reached a real phone.** Every defect so far was found by reading code, and two of them, a VAPID JWT missing its `sub` claim and a `subscribe()` call before the service worker activates, fail only on contact with a real push service. `docs/elfdans-device-test.md` is the plan; its phone-free half is automated and green, the half that needs a device has not run. Until it does, the quality requirements in section 10 of the design doc are designed for rather than demonstrated.

Four follow-ups are named in section 11 rather than hidden. The one I would not merge without discussing: a device token is still a full client token, so a paired phone can mint pairing codes for its own workspace. Contained by the workspace boundary and by revocation, but narrowing it is owed.

## Related issues

Refs #1346

## Checklist

- [x] Follows the conventions in AGENTS.md and CONTRIBUTING.md
- [x] Commit messages use Conventional Commits
- [x] Documentation updated: `docs/mobile-notifications-arc42.md`, `docs/relay-protocol.md`, `examples/relay/DEPLOY.md`, `site/docs/elfdans.html`


ingo-eichhorst:
@MBoegers: Thanks. That is amazing work. I'll have a deeper look asap.

ingo-eichhorst:
## First — thank you, @MBoegers. This is outstanding work. 🙏

122 files, ~19.5k lines, a Web Push implementation from the RFCs up, a PWA, an operator runbook, and a device-test rig — and the daemon genuinely untouched. I checked that last claim rather than taking it: `go list -deps ./core/cmd/irrlichd` names neither `notify` nor `webpush` nor the relay push package, and `go list -deps` on `notify` and `webpush` returns only themselves. The closed set you describe is real.

A few things stood out as better than the bar this repo usually asks for:

- **RFC 8291 Appendix A is transcribed as a byte-for-byte known-answer test**, and the test-package decryptor gets its own known answer. Two implementations that only validate each other can share a bug; these do not.
- **The mutation evidence is committed, not described.** `sw-contract.test.js` and `release-files.test.js` both drive their guards against deliberately-wrong fixtures *and* assert the mutation still applies, so a stale fixture fails instead of quietly passing. That is exactly the shape AGENTS.md asks for and rarely gets.
- **The PR description names what is not verified**, including two defects that only fail on contact with a real push service. That honesty made this review much faster.
- The `WEB_FILES` consolidation fixes a real latent bug: the relay image was serving a dashboard that 404'd its own module graph.

Below is a full review at `high` effort against `origin/main...HEAD`. Each finding survived three independent refutation attempts. `go test ./core/pkg/webpush/... ./core/domain/notify/... ./core/cmd/irrlichtrelay/... -race -count=1` passes on this branch.

**Two are blocking. Four are follow-ups.** Neither blocker shows up in the device test this PR is waiting on, which is why they are worth catching here.

---

## 🔴 Blocking

### 1 · The policy engine never forgets a session, and it re-reads them all on every event

`core/domain/notify/engine.go:97`

`Engine.sessions` loses an entry only via `sessionDelete` (line 444) or a rekey. But the relay never sends a delete when a daemon link drops — `hub.daemonDisconnected` evicts its own cache and fans out a daemon status only (#540, "fade, don't delete"). After a reconnect, `applyDaemonSnapshot` diffs against an empty `old`, so no delete comes from there either. **Every session that ends while its daemon is offline stays in the engine for the life of the relay.**

Separately, `fireDue` (line 211) walks the whole map on every event, and the observer's `run` loop calls `nextWake()` after each one — a second full walk.

The two compound: the pile that grows is the pile that gets re-read. Measured on an M4 (`go test ./core/domain/notify/ -bench Handle -benchtime 200x`):

| Retained sessions | One same-state update | `NextWake()` |
|---|---|---|
| 100 | 1.5 µs | — |
| 1 000 | 20 µs | — |
| 10 000 | 310 µs | 129 µs |

**Failure scenario.** A relay runs for months. A Mac disconnects with 30 live sessions; the user ends them offline; the Mac reconnects. Those 30 records become permanent, and the count only rises. At ~10 000 retained records one event costs ~0.44 ms on the single observer goroutine, capping the drain at ~2 300 events/s. The daemon emits `session_updated` on every metrics tick (the engine's own comment at line 300 says so), so the queue fills and `enqueue` drops. **Notifications stop, and the only sign is one throttled log line.**

A second defect rides on this: the drop log at `push_observer.go:238` states each dropped event "self-heals at the next update or snapshot". True for an update; **false for a `session_delete`** — the session is gone, so nothing ever names it again. A dropped delete leaks permanently. Per AGENTS.md's dismissals-carry-evidence rule, that line needs narrowing to the case it actually covers.

**Suggested fix**
- Stamp each record with a last-touched time and sweep idle records from the existing `Tick`. A 24 h idle window is generous — a live session that outlives it re-registers on its next update, silently, because first sightings already are.
- Keep a separate short list of sessions with a pending hold-down. `fireDue` and `NextWake` read that list instead of the full map, which makes per-event cost independent of how many sessions the relay has ever seen.
- Correct the drop log line so it does not dismiss a loss that does not recover.

A soak assertion would lock it: feed the engine 100 000 events across a disconnect/reconnect and assert the map stays bounded. Happy to hand over the probe I wrote to confirm this — it currently reports `10000` records still held after the daemon left and a day passed.

---

### 2 · Anyone can lock every phone out of pairing, with ten requests a minute

`core/cmd/irrlichtrelay/push/codes.go:101`

`POST /api/v1/push/pair` carries no bearer gate by design — the code is the credential, which is right. To bound online guessing, `Redeem` counts failures and refuses **every** redeem, correct codes included, once ten land inside a rolling minute. That counter is **one counter for the whole relay**, and there is no per-IP throttle or HTTP middleware in front of it (I checked `buildMux` and the whole `irrlichtrelay` package — only WebSocket connections are capped).

**Failure scenario.** An attacker who knows the relay's public origin POSTs ten wrong codes each minute. `len(s.failures)` stays ≥ 10 indefinitely. Every real phone gets 429 and can never pair. Cost to the attacker: ten unauthenticated HTTP requests a minute. §8.1 makes this reachable by design — "a push-capable relay is by definition reachable".

The same shape crosses tenants at line 73: `maxOutstandingCodes` caps 32 codes **relay-wide**. One workspace — or one paired phone, since a device token is a full client token — mints 32 and blocks every other tenant from minting.

**Suggested fix** — key the counters instead of sharing them:
- Count redeem failures per client IP, keeping a higher relay-wide ceiling as a backstop against a distributed attempt.
- Count outstanding codes per workspace rather than per relay.
- Keep the uniform 401 and the distinct 429 exactly as they are. The anti-oracle behaviour is correct and should not change.

---

## 🟡 Follow-ups

### 3 · The shutdown drain is described but never runs

`core/cmd/irrlichtrelay/push_observer.go:304`

`shutdown()` says `sendDrainTimeout` "sits well inside runServe's 5 s srv.Shutdown window, because the two run in sequence". They do not run in sequence: `runServe` closes `stop` (`main.go:379`) and calls `srv.Shutdown` immediately (`main.go:383`). Nothing reads `o.done` — it appears once in production code and in four test files, and nowhere else.

Since WS connections are hijacked and no HTTP request is typically in flight, `srv.Shutdown` returns at once and the process exits. The two-second bound never applies and the "abandoning them" line never prints.

The delivery impact is small — the phone recovers at the next event. The real cost is that **four tests assert against a path production never takes**, which is the same "green that was never red for the production path" shape this PR reports fixing for the dispatcher.

**Suggested fix.** Read `<-obs.done` in `runServe` before `srv.Shutdown`. One line, and it makes both the comment and the four tests describe production. The alternative — delete the drain and the channel, and state plainly that sends die with the process — is also fine. Keeping both is the only option that is not.

---

### 4 · A group notification says one word, four times

`core/domain/notify/engine.go:417`

`sessionEvent` sets `Label: s.Adapter` (`push_observer.go:210`), and `SessionState` carries no per-session title. Most people run several sessions of one agent.

So the burst banner reads **"4 agents need attention"** over the body **"claude-code, claude-code, claude-code, claude-code"**. The count is right; the list says nothing, and the six-label cap fills with repeats before a distinct adapter can reach it. A single-session banner has the same problem: three sessions in one project each produce "claude-code needs input" over the same project name, and the user cannot tell which one to open.

**Suggested fix.** Compose the label from more than one field in `sessionEvent` — adapter plus `GitBranch`, falling back to the short session id when the branch is empty. Mirror it in `ledgerRow` so the push fold and the live fold keep writing one dialect rather than two.

---

### 5 · Pairing overwrites the user's relay settings, and unpairing deletes them

`platforms/web/elfdans.js:527`

`renderSection` draws the mint control **and** the code entry on the same page, so a user can pair on the Mac dashboard. `configureLiveView` then writes `relayToken: deviceToken` over whatever token the user had configured, because `sameRelay(current.relayUrl, location.origin)` is true. `releaseLiveView` (line 546) later writes `{enableRelaySource: false, relayUrl: '', relayToken: ''}` — because `liveViewIsOurs` now recognises its own overwrite as ours.

**Failure scenario.** A user configures Settings → Sources against `https://relay.example`, opens that dashboard, pairs a code there, then presses **Unpair**. The Sources panel comes back empty, and the original URL and client token are gone with no warning.

**Suggested fix.** Record whether we wrote the token. Leave a `relayToken` we did not write alone, and surface that state in the health panel the way the "pointed elsewhere" case already does — `liveViewNoteText` is the right home for it. On unpair, restore the previous value rather than clear the source.

---

### 6 · A subagent can reach the home-screen badge

`platforms/web/elfdansDashboard.js:298`

`elfdansLedgerSessions` says "Top-level sessions only", and derives that from position in `g.agents` rather than from `parent_session_id`. But `insertNewSession` (`irrlicht.js:532`) puts a subagent into `group.agents` when `attachToParentIfPresent` cannot find the parent, and nothing re-parents it later. `applyDaemonSnapshot` fans sessions out in Go map order, which is randomised — so a child frame preceding its parent is routine, not exotic.

The badge can then count a subagent as needing attention, which contradicts the engine's deliberate rule that subagents never notify.

**Suggested fix.** One guard: `if (a.session_id && !a.parent_session_id)`.

---

## Two smaller notes

- **`examples/relay/irrlichtrelay.service` and `io.irrlicht.relay.plist`** tell the operator to run `cp platforms/web/*`. That is a fourth copy site — exactly what `WEB_FILES` exists to remove — it puts test files on a reachable relay, and it errors on the `node_modules` and `snapshots` directories. Pointing operators at the tarball's `Resources/web` would close it, and `release-files.test.js` could grade this site too.
- **The app's Unpair button leaves the device token valid on the relay.** `site/docs/elfdans.html:265` documents this clearly, so it is a decision and not a defect — but the person pressing the button gets no hint of it. One clause in the button's copy would settle it.

---

## On the device test

The PR is admirably honest that no notification has reached a real phone. Worth noting: **findings 1 and 2 would not have been caught by that test either.** One needs months of uptime, the other needs an attacker. So the device test is still the right gate for what it covers — it just is not the last remaining one.

Coverage of this review, for the record: I read every production file in the diff — `core/pkg/webpush`, `core/domain/notify`, the `irrlichtrelay` push package, handlers, observer, tokens, hub and main, and `platforms/web`'s `sw.js`, `elfdans.js`, `elfdansDashboard.js`, `irrlicht.js` and `index.html` — plus the build, install, Docker and unit-file changes. I sampled the ~8 000 lines of tests rather than reading all of them, and I did not review `tools/elfdans-drive`, `tools/elfdans-rig.sh`, or the arc42 and site documents beyond the claims I checked against code. Given the size and the crypto surface, a human `/code-review ultra` before merge is still worth the spend.

Thank you again — this is a large, careful, well-argued piece of work, and the two blockers are both narrow fixes in code that is otherwise a pleasure to read. Happy to open a follow-up branch with any of these if that is easier than carrying them yourself.

---
🤖 Review performed with [Claude Code](https://claude.com/claude-code)

MBoegers:
Also, some future work here. Setting up and sharing tokens between the Elfdanse relay, app and the Irrlicht workspace is still manual. We would need some kind of SSO (maybe GitHub) to identify users and then need a self-service setup to create tokens.

Furthermore, a security thing caused by defensive programming, `--workspace` defaults to `""`. If omitted in `irrlichtrelay token issue --label merlin-daemon` we connect to the same workspace and see the same events. We have to rethink the defaults here; it should be fine to error on empty.

ingo-eichhorst:
## Device test run — the "no notification has ever reached a real phone" gap is closed

Ran the device test against this branch (`4a6a6e4c`) on macOS, with an isolated relay
(`tools/elfdans-rig.sh`, state in `.build/elfdans-rig`), a separate dev daemon on `:7838`,
and the relay published over Tailscale Serve for a real HTTPS origin.

**A real Claude Code session going to `waiting` produced a lock-screen banner on an iPhone
via `web.push.apple.com`.** Not the diagnostic button, and not the synthetic driver — the
daemon observed a live session, the relay's policy engine passed it, and the phone buzzed.

### What passed

| Phase | Evidence |
|---|---|
| Rig assertions | All 10 of `elfdans-rig.sh check` green |
| 2 — daemon → relay | Live sessions from a real daemon appear in the relay's `/api/v1/sessions` |
| 0 — desktop | Chrome paired, subscription on `fcm.googleapis.com`, test push accepted, banner shown (tag `elfdans-test`) |
| 3 — phone | Installed from the home screen, paired, registered on `web.push.apple.com`, banners arrive |
| 5 — policy | All six `drive` scenarios, each verified as a real banner in Chrome: `waiting` immediate; `ready` at **+7.4 s**; hold-down cancel produced **no push at all**; burst → three singles plus one "4 agents need attention"; subagent silent; watchdog at **+60.3 s**; reconnect replaced silently |
| 7 — additivity | On the daemon origin `:7838`: no Elfdans section, no mint row, and **no service worker registered**. `/api/v1/push/info` answers 404 |

Both mock-proof defects named in the PR body are clear: a real push service accepted the
VAPID JWT, and `subscribe()` resolved after activation on a real iOS install.

### Not verified

- **Tap-to-open (R6).** Never exercised. Needs a real tap on a real banner.
- **The ~3 s latency claim.** Banners arrive; nobody timed them.
- **ADR-3's storage partition.** The Phase 3 check — pair in the Safari tab, confirm the
  installed app does not inherit it — has not run. Its outcome decides whether ADR-3 is
  load-bearing or should be simplified.
- **Phase 6 recovery** in full: revoke, un-pair, relay restart, self-heal, fresh-relay restore.

---

## Three findings worth fixing before merge

### 1. Every push-capable relay shows "Disconnected" on first load

The relay serves the dashboard, and the dashboard's default **Local** source connects
back to that same origin — but the client sends its bearer token only for relay-kind
sources:

```js
// platforms/web/irrlicht.js:2067
if (src.kind === 'relay' && settings.relayToken) hello.token = settings.relayToken;
```

The relay authenticates on the hello message and closes an empty token with
`CloseRevoked` (`core/cmd/irrlichtrelay/hub.go:249-255`). Push **requires** auth (§8.1),
so on every relay that can do push the Local source is always refused, and first contact is:

> Disconnected — no configured source is reachable.

The page the relay itself served has to be told, by hand, the URL and token of the relay
that served it. Suggested fix: on a relay-served origin the Local source should carry the
configured token, or the empty-token close should surface as "this relay needs a token"
rather than as a generic unreachable-source banner.

### 2. The mint row does not appear until the page is reloaded

`renderMacSide` builds the "Pair a phone…" control only when a client token already exists
at page-build time. Enter the token in Settings → Sources and the Elfdans panel keeps
showing only the redeem field. **There is no way to mint a code until you reload** — and
nothing says so. Combined with finding 1, first contact is a dashboard that says
"Disconnected" and a pairing panel with no way to pair.

### 3. `docs/elfdans-device-test.md` contradicts the spec it grades

Phase 5, Burst row:

> Four sessions into `waiting` inside 20s → One "N agents need attention", **not four**

arc42 §8.4 (line 398) specifies:

> `> 3 pushes within 20 s → single summary "N agents need attention"`

The implementation is correct and matches the spec: three individual banners, then a
summary for the fourth. Verified in Chrome. But an operator following the device test
counts four banners and marks a correct implementation as failed — on the document that
is the gate for calling this feature done.

---

## Smaller items, not blocking

- The mint hint tells the phone to open `location.origin`, which on the Mac is
  `http://127.0.0.1:7839/` — an address no phone can reach, printed at the exact moment
  it is needed.
- `elfdans.js:275` fails a non-secure origin with "This browser cannot receive push
  notifications". On a plain-HTTP LAN origin the browser is fine; the **origin** is not.
  Measured on `http://192.168.188.119:7901`: `isSecureContext:false`, `serviceWorker`
  absent, while `PushManager` and `Notification` still exist as globals.
  `window.isSecureContext` separates the two cases and the message should name HTTPS.
- `tools/elfdans-rig.sh up --serve` hangs indefinitely when Tailscale Serve is not enabled
  on the tailnet: `tailscale serve` prints an enable link and then polls, and the rig pipes
  that into `tail`, so the operator sees nothing.

## An addition worth considering later

Onboarding a phone is currently: type a tailnet hostname, add to home screen, mint a code,
type the code. A QR code on the Mac — menu bar app and web UI both, with the URL printed
underneath so copy-paste still works — would remove all the typing.

The one design note, since it is not obvious: `start_url` is `./`, so iOS drops a query
string at install time, and the storage partition (if ADR-3 holds) drops anything Safari
saved. Carrying the pairing code in the **path** (`/pair/<code>/`) survives both, because
`./` resolves relative to it. Worth its own issue rather than this PR.

Deriving the URL for such a QR needs its own small mechanism: a relay bound to loopback
cannot infer its public origin. Tailscale Serve forwards `Host` and `X-Forwarded-Proto`
(measured), and `tailscale serve status --json` maps local port to public host — but that
is one deployment among several, and plain-HTTP LAN cannot carry Web Push at all. An
explicit `--public-url`, plus a candidate list that labels which origins a phone can
actually use, covers every case.

MBoegers:
@ingo-eichhorst was just checking casually for the state an BOOM merge 🥳 thanks

ingo-eichhorst:
> @ingo-eichhorst was just checking casually for the state an BOOM merge 🥳 thanks

It just works :-). I now have it on my phone and I get notified when I have to speed up the coffee break 😄

A native experience would be even better, but that is for later. For now it's great. Will ship with the next version and Claude Desktop support.

PR #1903, opened 2026-09-04
WIP: fix(release): stage the whole dashboard web tree by rule, not a three-file list

### LINKED ISSUE #1900: Released dashboard is completely blank: build-release.sh ships only 3 of 12 web files, every ES-module import 404s
## Symptom

The web dashboard of the **released** build shows no subscription/quota chips. In fact it shows nothing at all: an empty page, the app-title strip, and a connection dot stuck at `connecting`.

Observed on the production daemon (`/Applications/Irrlicht.app`, `irrlichd version 0.6.2`) at `http://127.0.0.1:7837/`.

## Root cause

`tools/build-release.sh` copies a **hardcoded three-file list** into every release artifact:

```
cp platforms/web/index.html platforms/web/irrlicht.css platforms/web/irrlicht.js "$TARBALL_STAGING/web/"      # line 65  — darwin daemon tarball
cp platforms/web/index.html platforms/web/irrlicht.css platforms/web/irrlicht.js "$staging/web/"              # line 87  — linux amd64 + arm64 tarballs
cp platforms/web/index.html platforms/web/irrlicht.css platforms/web/irrlicht.js "$APP_CONTENTS/Resources/web/" # line 124 — macOS .app bundle
```

Since `refactor(web): extract a collapsed-groups store …` (c36c1384, 2026-06-20) and especially `refactor(web): split irrlicht.js's cohesive subsystems into …` (1c7faaa7, 2026-07-03), `irrlicht.js` is an **ES module that imports 10 sibling modules**:

`collapsedGroups.js`, `collapsedSet.js`, `collapsedSummaries.js`, `connectionProtocol.js`, `domReconcile.js`, `formatters.js`, `historyTab.js`, `permissionsWizard.js`, `quotaChips.js`, `sessionIdentity.js`

None of them are packaged. `registerUIRoutes` serves the staged directory with `http.FileServer`, so every import returns **404**. The module graph never evaluates, `rebuildSources()` never runs, no WebSocket opens, and no session row or quota chip is ever built.

The stale comment at `core/cmd/irrlichd/startup.go:412-414` ("the dashboard ships as three files") records the assumption that broke.

### Evidence

Installed bundle holds only three files:

```
$ ls /Applications/Irrlicht.app/Contents/Resources/web/
index.html  irrlicht.css  irrlicht.js
```

Every import 404s on the released daemon, while the dev daemon (repo `platforms/web/`) serves them:

```
$ for f in collapsedGroups collapsedSet collapsedSummaries connectionProtocol domReconcile \
           formatters historyTab permissionsWizard quotaChips sessionIdentity; do
    printf '%s %s %s\n' "$f.js" \
      "$(curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:7837/$f.js)" \
      "$(curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:7838/$f.js)"
  done
collapsedGroups.js     404 200
collapsedSet.js        404 200
collapsedSummaries.js  404 200
connectionProtocol.js  404 200
domReconcile.js        404 200
formatters.js          404 200
historyTab.js          404 200
permissionsWizard.js   404 200
quotaChips.js          404 200
sessionIdentity.js     404 200
irrlicht.js            200 200
```

Live DOM on the released dashboard (port 7837), after a full load:

```
{"headerClass":"","chips":0,"rows":0,"text":"Irrlicht\n—\nHistory\nContext\n⊟\n☾\n⚙\nconnecting"}
```

`connecting` is the *static* markup from `index.html:33`; the `.ws-status` tooltip is empty, which means the `sources` map is empty and `rebuildSources()` never ran.

### The chip code itself is correct

Feeding the live `/api/v1/sessions` payload from the production daemon (both providers present — `claude-code` with 5h+7d windows, `codex` with a 7d window) into the repo's web tree under vitest renders both chips:

```
n: 2
"quota-chip :: 5h38%7d33% :: title=5h: 38% used · 35pt under pace · resets in 1h 20m"
"quota-chip :: 7d46%  :: title=Prolite"
header: "has-quota-chips"
```

So this is a packaging defect only — no change is needed in `quotaChips.js`.

## Scope

All four release artifacts are affected, so **the dashboard is broken for every installed user on macOS and Linux**, not only the quota chips:

- macOS `.app` bundle (`Contents/Resources/web`)
- darwin daemon tarball (curl `--daemon-only` installer)
- linux amd64 + arm64 daemon tarballs (the web dashboard is the *only* UI on Linux)

## Fix

1. Stage the web tree by rule instead of by hand in `tools/build-release.sh` (all three sites). The runtime set is `platforms/web/*.html`, `*.css`, and `*.js` **excluding** `*.test.js`, `vitest.config.js`, `vitest.setup.js`, `node_modules/`, and `snapshots/`.
2. Fix the stale three-file comment in `core/cmd/irrlichd/startup.go:412-414`.

## Regression test (required)

A vitest test cannot catch this — the defect lives in the packaging step. Two guards are needed:

1. **Packaging guard (the real one).** A shell test that stages the web assets the same way `build-release.sh` does, then asserts every module `platforms/web/irrlicht.js` imports is present in the staging directory, and that no dev-only file (`*.test.js`, `vitest.*`) shipped. It must derive the required set from the actual `import … from './X.js'` statements, so a *future* module extraction fails the gate automatically instead of silently 404-ing in the next release. Per AGENTS.md, this guard has no "before the fix" to run red — mutate it (drop one module from the staged set) and confirm it goes red, and commit that mutation as a fixture.
   - **A verification mechanism must fail loudly when it cannot run**: if the import scan finds zero imports, or the staging directory is missing, the test must fail rather than pass vacuously.
2. **Smoke check (optional but cheap).** Boot the staged daemon against the staged web dir and assert every imported module returns 200 — end-to-end proof the served tree is complete.

Wire the packaging guard into `tools/preflight.sh` so it runs before a release.

ingo-eichhorst:
> **⚠️ BLOCKED ON A DESIGN DECISION — do not merge as-is.**
>
> While this PR was in review, **#1847** merged and independently fixed the same
> defect with a **different design**: a single `WEB_FILES=( … )` array plus a
> `copy_web_files` helper in `tools/build-release.sh`, a fourth staging site
> (the Linux relay tarball), fixes to `site/install.sh` and
> `examples/relay/Dockerfile`, and a ~400-line vitest guard
> (`platforms/web/release-files.test.js` + `shippedFiles.testutil.js`) with its
> own committed mutation fixtures.
>
> This PR replaces that **list** with a **rule** (`tools/lib/stage-web.sh`), so
> the two cannot both stand. Measured, not assumed: applying this PR's staging
> to `origin/main`'s `build-release.sh` and running main's own suite gives
> **5 failed | 57 passed (62)** — the arms that assert "build-release.sh defines
> a single WEB_FILES list", "every copy site draws from that list", and "the
> relay tarball must inherit WEB_FILES via copy_web_files".
>
> Landing this PR therefore means rewriting #1847's just-merged test suite to
> grade the rule instead of the list. That is a maintainer call, not an
> execution detail, so `ir:exec` stopped here rather than choosing.
>
> Note also that #1847 did **not** close #1900, and #1900 is still open — the
> symptom is fixed on main, the design question is what is open.

---

Closes #1900

## What was broken

`tools/build-release.sh` staged the dashboard from a hand-written three-file `cp` list (lines 65 / 87 / 124) into all four release artifacts. The list was complete when written (#418) and stopped being complete in #712 / #820, when `irrlicht.js` became an ES module importing ten siblings. None were packaged, so every import 404'd, the module graph never evaluated, and the **released** dashboard rendered a blank page for every installed user on macOS and Linux — where the web dashboard is the only UI.

## What changed

- **`tools/lib/stage-web.sh` (new)** — the runtime asset set as a *rule*, defined once: every `*.html` / `*.css` / `*.js` directly in the web tree, minus `*.test.js`, `vitest.config.js`, `vitest.setup.js`. The glob does not recurse, so `node_modules/` and `snapshots/` cannot be swept in, and `package*.json` is not matched. Naming the set by extension and subtracting the dev files — rather than listing the runtime files — is what makes a newly extracted module ship by default.
- **`tools/build-release.sh`** — sources that library and calls `stage_web` at all three staging sites.
- **`tools/web-release-assets-guard.sh` (new)** — *executes* that rule (not a copy) into a temp dir, walks the transitive closure from `index.html`'s `<script type=module>` / `<link rel=stylesheet>` with a visited set, and asserts both directions: everything reachable is staged, and nothing dev-only is. It also asserts `build-release.sh` still routes every `platforms/web` copy through `stage_web`, so the two cannot drift apart again.
- **`core/cmd/irrlichd/ui_release_tree_test.go` (new)** — the served-side half: stage with the real rule, walk the real import graph, require 200 for every reachable asset through `registerUIRoutes`. The two sides come from *different* sources on purpose — a test that asked the staged directory what to request could never reproduce the defect.
- **`core/cmd/irrlichd/startup.go`** — the stale "ships as three files" comment, which is where the broken assumption was recorded. It now states the rule (a blanket `http.FileServer`), not a file count. Also fixes a stale `main.go` → `paths.go` reference beside the staging site.
- **`tools/preflight.sh`** — the guard joins the `tools` group, and `^platforms/web/` joins both that gate's trigger and the shell-lib-tests trigger.

### Three properties are load-bearing

1. **The walk is transitive.** `irrlicht.js` imports 9 siblings directly; `collapsedSet.js` is reachable only via `collapsedGroups.js` / `collapsedSummaries.js`. A direct-import scan finds 9 of 10 and reports completeness.
2. **It is cycle-tolerant.** `permissionsWizard.js` and `quotaChips.js` both import back from `./irrlicht.js`.
3. **It fails loudly when it cannot look.** `irrlicht.js` and `sessionIdentity.js` carry literal NUL bytes; a binary-detecting grep answers `Binary file … matches` and yields no edges, which would compute an empty closure and pass. Every read is `grep -a`, and a closure with zero import edges REFUSEs.

## Test evidence

**Red-first (a real "before the fix" — the defect ships today):**

- The guard's closure run against `origin/main`'s own line 65 `cp` command names all 10 missing modules — the same 10 the issue measured as 404 on the released daemon.
- `TestRegisterUIRoutesServesTheStagedReleaseTree` under `origin/main`'s three-file rule reports `GET /<module>.js = 404, want 200` for all ten. Reproduce (also recorded in the test's doc comment):
  ```
  tools/mutate.sh tools/lib/stage-web.sh \
    '    for f in "$src"/*.html "$src"/*.css "$src"/*.js; do' \
    '    for f in "$src"/index.html "$src"/irrlicht.css "$src"/irrlicht.js; do' \
    bash -c 'cd core && go test ./cmd/irrlichd/ -run TestRegisterUIRoutesServesTheStagedReleaseTree -count=1'
  ```

**Mutation-proved (`tools/lib/web-release-assets-guard-mutations_test.sh`, 8 committed fixtures, all seen red then restored):** the staging rule dropping `collapsedSet.js` (the transitive-only module); dropping `formatters.js` (a first-hop import); shipping `*.test.js`; the scan losing `grep -a` on a NUL-containing module; the zero-import-edge refusal disarmed; `build-release.sh` staging by hand again; and `^platforms/web/` leaving either of the two preflight triggers. The NUL row asserts its own precondition and *fails* — not skips — if the local `grep` cannot binary-detect.

**Locks (green by construction, not evidence):** `tools/lib/web-release-assets-guard_test.sh`'s happy path and its 7 fail-loudly refusal fixtures; `platforms/web/` vitest (no web source is touched); `tools/lib/shell-lib-errexit_test.sh` (extended with 6 rows for the new library — it auto-discovers `tools/lib/*.sh`).

**Gate parity** (the thing this issue is about, one level up): the lock test runs preflight's *own* trigger regexes through the same `grep -qE` against `platforms/web/newlyExtractedModule.js` and fails if either gate would be SKIPped — so a commit that extracts a new module cannot skip the gate that catches it.

**Preflight:** `go`, `web`, `arch`, `tools`, `skills`, `posix`, `bash` all PASS run individually; the pre-push run (`PREPUSH_BUDGET=2400`) reported PASS for every selected gate including `security` and `swift`. An earlier unbounded `--only security` run failed *only* on `npm audit` network timeouts to `registry.npmjs.org` (the gate says so itself: "this is NOT a vulnerability finding"); re-running `npm audit` in both web trees returned `found 0 vulnerabilities`, and this diff touches no `package.json` or lockfile.

## Design deviation from triage

Triage said "a single shell function **in** `tools/build-release.sh`". The function lives in `tools/lib/stage-web.sh` instead, sourced by `build-release.sh`. Evidence: `build-release.sh` runs a full signed universal build at top level under `set -e`, so it cannot be sourced — a guard could only re-parse its text, and triage's own requirement was that the guard execute the *real* rule. `tools/lib/` is this repo's established home for sourced shell libraries, and `tools/lib/shell-lib-errexit_test.sh` grades every function there by discovering it. The loop back to the shipping script is closed by assertion (c) in the guard, which fails if `build-release.sh` touches `platforms/web` outside `stage_web`.

## Human gate

Per triage, the release build + notarization is the maintainer's (`/ir:release`). After merge: `ls "$BUILD_DIR/Irrlicht.app/Contents/Resources/web/"` should show 13 files.

https://claude.ai/code/session_01615niHdVNthGbskmVG4JVt

### DIFF
--- a/.claude/skills/ir:release/SKILL.md
+++ b/.claude/skills/ir:release/SKILL.md
@@ -416,7 +416,9 @@ from `gh api`, the token is missing the `security_events` scope:
 > `killStaleDaemons` skip the pkill. The app's embedded daemon spawn can be
 > flaky under a bare exec; if so, validate the bundle's daemon directly:
 > `IRRLICHT_BIND_ADDR=127.0.0.1:7839 IRRLICHT_HOME=… .build/Irrlicht.app/Contents/MacOS/irrlichd`
-> then `curl 127.0.0.1:7839/ | grep '<title>'` (covers the v0.4.4 missing-web bug).
+> then `curl 127.0.0.1:7839/ | grep '<title>'` (covers the v0.4.4 missing-web
+> bug) **and** `curl -fsS 127.0.0.1:7839/irrlicht.js >/dev/null` plus one of its
+> sibling modules (covers #1900, which the `<title>` check alone cannot see).
 
 ### Go daemon (universal binary + tarball)
 The daemon reads `platforms/web/index.html` from disk at runtime; no embed.
@@ -432,7 +434,11 @@ lipo -create /tmp/irrlichd-arm64 /tmp/irrlichd-amd64 -output /tmp/irrlichd-darwi
 # Tarball for the curl --daemon-only installer
 rm -rf /tmp/irrlichd-tarball && mkdir -p /tmp/irrlichd-tarball/web
 cp /tmp/irrlichd-darwin-universal /tmp/irrlichd-tarball/irrlichd
-cp /Users/ingo/projects/irrlicht/platforms/web/index.html /tmp/irrlichd-tarball/web/index.html
+# The dashboard is staged BY RULE, never by a hand-written file list — see
+# tools/lib/stage-web.sh. #1900: a three-file list here shipped a blank
+# dashboard to every install once irrlicht.js grew sibling modules.
+( cd /Users/ingo/projects/irrlicht && . tools/lib/stage-web.sh &&
+  stage_web platforms/web /tmp/irrlichd-tarball/web )
 tar -czf /tmp/irrlichd-darwin-universal.tar.gz -C /tmp/irrlichd-tarball .
 ```
 
@@ -441,8 +447,9 @@ tar -czf /tmp/irrlichd-darwin-universal.tar.gz -C /tmp/irrlichd-tarball .
 These ship the Linux curl install path (`site/install.sh` auto-detects Linux
 and downloads `irrlichd-linux-<arch>.tar.gz`). Pure cross-compile from macOS,
 no cgo. **Required** — omitting them makes every Linux `curl … | sh` 404 on
-the asset. Each Linux tarball carries all three web files (index.html + css +
-js), because the Linux installer installs all three.
+the asset. Each Linux tarball carries the whole dashboard — `index.html`, the
+stylesheet and every ES module `irrlicht.js` imports — staged by
+`stage_web`, because on Linux the web dashboard is the only UI there is.
 
 ```bash
 cd /Users/ingo/projects/irrlicht/core
@@ -451,10 +458,8 @@ for arch in amd64 arm64; do
     -o "/tmp/irrlichd-linux-$arch" ./cmd/irrlichd
   rm -rf "/tmp/irrlichd-linux-tarball-$arch" && mkdir -p "/tmp/irrlichd-linux-tarball-$arch/web"
   cp "/tmp/irrlichd-linux-$arch" "/tmp/irrlichd-linux-tarball-$arch/irrlichd"
-  cp /Users/ingo/projects/irrlicht/platforms/web/index.html \
-     /Users/ingo/projects/irrlicht/platforms/web/irrlicht.css \
-     /Users/ingo/projects/irrlicht/platforms/web/irrlicht.js \
-     "/tmp/irrlichd-linux-tarball-$arch/web/"
+  ( cd /Users/ingo/projects/irrlicht && . tools/lib/stage-web.sh &&
+    stage_web platforms/web "/tmp/irrlichd-linux-tarball-$arch/web" )
   tar -czf "/tmp/irrlichd-linux-$arch.tar.gz" -C "/tmp/irrlichd-linux-tarball-$arch" .
 done
 ```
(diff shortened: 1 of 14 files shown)


ingo-eichhorst:
Closing in favour of #1906, per the maintainer's call.

**Why.** #1847 merged at 13:13 UTC while this branch was being built, and independently fixed #1900 with a different design: one `WEB_FILES` array plus a `copy_web_files` helper across four staging sites (including a Linux relay tarball this branch did not know about), fixes to `site/install.sh` and `examples/relay/Dockerfile`, and a 504-line guard suite in `platforms/web/release-files.test.js`.

This branch replaces that **list** with a **rule** (`tools/lib/stage-web.sh`). The two cannot both stand — applying this staging to current main gives 5 failed / 57 passed against main's own suite.

**Main's fix was verified before closing this, not assumed:**
- Every runtime file in `platforms/web/` appears in `WEB_FILES`; the only unlisted file is `shippedFiles.testutil.js`, a test helper.
- `release-files.test.js`: 62 passed.
- Staged a tree with main's own rule, booted a real `irrlichd` against it, polled every entry: 18/18 return 200 (`index.html` at `/`).

The two `site/install.sh` and `examples/relay/Dockerfile` misses this branch found were real, and #1847 fixed both independently.

**What survives, in #1906:** the served-side assertion (main's guard is static and never asks a server), adapted to execute main's `WEB_FILES` + `copy_web_files` instead of `stage-web.sh`, with its mutation fixtures; and the two stale "ships as three files" comments in `main.go` and `startup.go`.

The triage finding about `tools/preflight.sh`'s trigger turned out to be moot: line 676 already fires the `web` gate on `^platforms/web/`, and main's guard lives in that vitest suite.

The branch stays on the remote; nothing is deleted.

Issue #1905, opened 2026-09-04
Autonomy spans: show how long a session ran without needing a human


ingo-eichhorst:
## Idea

Show how long a session ran **autonomously** — the longest unbroken stretch of
`working` before it needed a human.

The appeal: it is the one number that says how much the agent actually carried on
its own, as opposed to cost or tokens, which say how much it consumed.

## Decision — where it lives

**Autonomy is a new top-level History section, a sibling of Usage / Metrics /
Quota — not another entry in an existing chart picker.**

- macOS: a fifth case on `HistoryTab`
  (`platforms/macos/Irrlicht/Views/HistoryView.swift:20`, today
  `usage, activity, metrics, quota`), with its own control bar. The enum's own
  doc comment already states the rule this follows: *"Each concern is too
  different from the others to share one control bar."* Autonomy qualifies —
  neither of its two elements uses `Range`, `Group`, or the drilldown filters.
- Web: the dashboard has no top-level tab strip, so the equivalent is a third
  segmented group in the History control bar, beside `#history-chart-sel` and
  `#history-metrics-sel` (`platforms/web/index.html:253` and `:264`). Note the
  two surfaces already differ here — the web History bar has no Quota group at
  all (quota lives in the header strip, `platforms/web/quotaChips.js`). Closing
  that gap is not part of this issue.

The section holds **two** elements, both described below. Both are new; only the
second resembles an earlier sketch.

## The constraint that shapes the design

The obvious implementation — mine the spans out of the state history — does not
work today.

`chart=state` (#981, `core/cmd/irrlichd/handlers.go:727`) is served by
`ConcurrencyTracker`, which *"reconstructs concurrent-agent counts over time from
the lifecycle recordings irrlichd writes under `<dataDir>/recordings`"*
(`core/adapters/outbound/filesystem/concurrency_tracker.go:49`). Those recordings
are **opt-in**: `recordEnabled := hasFlag("--record") || os.Getenv("IRRLICHT_RECORD") == "1"`
(`core/cmd/irrlichd/main.go:567`), and the packaged app's daemon runs without
either — observed during this investigation as
`/Applications/Irrlicht.app/Contents/MacOS/irrlichd` with no `--record` in argv.

So for a normal user that history is empty, and a history-derived "longest span"
would silently report nothing — or a few minutes when the real answer was hours.
A wrong number here is worse than no number, because nothing on screen would say
it was wrong.

[...]

PR #1914, opened 2026-09-05
feat(history): a MARKED historical back-fill for Autonomy (#1905)

ingo-eichhorst:
A **marked** historical back-fill for the Autonomy section (#1905, shipped in #1912), plus the `source` / `unknown` handling that ships in the daemon and both clients so a reconstructed figure can never be read as a measured one.

## Why there is anything to back-fill

The span log starts collecting the day the feature ships. On this machine two logs that were already on disk reach back much further:

| source | period | what it can honestly claim |
|---|---|---|
| `logs/events.log*` | 18 Aug → now | its `session-detector` entries carry the **real transition**, so a span from it has a real end reason |
| `cost/*.jsonl` | 18 Apr → now | records **when** a session was consuming tokens; nothing at all about **why** a run stopped |

#1905 named the lifecycle recordings as the back-fill source. **They do not exist on this machine** — `~/Library/Application Support/Irrlicht/` holds `cost`, `instances`, `logs`, `schema`, `sessions` and no `recordings`, because the packaged app runs without `--record`. Hence the two logs above.

## The exact command that back-fills this machine

```
go run ./tools/autonomy-backfill "$HOME/Library/Application Support/Irrlicht" --apply
```

Drop `--apply` for a dry run — that is the default, and it writes nothing. Everything below came from

```
go run ./tools/autonomy-backfill <a temp copy of that directory>
```

run at 2026-09-05 15:26 local. The daemon is live and both logs keep growing, so a later run reports slightly larger counts.

## Measured span counts per source

```
SOURCES
  event log : 6 file(s), 228247 lines, 228247 parsed, 7127 used, 0 malformed (0.000%)
  cost log  : 193 file(s), 78228 lines, 78228 parsed, 78228 used, 0 malformed (0.000%)
  boundary  : 2026-08-18 01:07   (the event log's first transition; the cost era stops there)
  live from : «none»             (nothing measured yet on this copy)

RECONSTRUCTED
  log      1755 spans  2026-08-18 22:00 → 2026-09-05 15:26  [error 32  ready 917  waiting 806]
  cost     8346 spans  2026-04-18 22:38 → 2026-08-18 01:07  [unknown 8346]
  TOTAL   10101 spans

DELIBERATELY DROPPED (an undercount, and the safe direction)
      52  daemon restarts in the event log (clustered from its `startup` entries)
      58  log spans straddling a restart — several runs merged, dropped not split
      82  closes with no matching open (the open half predates the retained log)
     550  `ready` with no open span (provisional; emits nothing either way)
      85  `→ working` while already working (the live detector replaces the start too)
       9  log spans still open when the log ends (no measured end; none invented)
       1  log spans with no project to file them under
       0  spans reaching into the era the daemon already MEASURED (never counted twice)
    1842  cost stretches at or after the boundary (the event log speaks for that era)
    5296  cost stretches of a single row — an instant, with no duration to measure
     141  log spans whose end did not follow their start
```

Two numbers differ from the ones in the brief, both explainable:

- **restarts: 52, not 400.** 400 is the count of `startup` **entries**; one boot writes several. They are clustered to one instant per boot (`restartClusterSeconds`, 30 s), taking the earliest line of each cluster — the daemon was down *before* that timestamp.
- **restart-straddlers: 58, not 63.** Same cause: five spans straddled only a *later* line of a boot they were entirely inside. Clustering to the boot's first line correctly stops calling those straddlers.

## Cost-log gap sensitivity

Printed on every run, dry or not; only `source=cost` spans are affected.

```
  gap           spans        p50        p95
  120s           9395      2m43s     15m02s
  180s           8346      3m07s     20m11s  ← in force
  300s           7069      4m15s     32m14s
```

`costGapSeconds = 180` carries the reasoning at the constant: 3× the cost log's 60 s write interval, long enough to ride out a tool call that produces no tokens, short enough that a human answering in a minute is not absorbed into the run they interrupted.

## The honesty rules, and where each one lives

- **`source` on the row.** Absent means measured live, so **not one row already on disk is rewritten** and there is no migration. `log` and `cost` are the two reconstructions. Any *unrecognized* source still reads as reconstructed — a row from a newer back-fill must never read back as measured.
- **`unknown` for the cost era, never a guess.** It is deliberately **not** a session state: absent from `CanonicalStates()`, refused by `IsAutonomyEndReason`, ranked below every measured reason on the strip's collapse ladder, and drawn in the neutral colour both surfaces already use. Assuming `ready` would paint four months of the strip green under a claim nobody measured — and the duration chart, the section's main element, never reads the reason at all, so telling the truth costs nothing.
- **Restart-straddlers are dropped, not split.** Splitting would invent the boundaries the log does not contain.
- **Closes with no open are dropped.** No start means no duration.
- **Nothing is reconstructed into the era the daemon already measured.** The event log keeps being written after the feature ships, so the two eras overlap; a run present in both would be counted twice, which is the one back-fill error that looks like productivity. The live rows win.
- **The tool fails loudly when it cannot read.** Every unparseable line is counted, and `--apply` is refused above a 1 % unreadable share (`--allow-malformed` overrides, and says what it is overriding). Fewer spans and an unreadable source are otherwise the same result.
- **A second `--apply` is refused.** There is no undo; appending twice doubles every figure with no way to tell afterwards. `--force` overrides.

## Local-only

`tools/autonomy-backfill` is a manual CLI run by hand against an explicit data directory. It is **not** wired into daemon startup, migrations or first-run logic, and `TestBackfillIsNotWiredIntoTheDaemon` fails the build if anything under `core/` imports or execs it. Every other user's Autonomy view correctly starts empty on the day they upgrade, and the shipped empty state already says so.

## The UI marking

Both panels gain one sentence, in the register the shipped empty state uses, **silent when nothing in view was reconstructed**:

> 412 of 507 runs in view were reconstructed from logs this machine already had, not measured as they happened. Everything before Sep 6, 2026 is reconstructed. 96 of them come from the cost log, which records when a session was working and never why it stopped, so their end reason is unknown — not assumed.

(Illustrative shape — the counts are window-scoped, so the real ones depend on the range picked. The date is `live_since`: the earliest span the daemon measured.)

`live_since == 0` prints "Nothing here was measured live" rather than a fabricated 1970 date. The strip legend gains the neutral `unknown` entry, **only when the window actually holds a run with no nameable reason** (which covers both a cost-derived span and an old row whose reason was never recorded). The web panel's green/purple/orange p95/p50/p5 colour key is untouched.

## Which mutation fixture proves each new check

`tools/lib/autonomy-backfill-mutations_test.sh` drives the real `tools/mutate.sh` (stale-anchor guard, no-op refusal, byte-for-byte restore) and runs in preflight's `tools` group. All five were seen red:

| check | mutation | test that goes red |
|---|---|---|
| a reconstructed span is marked, a live one is not | `IsAutonomyReconstructed` always returns `false` | `TestAutonomySpanTracker_MarksReconstructedRowsAndLeavesLiveOnesBare`, `…_UnknownSourceStillCountsAsReconstructed` |
| a restart-straddling span is dropped | `straddlesRestart` always returns `false` | `TestRestartStraddlingSpanIsDropped` |
| a close with no open is dropped | the orphan branch emits a span instead of counting | `TestSpanRules/a_close_with_no_open_emits_nothing` |
| a cost-derived span gets `unknown`, never a guess | `AutonomyReasonUnknown` → `StateReady` | `TestCostSpansCarryUnknownAndNeverAGuessedReason` |
| nothing lands on top of a measured run | the live-floor bound stops bounding | `TestSpansReachingIntoTheMeasuredEraAreDropped`, `TestApplyNeverWritesIntoTheMeasuredEra` |

```
$ MUTATION_FIXTURES_STRICT=1 bash tools/lib/autonomy-backfill-mutations_test.sh
ok  a reconstructed span still reads back as reconstructed
ok  a span crossing a daemon restart is dropped
ok  a close with no matching open emits nothing
ok  a cost-derived span carries `unknown` and never a guessed reason
ok  a span reaching into the era the daemon already measured is dropped
autonomy-backfill-mutations: ALL PASS
```

The sixth check — **the panel states provenance when any span in view is reconstructed and says nothing when none is** — is proven by *committed in-language mutants* instead, in `platforms/web/irrlicht.history.autonomy.backfill.test.js` and `platforms/macos/Tests/HistoryAutonomyBackfillTests.swift`, following the idiom those suites already use (`mergedResolver`, `denseBuckets`): an always-speaks build and a never-speaks build answer identically for both fixtures, and production must not. Driving vitest and `swift test` from the shell fixture would have made it depend on an installed `node_modules` and a multi-minute Swift build, and a fixture that cannot run is worse than one scoped to what it can — that trade-off is written down in the fixture's header.

Both log parsers get table-driven fixtures including malformed lines: truncated JSON, non-JSON, an unreadable timestamp (which must **not** fall back to 1970), and cost rows missing the fields this tool needs. The `destinationState` cases are the nineteen real message shapes a census of this machine's log produced, which is why it takes the **last** arrow match — a third of them name a state on both sides of the arrow.

## Preflight

All nine groups run as separate foreground invocations, on the committed tree:

`go` PASS · `tools` PASS · `web` PASS · `swift` PASS · `security` PASS · `arch` PASS · `skills` PASS · `bash` PASS · `posix` PASS

`--only linux` was not run (needs Docker).

`tools/autonomy-backfill` is registered in `go.work`, `.github/dependabot.yml`, `tools/preflight.sh`'s `go` group and `test.yml` — otherwise its tests are a suite nothing runs.

## Not merged

Left for QA + screenshots. To see it populated without touching real data: copy `logs/` and `cost/` into a temp directory and run the command above against that.

---

https://claude.ai/code/session_01Rw2jkcKiLcYdGavEWDeRRB


ingo-eichhorst:
## QA-1 and QA-2 fixed — b273d311

### QA-1 · the run strip had no row cap

Both surfaces now draw a **prefix of the daemon's ranking** and state the omission:

- web: 12 rows, then `+83 more projects, each with less autonomous time than the rows above`
- macOS: 6 rows (380 pt popover, under a 190 pt chart), same line

The daemon already orders `projects` by **total autonomous seconds**, so the cap keeps what matters — one four-hour run outranks forty three-second ones — and no ranking is re-derived client-side. The two caps differ **because each surface picks what it can show**; since both take a prefix of one shared order, that is not a disagreement, and a test on each side pins the prefix property so a future "optimisation" cannot start sorting locally.

### QA-2 · a measurement artifact read as a behaviour change

The chart now marks each instant where the data's source changes: a **dashed hairline drawn under the lines**, captioned `← cost log · 60s resolution`. The arrow is the load-bearing part — it makes the caption describe the data *before* the rule rather than the rule itself. `.secondary` / `--muted` at 45 % so both themes resolve it; findable when looked for, invisible when not.

Drawn **only when the visible range straddles a boundary**, and never when a boundary lands on the first or last bucket — a rule on the axis marks nothing and reads as a chart border.

**One mechanism, not a case per pair.** The store now reports the earliest start per *provenance* — `EraStarts`, keyed by the row's own `source` field, `""` for measured — and the daemon sorts those eras and emits a handover between each adjacent pair. `cost→log` and `log→live` fall out of the same loop, and so does a source this build has never heard of (there is a test for exactly that). `live` is the wire's name for a measured row's absent source and is never written to a row; a test asserts it stays out of `session.AutonomySources()`.

### Tests

New checks on both surfaces, both directions:

| check | web | macOS |
|---|---|---|
| more projects than the cap → exactly N rows + "+N more" | `a 95-project window draws exactly the cap and accounts for the rest` | `testStripDrawsExactlyTheCapAndAccountsForTheRest` |
| projects fit → nothing said | `nothing is said when every project fits` | `testStripSaysNothingWhenEveryProjectFits` |
| range straddles a boundary → marker drawn | `a range that straddles a boundary marks it, at the right fraction` | `testARangeStraddlingABoundaryMarksIt` |
| range does not → nothing drawn | `a range that does not reach the boundary marks nothing` | `testARangeThatDoesNotStraddleABoundaryMarksNothing` |

Plus: a boundary on either edge is not drawn, both handovers come from the one mechanism, a degenerate domain cannot divide by zero, an unknown era still gets a label, and `testBoundaryLabelsMatchTheWebs` pins the two caption tables together so one section cannot show two explanations of one artefact.

A sixth `mutate.sh` fixture was added — the boundary loop stops emitting, and `TestAutonomyBoundaries*` goes red:

```
ok  a reconstructed span still reads back as reconstructed
ok  a span crossing a daemon restart is dropped
ok  a close with no matching open emits nothing
ok  a cost-derived span carries `unknown` and never a guessed reason
ok  a span reaching into the era the daemon already measured is dropped
ok  a handover between two sources is reported
autonomy-backfill-mutations: ALL PASS
```

The two client-side checks carry committed in-language mutants, same idiom as before: an always-speaks build and a never-speaks build answer identically for both fixtures of each pair, and production must not.

### Preflight

`web` PASS · `swift` PASS (the two asked for), and since the ports/adapter/daemon layers changed too: `go` PASS · `tools` PASS · `arch` PASS. Each its own foreground invocation.

Not merged.

Issue #1825, opened 2026-08-25
Recording drivers assert process death instead of observing it — every exit_clean run leaks a live claude + tmux session


ingo-eichhorst:
## Summary

Every onboarding-factory recording whose recipe ends in `exit_clean` leaks a live
`claude` process plus its detached tmux session. Nine such runs today left nine
leaked sessions on the dev machine — eight from the 07:00–09:07 batch, one more
from a 12:32 run — each showing up as a real session row in the Irrlicht menu bar
and crowding out the two sessions actually in use.

The daemon is not at fault: every leaked PID answered `ps`, and the `readyTTL`
sweep correctly skips a stale `ready` session whose process is alive. Killing one
leaked pane dropped the menu-bar count from 11 rows to 10, so the two numbers are
the same number.

## Root cause

`replaydata/agents/claudecode/driver-interactive.sh:440-449`

```bash
step_exit_clean() {
  tmux send-keys -t "$CURRENT_TMUX" C-d   # one press
  sleep 1                                 # waits, never looks
  SES_ALIVE[$ACTIVE]=0                    # asserts death unconditionally
  echo "[driver] exit_clean: sent Ctrl-D to $CURRENT_TMUX" >&2
}
```

Claude Code answers a single `Ctrl-D` with `Press Ctrl-D again to exit`, then lets
that confirmation expire. The driver never checks, and records the slot as dead.

The end-of-run teardown loop — the one net that would have caught this — is gated
on the flag the failing step just cleared:

`replaydata/agents/claudecode/driver-interactive.sh:616-621`

```bash
for (( i = 1; i <= N_SLOTS; i++ )); do
  if [[ "${SES_ALIVE[$i]}" == "1" ]]; then   # false, so the kill never runs
    tmux kill-session -t "${SES_TMUX[$i]}" 2>/dev/null || true
  fi
done
```

The run then writes `driver.exit-reason=ok` and `run-cell.sh` writes a `STAGED`
manifest with `completeness: complete`.

### Verified by experiment

On a leaked pane (`claudecode-onboard-1787641617-95883`, pid 95895):

- one `Ctrl-D` → process alive, pane unchanged;
- a second `Ctrl-D` sent 1.5 s later → still alive (the confirmation window had
  already closed);
- `tmux send-keys ... C-d C-d` in a single call → `PID 95895 GONE`, `tmux session: GONE`.

Typing `zz` into the pane between attempts confirmed keystrokes were reaching the
[...]

PR #1917, opened 2026-09-05
Recording-rig follow-ups from #1825: name the slot flag for ownership, refuse double-run suites, split the pid verdict (#1828)

### LINKED ISSUE #1828: Recording-rig follow-ups from #1825: retire SES_ALIVE, refuse double-run suites, and four smaller gaps
Surfaced while fixing #1825 (PR #1827). Each is real, each was deliberately left
out of that PR to stop it growing further, and each is independently landable.

## 1. `SES_ALIVE` is named like an observation and used like an intent flag

#1825's root cause was a teardown gated on `SES_ALIVE`, which nothing anywhere
re-derives from `tmux has-session`. PR #1827 removed its last load-bearing uses
but left the name, so four drivers now carry a
`# shellcheck disable=SC2034` saying the flag is write-only in that file
(`git diff` count: `claudecode`, `codex`, `gemini-cli`, `pi`).

A `slot_session_alive <i>` helper in `replaydata/_lib/drive/slots.sh` that
actually asks `tmux has-session` would make "alive" an observation fleet-wide and
retire both the array and all four annotations. `kiro-cli` and `antigravity`
genuinely branch on their copies (slot-aliasing guards after `reset_session`), so
those call sites become the helper's first real consumers rather than obstacles.

## 2. Nothing refuses a test suite that runs twice

`tools/onboarding-factory/scripts/smoke-test.sh` discovers suites via
`shell_lib_suite_run`, but PR #1827 initially re-added two hand-written
registration blocks four lines below the glob that already found them. **Both
suites ran twice for a full PR cycle and no gate noticed** — it was caught by a
reviewer counting `ALL PASS` lines.

The `shell-lib-suite` census line (`found 18, skipped 0, ran 18, failed 0`) is
where a "no file executed twice in one run" refusal belongs. That makes the
hand-listed shape mechanically un-reintroducible instead of relying on someone
counting.

## 3. A `driver.pid` write failure classifies as a tmux problem

PR #1827 made the *message* distinguish three cases — file absent (the wrapper
never ran, staging unwritable), present-but-empty, and present-but-non-numeric —
but all three still classify as `driver_teardown_unverifiable`.

A distinct `driver_pid_unrecorded` arm in
`tools/onboarding-factory/scripts/lib/classify-failure.sh` would let the
classifier give the right first move ("check `$STAGING` is writable; the driver
never started") instead of "the tmux-teardown check could not be made". The
bidirectional writer⟺arm tripwire added in #1827 will keep the new code in sync
automatically.

## 4. The teardown deadline clamp is duplicated

`clamp(lifetime/10, 1, 5)` is computed in both `run-cell.sh` and
`run-cell-multi.sh`. A `tmux_teardown_deadline_for <lifetime>` helper belongs in
`lib/tmux-teardown-check.sh` beside the check that consumes it; the duplication
exists only because that file was outside the implementing agent's scope.

## 5. `DRIVE_EXIT_CLEAN_CAP_S=15` is a bound, not a measurement

Stated honestly in its own comment, and deliberately so: six adapters were raised
from an inherited `2` to a uniform 15 s because the strict poll turned that
number from a settle budget into a hard deadline that SIGHUPs a pane mid-flush.

A measured per-adapter table would be better, and needs live recording runs on a
loaded host, repeated to see the tail. Blocked in part on gemini-cli, which is
not drivable here (the account is tier-ineligible — see the note in its own
driver). Worth doing when that changes; the poll returns on first observation, so
the current value costs nothing on the happy path.

ingo-eichhorst:
Closes #1828 (items 1-4; item 5 is parked with its blocker, see the [decision
record](https://github.com/ingo-eichhorst/Irrlicht/issues/1828#issuecomment-5553495445)).

Three commits, one per item group. The issue was `needs-info` on two questions;
both are answered in that comment.

## What changed

**1 — `SES_ALIVE` → `SES_OWNED`** (`c6739f3`). The issue proposed retiring the
array behind a `tmux has-session` helper. That is not safe: [verified by read]
`kiro-cli/driver-interactive.sh:558-566` — `reset_session` reuses the SAME tmux
session for its successor slot, so a retired slot number can alias a session a
still-live slot now owns, and `has-session` reports "alive" there. The array is
an ownership record and stays; the name was the defect. No helper was added —
[verified by read] `teardown.sh` already ships `wait_tmux_session_gone` and
`require_tmux_session_gone`. The four `SC2034` waivers shrink to one line each.

**2 — a test file that runs twice is now a refusal** (`12b74d4`).
`shell_lib_suite_exec` runs each file through a run-scoped ledger keyed on a
canonicalized absolute path. `shell_lib_suite_run`'s loop and smoke-test.sh's
three out-of-tree files both go through it, so a re-added duplicate is caught
wherever it is added. `cd … && pwd -P`, not `readlink -f` — stock macOS has
neither that flag nor a guaranteed `realpath`. A directory that cannot be
entered is a refusal, never a fallback to the raw string.

**3 — `driver_pid_unrecorded`** (`1492e37`). A `driver.pid` that was never
written no longer classifies as a tmux-teardown problem. The other two pid
failures keep the teardown code: they mean the driver DID start.

**4 — `tmux_teardown_deadline_for`** (`1492e37`). The `clamp(lifetime/10, 1, 5)`
duplicated in `run-cell.sh` and `run-cell-multi.sh` moves beside the check that
consumes it.

## Three defects found along the way

None is in the issue. All three were invisible, and all three are now held by
tests rather than by care:

1. **INV-1 went blind.** `driverteardown` identified the forbidden gate by
   matching the substring `alive`. The rename removed it, so the checker stopped
   detecting a driver gating teardown on the fleet's own array — the exact
   anti-pattern it exists to catch — while every fixture stayed green. Proven by
   mutation (0 findings). The pattern now covers both spellings, and
   `TestAliveGateMatchesTheSharedOwnershipFlag` binds it to the names `slots.sh`
   declares.
2. **The generator template drifted.** `templates/drive-interactive.sh.tmpl` was
   left declaring `SES_ALIVE` while `slots.sh` wrote `SES_OWNED`. The next
   generated adapter would have carried a dead array and no declaration for the
   live one; both files lint clean either way.
   `TestTemplateDeclaresExactlyTheSharedSlotArrays` requires the sets to be equal.
3. **Item 2's own implementation truncated runs.** `shell_lib_suite_exec`
   returned 2 for both "the ledger refused" and "the test file exited 2", so one
   such file aborted the run before its census line, every file behind it
   silently unrun. That is a truncated run reading as a clean one — the defect
   that library exists to remove. The refusal now travels out-of-band.

## Verification

Every item is an addition, so none has a "before the fix" red run. Each is
proven by a mutation seen RED before the fix and green after, and each mutation
is committed as a fixture:

| Mutation | Red output |
|---|---|
| delete the classifier arm | `expected [driver_pid_unrecorded] got [unknown]` |
| gate `good_driver.sh`'s sweep on `${SES_OWNED[$i]}` | INV-1 fires at `:52` (0 findings before the widening) |
| narrow the pattern back to `(?i)alive` | binding test names all seven arrays |
| revert the template to `SES_ALIVE` | drift test prints both sets |
| restore the conflated `exec_rc -eq 2` check | `zzz_test.sh` never ran, no census line |
| re-inline the clamp | textual guard flags it |

One honest limit, stated in the test's own comment: the clamp guard is a narrow
text match, not a general duplication detector. A third call site under other
variable names would escape it.

Gates, all local, all PASS: `tools/preflight.sh --only bash` / `--only posix` /
`--only skills` / `--only tools` / `--only go` (which covers `replaydata
validate`, `recording-rig smoke test` and `replay fixtures`). The smoke test now
prints exactly one census line; it printed two.

https://claude.ai/code/session_01615niHdVNthGbskmVG4JVt


ingo-eichhorst:
## Item 5 is now in this PR too (`0098900`)

The blocker was real but only covered half the item. The measurement needs no
campaign: `require_tmux_session_gone` already counted the wall-clock the cap's
own comment asked for, in its tick variable, and threw it away on every run the
rig has ever done.

It now publishes `TMUX_SESSION_GONE_ELAPSED_S` and appends a row per teardown to
`$DRIVE_TEARDOWN_TIMINGS`. Both rigs set that path and fold it into the run
manifest, so every ordinary recording contributes a data point. Env-driven, so
an adapter added later is instrumented by sourcing the library — no per-driver
line, which is the duplication item 4 removed.

Two design points worth review:

- **Three outcomes, not two.** "No rows" is legitimate (a recipe with no
  `exit_clean`), so it must not read the same as "the rig never set the path".
  Both rigs pre-create the file; `teardown_timings_json` reports `unrecorded` /
  `no_exit_clean` / `recorded`.
- **Capped rows are counted apart from `max_s`.** A capped row is censored — the
  session was still alive when the deadline fired, so its true time is unknown
  and greater than the cap. Averaging it in as a measurement is how a fitted
  table ends up tighter than the behaviour it describes.

Mutations, committed as fixtures: dropping the recording call leaves the elapsed
stale at a previous run's value (`expected [0.6] got [1.0]`) and writes one row
instead of three; making an absent file report `no_exit_clean` collapses the
outage into the legitimate answer.

**What is still open, so the collection is not read as the conclusion.** Nothing
is fitted. 15 stays a bound. This host contributes no gemini-cli rows at all —
[verified] `gemini-cli/driver-interactive.sh:205`, the account is
tier-ineligible — and gemini-cli is one of the two Ink/node TUIs whose teardown
is the latency being bounded. [verified] PR #1725 merged 2026-08-20 and
schedules that adapter's deprecation at v0.7, so the hole may close by
retirement rather than by measurement.

Gates re-run after this commit: `--only bash` PASS, `--only go` PASS (covering
`core module tests`, `onboarding-factory tests`, `replaydata validate`,
`recording-rig smoke test`, `replay fixtures`).

Issue #1893, opened 2026-09-04
test(onboard): record Claude Desktop agent and metrics scenarios


ingo-eichhorst:
Part of #974 — Desktop recording campaign D. Depends on #1886 and #1888.

## Scope

Run these current `claudecode` catalog cells through Claude Code Desktop Local on macOS:

- section 3: `3.1` through `3.5`;
- section 5: `5.1` through `5.8`.

## Required result for each cell

Record exactly one explicit Desktop result: observed passing, observed failure, not applicable, unobservable with evidence, or not runnable with the missing Desktop control named.

For observed cells, preserve raw Desktop registry, transcript, hook, process, and Irrlicht evidence. Record the Desktop app, bundled Claude Code, and Irrlicht versions. Preserve each existing recipe intent. Do not reuse CLI or synthetic evidence.

## Acceptance

- All thirteen named cells have a `desktop-local` result.
- Applicable recordings pass replay and expected-result checks, or carry an observed failure classification.
- Subagent, background-process, cleanup, token, model, subscription, and task-estimate behavior is classified from Desktop evidence.
- `5.5 provider-failover-midturn` and `5.7 quota-burndown` keep evidence-based unobservable or not-runnable reasons unless Desktop supplies a new observable signal.
- The Desktop completeness check passes for this assigned subset.

## Exclusions

- Do not change daemon behavior to make a scenario pass.
- Do not change CLI Local results.
- Do not record cells outside sections 3 and 5.

ingo-eichhorst:
> *This was generated by AI during triage.*

## Triage Assessment

**Category:** enhancement — adds Desktop Local evidence for a bounded catalog subset.
**Scope:** scope:claudecode, scope:replay — thirteen subagent, background, model, and metrics cells receive separate Desktop results.
**Complexity:** 6 (base 6, uncertainty +0) — thirteen established recipes require live runs, evidence review, and fixture promotion.

**Readiness:**
- ✓ **Scope** — sections 3 and 5 define exactly thirteen cells.
- ✓ **Specification** — allowed results, evidence, versions, and special unobservable cases are fixed.
- ✓ **Verifiability** — replay, expected-result checks, identity validation, and subset completeness give concrete results.
- ✗ **Observability** — Desktop recordings do not exist yet; the planned driver supplies them at Low run cost once its prerequisites land.
- ✓ **Context** — the existing recipes identify subagent, background, model, subscription, and metrics behavior.
[...]

Issue #1892, opened 2026-09-04
test(onboard): record Claude Desktop turn scenarios 2.14 through 2.27


ingo-eichhorst:
Part of #974 — Desktop recording campaign C. Depends on #1886 and #1888.

## Scope

Run Claude Code catalog cells `2.14` through `2.27` through Claude Code Desktop Local on macOS.

## Required result for each cell

Record exactly one explicit Desktop result: observed passing, observed failure, not applicable, unobservable with evidence, or not runnable with the missing Desktop control named.

For observed cells, preserve raw Desktop registry, transcript, hook, process, and Irrlicht evidence. Record the Desktop app, bundled Claude Code, and Irrlicht versions. Preserve each existing recipe intent. Do not reuse CLI or synthetic evidence.

## Acceptance

- All fourteen named cells have a `desktop-local` result.
- Applicable recordings pass replay and expected-result checks, or carry an observed failure classification.
- Error, interrupt, streaming, permission, and blocking-dialog paths are exercised through Desktop where runnable.
- Cells `2.26` and `2.27` are re-run against current `main`. Their prior `blocked-daemon` result does not predetermine the Desktop result because #1861 was closed by #1867.
- The Desktop completeness check passes for this assigned subset.

## Exclusions

- Do not change daemon behavior to make a scenario pass.
- Do not change CLI Local results.
- Do not record any cell before `2.14` or after `2.27`.

ingo-eichhorst:
> *This was generated by AI during triage.*

## Triage Assessment

**Category:** enhancement — adds Desktop Local evidence for a bounded catalog subset.
**Scope:** scope:claudecode, scope:replay — fourteen error, interrupt, and blocking cells receive separate Desktop results.
**Complexity:** 6 (base 6, uncertainty +0) — fourteen established recipes require live runs, evidence review, and fixture promotion.

**Readiness:**
- ✓ **Scope** — cells 2.14 through 2.27 form an exact boundary.
- ✓ **Specification** — allowed results, revalidation of 2.26 and 2.27, evidence, and exclusions are fixed.
- ✓ **Verifiability** — replay, expected-result checks, identity validation, and subset completeness give concrete results.
- ✗ **Observability** — Desktop recordings do not exist yet; the planned driver supplies them at Low run cost once its prerequisites land.
- ✓ **Context** — each named cell has a current recipe or explicit unobservable condition; PR #1867 verifies why old daemon-blocked labels must not predetermine results.
[...]

Issue #1891, opened 2026-09-04
test(onboard): record Claude Desktop turn scenarios 2.1 through 2.13


ingo-eichhorst:
Part of #974 — Desktop recording campaign B. Depends on #1886 and #1888.

## Scope

Run Claude Code catalog cells `2.1` through `2.13` through Claude Code Desktop Local on macOS.

## Required result for each cell

Record exactly one explicit Desktop result: observed passing, observed failure, not applicable, unobservable with evidence, or not runnable with the missing Desktop control named.

For observed cells, preserve raw Desktop registry, transcript, hook, process, and Irrlicht evidence. Record the Desktop app, bundled Claude Code, and Irrlicht versions. Preserve each existing recipe intent. Do not reuse CLI or synthetic evidence.

## Acceptance

- All thirteen named cells have a `desktop-local` result.
- Applicable recordings pass replay and expected-result checks, or carry an observed failure classification.
- Basic turn, tool use, task list, slash command, long run, queued input, permission behavior, compaction, and terminal-text behavior are each exercised through the Desktop entry point.
- The Desktop completeness check passes for this assigned subset.

## Exclusions

- Do not change daemon behavior to make a scenario pass.
- Do not change CLI Local results.
- Do not record any cell after `2.13`.

ingo-eichhorst:
> *This was generated by AI during triage.*

## Triage Assessment

**Category:** enhancement — adds Desktop Local evidence for a bounded catalog subset.
**Scope:** scope:claudecode, scope:replay — thirteen Claude Code turn cells receive separate Desktop results.
**Complexity:** 6 (base 6, uncertainty +0) — thirteen established recipes require live runs, evidence review, and fixture promotion.

**Readiness:**
- ✓ **Scope** — cells 2.1 through 2.13 form an exact boundary.
- ✓ **Specification** — allowed results, preserved intent, evidence, and exclusions are fixed.
- ✓ **Verifiability** — replay, expected-result checks, identity validation, and subset completeness give concrete results.
- ✗ **Observability** — Desktop recordings do not exist yet; the planned driver supplies them at Low run cost once its prerequisites land.
- ✓ **Context** — each named cell already has a Claude Code recipe and expected behavior.
- ✗ **Independence** — recording requires #1886 and #1888.
- ✓ **Reversibility** — recording data and classifications do not change daemon behavior.

**Verdict:** needs-info

**Blocked on:**
[...]

Issue #1887, opened 2026-09-04
feat(onboard): drive a basic Claude Code Desktop Local session


ingo-eichhorst:
Part of #974 — Desktop driver core. Depends on #1884 and #1885.

## Goal

Add the first safe end-to-end onboarding driver for Claude Code Desktop with the Local environment on macOS.

## Required behavior

- Use the official `claude://code/new` route to open a new Code composer.
- Select Local and the exact staging workspace.
- Use an isolated Desktop profile when the application supports one. Otherwise use a verified snapshot-and-restore boundary for affected app-wide settings, hooks, MCP servers, skills, and plugins.
- Restore prior state after success, failure, interruption, and timeout.
- Refuse the run when safe restoration is not possible.
- Set and submit a no-tool prompt through the Desktop UI.
- Wait on observed UI, registry, transcript, process, hook, and Irrlicht conditions with deadlines.
- Map the Desktop `local_...` session ID to the Claude Code transcript ID.
- Verify the registry `cwd`, transcript `cwd`, and requested workspace.
- Verify `launcher.host_bundle_id: "com.anthropic.claudefordesktop"` and transcript `entrypoint: "claude-desktop"`.
- Archive only the session that this driver created.
- Leave the user worktree and normal Desktop sessions unchanged.
- Record the Desktop app, bundled Claude Code, and Irrlicht versions.

## Acceptance

- A live basic-turn smoke run reaches `ready → working → ready` through Claude Desktop.
- A live archive smoke run observes the created Claude process exit and Irrlicht remove that session.
- A nested staging workspace that opens as **No folder** is rejected by the `cwd` checks.
- Every wait has a deadline and names the missing condition on failure.
- Interruption and timeout tests prove restoration and owned-session cleanup.
- The driver emits the execution-profile and evidence files required by #1884.

## Runtime prerequisites

- A supported Claude Desktop version is installed.
- Claude Code access and an active login are available.
- The driver process has macOS Accessibility access.
- The repository helper from #1885 is available.
- The run has exclusive control of Claude Desktop.

## Exclusions

- Do not add advanced recipe actions such as model changes or multiple sessions.
- Do not run the complete scenario catalog.
- Do not change the public support page.

## Evidence requirements

The nested-workspace check is a defect test and must be observed failing without the registry/transcript `cwd` guard. Restoration and ownership checks need mutation proof. The two live smoke runs are manual macOS evidence.

ingo-eichhorst:
[...]

PR #1932, opened 2026-09-06
fix(onboard): unblock the remaining Claude Desktop cells

ingo-eichhorst:
## Summary

- Re-resolve Desktop controls after each accessibility-tree change.
- Stop Send retries after the first click can have started a turn.
- Re-resolve the owned archive target on every attempt.
- Give agent turns the full cell timeout.
- Accept `waiting` as a completed user-blocking turn.
- Guard later prompts with the selected owned conversation.
- Resolve Stop directly while Send is absent.

## Red-first evidence

The helper postcondition tests failed before the helper fix. A transient post-click tree read escaped as `actionFailed` instead of `postconditionFailed`.

The driver regression tests failed before the state-aware fixes:

```
go test ./tools/onboarding-factory/internal/desktopdriver -run "TestInterruptClicksStop|TestSetPromptResolves|TestWaitingRequires" -count=1
```

The failures reproduced all three gaps. Interrupt searched for an absent Send control. SetPrompt did not retry a missed value postcondition. A live `waiting` state passed without a recorded `working -> waiting` transition.

The ownership mutation test also failed when its selected-conversation guard was disabled. The restored guard passes.

## Verification

- `go test ./core/... -race -count=1`
- `go test ./tools/onboarding-factory/... -race -count=1`
- `tools/replay-fixtures.sh`
- `tools/state-vocabulary-lint.sh`
- `tools/preflight.sh --only go`
- `tools/preflight.sh --only arch`
- `tools/preflight.sh --changed --budget 540` through the pre-push hook

Refs #1887
Refs #1891
Refs #1892
Refs #1893


ingo-eichhorst:
Follow-up from live Desktop evidence: cell 2.17 completed its turn but cleanup saw a Claude Code CLI cache-sweep marker under `.claude/plugins/cache`. Commit `cbafcff8` excludes only the CLI-derived plugin cache paths. It keeps the installed plugin set and marketplace definitions guarded. The red-first cache test and two guard-lock tests pass. Commit `388aba04` also splits the methods named by the first CodeScene report.

PR #1966, opened 2026-09-15
fix(homebrew): drop deprecated verified: from the cask url stanza

fhoef:
## Summary

Every `brew` invocation that loads the Irrlicht cask prints a deprecation warning for the `verified:` parameter on the `url` stanza. Homebrew 7 lists `:verified` in `Cask::URL::DEPRECATED_URL_SPECS` and accepts it only "as a no-op for compatibility" — it no longer does anything, since Homebrew verifies the URL's origin itself now. Dropping the line removes the warning with no behavior change.

## Changes

- `tools/homebrew-tap/Casks/irrlicht.rb`: remove `verified:` from the `url` stanza (and the now-trailing comma on the url line).

Fixed in the canonical template rather than in the tap repo: `tools/homebrew-tap/update-cask.sh` copies this file into `ingo-eichhorst/homebrew-irrlicht` on release and rewrites only `version` and `sha256`, so a tap-side fix would be overwritten on the next release run.

## Test plan

- [x] `ruby -c tools/homebrew-tap/Casks/irrlicht.rb` -> Syntax OK
- [x] `brew style tools/homebrew-tap/Casks/irrlicht.rb` -> 1 file inspected, no offenses detected
- [x] Confirmed `verified:` appears nowhere else in a Homebrew context in this repo
- [x] Confirmed `update-cask.sh` does not rewrite the `url` stanza, so the template change survives the next release run
- [ ] Full test suite not run — this PR touches only the Homebrew cask template, no code in `core/`, `platforms/`, or `tools/onboarding-factory/`

## Related issues

Follow-up to d0172d75 (`fix(homebrew): use symbol form for depends_on macos`) — same file, same class of Homebrew 7 cleanup.

## Checklist

- [x] Follows the conventions in [AGENTS.md](../AGENTS.md) and [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] Commit messages use [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Documentation updated if behavior changed — n/a, no user-visible behavior change
- [x] No `cancelled` session state introduced — three states only: `working`, `waiting`, `ready`


ingo-eichhorst:
Thank you — this is a lovely little PR. 🙏

I re-verified your claims locally and they all hold:

- `Library/Homebrew/cask/url.rb:41` — `DEPRECATED_URL_SPECS = [:verified].freeze`
- `cask/dsl.rb:450` builds the URL with `options.except(*URL::DEPRECATED_URL_SPECS)`, so the value never reaches `URL.new` — a true no-op, exactly as you wrote.
- `update-cask.sh` rewrites only `version` and `sha256` (two `re.sub` calls), so your fix in the canonical template survives the next release run. Choosing the template over a tap-side patch was the right call.

One detail that strengthens the case further: the deprecation is already live on Homebrew 6.0.22, not only on 7 — and `odeprecated` raises a hard error under `HOMEBREW_DEVELOPER=1`, so this was more than cosmetic for anyone with a developer install.

The live tap still carries the old stanza; it picks up your change on the next release run.

Thanks also for the precise test plan and for saying plainly which checks you did not run. That made the review fast. Merging now. 🎉

